PR #55, opened 2014-10-29
Make media conversion work on file sensitive system

SkySkimmer:
My AoE installation has directories like DATA where openage expects Data. This pull request should solve the general case.

### DIFF
--- a/py/openage/convert/util.py
+++ b/py/openage/convert/util.py
@@ -144,7 +144,7 @@ def file_get_path(fname, write=False, mkdir=False):
         basedir = readpath
 
     path = path_insensitive(os.path.join(basedir, fname))
-	
+
     if mkdir:
         dirname = os.path.dirname(path)
         mkdirs(dirname)


Almin17:
I have the same problems!
It's the Collectors Edition.
http://www.amazon.de/Age-of-Empires-Collectors-Edition/dp/B000X23MIM/ref=sr_1_2?ie=UTF8&qid=1420122563&sr=8-2&keywords=Age+Of+Empires+Collectors+Edition
http://www.amazon.de/Ubisoft-Age-Empires-Collectors-PC/dp/B00BB629T2/ref=sr_1_1?ie=UTF8&qid=1420122563&sr=8-1&keywords=Age+Of+Empires+Collectors+Edition
http://www.amazon.de/Age-of-Empires-Collectors-Edition/dp/B00004Z3L8/ref=sr_1_3?ie=UTF8&qid=1420122563&sr=8-3&keywords=Age+Of+Empires+Collectors+Edition

also language.dll is LANGUAGE.DLL
and there is no Bin folder in agedir.

goto-bus-stop:
`Bin/` is an HD Edition folder, but the convert script looks for it when it can't find `language.dll`.

> One should determine a finite set of possible paths (in this case "Data" and "DATA") and only check for all wellknown paths if one of them exists.

I guess that makes `DATA` → `Data` and `LANGUAGE.DLL` → `language.dll`. I don't know if there's any more different versions, maybe just check for `filename` and `uppercase(filename)`?

franciscod:
http://www.brain-dump.org/projects/ciopfs/

franciscod:
what if we rename all things to lowercase before starting media conversion?

StepS-:
what if the things are not writable?

franciscod:
make symlinks on /tmp, rename symlinks?

PR #150, opened 2014-11-12
Added block allocator

schets:
I've written up a block allocator for single objects with cheap new/delete variants and pretty decent cache characteristics. It's been tested with the linked list and pairing heap. It was originally going to go with some pathfinding/unit stuff, but Jon0 wayyyyy beat me to that punch. Figured it could still be helpful though.

Also added a hash function for tiles and made some pointers into std::vectors in the terrain class.


TheJJ:
continued in #196 and #232

PR #165, opened 2014-11-19
WIP - load and save maps for testing

Jon0:
Pull my other branch first. This is currently being based off of that work.

some save (f5) and load (f9) map functions. It takes a while to setup a map with the right objects for testing so this just makes it easier. this is not a complete save and load -- units current state is dropped.
uses a filename "/tmp/saved_map.txt" to save and load.


franciscod:
rebase and :ship: :it:

PR #182, opened 2014-12-28
Fixing memory leaks and unsafe attribute casting

schets:
There are some memory leaks in the unit texture outlines and locations, and unsafe operations with the attributes. This fixes the leaks and puts compile time restrictions on attribute typing


TheJJ:
most (if not all) of this is fixed now, and the codebase has evolved significantly.

PR #185, opened 2015-01-03
Conversion: support for some uppercase file names

goto-bus-stop:
The AoK:TC Collector's Edition has uppercase folder and file names for a bunch of base game files. (i.e. AoK, not AoC files.)

This attempts to fix with a different approach than #55, by just special casing for the uppercase names in addition to the normal names.

Supports `DATA/` in addition to `Data/`, and `LANGUAGE.DLL` in addition to `language.dll`.

I only this with my AoK-only disk, which also has those uppercase names, but they might not quite be the same on the Collector's Edition. So if somebody with a CE disk could confirm, would be much appreciated :)

### DIFF
--- a/py/openage/convert/mediafile.py
+++ b/py/openage/convert/mediafile.py
@@ -1,4 +1,4 @@
-# Copyright 2013-2014 the openage authors. See copying.md for legal info.
+# Copyright 2013-2015 the openage authors. See copying.md for legal info.
 
 # media files conversion stuff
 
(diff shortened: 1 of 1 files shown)


LordAro:
should use .format() rather than %, surely?

goto-bus-stop:
Probably, but I went with `%` because that's what the rest of the convert script uses

TheJJ:
Thanks for another attempt of hardcoding version diffferences :smile: .
Please meld the commits, after that we can merge the pull :+1:

goto-bus-stop:
Maybe @Almin17 can confirm if this works on their Collector's Edition version too, before merging?

Almin17:
Yes, it does work!

StepS-:
My data directory is all-lowercase "data" instead. Shouldn't there be an unified way to fix these things?

franciscod:
ciopfs, lol

goto-bus-stop:
@StepS- Yes, in #55. I guess that might be a better option after all.

PR #187, opened 2015-01-04
convert: fixes strings with garbage data past their \0 terminators

goto-bus-stop:
Real world case: empires2_x1.dat has a string `b'M\x00\xf8\x004"\xf8\x00\x00\x00\x00\x00\x00'` for one of the graphic `name1`s, which has a bunch of garbage after its terminator. `data.decode('utf-8')` trips up on this because `\xf8` is not a valid unicode char.

This patches `zstr` to truncate data to right before the first \0, or to keep it entirely if there is no \0, before utf8-decoding it.

Thanks to rkreis on #sfttech IRC for the report & patch hint :)

### DIFF
--- a/py/openage/convert/util.py
+++ b/py/openage/convert/util.py
@@ -1,4 +1,4 @@
-# Copyright 2013-2014 the openage authors. See copying.md for legal info.
+# Copyright 2013-2015 the openage authors. See copying.md for legal info.
 
 import os
 import os.path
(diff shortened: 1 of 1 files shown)


goto-bus-stop:
Oh. Should probably fix copyright years throughout.

mic-e:
Looking good (apart from the stuff that travis complains about) :+1:

goto-bus-stop:
Better :)

TheJJ:
mkay, this should help indeed. merged :+1: etc

PR #186, opened 2015-01-04
Changes to AssetManager

Ethon:
Everything is described in the commit description.

PR #190, opened 2015-01-08
convert script updates and memory fixes

TheJJ:
Includes the #186 changes for loading the "missing" texure lazily.
Additionally, fixed pointer owners by smart pointers (`unique_ptr`, `shared_ptr`).

More convert script updates: Variables for tech, units, terrain, graphics and sounds uncovered.

Fixes to the codecompliance script to be compatible with `git rebase` and `cherry-pick`.

Also we got some new ideas from 31c3 people :grinning:

### DIFF
--- a/cpp/assetmanager.h
+++ b/cpp/assetmanager.h
@@ -1,4 +1,4 @@
-// Copyright 2014-2014 the openage authors. See copying.md for legal info.
+// Copyright 2014-2015 the openage authors. See copying.md for legal info.
 
 #ifndef OPENAGE_ASSETMANAGER_H_
 #define OPENAGE_ASSETMANAGER_H_
@@ -7,6 +7,7 @@
 
 #include <unordered_map>
 #include <string>
+#include <memory>
 
 #include "texture.h"
 
@@ -16,10 +17,9 @@ namespace openage {
  * Container class for all available assets.
  * Responsible for loading, providing and updating requested files.
  */
-class AssetManager {
+class AssetManager final {
 public:
 	AssetManager(util::Dir *root);
-	virtual ~AssetManager();
 
 	/**
 	 * Test whether a requested asset filename can be loaded.
@@ -46,7 +46,12 @@ protected:
 	/**
 	 * Create an internal texture handle.
 	 */
-	Texture *load_texture(const std::string &name);
+	std::shared_ptr<Texture> load_texture(const std::string &name);
+
+	/**
+	 * Retrieves the texture for missing textures.
+	 */
+	std::shared_ptr<Texture> get_missing_tex();
 
 private:
 	/**
@@ -57,12 +62,12 @@ private:
 	/**
 	 * The replacement texture for missing textures.
 	 */
-	Texture *missing_tex;
+	std::shared_ptr<Texture> missing_tex;
 
 	/**
 	 * Map from texture filename to texture instance ptr.
 	 */
-	std::unordered_map<std::string, Texture *> textures;
+	std::unordered_map<std::string, std::shared_ptr<Texture>> textures;
 
 #if WITH_INOTIFY
 	/**
@@ -74,7 +79,7 @@ private:
 	 * Map from inotify watch handle fd to texture instance ptr.
 	 * The kernel returns the handle fd when events are triggered.
 	 */
-	std::unordered_map<int, Texture *> watch_fds;
+	std::unordered_map<int, std::shared_ptr<Texture>> watch_fds;
 #endif
 };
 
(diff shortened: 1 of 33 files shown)


TheJJ:
@Ethon are you ok with this? I also replaced the hardcoded member-fonts with a hashmap (see 
5d1ab2d).

cpp/assetmanager.h
```diff
@@ -74,7 +79,7 @@ class AssetManager {
 	 * Map from inotify watch handle fd to texture instance ptr.
 	 * The kernel returns the handle fd when events are triggered.
 	 */
-	std::unordered_map<int, Texture *> watch_fds;
+	std::unordered_map<int, std::weak_ptr<Texture>> watch_fds;
```
fr3akout: I would use a shared_ptr in this map, as well. The weak_ptr does not avoid any circular references and with every usage it is converted to a shared_ptr using the lock-function anyway.

cpp/game_main.h
```diff
@@ -89,8 +95,7 @@ class GameMain :
 	/**
 	 * map graphic id to gamedata graphic.
 	 */
-	std::unordered_map<int, gamedata::graphic *> graphics;
-
+	std::unordered_map<ssize_t, gamedata::graphic *> graphics;
```
fr3akout: Is there a specific reason for using ssize_t as index type. I thought (s)size_t is only used for machine-width-integers for memory management.

cpp/terrain/terrain.cpp
```diff
@@ -46,16 +47,15 @@ Terrain::Terrain(AssetManager &assetmanager,
...
-	this->textures                 = new Texture*[this->terrain_id_count];
-	this->blending_masks           = new Texture*[this->blendmode_count];
-	this->terrain_id_priority_map  = new int[this->terrain_id_count];
-	this->terrain_id_blendmode_map = new int[this->terrain_id_count];
-	this->influences_buf           = new struct influence[this->terrain_id_count];
+	this->textures.reserve(this->terrain_id_count);
+	this->blending_masks.reserve(this->blendmode_count);
+	this->terrain_id_priority_map  = util::make_unique<int[]>(this->terrain_id_count);
+	this->terrain_id_blendmode_map = util::make_unique<int[]>(this->terrain_id_count);
+	this->influences_buf           = util::make_unique<struct influence[]>(this->terrain_id_count);
```
fr3akout: Why don't use a std::vector instead of unique_ptrs to arrays. Those vectors can be passed to all functions, that use them in a readonly way, as a const reference. Therefore unnecessary copies will be avoided.

cpp/terrain/terrain.cpp
```diff
@@ -482,11 +469,13 @@ struct tile_draw_data Terrain::create_tile_advice(coord::tile position) {
 		struct neighbor_tile neigh_data[8];
 
 		// get all neighbor tiles around position, reset the influence directions.
-		this->get_neighbors(position, neigh_data, this->influences_buf);
+		this->get_neighbors(position, neigh_data, this->influences_buf.get());
```
fr3akout: get_neighbors should return a std::vector<struct neighbor_tile> instead of writing the result to a preallocated array.

cpp/terrain/terrain.h
```diff
@@ -364,13 +372,13 @@ class Terrain {
...
+	std::vector<Texture *> textures;
+	std::vector<Texture *> blending_masks;
 
-	int *terrain_id_priority_map;
-	int *terrain_id_blendmode_map;
+	std::unique_ptr<int[]> terrain_id_priority_map;
+	std::unique_ptr<int[]> terrain_id_blendmode_map;
 
-	struct influence *influences_buf;
+	std::unique_ptr<influence[]> influences_buf;
```
fr3akout: Replace all unique_ptrs to arrays with std::vectors.

cpp/terrain/terrain_outline.cpp
```diff
@@ -25,12 +28,10 @@ Texture *square_outline(coord::tile_delta foundation_size) {
 		}
 	}
 
-	Texture *ret = new Texture(width, height, image_data);
-	delete[] image_data;
-	return ret;
+	return std::make_shared<Texture>(width, height, image_data.get());
```
fr3akout: Instead of passing a raw pointer to the image data, just add a Texture constructor that expects a unique_ptr to the image data and move image_data into the constructor. This avoids copying the complete image data within the constructor. The Texture's internal image data should be managed as unique_ptr anyway.

cpp/terrain/terrain_outline.cpp
```diff
@@ -54,9 +56,7 @@ Texture *radial_outline(float radius) {
...
 }
```
fr3akout: See above.

fr3akout:
Except from the above comments, the code looks ok.

TheJJ:
i'll fix the terrain memory stuff in a seperate pullreq, as the whole renderer and terrain will be overhauled..

PR #191, opened 2015-01-10
This is a simple to use and efficient rng.

schets:
It provides an easy and reliable way of reproducible random numbers,
along with platform-independent serialization. There is also a thread-safe global rng, although it makes no promises about the order in which requests will be processed.

### DIFF
--- a/cpp/rng/CMakeLists.txt
+++ b/cpp/rng/CMakeLists.txt
@@ -1,0 +1,8 @@
+add_sources(${PROJECT_NAME}
+	rng.cpp
+	global_rng.cpp
+	rng_tests.cpp
+)
+
+#[[Some basic distribution and uniformity tests are applied to the rng as a basic sanity check. Of course, this isn't a full rng test but it ensures that nothing terrible is happening with the implementation. It is feasible that a working rng stream wouldn't actually satisfy the uniformity requirements, since a truly random stream may never return a certain result, or at least return it very little. This is incredibly unlikely]]
+add_test_cpp(openage::rng::tests::rng_tests "test functionality of rng - this *may* fail due to a statistical anomaly, but is extremely unlikely")
--- a/cpp/rng/rng.cpp
+++ b/cpp/rng/rng.cpp
@@ -1,0 +1,203 @@
+// Copyright 2015-2015 the openage authors. See copying.md for legal info.
+#include "rng.h"
+#include "../util/error.h"
+#include "../log.h"
+
+#include <chrono>
+#include <random>
+#include <sstream>
+#include <string.h>
+
+namespace openage {
+namespace rng {
+
+rng::rng(uint64_t v1) {
+	this->seed(v1);
+}
+
+rng::rng(const void *data, size_t len) {
+	this->seed(data, len);
+}
+
+rng::rng(const std::string &instr) {
+	this->from_string(instr);
+}
+
+rng::rng(std::istream &instream) {
+	this->from_stream(instream);
+}
+
+void rng::seed(uint64_t val) {
+	this->seed(&val, sizeof(val));
+}
+
+//murmurhash3 bit avalancher - output is uniformly distributed
+//and changing one bit of input results in ~1/2 of output bits changed
+//This function is a good way of mixing up an input,
+//or mixing together two results with hash64(val1 ^ val2)
+static uint64_t hash64(uint64_t val) {
+	val ^= val >> 33;
+	val *= 0xff51afd7ed558ccd;
+	val ^= val >> 33;
+	val *= 0xc4ceb9fe1a85ec53;
+	return val ^ (val >> 33);
+}
+
+//translates a char* into a uint64_t
+static inline uint64_t from_cptr(const char* data) {
+	return *reinterpret_cast<const uint64_t *>(data);
+}
+
+//!hashes an initial seed into the base state
+static void init_hash(uint64_t *state, uint64_t seed) {
+	//Mix the initial seed into data for each state value
+	state[0] = hash64(seed);
+	state[1] = hash64(seed ^ state[0]);
+}
+
+//For each 8-byte block after the initial seed,
+//mix it with state[0] and set state[0] equal to the result.
+//Then, set state[1] to the mixture of state[1] and state[0]
+//This lets each bit from the input play a role in the output
+//w/o requiring that the size be a multiple of 8 or having extra
+//code for the non multiple-of-8 sections
+static void hash_bytes(uint64_t *state, const char *data, size_t count) {
+	for(size_t i = 1; i <= count; i++) {
+		state[0] = hash64(state[0] ^ from_cptr(data+i));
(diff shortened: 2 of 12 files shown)


fr3akout:
Is there a specific reason to implement a new rng instead of using one from <random>. Further the Travis build fails.

schets:
Building in provides stronger guarantees of reproducibility, as well as guarantees that serialization is the same.

According to the standard, the stl rng implementations themselves should all have identical output. GCC and Clang both follow that atm, but I don't know about ICC or VC++. I wasn't able to find any information governing the serialization of generators. The standard ones all support stream output, but I don't know how well that translates across implementation.

Also, pretty much everything that's implemented in the class besides random, discard, and seed would need to be implemented anyways. The distributions that are provided by the stl to go from random uint64 -> something useful are not standardized and do actually differ across implementations, and I couldn't find anything about serialization portability.

cpp/rng/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+add_sources(${PROJECT_NAME}
+	rng.cpp
+	global_rng.cpp
+	rng_tests.cpp
+)
+
+add_test_cpp(openage::rng::tests::rng_tests "test functionality of rng - this *may* fail due to a statistical anomaly, but is extremely unlikely")
```
TheJJ: lolwhat :laughing: 
could you add some comment what could fail there?

cpp/rng/rng.cpp
```diff
@@ -0,0 +1,171 @@
...
+	val *= 0xff51afd7ed558ccd;
+	val ^= val >> 33;
+	val *= 0xc4ceb9fe1a85ec53;
+	return val ^ (val >> 33);
+}
+
+static void hash_seed_from_blob(const void *dat, uint64_t *state, int64_t bytes) {
+	const char *data = static_cast<const char *>(dat);
+	uint64_t uint64 = 0;
+	memcpy(&uint64, data, bytes <= 8 ? bytes : 8);
```
TheJJ: maybe this is better readable/understandable with
```cpp
uint64_t state_init = 0;
size_t read_count;
if (bytes <= 8) {
    read_count = bytes;
} else {
    read_count = sizeof(state_init);
}
memcpy...

File: cpp/rng/rng.cpp [hunk 3]
```diff
@@ -0,0 +1,171 @@
...
+	state[0] = hash64(uint64);
+	state[1] = hash64(uint64 ^ state[0]);
+	for(int64_t i = 1; i <= bytes - 8; i++) {
+		state[0] = hash64(state[0] ^ *reinterpret_cast<const uint64_t *>(data + i));
+		state[1] = hash64(state[0] ^ state[1]);
+	}
+	
+	if(!(state[0] || state[1])) {
+		state[0] = 0x0123456789abcdef;
```
TheJJ: uh what? please also comment this^^

cpp/rng/rng.cpp
```diff
@@ -0,0 +1,171 @@
...
+	}
+}
+
+void rng::seed(const void *data, uint32_t bytes) {
```
TheJJ: shouldn't `bytes` be named `count` and be `size_t`?

cpp/rng/rng.cpp
```diff
@@ -0,0 +1,171 @@
...
+	if(bytes == 0) {
+		throw util::Error("0 bytes supplied as seed data");
+	}
+	hash_seed_from_blob(data, this->state, bytes);
+	this->discard(discard_on_seed);
+}
+
+//http://en.wikipedia.org/wiki/Xorshift
+//!actually does the number crunching
+static inline uint64_t do_rng(uint64_t &s0, uint64_t &s1) {
```
TheJJ: Nice, inlined. Is it intended that `s0` and `s1` get modified? If not, can `&s0` and `&s1` be `const &`?

cpp/rng/rng.cpp
```diff
@@ -0,0 +1,171 @@
...
+	std::stringstream strstr;
+	return (strstr << *this).str();
+}
+
+void rng::from_string(const std::string &instr) {
+	std::stringstream strstr(instr);
+	strstr >> *this;
+}
+
+using namespace std::chrono;
```
TheJJ: hm, you sure that's a good idea? might be better to use explicitly use the `std::chrono` namespace.

cpp/rng/rng.cpp
```diff
@@ -0,0 +1,171 @@
...
+static uint64_t do_time_seed(bool do_rand) {
+	uint64_t rand1 = 0;
+	if(do_rand) {
+		try {
+			std::random_device rnd;
+			if(rnd.entropy() > 0){
+				rand1 = rnd() + (uint64_t(rnd()) << 32);
+			}
+		}
+		catch (...) {}
```
TheJJ: what is catched here?

cpp/rng/rng_tests.cpp
```diff
@@ -0,0 +1,178 @@
...
+	test.seed(initstr, sizeof(initstr));
+	test2.seed(initstr, sizeof(initstr));
+	ret_if_neq(test, test2, 1)
+
+		char initstr2[] = "abcqq";
+	test.seed(initstr, 3);
+	test2.seed(initstr2, 3);
+	ret_if_neq(test, test2, 2)
+
+		return -1;
```
TheJJ: indentation seems a bit strange here.

cpp/rng/rng_tests.cpp
```diff
@@ -0,0 +1,178 @@
...
+			return 1;
+			test.from_string("");
+			return 2;
+			test.from_string("100");
+			return 3;
+			test.from_string("100 ");
+			return 4;
+		}
+		catch (util::Error) {}
+	return -1;
```
TheJJ: indentation is strange again, maybe `{}` after the macro could clarify this?

TheJJ:
Apart from the above comments, awesome work, code, docs and tests are pretty easy to read and understand. Nicely done.

cpp/rng/rng.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+void rng::fill_real(double *dat, size_t len) {
+	act_fill(dat, this->state, len, [](double v, double *d, size_t i) {
+			d[i] = v / rng::max();
+		});
+}
+
+std::ostream &rng::to_stream(std::ostream &out) const {
+	if((out << this->state[0] << " ").good() &&
+	   !(out << this->state[1]).fail()) {
```
fr3akout: Why are you using !(...).fail() instead of good()?

fr3akout:
The code looks ok. Nevertheless it would be nice to have some comments about the mathematical background.

TheJJ:
<vortigaunt>
agreeeed.
the sams must explain the algorithm in comments.
also the sams must rebase to resolve merge conflict.
</vortigaunt>

TheJJ:
i'm continuing the work in #221, thanks again.

PR #194, opened 2015-01-22
codecompliance: fixed error messages for legal header copyright years

zuntrax:
The compliance checker printed only the first digit of the expected year when one of the copyright years in the legal header was wrong. This should now work correctly.
Also the "from git history" part still assumed that the topmost git log entry is the most recent one (this is no longer true when someone swaps commits during a rebase, see Jonas' previous commit on that file). I deleted that part since I didn't come up with an easy fix immediately (and I don't think it is useful, just use git log if there are problems).


fr3akout:
Looks good :+1:

TheJJ:
bugfix for f47370c695b8f8f237a4b9562c6621388e8252e9, lookin good :+1:

PR #195, opened 2015-01-23
Codestyle updates and simplifications

TheJJ:
The codestyle info files needed some updates, please complain about obscurities.

### DIFF
--- a/doc/code_style/mom.cpp
+++ b/doc/code_style/mom.cpp
@@ -4,10 +4,9 @@
 #include <cstdio>
 #include <SDL/SDL.h>
 
-#include "../whatever.h"
-#include "lolinclude.h"
-#include "roflinclude.h"
-#include "wtfinclude.h"
+#include "../valve.h"
+#include "crossplattform/opengl.h"
+#include "half_life.h"
 
 // [SFT]tech coding guidelines
 // ---------------------------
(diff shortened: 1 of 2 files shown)


LordAro:
Still disagree (reasonably strongly) with your insistence on using "not", "and" & "or"
This isn't python...

doc/code_style/mom.cpp
```diff
@@ -17,155 +16,192 @@
...
 		// spaces here:
 		//        v  v
-		this->mass *= 2;
+		this->mass += 1;
 	}
 
 	// spaces again:
 	// v      v v  v v v     v    v
-	for (int i = 0; i < 1337; i++) {
-		this->mass += 1;
```
janisozaur: doesn't have line above updated

doc/code_style/mom.cpp
```diff
@@ -17,155 +16,192 @@
...
+		}
 	}
 
-	switch (this->mass) {
-	// the case label is at the same indentation level as its switch statement:
+	switch (this->stupidity) {
+	// the case label is at the same indentation level
+	// as its switch statement:
+	// vvv
 	case 0:
```
janisozaur: can cases declare new variables? if so, how would blocks here look like?

janisozaur:
how about adding a rule file for http://clang.llvm.org/docs/ClangFormat.html or some such?

mic-e:
I agree with @LordAro; we should keep down the WTFs/minute. Any other opinions on the `not`/`and`/`or` question?

janisozaur:
If my vote means anything, then +1 for regular (&&, ||, !) operands.

TheJJ:
I really like the `not` `and` `or` keywords and I don't think they'll decrease the readability.
`clang-format` sounds nice, we should integrate that.

shinyquagsire23:
I personally find && || and ! to be more readable, to be honest I didn't even know that C++ *had* a "not" "or" or "and" before this.

doc/code_style/mom.cpp
TheJJ: `case 8:` should provide an example now.

mic-e:
@TheJJ: Yes, `not` etc are more readable, especially when you're using Python as well as C++, but nobody in the C++ community seems to know about them, and following common practice is often the good thing. The keyword operators are bound to increase the [WTF/m](http://images.cryhavok.org/d/2471-1/WTFs+per+Minute.jpg).

LordAro:
The thing with clang-format is that it forces you to do what it wants you to do regardless of the situation. For sure, add a make rule, but I wouldn't recommend making it part of the build system/travis. Keep it as a general guideline instead

mic-e:
I agree with @LordAro. An auto-formatter _might_ _just_ work for Python, but C++ is far too complex, and far too quickly evolving at that. It has multiple levels of turing-completeness and there's just no way a utility will format it right.

fr3akout:
Although the discussion about `and`, `or` and `not` has not been finished yet, the remaining codestyle updates are ok, so ready for merge. The discussion can be continued later on anyway.

PR #200, opened 2015-01-29
Add a move physics camera in Engine class.

lndl:
This is a simple refactor which points to standarize the way
to access to engine camera. For instance, scrolling with keyboard
arrows and mouse middle button was modificated to call this single
method instead of have duplicated code.


TheJJ:
Yup, looking good and still works. merge: :+1:

fr3akout:
:+1:

PR #208, opened 2015-02-04
define _USE_MATH_CONSTANTS for cmath

franciscod:
https://msdn.microsoft.com/en-us/library/4hwaceh6.aspx

Math Constants are not defined in Standard C/C++. To use them, you must
first define _USE_MATH_DEFINES and then include cmath or math.h.

### DIFF
--- a/cpp/pathfinding/a_star.cpp
+++ b/cpp/pathfinding/a_star.cpp
@@ -1,4 +1,4 @@
-// Copyright 2014-2014 the openage authors. See copying.md for legal info.
+// Copyright 2014-2015 the openage authors. See copying.md for legal info.
 
 /** @file
  *
@@ -10,9 +10,10 @@
  * Cybernetics, IEEE Transactions on 4, no. 2 (1968): 100-107.
  */
 
-#include "a_star.h"
-
+#define _USE_MATH_DEFINES
 #include <cmath>
+
+#include "a_star.h"
 
 #include "../datastructure/pairing_heap.h"
 #include "../log.h"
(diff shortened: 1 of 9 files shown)


franciscod:
(is this the preferred way of setting a macro?)

franciscod:
fak u lawyer travis

TheJJ:
Yeah, seems to be alright. It's the same crap in OpenGL if one doesn't use glew, the prototypes need to be activated by macro as well... 

It seems that we only use one constant in `cpp/unit/unit.cpp`, right? Probably we should introduce a math constant header in `cpp/crossplattform/` to avoid the fakking `define _USE_MATH_CONSTANTS`?

This header could assign `constexpr double PI = M_PI;` or similar.

franciscod:
http://stackoverflow.com/a/9017635
>CMake has the add_definitions command for this purpose. An example:
>```add_definitions(-DOPENCV_VERSION=${OpenCV_VERSION})
add_definitions(-DWITH_OPENCV2)```

Maybe we should do that?

TheJJ:
uh, i don't think the buildsystem should be involved in any of this, this is pure c++ and preprocessor stuff.

franciscod:
yep, true

TheJJ:
I implemented it with a crossplattform header here: https://github.com/TheJJ/openage/tree/math_constants

This is the sane solution in my opinion, you can push my branch to this pullrequest or let me submit a separate request.

franciscod:
:+1: much better! make a separate version! thanks :D

PR #209, opened 2015-02-04
crossplatform: use _WIN32 instead of WIN32

franciscod:
While Clang/LLVM, GCC, and Portland Group compilers define a lot of WIN32 and WIN64 macros with various numbers of underscores, the only macros that matter are those that are compatible with Microsoft's Visual Studio: _WIN32 and _WIN64.

http://nadeausoftware.com/articles/2012/01/c_c_tip_how_use_compiler_predefined_macros_detect_operating_system#WindowsCygwinnonPOSIXandMinGW


franciscod:
fak u attorney ci

TheJJ:
kay, :+1:

fr3akout:
:+1:

PR #210, opened 2015-02-04
use unsigned int instead of nonstandard uint

franciscod:
https://stackoverflow.com/questions/5678049/difference-between-uint-and-unsigned-int


franciscod:
http://thetravislawfirm.net/

TheJJ:
lul, looking good :+1:

fr3akout:
:+1:

PR #202, opened 2015-01-30
improved shader for team colors

ptolstoi:

Issue #204, opened 2015-01-30
Media file versioning


TheJJ:
When changes to the media files are done, the engine does not realize the media files need to be reconverted and strange errors may occur. E.g. #202 changes the alpha values of the sprites and thus requires a reconversion. We can't enforce it currently, this is why some versioning needs to be introduced.

Proposal: introduce some `media_version` variable which hast to be incremented each time a reconversion needs to be done. This version is stored in the `assets/` folder and checked by the engine.

Other suggestions?

* [X] asset versioning (`openage/convert/changelog.py`)
* [X] gamespec format hashing

franciscod:
If the reconversion only affects, say, sounds, it should be nice NOT to reconvert ALL THE STUFF. We might want to fragment the version based on media type.

TheJJ:
Indeed, so this is closely related to #79 and #188.

franciscod:
Maybe we should have something like this:

```python

class MediaVersion():
    graphics = 1
    terrain = 1
    sounds = 1
    gamedata = 1
    interface = 1
```

This might be improved when nyan arrives

franciscod:
this numeric approach has a caveat:

- let's say at master we have graphics version 5.
- some branch A modifies something about graphics, sets their asset version to 6
- another branch B modifies another graphic thing, also sets version 6
- A gets merged
- when merging B, we should really check that the graphics gets bumped to 7 at least

TheJJ:
We should hash the version whereever possible, this is persistent across merges.
For non-hashable versions like graphics, the numeric approach is, as you proved, bad.

The problem we have to solve is that some changes to the algorithms used don't require reconversion, but others do. How do we separate between them automatically (probably we'll hit the halting problem..)?

Anyone any other ideas?

iNaKoll:
This is how I would do it :

Let say we have a file containing hash (sha1?) of all the converted media files. This hash file is under source control and is merged automagically. 

1. Let say this file is up to date on all branches.
2. The convert script could read this file compute the hash and only reconvert necessary files.
3. If the hash is still bad after the reconversion, everything needs to be re converted and the hash file should be updated and committed. 
4. Some trusted people should maintain this hash file up to date each time it is really necessary on master. 
5. Others don't care about media files and let the build system figure it out.
6. The "version" of the media could simply be the hash of this special file.

TheJJ:
[...]

PR #214, opened 2015-02-07
improved shader for team colored textures

ptolstoi:
resubmission: #202

### DIFF
--- a/assets/shaders/teamcolors.frag.glsl
+++ b/assets/shaders/teamcolors.frag.glsl
@@ -1,79 +1,39 @@
 #version 120
 
-//team color replacement shader
+// team color replacement shader
 //
-//looks for an alpha value specified by 'alpha_marker'
-//and then replaces this pixel with the desired color,
-//tinted for player_number, and using the given color as base.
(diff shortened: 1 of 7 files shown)


TheJJ:
I think this is ready for merge, but we have to keep #204 in mind. Otherwise many users will totally get confused why the game looks wrong for them after this is merged.

franciscod:
This seems to break the colors a bit, have a look:

openage_2015-02-13_03-25-42_00

openage_2015-02-13_03-29-43_00

Can anyone repro?

TheJJ:
that's why we need #204 to trigger a media reconversion :wink:

franciscod:
after `make media`:
openage_2015-02-13_14-06-46_00

seems nice! I like the alpha on the "ghost" player when building is not yet placed. :+1: :ship: :it:

mic-e:
What's the status of this?

@TheJJ: Is this still valid with #287 in progress?

TheJJ:
it'll be integrated in #287 but also requires #265 / #204 to be completed.

PR #218, opened 2015-02-07
crossplattform: enable math constants on all plattforms

TheJJ:
This introduces namespaced math constants available on all plattforms.

closes #208.

### DIFF
--- a/cpp/crossplatform/math_constants.h
+++ b/cpp/crossplatform/math_constants.h
@@ -1,0 +1,30 @@
+// Copyright 2015-2015 the openage authors. See copying.md for legal info.
+
+#ifndef OPENAGE_CROSSPLATFORM_MATH_CONSTANTS_H_
+#define OPENAGE_CROSSPLATFORM_MATH_CONSTANTS_H_
+
+// instruct the windows math header to define the constants.
+#define _USE_MATH_DEFINES
+#include <cmath>
+
+namespace openage {
+namespace math {
+
+constexpr double E            = M_E;        //!< e
+constexpr double LOG2E        = M_LOG2E;    //!< log_2 e
+constexpr double LOG10E       = M_LOG10E;   //!< log_10 e
+constexpr double LN2          = M_LN2;      //!< log_e 2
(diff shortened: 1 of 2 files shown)


franciscod:
awesome! :+1:

franciscod:
see the previous comment. I think it breaks because cmath gets imported earlier in the build process without the `_USE_MATH_DEFINES` macro defined.

franciscod:
however, on linux it builds fine, lol

franciscod:
i'm poor man's travis

TheJJ:
Hmm, but then i see no other solution than defining the constants manually to the actual numbers.

We can't manipulate the inclusion order and devs will want to include `<cmath>` without thinking about breaking the constant definitions..

franciscod:
These guys were defining that macro on CMakeLists, but since then migrated too boost.

https://github.com/ddemidov/vexcl/commit/696d74a44d406e979ebbe37dc4a1bd924e0d8358
---

>The math library normally defines M_PI to a double approximation of pi. If strict ISO and/or POSIX >compliance are requested this constant is not defined, but you can easily define it yourself:
>
>`#define M_PI 3.14159265358979323846264338327`
>You can also compute the value of pi with the expression acos (-1.0).

http://www.sbin.org/doc/glibc/libc_19.html

---

My vote is also hardcode the constants without relying on the cmath behaviour.

franciscod:
https://github.com/TheJJ/openage/pull/1

[!yo](http://imgur.com/JCoTMVx)

franciscod:
typo on crossplat**t**form!

TheJJ:
oh lol :facepunch:

i think it should be with one `t`, as far as http://www.dict.cc/?s=platform is telling me

franciscod:
builds fine on msys2 now, :+1:

franciscod:
half typos remaining!
```
crossplatform: namespaced math constants on all plattforms
        ^                                          ^
        |                                          |-NOT OK
        |-OK                                          
```

fr3akout:
:+1:

PR #212, opened 2015-02-05
buildsystem: fix opusfile, make it crossplatform

franciscod:
Look at this, it's beautiful! Copied the hard work of Unvanquished:
https://github.com/Unvanquished/Unvanquished/blob/master/cmake/FindOpus.cmake

works perfectly on GNU/Linux and MSYS2!!!

Issue #164, opened 2014-11-19
opusfile.h missing at fedora 20


Pouja:
With the given dependecies for fedora 20 it will not install the opus header file on the system. This results in the following error when ./configure is run:

```
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:108 (message):
  Could NOT find Opusfile (missing: OPUSFILE_LIBRARY OPUSFILE_INCLUDE_DIR)
Call Stack (most recent call first):
  /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:315 (_FPHSA_FAILURE_MESSAGE)
  buildsystem/modules/FindOpusfile.cmake:27 (find_package_handle_standard_args)
  cpp/CMakeLists.txt:57 (find_package)
```

This can be solved by `yum install opusfile-devel`.

PR #219, opened 2015-02-07
find opusfile via cmake, improved edition™

TheJJ:
improves portability of the opus location, resubmit of #212.

Probably still works for fedora (#164), somebody could test it if we're in doubt.


fr3akout:
:+1:

PR #221, opened 2015-02-07
rng: implemented a random number generator

TheJJ:
It provides an easy and reliable way of reproducible random numbers,
along with platform-independent serialization.

Sanitized and updated version of #191. Thanks @schets !

### DIFF
--- a/cpp/rng/global_rng.h
+++ b/cpp/rng/global_rng.h
@@ -1,0 +1,33 @@
+// Copyright 2015-2015 the openage authors. See copying.md for legal info.
+#ifndef OPENAGE_RNG_GLOBAL_RNG_H_
+#define OPENAGE_RNG_GLOBAL_RNG_H_
+
+/** @file
+ * This file contains functions for the global random number generator.
+ *
+ * Please avoid using it, instead, create rng objects for your current
+ * class as members.
+ */
+
+
+namespace openage {
+namespace rng {
+
+/**  Returns the value of the global seed. */
+uint64_t global_seed();
+
+/** Returns a random number from the global rng. */
+uint64_t random();
+
+/** Returns a random number n [lower, upper) from the global rng. */
+uint64_t random_range(uint64_t lower, uint64_t upper);
+
+/** Returns a double in [0, 1) from the global rng. */
+double real();
+
+/** Returns a double in [lower, upper) from the global rng. */
+double real_range(double lower, double upper);
+
+} // namespace rng
+} // namespace openage
+#endif
--- a/cpp/rng/rng.cpp
+++ b/cpp/rng/rng.cpp
@@ -1,0 +1,229 @@
+// Copyright 2015-2015 the openage authors. See copying.md for legal info.
+
+#include "rng.h"
+#include "../crossplatform/timing.h"
+#include "../util/error.h"
+
+#include <random>
+#include <sstream>
+#include <string.h>
+
+namespace openage {
+
+/**
+ * Contains an easy to use and fast random number generator.
+ * It is not safe for cryptography, but good for all kinds of randomness
+ * e.g. for terrain generation, ai decisions, ...
+ */
+namespace rng {
+
+RNG::RNG(uint64_t v1) {
+	this->seed(v1);
+}
+
+RNG::RNG(const void *data, size_t len) {
+	this->seed(data, len);
+}
+
+RNG::RNG(const std::string &instr) {
+	this->from_string(instr);
+}
+
+RNG::RNG(std::istream &instream) {
+	this->from_stream(instream);
+}
+
+void RNG::seed(uint64_t val) {
+	this->seed(&val, sizeof(val));
+}
+
+/*
+ * murmurhash3 bit avalancher - output is uniformly distributed
+ * and changing one bit of input results in ~1/2 of output bits changed
+ * This function is a good way of mixing up an input,
+ * or mixing together two results with hash64(val1 ^ val2)
+ *
+ * https://en.wikipedia.org/wiki/MurmurHash
+ */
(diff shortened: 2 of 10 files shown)


cpp/rng/global_rng.h
```diff
@@ -0,0 +1,37 @@
...
+#define OPENAGE_RNG_GLOBAL_RNG_H_
+
+#include <cstdint>
+
+namespace openage {
+namespace rng {
+
+/**
+ * Returns the value of the global seed.
+ */
```
fr3akout: It would be nice to have single line comments if possible.
/** Returns ... */

franciscod:
I've read "don't roll your own crypto" a lot of times. Isn't there any tested, used, "standard" rng for our use case?

anyway the name `murmurhash3` rocks

TheJJ:
The advantage of this is that it's pretty easyly understandable, and supports very simple serialization. I think it's ready for merge, so :+1: etc.

franciscod:
:+1:, the rng can be improved later if any crypto issues appear

fr3akout:
I would like if somebody did a bit more research about the standard library of C++11, which already provides a pretty big set of random-utilities. If the standard library is sufficient for the use cases of openage, there is no need for this rng. Nevertheless this rng is not that critical, as it is not used for any crypto stuff at all.

fr3akout:
I think this answers the serialization question: http://stackoverflow.com/questions/18361050/saving-random-number-generator-state-in-c11

cpp/rng/rng.h
```diff
@@ -0,0 +1,200 @@
...
+#include <iostream>
+#include <string>
+
+namespace openage {
+namespace rng {
+
+/** \class rng
+ * This class holds an xorshift128+ rng, as well as
+ * utility generator functions
+ */
```
mic-e: why is this class name lower-case?

cpp/rng/rng.cpp
```diff
@@ -0,0 +1,224 @@
...
+		if (rnd.entropy() > 0){
+			rand1 = rnd() + (uint64_t(rnd()) << 32);
+		}
+	}
+	catch (...) {
+		throw util::Error(MSG(err) << "failed using std::random_device!");
+	}
+	return rand1;
+}
+
```
mic-e: How about seeding from `openage::timing::get_monotonic_time()`, `openage::timing::get_real_time()`, or, for added unpredictability, both?

cpp/rng/rng.cpp
```diff
@@ -0,0 +1,224 @@
...
+ * murmurhash3 bit avalancher - output is uniformly distributed
+ * and changing one bit of input results in ~1/2 of output bits changed
+ * This function is a good way of mixing up an input,
+ * or mixing together two results with hash64(val1 ^ val2)
+ */
+static uint64_t hash64(uint64_t val) {
+	val ^= val >> 33;
+	val *= 0xff51afd7ed558ccd;
+	val ^= val >> 33;
+	val *= 0xc4ceb9fe1a85ec53;
```
mic-e: yay, magic numbers!
(maybe you could link the wikipedia article that explains the choice)

cpp/rng/rng.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+static uint64_t try_random_device() {
+	uint64_t rand1 = 0;
+	// std::random_device constructor may throw if it isn't available
+	try {
+		std::random_device rnd;
+		if (rnd.entropy() > 0){
+			rand1 = rnd() + (uint64_t(rnd()) << 32);
+		}
+	}
```
mic-e: I don't approve of your pokémon exception handling.

Please catch as `std::exception &e` and include `e.what()` in the MSG.

cpp/rng/rng_tests.cpp
```diff
@@ -0,0 +1,291 @@
...
+	int (*test_fnc)();
+	const char *name;
+};
+
+static void perform_tests(std::vector<data_test> tests) {
+	int ret;
+	for (data_test test : tests) {
+		if ((ret = test.test_fnc()) != -1) {
+			log::log(MSG(err) << "Testing " << test.name << " failed at stage " << ret);
+			throw "failed pairing heap tests";
```
mic-e: sure, pairing heap tests...

cpp/rng/rng_tests.cpp
```diff
@@ -0,0 +1,291 @@
...
+	}
+	return 5;
+}
+
+static int rng_real_fill() {
+	constexpr size_t n = 1 << 7;
+	rng test{time_seed()};
+	rng test2{test};
+	double data[n];
+
```
mic-e: I think `sizeof(data)` might be safer/have less redundancy.

mic-e:
The code looks pretty nice and clean; in fact, I think it looks even cleaner due to the fact that the `rng` class has an all-lower-case name. Unfortunately, we agreed on upper-case letters in class names.

Since people often seem to associate RNGs with crypto, we should add a comment that states that this RNG was neither designed, nor tested, for use in crypto applications, and only morons would roll their own crypto.

cpp/rng/rng.cpp
TheJJ: this is done in `time_seed`

cpp/rng/rng.cpp
TheJJ: if you want your comments auto-outdate, you need to make it after the change you want.
The comment is associated with the 15 lines above it i think.

cpp/rng/rng.cpp
mic-e: unfortunately the wikipedia article doesn't mention those magic numbers, it uses other ones...

PR #224, opened 2015-02-09
etc: removed the sdl memleak example

TheJJ:
we can safely remove this, it is a valgrind detected memory leak in SDL which probably is a false positive.


franciscod:
:+1: :ship: :it:

PR #225, opened 2015-02-10
fix judgement of child errno

ShionRyuu:
The call function of module subprocess will not increase var total, so the judgement of "if (total > 0)" will always fail.


franciscod:
explain pls?

TheJJ:
good find, seems to be alright. :+1:

fr3akout:
:+1:

PR #228, opened 2015-02-11
delta graphics offset fixed

TheJJ:
fixes wrong start offset for delta graphic definitions.

### DIFF
--- a/py/openage/convert/gamedata/graphic.py
+++ b/py/openage/convert/gamedata/graphic.py
@@ -52,13 +52,13 @@ class Graphic(Exportable):
     struct_description = "metadata for ingame graphics files."
 
     data_format = (
-        (READ_EXPORT, "name0", "char[21]"),
+        (READ_EXPORT, "name0", "char[21]"),             # internal name: e.g. ARRG2NNE = archery range feudal Age north european
         (READ_EXPORT, "name1", "char[13]"),
-        (READ_EXPORT, "slp_id", "int32_t"),
+        (READ_EXPORT, "slp_id", "int32_t"),             # id of the graphics file in the drs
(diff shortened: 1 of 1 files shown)


franciscod:
I didn't notice the moved line. Nice job, how did you spot it? :+1:

TheJJ:
@Jon0 uses the deltas in his code and noticed the regression

iNaKoll:
I'm sorry to ask this question here but what is a "delta"? I understand that the games assets are packed in a proprietary binary file format. I just never heard the word "delta" in this context. I've first seen this word in the nyan doc ("runtime delta") so my next questions are : 

- What is a "runtime delta" ?

- Does Nyan would be used in this particular case 
here?

- What are the contexts and use cases of Nyan?


Thanks

TheJJ:
@iNaKoll Unfortunately, this has nothing to do with nyan. These are linked graphics for some other graphic, e.g. shadows for some building are stored as a delta for the normal graphic.

The nyan runtime delta are small "mods", which are applied at runtime.
This means you can imagine having some files describing some unit costs etc, and as nyan supports modding, some files will have updates for the stock files (i.e. `villager hp += 10`).
As these changes can be accumulated at load-time, this will just create the villager with `+= 10 hp` in the first place, the engine will never know the villager ever had less.

Runtime deltas are actually stored and applied after the initialization, they are needed for tech upgrades that modify values after the unit has had some other values.


And for the usage of nyan: it is a storage format aimed for human readability and moddabilty. You can simply define new units or make changes to existing ones in the form of patches.
This allows everybody having the converted original game files and metadata, but the changes can be shared as simple patches for introducing a tentacle monster or changing unit properties.

PR #232, opened 2015-02-12
Foreach for containers and other phys3 hash relocation

TheJJ:
* foreach added
* reloacte phys3 hashing to correct file.

### DIFF
--- a/cpp/coord/phys3.h
+++ b/cpp/coord/phys3.h
@@ -1,9 +1,12 @@
-// Copyright 2013-2014 the openage authors. See copying.md for legal info.
+// Copyright 2013-2015 the openage authors. See copying.md for legal info.
 
 #ifndef OPENAGE_COORD_PHYS3_H_
 #define OPENAGE_COORD_PHYS3_H_
 
+#include <functional>
+
 #include "decl.h"
+#include "../util/misc.h"
 
 #define MEMBERS ne, se, up
 #define SCALAR_TYPE phys_t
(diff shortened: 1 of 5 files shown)


cpp/util/algorithm.h
```diff
@@ -0,0 +1,38 @@
...
+
+// TODO: make general transformation from iterator algorithm to container
+// algorithm. so much metaprogramming...
+
+namespace openage {
+namespace util {
+
+/**
+ * std::for_each except just on containers.
+ */
```
fr3akout: please use one line comments again

cpp/util/compiler.h
```diff
@@ -12,4 +12,6 @@
 #define likely(x)    __builtin_expect(!!(x), 1)
 #define unlikely(x)  __builtin_expect(!!(x), 0)
 
+#define no_inline __attribute__((noinline))
```
fr3akout: is there a place where we use no_inline with a proper reason?

cpp/coord/phys3.h
```diff
@@ -1,9 +1,12 @@
-// Copyright 2013-2014 the openage authors. See copying.md for legal info.
+// Copyright 2013-2015 the openage authors. See copying.md for legal info.
 
 #ifndef OPENAGE_COORD_PHYS3_H_
 #define OPENAGE_COORD_PHYS3_H_
 
+#include <functional>
```
fr3akout: what is used from functional?

cpp/util/compiler.h
franciscod: +1 to fr3akout's question

cpp/util/algorithm.h
```diff
@@ -0,0 +1,38 @@
...
+}
+
+/**
+ * Filters items from a container which satisfy a certain predicate.
+ */
+template<class Container, class Function>
+inline void remove_from(Container &container, Function &&func) {
+	container.erase(std::remove_if(std::begin(container),
+	                               std::end(container),
+	                               std::forward<Function>(func)));
```
iNaKoll: Be VERY careful here. I don't think this code is doing what you think it does! This will only erase ONE element from the container if the predicate is satisfied at least once. If more than one element satisfies the predicate the remaining elements would be moved at the end of the container and will remain there. If no element satisfies the predicate this code would try to erase std::end(container) which is an undefined behavior!

The correct code would be : 
container.erase(std::remove_if(std::begin(container),
	                               std::end(container),
	                               std::forward<Function>(func))
                          std::end(container)); // Note the additional parameter here!
This would call the correct erase overload for an std::vector but unfortunately this would not work for std::list, std::map and other containers and there is no uniform way (yet) to do this for all containers.
For more details : 
- http://stackoverflow.com/questions/16013545/how-do-i-erase-elements-from-stl-containers
- http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4161.htm

TheJJ: Thanks very much for the hint, I updated the code.

TheJJ:
Do i get my +1 now?^^

StepS-:
:+1:

franciscod:
:+1:, but what about the comment style and @fr3akout's suggestion?

TheJJ:
I tried both styles here and came to the conclusion that the 3 line variant provides more visual separation and was more readable.

franciscod:
yup, plus it's consistent with the multiline comments

fr3akout:
I am not really happy with the 3 line variant, as I don't get the feeling that it makes the code more readable. It just needs more space. But if everyone else likes it: :+1:

PR #207, opened 2015-02-04
Windows support on MSYS2, stubs unimplemented unused stuff when on WIN32

franciscod:
**heavily** based on @mappu guide on #9. This aims to close that issue, and enables openage to be built on MSYS2 with traditional `cmake` by stubbing not yet implemented (and coincidentally, unused) features under WIN32. 

A better solution is to implement the features (I think this is desirable on os.cpp functions), or strip the entire feature from all platforms if it's not useful (I never used the console tests, they seem dated).

Needs #218 so it should be merged first, and this branch should be rebased. If you want to test, just pull TheJJ/math-constants into master, and then this branch into the result, it should build.

In contrast with mappu's guide, this doesn't modify`./configure` nor `buildsystem/simple`. Also, the majority of the patches that mappu required are no longer necessary due to better code quality :) Oh, and we now have opusfile on the MSYS2 repos, so no manual compiling of that dep.

TL;DR: 
```
# convert assets or copy them already converted
mkdir build && cd build
cmake -G "MSYS Makefiles" .. #configures flawlessly
make #compiles flawlessly
cd .. #changes directory flawlessly
./build/cpp/openage #runs flawlessly
```

---

This was on a Windows7 32bit VM. This guide should work without issues, please try it out and comment about your experience!

- Install [msys2](https://msys2.github.io/) on C:\msys2. When the wizard asks, don't open the shell. 

`/` means C:\msys2 and `~` is `/home/<username>`.

- Move AoE2 directory to ~, assuming "~/age2/Data" is the right directory (contains a lot of dat an drs).

From now on, you need to use mingw32_shell.bat from C:\msys2. Open up a shell and go:
- `pacman -Syu`
- copypasta this package installing havoc:
```
pacman -S git mingw-w64-i686-{gcc,pkg-config,make,openssl,libogg,opus{,file,-tools},SDL2{,_image},glew,freetype,ftgl,python3{,-{,numpy,Pillow}}}
```
- copy mingw32-make to make, so standard tools don't whine:
```
cp /mingw32/bin/{mingw32-,}make.exe
```
- get the windows branch from my repo:
```
cd ~
git clone https://github.com/franciscod/openage
cd openage
git checkout windows
```
- cmake magic!:
```
mkdir build && cd build
cmake -G "MSYS Makefiles" ..
make
cd ..
```
- you might need to install dejavu serif from http://sourceforge.net/projects/dejavu/files/dejavu/2.34/dejavu-fonts-ttf-2.34.zip
[...]

PR #234, opened 2015-02-13
Stub some unused, unimplemented stuff on WIN32 so it can compile

franciscod:
Cleaner, fresher form of #207. Won't harm current platforms :)

### DIFF
--- a/libopenage/console/tests.cpp
+++ b/libopenage/console/tests.cpp
@@ -1,4 +1,4 @@
-// Copyright 2014-2015 the openage authors. See copying.md for legal info.
+// Copyright 2014-2016 the openage authors. See copying.md for legal info.
 
 #include <vector>
 #include <string>
@@ -43,6 +43,8 @@ void render() {
 
 
 void interactive() {
+	#ifndef _WIN32
+
 	console::Buf buf{{80, 25}, 1337, 80};
 	struct winsize ws;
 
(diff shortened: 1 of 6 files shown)


TheJJ:
although stubbing stuff is probably the way to go for now, we might add plattform-specific demos and tests so that the incompatible functions don't even show up when under an incompatible plattform.

franciscod:
We could have a cpp/crossplatform/os.h that simply does an include to os_apple.h or os_linux.h or os_windows.h, and the platform-specific .h and .cpp would have #ifdef PLATFORM :)

franciscod:
We could have a cpp/crossplatform/os.h that simply selects os_apple.h or os_linux.h or os_windows.h, using preprocessor ifdefs and platform-dependant macros. With that scheme, the platform-specific .h and .cpp files would be clean of ifdefs :)

TheJJ:
This also needs updating to the new buildsystem updates and project layout.

franciscod:
I'll take a look :)

franciscod:
okay this is now rebased on top of my [msys2 random tweaks pull request™](https://github.com/SFTtech/openage/pull/476)

I'd say we merge this so we have MSYS support ASAP, and then improve the stubs with the real functions.

Keep in mind these stubbed functions are not used in the game yet!

franciscod:
oh i'm rebasing commits from a year ago and travis yells me "2016 is wrong, use 2015!!!"

Issue #28, opened 2014-10-27
create nyan - yet another notation


TheJJ:
_nyan_ is the future data storage format for openage.

It's designed for moddability and human-readability.

Development takes place here:
https://github.com/SFTtech/nyan

Tasks:
- [x] format specification
- [x] implement the [nyan database](https://github.com/SFTtech/nyan)

TheJJ:
Soooo, we may have a specification:

https://github.com/SFTtech/nyan/blob/master/doc/nyan.md

head over there and improve it if things are missing :)

TheJJ:
To "officially" announce it, my master's thesis is the design and implementation of nyan :tada:

franciscod:
niceee!! openage goes to academia!

elnabo:
Quick question, what is a master thesis and how long does it last ?

janisozaur:
@elnabo (master) thesis is a work one has to create to be awarded a degree in a field of science. "Thesis" means the text document, but commonly the term also encompasses all of the derived products: source code, hardware platforms or other things, especially in other fields of studies. It "lasts" forever, as it gets archived by university and sometimes is made available to the public, as per agreement with the school. Perhaps you wanted to ask "how long does it take to create one?": in that case it's usually one or two terms (i.e. half a year or a year), but may, depending on the author, take also less or more time.

See also https://en.wikipedia.org/wiki/Master_of_Science (especially the pat about Germany, where I believe @TheJJ is) and https://en.wikipedia.org/wiki/Thesis

elnabo:
Ok thanks, I was wondering what it was, as I think there is no equivalent in France or at least in my university.

TheJJ:
The core nyan implementation should be done by now, some TODOs and mainly optimizations are left.
Next we have to come up with a step-by-step transition plan to integrate it.

PR #239, opened 2015-02-23
[WIP] nyan specification

iNaKoll:
Before I forget, I am committed to write down what @TheJJ explained to me in #228 about nyan.
So here it is, an introduction to nyan (for newbies). At least this is my understanding.
This is a work in progress and I will see if I am able to do more specifications.
In the meantime I'll accept comments, corrections, suggestions..

### DIFF
--- a/doc/nyan/nyan.md
+++ b/doc/nyan/nyan.md
@@ -62,7 +62,7 @@ of sets and not of bags. Therefore no duplicates are allowed within a set and no
 order is guaranteed. For each attribute a default value can be specified after
 the attribute type separated by an equals sign.
 
-### nyan interfaces
+### nyanspec - nyan interfaces definition
 
 A nyan set can store objects of heterogeneous types thanks to implicit interface.
 A nyan interface is simply a nyan type that define a minimum set of attributes
@@ -95,14 +95,14 @@ implementation of the ABILITY interface :
     	int : construction_speed
     }
 
-### Runtime deltas
+### nyanspec - Runtime deltas declaration
 
 Currently we are able to define nyan types that consist of named attributes and
 dynamic attributes. Thus we are able to define types with static attribute
 values and relations. But this is not always the case in **openage**.
 Ingame technologies should have the ability to update other nyan objects
 attributes during runtime. For example the _loom_ technology should update the
-_villager_'s health points. Those so called _runtime-deltas_ can be specified in
+_villager_'s health points. Those so called _runtime-deltas_ can be declared in
 the **nyanspec** the following way:
 
     TECHNOLOGY {
@@ -124,7 +124,7 @@ In contrast to the **nyanspec**, which describes all existing types, their
 relations and attributes, the nyan data files, describe the concrete nyan
 objects. Nyan data files have the file extension `*.nyan`.
 
-### Basic object definition
+### nyan - Basic object definition
 
 All objects that are defined in nyan data files must be of a type defined in the
 **nyanspec**. Hereafter we will use the following **nyanspec** as a base for all
@@ -132,7 +132,17 @@ nyan data files.
 
     ABILITY {
     	description : string,
-    	...
+    }
+    
+    MOVE_ABILITY {
+    	description : string,
+    	speed : float
+    }
+    
+    FIGHT_ABILITY {
+    	description : string,
+    	damage : int,
+    	ranged : bool
     }
     
     UNITTYPE {
@@ -147,12 +157,12 @@ nyan data files.
 
 Firstly we define a few abilites and a unit type that makes use of them.
 
-    +ABILITY MOVE {
+    +MOVE_ABILITY MOVE {
     	description = "Allows units to move.",
     	speed = 10.0
     }
     
-    +ABILITY FIGHT {
+    +FIGHT_ABILITY FIGHT {
     	description = "Allows units to fight.",
     	damage = 20,
     	ranged = false
@@ -162,7 +172,7 @@ Firstly we define a few abilites and a unit type that makes use of them.
     	health_points = 50,
     	abilites = [ MOVE, FIGHT ]
     }
-
+    
 New objects can be introduced using the `+`-operator combined with the object's
 nyan type. Each object must have a unique name within the whole nyan context
 (which will be explained later). Attribute values are assigned using `=` with a
@@ -177,19 +187,39 @@ built-in types look like the following:
 Sets are defined using brackets. Values within the set are separated by commas.
 
 ### Anonymous object definition
+TODO: Do we really need this? Why?
 
-### Runtime-delta definition
+### nyan - Delta definition
+The syntax used to defined deltas is very simple. A delta is an overlay over an pre-existring objects.
+Suppose that a VILLAGER object is defined (with variaous attributes and values). To define a delta that updates the villager properties, we need to refer to the previously defined VILLAGER object with the following syntax using the @ character :
+@UNIT VILLAGER {
+	health_points *= 1.1
+}
 
-### Delta definition and application
+The previous definition defines a delta over the VILLAGER object that increase the villager health_points by 10%. This delta would be applied at the game initialization.
+
+### nyan - Runtime-delta definition
+A runtime-delta is simply a delta that is defined inside the scope of another object (typically a technology object). The game runtime can then decides when to apply the delta associated with a particular object.
+The following example defines the LOOM technology that upgrades the VILLAGER objects healh_points attribute :
++TECHNOLOGY LOOM {
+	name = "Loom"
+	@UNIT VILLAGER {
+		health_points = 50
+	}
+}
 
 nyanspec compiler
 -----------------
+The nyanspec compiler is used to generate C++ code from a nyanspec file. The generated C++ code consists of structs and parsers associated with the nyan types defined in the nyanspec file. The nyanspec compiler itself is written in Python and resides in py/openage/nyan.
 
-Using the generated code
+Using the generated code with _libnyan_
 ------------------------
+The generated C++ code is used by openage to parse the nyan files and to create an in-memory game data model. libnyan is written in C++, it manages and combines all the informations contained in the game data model. During the game, libnyan is queried by the game engine and can be considered as a _view_ over the game data model.
 
 Questions
 ---------
 
 Why is there no inheritance between nyan types? This would avoid redefining same
 attributes for multiple types.
+
+Why not generate the nyanspec and nyan files directly by the convert script ? Then generate the associated C++ code (structs and parsers) used by libnyan ?
--- a/doc/nyan/nyan.md
+++ b/doc/nyan/nyan.md
@@ -49,6 +49,7 @@ part of a nyanspec describes a simple unit-type in **openage**.
     
 For sake of completeness, here is a *very* basic (and not realistic) usage example
 of this type in nyan data file :
+
     +UNITTYPE VILLAGER {
 	    health_points = 10,
 	    speed = 1,
@@ -168,7 +169,7 @@ the following way:
 		armor_melee = 0
     }
     
-   +TECH BLOODLINES {
+	+TECH BLOODLINES {
 	   description = "through centuries of in-breeding horses and knights, you manage to somehow increase their hp."
 	   cost_food = 300
 	   cost_gold = 200
@@ -255,15 +256,17 @@ TODO: Do we really need this? Why?
 nyan deltas enable the user to modify the game base properties and to patch existing objects.
 The syntax used to defined deltas is very simple. A delta is an overlay over a pre-existring objects.
 Suppose that a VILLAGER object is defined (with variaous attributes and values). To define a delta that updates the villager properties, we need to refer to the previously defined VILLAGER object with the following syntax using the @ character :
-@UNIT VILLAGER {
-	health_points *= 1.1
-}
+
+	@UNIT VILLAGER {
+		health_points *= 1.1
+	}
 
 The previous definition defines a delta over the VILLAGER object that increase the villager health_points by 10%. This delta would be applied at the game initialization.
 
 ### nyan - Runtime-delta definition
 A runtime-delta is simply a delta that is defined inside the scope of another object (typically a technology object). The game runtime can then decides when to apply the delta associated with a particular object.
 The following example defines the LOOM technology that upgrades the VILLAGER objects health_points attribute :
+	
 	+TECH LOOM {
 		description = "Loom"
 		^UNITTYPE VILLAGER {
@@ -273,6 +276,7 @@ The following example defines the LOOM technology that upgrades the VILLAGER obj
 
 A runtime delta can also be part of a nyan delta. For example, The previously defined LOOM technology can be
 patched the following way :
+	
 	@TECH LOOM {
 		^UNITTYPE VILLAGER {
 			health_points += 50    # The loom technologie now gives 100 more health points to the villager
@@ -299,6 +303,7 @@ nyan objects can inherit attribute values *from objects of the same type*. When 
 	}
 
 When a subobject inherits from an object, the subobject is still concerned by deltas and runtime deltas primary applied to the original object. Continuing the previous example :
+	
 	# nyan base properties
 	+TECH BLOODLINES {
 		description = "through centuries of in-breeding horses and knights, you manage to somehow increase their hp."
--- a/doc/nyan/nyan.md
+++ b/doc/nyan/nyan.md
@@ -1,11 +1,32 @@
 nyan data format
 ================
 
-**nyan** is the recursive acronym for **nyan: yet another notation**. It
-consists of two separate, but very similar file formats. Firstly, there is the
-nyan specification (short: nyanspec) which describes a domain of object types.
-Secondly, there is the data file format itself, which defines a set of objects,
-with their attribute values.
+**nyan** is the recursive acronym for **nyan: yet another notation**. *nyan* is
+a human-readable text data format used in openage to represent game assets 
+informations. *nyan* is the data format in which game developers specify game 
+entities characteristics such as units health points, abilities, technologies, etc.
+*nyan* is designed for human readability, simplicity and moddability. It enforce
+a minimum set of grammatical rules to achieve its goal.
+
+By default, the game engine does not know anything about entities properties.
+The nyan reader provides all those characteristics to the game engine. nyan is used
+in different contexts :
+
+- definition of game entities (unit, building) base properties such as health points, 
+ressource cost, abilities, civilization informations, etc...
+- definition of game modes taking the form of patch applied over the game entities 
+base properties. Those patch or "deltas" are applied by the game engine during the 
+game initialization. Modes consists of a bunch of patches and can be exchanged
+without tweaking the game base properties.
+- definition of technologies properties such as base entities properties enhancement
+and ressource cost. Those technologies benefits are applied at runtime (after the game
+initialization) and may be refered as "runtime deltas". Like the game entities base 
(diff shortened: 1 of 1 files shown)


TheJJ:
Very nice, it's actually understandable now.

Another thing that might be useful understanding the upcoming deep integration of nyan:

The nyanspec is relevant at compiletime, because it will specify all static types and structs for running and reading nyan files later.

That means: The nyan spec is the single source of format specification. The specification is used to generate code:
* structs and other data formats are created in C++
  * the engine can access all file contents with type safety
  * the compiler can optimize everything better as there are no string-names as there would be with `yaml` or `json`
* a parser is generated (by `py/openage/nyan`) to create a C++ nyan reader (and later writer) for specified data format
  * basically fills up the structs generated above
  * can read any `.nyan` file at runtime and make calls to `libnyan` (see below)

`libnyan` is written in C++ (at least i think so) and manages all the patching stuff.
* The dynamic data overlaying is not dependent on the data format, but on the generated structs which are simply used then.
* `libnyan` is queried by the engine for providing all required information ("melding start-deltas"), trigger updates ("runtime deltas"), overlays etc.


And we even push this further:
* `py/openage/convert/gamedata/` currently specifies the format of `empires_x1_p1.dat`
  * this format allows reading the binary file and storing it into tables
  * these tables and the original format suck, information is redundant and we can heavily improve it
  * buuut: the file format is given (every original game has it.)
  * :arrow_right: data transformation.
* the original file will be read according to the original format (we can't help it)
* a transformation is specified
  * mapping:   original format -> our awesome format
  * redundancy is purged (e.g. no more double huskarl unit (1x castle, 1x barracs))
* export transformation destionation format as *nyanspec*
* export actual gamedata from the dat file as `.nyan` files
* create C++ code according to this specification
* :arrow_right:
  * engine gets structs (defined in python `py/openage/convert/gamedata/` -> converted to a nyanspec -> the spec is converted to the structs and parsers by python `py/openage/nyan`)
  * locally converted data is stored in many `.nyan` files, read by the generated parser at runtime
  * `libnyan` is used for aggregating and manipulating all data from loaded `.nyan` files


(oh gawd we really have to write these plans down, i guess if i get run over by a bus nobody would ever remember) (whee i just wrote them down)

(if you like you can just copy-paste my explanations, but feel free to optimize the understandability :grin:)


tl;dr: jj just tried to explain the nyan master plan for storing data without any redundancy

iNaKoll:
I updated the documentation and tried to keep it simple. I am not sure I was able to optimize the understandability though

fr3akout:
@iNaKoll Although I appreciate, that you are trying to improve the nyan documentation, it would be nice, if you keep the sections I have wrote or ask before removing them. The interfaces that you describe are not part of nyan and can't replace the dynamic attributes, as with interfaces you are still limited to a fixed predefined set of types at compile time, while dynamic attributes make it possible to add objects of a type with completely new attributes within nyan data files.

iNaKoll:
I've asked on IRC before proposing this here. I'm not trying to impose anything but rather write my understanding of what JJ said :

"the compiler can optimize everything better as there are no string-names as there would be with yaml or json"

The benefits of static types are better performance and better compiler diagnosis. Dynamic attributes reduce the advantages of nyanspec, static types and c++ code generation.

I don't have any clue of what is really needed but I just can see that there is a real development effort here and that nyanspec and dynamic attributes doesn't seem to fit well together. I thought nyan interfaces could be a solution but if I were wrong I have absolutely no problem with putting everything back the way it was. :)

TheJJ:
@iNaKoll no problem, at least our ideas no longer exist just in some of these brains :wink:.

The `...` is actually part of the syntax, it means that this nyan type can have dynamic attributes.
How they are typechecked is unknown to me, but their type is defined in the nyan spec file.

The libnyan will also probably never exist (my misunderstanding) because just defining entry functions that call to generated code is pretty useless. So the whole nyan-framework usable by the game engine is generated from the format specification.

Mods must adhere this format, otherwise they can't load. Static typing ftw!

iNaKoll:
If types containing dynamic attributes are partially know by the game engine, type checking is partially irrelevant for types containing one of those "...".

For example an ability defined as 
```
ABILITY {	
    name : string,
    ...
}
```

would be generated as (something like) :
```
class Ability {
public:
    std::string name();
    std::map <std::string, util::variant<int, bool, double, std::string, util::set>> dynamic_ attributes();
};
```

The "dynamic_attributes()" content just can't be statically checked at all or am I missing something? As long as there is no dynamic attributes everywhere or as long as they are not accessed from performance critical code it should be OK.

What is bothering me is that if for example we define the MOVE ability like this : 
```
+ABILITY MOVE {
     	description = "Allows units to move.",
     	speed = 10.0     # with speed being part of the dynamic attributes of some ABILITY instance
}
```
at some point in the C++ code we really need to know that there is a MOVE ability with a speed attribute so at some point we would have to write something like : 
```
    if(some_ability->is_move_ability()) { // or some other test
        do_something(some_ability->dynamic_attributes()["speed"]); // virtual function call + map access
    }
} catch (...) // oops is_move_ability implementation is buggy or throw or there is no "speed" attribute in some_ability
```
This could be done once at startup or maybe somewhere in the game loop and is really error-prone. The real problem is that the MOVE ability type is not known at compile time. One solution could be to provide the code generator with both a nyanspec file and a nyan file containing some base game properties. Thus the MOVE ability defined above would be generated in C++ code as : 
```
class MOVE_ABILITY { // The MOVE ability has its own implicit type!
public:
    std::string name();
    double speed() noexcept; // Great!
};
```
If everything i wrote here is obvious to you then I've just understood something. Anyway, what do you thing about it? As long as the user/modder does not extend from ABILITY but overrides existing abilities, it should be OK.

TheJJ:
A mod must be able to extend the abilities dynamically.

Imagine: You have a japanese tentacle monster mod, which adds the ability `vomit` to a villager, something that didn't exist before. You can't regenerate and recompile all engine/nyan code, this attributes must be able to load dynamically. As far as I understood our idea, the ability can only modify known attributes (as these are generated struct members/classes/etc in the engine which can be manipulated).

If mods need to go deeper, than the engine actually needs to be extended, the nyanspec updated and we have to merge the change upstream i'd say. But @fr3akout will surely know more about that.

dbrgn:
Can someone explain what differentiates this format from other formats like JSON, YAML, TOML or S-Expressions?

In my eyes, introducing a new format always introduces a greater contributing barrier. New contributors need to learn a new format first. And parsers need to be written and optimized.

Does nyan have a unique property that makes it exceptionally fit for this task? :) (The USP, in marketing speech.)

TheJJ:
Lemme try:
* nyan is typesafe
  * all allowed members, types and assignments are defined in the nyanspec
  * this specification is then used to generate the nyan parser and C structs (used in the engine)
  * this ensures the C++ engine doesn't need to be recompiled when mods change data
* nyan allows easy modding
  * data packs ship configuration data, all units available, etc.
  * modpacks can update the present information easily, by applying config "patches"
  * these "patches" can be applied when other triggers occur, this is the way technologies will work
* nyan is human readable
  * everything is written whitespace-insensitive (except in strings obv.)
* nyan is invented here™
  * we can change the specification to our needs whenever we want

mic-e:
Basically what @TheJJ said. In addition, well, nyan tries to be a general language, but it _was_ designed with openage in mind, so it has facilities to easily represent some of the more complex information, especially how a definition from one datapack updates the definition from an other one, and how a technology will modify the attributes of a unit.

dbrgn:
OK, so you have a special patching syntax? Or inheritance?

mic-e:
@dbrgn: Unfortunately, the doc seems to lack any larger example code, which I'd find far more useful than theoretical discussions about model-view controllers.

IIRC, the syntax would look something like this:

    +UNITTYPE SCOUT {
        name="scout cavalry"
        hp=40
        damage=3
        armor_pierce=1
        armor_melee=0
    }

    +UNITTYPE LIGHTCAVALRY : SCOUT {
        name="light calvalry"
        hp+=20
        damage+=2
        // inherits all other properties from SCOUT
    }

    +TECH BLOODLINES {
       description="through centuries of in-breeding horses and knights, you manage to somehow increase their hp."
       cost_food=300
       cost_gold=200

       ^UNITTYPE SCOUT {
           hp *= 1.2                        // also affects LIGHTCAVALRY
       }
    }

    +TECH LIGHTCAVALRY {
         description="upgrades your scouts to light cavalry."
         cost_gold=400
         upgrade_from = SCOUT
         upgrade_to = LIGHTCAVALRY
    }


A mod that would increase the hp for light cav and increase the effectiveness of bloodlines to 30% would look like this:

    @UNIT LIGHTCAVALRY {
        hp+=5
    }

    @TECH BLOODLINES {
        cost_gold *= 1.5
        ^UNIT SCOUT {
            hp *= 1.3
        }
    }

Regarding the execution order of the change:

All enabled datapacks are read first; after reading base, SCOUT has 40hp and LIGHTCAVALRY has +=20hp.; after reading the mod, SCOUT has 45 hp and LIGHTCAVALRY has +=20hp.

Once all the datapacks have been read, inheritance is resolved: SCOUT has 45hp, LIGHTCAVALRY has 65hp, and BLOODLINES applies to SCOUT and LIGHTCAVALRY. This is the information (unit types) that will be available at game start.

Now when the player researches bloodlines in-game, that player's SCOUT and LIGHTCAVALRY unit types get modified to have 59hp and 85hp.

If the player researches the light cavalry tech, all the player's units that have unit type SCOUT get changed to unit type LIGHTCAVALRY.

dbrgn:
@mic-e thanks, looks very interesting :)

mic-e:
@iNaKoll: Maybe you could include my example as well?

iNaKoll:
@mic-e Done!

I think there is still room for some improvements

doc/nyan/nyan.md
```diff
@@ -25,68 +46,149 @@ part of a nyanspec describes a simple unit-type in **openage**.
...
+when an object is instanciated.
+
+Imagine a villager unit in **openage** has multiple abilities. This would lead to the
+following (not realistic) example:
+
+    # nyanspec
     ABILITY {
-	    name : string
+	    name : string	# Regular attribute specification
+	    ...			# Dynamic attributes are allowed for this type
```
TheJJ: This `...` is driving me nuts, we have to come up with a better alternative. Nobody will understand that, everybody will asume that the example will just go on.

dbrgn: Even Python has it :D

```python
>>> ...
Ellipsis
>>> type(...)
<class 'ellipsis'>
```

franciscod: why don't we allow dynamic attributes everywhere?

mic-e:
The reason why we might or might not need dynamic attributes is because different units have different abilities.

For example, a Militia doesn't have a ranged attack and thus doesn't need a `range` attribute. A building can't walk and thus doesn't need a `speed`because it doesn't have a ranged attack.

franciscod:
```
+UNITTYPE archer {
    health=30
    ABILITY move { speed=5}
    ABILITY ranged_attack { attack=3, range=5 }
}
```
i like this! 

for example, putting aside damage weakness/resistance, a militia isn't too far from a cavalry, or a camel, or an elephant: all of them are melee combat units that can move and do melee attacks

mic-e:
Notwithstanding the above, I'd actually prefer a braceless syntax:

    UNITTYPE archer:
        health=30
        ABILITY move: speed=5
        ABILITY ranged_attack: attack=3, range=5

dbrgn:
I'd strongly support that.

franciscod:
https://github.com/SFTtech/openage/issues/28

here, @fr3akout said that commas make the parser easier to write. i assume braces are along the same lines. but it'd be awesome to have a pythonish syntax! 

...what about translating nyandata into python and then parsing it?

mic-e:
What's a simple parser compared to an awesome-to-use language...

dbrgn:
Yep, I'd create a custom parser.

This shouldn't be too hard. An indentation starts a new scope. De-indentation resets it. A simple state machine should suffice.

Is there already some kind of implementation? Are you going to use a parser generator?

mic-e:
A parser generator is usually overkill; a lexer generator might make sense.
@fr3akout has some code in his `nyan` branch [here](https://github.com/fr3akout/openage/tree/nyan).

iNaKoll:
OK so following the "historical module" approach presented above by @mic_e, I tried to simplify the type system and reinforce the syntax consistency.
Disclaimer:
This proposal is a long shot... With this proposal the barrier between type and objects is kind of dropped. Nevertheless, objects attributes are well-defined and exclusively specified in nyanspec.

    ## nyanspec - proposal #3 - Object Oriented
    # philosophy :
    # - base types are exclusively defined in nyanspec
    # - types are composable in nyanspec AND in nyan (at runtime)
    # - as a consequence each object has its own unique type (a composition of base types)
    # - unknown object properties cannot be introduced in the nyan data file
    # - type inheritance is handled at runtime ('+' sign syntax)
    # - the '+' sign marks the allocation of a new subobject
    # - '@' refer to types/objects for static delta and overlays
    # - '^' refer to types/objects for runtime delta
    # - deltas are contininously applied following the inheritance order
    # - runtime deltas are trigger by the game engine and/or a python script provided by a mod

    MOVE {
        speed : int
    }

    WALK : +MOVE {}     
    SAIL : +MOVE {}

    # the WALK ability inherits from MOVE. Walking units obviously can move.
    # WALK is a subtype of MOVE
    # the SAIL ability also inherits from MOVE but is a different from the WALK

    ATTACK {
        damage : int,
        target : list()      # A list of target objects of any type
    }

    RANGE_ATTACK : +ATTACK {
        range : int
    }

    ALIVE {
        health_points : int
    }

    INFANTRY :
        +WALK,
        +ALIVE,
        +ATTACK
    {}

    ARCHER :
        +WALK,
        +ALIVE,
        +RANGE_ATTACK
    {}

    BUILD_ABILITY {
        speed : int,
        building : list()  # A list of building objects of any type
    }

    VILLAGER : 
        +WALK,
        +ALIVE,
        +ATTACK,
        +RANGE_ATTACK,
        +BUILD_ABILITY
    {}

    VILLAGER_TECH
    {
        ^VILLAGER
    }

    LOOM :
        +VILLAGER_TECH 
    {
        hp_bonus : int
    }

    ## openage nyan datapack

    # Set default properties for all villagers :
    @VILLAGER {
        @WALK : speed = 10,             # inline syntax
        @ATTACK : {                     # or brace syntax?
            damage = 1,
            targets = [ANIMAL, VILLAGER, INFANTRY, ARCHER]
        }
        @RANGE_ATTACK : damage = 2, target = [ANIMAL],
        @BUILD_ABILITY : building = [ ## list all buildings here ## ]
    }

    @LOOM {
        ^VILLAGER @ALIVE : health_points += hp_bonus
    }

    @INFANTRY {
        # Infantry units can attack the following nyan objects
        @ATTACK.targets = [PHYS_COORD, ANIMAL, VILLAGER, INFANTRY, ARCHER]
    }

    @ARCHER {
        @RANGE_ATTACK.targets = [PHYS_COORD, ANIMAL, VILLAGER, INFANTRY, ARCHER]
    }

    VILLAGER_MALE :
        @VILLAGER 
    {
        # VILLAGER_MALE is a static delta/overlay over the VILLAGER object
        # Load male villager graphics here
    }

    +ARCHER LONGBOWMAN {
        # Load LONGBOWMAN archer graphics and attack properties here
    }
    
    ## battlefield villager mod nyan datapack example
    
    BATTLEFIELD_VILLAGER :
        @VILLAGER
    {
        # BATTLEFIELD_VILLAGER is a static delta of VILLAGER
        @WALK : speed *= 1.1,
        # This unit can range-attack ennemy units
        @RANGE_ATTACK : damage = 2, target += [VILLAGER, INFANTRY, ARCHER]
    }

The nyan parser/generator would generate C++ code for leaf types defined in the nyanspec. Other types would simply be a dynamic composition of leaf types (e.g. std::vector<nyantype*>). The C++ should then handle the copy/reference relation between objects. A '+' char in the nyanspec suppose a copy semantic in the C++ code. An '@' char in the nyanspec suppose a reference semantic in the C++ code.

fr3akout:
@iNaKoll So actually you want to use traits for composing nyan types?

iNaKoll:
@fr3akout Yes, that sums up everything.

I knew the C++ traits class technique but traits are a more general concept. So, this proposal is about traits for nyan.

dbrgn:
Are traits something similar to mixins?

iNaKoll:
Since nyan mixins define data (or state) they are mixins not traits, good point! :)

iNaKoll:
To answer @mic-e IRC concerns. The SCOUT and LIGHTCAVALERY units would be defined as follows :

    CAVALRY :
        +WALK,
        +ALIVE,
        +ATTACK,
        +ARMOR,
        +COST
    {}

    SCOUT :
        +CAVALRY   #  scout is a new independent cavalry unit
    {
        name = "scout cavalry"
        health_points = 40   # @ALIVE.health_points implicit??
        damage = 3
        armor_pierce = 1
        armor_melee = 0
}

    LIGHTCAVALRY : 
        @SCOUT  # Light cavalry is an overlay over SCOUT
    {
        name = "light calvalry"
        @ALIVE.health_points += 20
        @ATTACK.damage += 2
        # inherits all other properties from SCOUT
    }

mic-e:
what about

    UNITTYPE BOWMAN {
        {+RANGEDATTACK}

        health=50
        range=4
        rangeddamage=3
        rangedfrequency=1

        abilities = [walk, rangedattack]
    }

    UNITTYPE CROSSBOWMAN : BOWMAN {
        rangeddamage *= 2
        rangedfrequency *= 0.8
    }

    UNITTYPE SUPERMELEECROSSBOWMAN : CROSSBOWMAN {
        [+MELEEATTACK]
        health += 40
        meleedamage = 10
        meleefrequency = 1
        abilities += [meleeattack]
    }

detrumi:
A few points:
- Try to keep the syntax simple. I like mic-e's example, but you could change `{+RANGEDATTACK}` to `+RANGEDATTACK`, for example. Also cut down all the special characters, keywords might be better for clarity.
- Maybe it's a good idea to implement the basic idea first? Hard-code some datatypes for the units first before we implement the codegen, so we can start using them. Inheritance is nice but not needed to get things to work, and those dynamic things aren't really needed until upgrades or civilizations.

TheJJ:
Current version (#8), non final of course but very interesting ideas in it:
https://pad.stusta.mhn.de/p/openage_nyan_proposal_

TheJJ:
also very valuable ideas in here, although the previous link has the latest proposal.
https://pad.stusta.mhn.de/p/nyan-format

franciscod:
Can we start defining some kind of printer/parser/lexer/formalgrammar for the nyan format?

TheJJ:
No, even the grammar is not fix yet because the languate is far from being specified. See the latest pad for design information, @mic-e  and I need to gather some motivation to continue designing this, but it's very complicated :)

ChipmunkV:
Ow, I've seen a similar language of one big RTS: don't even think about the hierarchy for the units. Make them out of independent bricks.

Funny observation about the attribute/dynamic attribute system: it's like a straight Qt QObject with its properties and additional dynamic properties that can appear at the runtime.

TheJJ:
Soooo, we may have a specification:

https://github.com/SFTtech/nyan/blob/master/doc/nyan.md

head over there and improve it if things are missing :)

TheJJ:
Thanks for all the contributions, I think we can safely close this now.

Once the implementation is more or less working, we need to write the openage nyan library to provide all the NyanObjects the engine can handle. This will turn out totally awesome I hope! :dancers:

Issue #189, opened 2015-01-08
Create test cases for existing subsystems


TheJJ:
Because reasons we should add more tests to the code.

Missing systems without any case yet:

[ ] Pathfinding
[ ] Jobmanager
[ ] Terrain blending
[ ] Coordinate system
[ ] Asset manager

Please create some creative tests. An example how that could look like can be seen in `cpp/datastructure/`.

sreejithr:
Hey, I can write unit tests for you. I'm still reading up the codebase. But I'll start right away. Anything in particular you guys want from me?

TheJJ:
@sreejithr: Whatever test you want to implement, go for it :smile:. If you need any help on the codebase the best would be joining our IRC and asking there.

TheJJ:
#197 will benefit greatly from this.

jgelderloos:
Created tests for pathfinding logic in pull request #241

kukinsula:
Hello everyone ! You asked for subsystems tests coverage and I was wondering if you would like to have more tests for the log subsystem (I know it's not in your list that's why I'm asking).

I've been reading up the whole log subsystem and I realized that its test file only tests the log messages through a console LogSink but not through a FileSink. So I managed to also test it and write a few log messages in /tmp/openage-log-test. So now the log's test launches the previous test and the one I just created and it works fine.

Would you guys accept this test ?

TheJJ:
@kukinsula Hi! Just submit your attempt, it's probably good, but we can only tell if we actually see the code :smile:. Thanks for contributing!

kukinsula:
@TheJJ pull request created here: https://github.com/SFTtech/openage/pull/309

sreejithr:
Made existing testcases for JobManager test with atleast 10 jobs in parallel (https://github.com/SFTtech/openage/pull/310).

sreejithr:
Forgot to add myself to copying.md. Oops :) Updated the pull request.

sreejithr:
Cleaned up some whitespaces emacs left. I'm working on a testcase for abortable jobs. Meanwhile, you can check out #310.

sreejithr:
I'm make a test case for abortable jobs. I spawn 20 jobs and stop the JobManager in between (at ~12jobs). Each job has a delay of 1sec. I wait for 20 sec (to make sure that if aborting doesn't work, all the jobs will get completed). After this, I check how many got finished. Hopefully, <20 jobs would be finished which implies aborting worked.

[...]

Issue #230, opened 2015-02-12
Avoid merge conflicts in copying.md author list


TheJJ:
The most common merge conflict are names in copying.md.

The only solution for this i can think of is creating a `contributors/` directory containing one file for each contributor, which contains further info in a yet-to-be-determined format.

This will finally stop all the merge conflicts in the author list. What do you think? Other ideas?

janisozaur:
I think that, while perhaps common, it is the easiest merge conflict to solve.
Other projects, say scummvm, openrct2, etc, all have just one file with all the authors listed in it as well.
Once you reach a certain threshold, this problem will eventually go away, or at least no longer be of great importance.

TheJJ:
this problem will never go away as long as users append themselves to the end of the list.

franciscod:
http://git-scm.com/docs/gitattributes

union

    Run 3-way file level merge for text files, but take lines from both
versions, instead of leaving conflict markers.

this problem will never go away as long as users append themselves to the
end of the list.

—
Reply to this email directly or view it on GitHub
<https://github.com/SFTtech/openage/issues/230#issuecomment-74161132>.

franciscod:
boo, github doesn't support this. made an experiment: https://github.com/franciscod/github-union-merge-test

franciscod:
i've mailed a feature request to github :)

franciscod:
a possible format for the `contributors/` directory idea:

```
contributors/Jonas Jelten:
    aliases: TheJJ
    email: jj@sft.mx

contributors/Matthias Bogad:
    aliases: delirium, masterofjellyfish
    email: matthias@bogad.at
```

TheJJ:
How about underscores in the author filename? Having spaces will require ugly backslashes when referring to the files in a shell. But yeah, otherwise this is what i thought of.

franciscod:
or simply drop the whitespace: `JonasJelten` `FranciscoDemartino` `MichaelEnsslin`

zuntrax:
I'd rather have to merge pull requests with the CLI than have separate files for each author.

rkreis:
I'm not quite happy with having one file per author. I think it would add too much complexity with too little benefit. If we just agree on keeping the list sorted, merge conflicts will eventually become quite rare (and easily resolved).

TheJJ:
[...]

PR #241, opened 2015-03-02
Pathfinding testing #189

jgelderloos:
Added tests for the basic pathfinding methods.


TheJJ:
Thanks!

First of all, please indent with tabs instead of two spaces.

You can just remove the empty header file if it doesn't have content.

Last, please comment the functions in doxygen-compatible style (`/** blabla */`) so one can quickly see what the function is testing and why. The actual length of comments is up2you.

jgelderloos:
OK, I cleaned it up and added doxygen comments.

TheJJ:
i fear you have to rebase the commit and fix the conflict in the authors list.. (#230....)

PR #238, opened 2015-02-20
[WIP] convert script cleanup

zuntrax:
WIP. Do not merge.
This is just for easier progress tracking.
Includes stuff from #237.

PR #237, opened 2015-02-20
[WIP] media version detection

zuntrax:
WIP. Do not merge yet.
This is just for easier progress tracking.

PR #243, opened 2015-03-04
Ignore 'Bin' directory if non-existent

dopsi:
In my installation the 'Bin' directory does not exist, however everything runs fine.
Does this cause any trouble ?

### DIFF
--- a/py/openage/convert/mediafile.py
+++ b/py/openage/convert/mediafile.py
@@ -143,17 +143,21 @@ def media_convert(args):
             # stringres.fill_from(PEFile("Games/Forgotten Empires/Data/"
             #                            "language_x1_p1.dll"))
         else:
-            from .hdlanguagefile import HDLanguageFile
-            for lang in os.listdir(util.file_get_path("Bin")):
-                langfile = "Bin/%s/%s-language.txt" % (lang, lang)
+            try:
+                from .hdlanguagefile import HDLanguageFile
+                for lang in os.listdir(util.file_get_path("Bin")):
+                    langfile = "Bin/%s/%s-language.txt" % (lang, lang)
 
-                # there is some "base language" files in HD that we don't
-                # need and only the dir for the language that's currently in
-                # use contains a language file
-                if os.path.isdir(util.file_get_path("Bin/%s" % (lang)))\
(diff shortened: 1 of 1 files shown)


zuntrax:
Which version of AoE2 do you have? If it's the HD edition, you should have a `Bin` directory. Assuming you didn't tamper with the original file structure and get errors during media conversion, something weird might be going on. I'd like to see a `ls -l` of your `AGE2DIR`
@TheJJ are we aware of cases where `language.dll` is missing in non-HD installations?

The code block you edited will be removed in #237 and #238 and replaced by a more sane version detection. However, you would probably still get an exception because your files don't match what we expect.

The only thing we export from the `Bin` directory are the language files, but afaik we don't use them on the master branch. That's why you don't get errors this way, but as soon as we start displaying unit names or similar stuff, you would run into problems anyway.

goto-bus-stop:
If your language file is called eg. `LANGUAGE.DLL`, the script won't find `language.dll`, and assume HD Edition. See #55…

zuntrax:
That might indeed be the cause.
Seems like I'll have to write some case-insensitive file access helper functions. :fu:

TheJJ:
Hnnng case sensitivity.. :cactus: -> :point_up: -> :astonished:

A solution could be:
* listing all files in a relevant directory
* index them by number
* make a copy of the `idx->filename` dict and do  `.lower()`,
* select the one we wanna have
* and return the actual filename by index

This will probably be readable and portable.

TheJJ:
I'm just implementing such a function, eta 20min.

zuntrax:
Something like that, but we'd also need to reproduce the functionality of `glob.glob()`. `fnmatch` will help, but i don't look forward to it.

dopsi:
Here's the `ls -l` output : http://pastebin.com/vXtacALr
I did not mess with the hierachy, just copied it from my win7 partition to my home.

goto-bus-stop:
@dopsi For the time being, you can rename LANGUAGE.DLL to language.dll, and you should be good to go.

TheJJ:
My attempt is #248 for this, please try whether it works.

PR #247, opened 2015-03-04
buildsystem: improved check for c++14 and clang

JonasBorchelt:


franciscod:
:+1: :ship: :it:

TheJJ:
although i don't like the reformat of the authors list, we probably just have to accept it.

the update itself: :+1:

franciscod:
:confetti_ball: :fireworks: :tada:

PR #248, opened 2015-03-05
Casesensitivity fixes

TheJJ:
Probably closes #243 and finally resolves #55.

Implements case insensitive file name fetching for the whole conversion process.

Please test on all plattforms and game versions, especially @dopsi @Almin17 and @SkySkimmer are encouraged to try whether this branch works for them.

@goto-bus-stop, please review the HD support of this, I can't test it because of a lacking installation.

### DIFF
--- a/py/openage/util.py
+++ b/py/openage/util.py
@@ -1,4 +1,4 @@
-# Copyright 2014-2014 the openage authors. See copying.md for legal info.
+# Copyright 2014-2015 the openage authors. See copying.md for legal info.
 
 
 class NamedObject:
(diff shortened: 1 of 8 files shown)


py/openage/util.py
```diff
@@ -48,3 +57,66 @@ def gen_dict_key2lists(keys):
...
+    a file of any case.
+    """
+
+    import os
+
+    # get actual file names
+    files = os.listdir(directory)
+
+    # create idx->filename dict
+    idxmap = dict(zip((i for i in range(len(files))), files))
```
zuntrax: idxmap = dict(enumerate(files))

goto-bus-stop:
There's still a file_get_path call in HDLanguageFile:
https://github.com/TheJJ/openage/blob/casesensitivity/py/openage/convert/hdlanguagefile.py#L22

It can be removed entirely, because the path is accurate before any HDLanguageFile gets instantiated.

Without that line, I get another error:
```
  File "/usr/lib/python3.4/runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.4/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/tmp/openage/py/openage/convert/__main__.py", line 71, in <module>
    main()
  File "/tmp/openage/py/openage/convert/__main__.py", line 68, in main
    args.handler(args)
  File "/tmp/openage/py/openage/convert/mediafile.py", line 317, in media_convert
    with open(fname, "wb") as f:
FileNotFoundError: [Errno 2] No such file or directory: 'assets/converted/Data/gamedata_x1.drs/60012.bin'
```

Which might just as well be on my side. I'll have a better look later today.

py/openage/util.py
TheJJ: jup, thx

TheJJ:
updated, probably will work for you now @goto-bus-stop, mkdirs was missing for the output folders

goto-bus-stop:
Yep, all good!

dopsi:
Works well now, thanks :+1:

PR #250, opened 2015-03-08
pathfinding: basic tests for Path::Node

TheJJ:
fixup of #241 by me, implements tests for the path node.


TheJJ:
orig pull was updated.

PR #252, opened 2015-03-17
ci caching

TheJJ:
probably speeds up the ci run

i also removed some leftover config sections for coverity.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,3 +55,8 @@ notifications:
       - "%{repository}/%{branch} (%{commit} - %{author}): %{message}"
     use_notice: true
     skip_join: true
+
+cache:
+  directories:
+    - .cicache
+  apt: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-#travis ci config script
+# travis ci config script for openage
 
 
 language: cpp
@@ -45,9 +45,6 @@ after_script:
   - echo test build exited
 
 notifications:
-#  recipients:
-#    - lol@mail.ie
-#    - rofl@liam.ei
   email:
     on_success: change
     on_failure: always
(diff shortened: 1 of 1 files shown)


franciscod:
why the hell not? :+1:

janisozaur:
why was coverity disabled?

TheJJ:
iI was never enabled except for manual uploads.
The problem is that coverty limits the amount of submissions. If we have much activity, each master branch push would submit to coverty, and we won't have a separate coverty branch to push to. So this was just removal of never-enabled code.

janisozaur:
Hmm, pulling a few sources together I got following picture:
1. Coverity was disabled because it didn't support gcc 4.8 [1], likely implying c++11 used extensively throughout the project
2. As you just indicated, coverity limits amount of builds possible, and you don't want another branch just for that.

Regarding point 1: a lot of time has passed since [1] and coverity scan is now at version 7.7.0 [2], where release 7.0 was one adding support for c++11 [3]
Regarding point 2: Coverity only rejects subsequent requests and does not ban you [4], so you could try pushing every build and only the ones falling within allowances would be analyzed.

That's 100% more analysis than none.

[1] https://github.com/TheJJ/openage/commit/c5ee5a51ef34e96035b29d8b4cda5b40f262dd01
[2] https://scan.coverity.com/models
[3] https://communities.coverity.com/docs/DOC-1571
[4] https://scan.coverity.com/faq#frequency, "Once a project reaches the maximum builds per week, additional build requests will be rejected. You will be able to re-submit the build request the following week."

TheJJ:
When `SFTtech/kevin` works and replaces travis for us, this is obsolete anyway. We can then add a hook to the `.kevin` file to submit to coverty. We're using c++14 currently, dunno whether coverty already supports that.

Issue #159, opened 2014-11-17
Slowly implement game_main mess as sane engine features


mic-e:
The `Game` object should be initialized directly from `int main()`, with `--no-video` and `--no-audio` arguments.
##

franciscod:
refactoring, yummy!

franciscod:
progress: https://github.com/franciscod/openage/tree/gamemain-engine-merge

franciscod:
I'm bringing some comments from the now closed #246, which naively merged both classes:

>hm, having a look at it, it doesn't seem like a good idea merging the utterly broken GameMain with the hopefully clean Engine. The better way, as @mic-e hopefully meant in #159, is to migrate all the contents of gamemain to the engine class in a sane way.
>
>All features instanced and "tested" in GameMain should become sane engine features someday. Just merging the contents directly doesn't address the actual problem.
>
> *@TheJJ [said here](https://github.com/SFTtech/openage/pull/246#issuecomment-77269123)*

>Seperation does make some sense. Think of the game menus being rendered before a game is started. The engine is always running to handle any kind of rendering (including ui / menus) even without a game, allowing you to start one game quit back to the main menu and then start a new game.
>
> *@Jon0 [said here](https://github.com/SFTtech/openage/pull/246#issuecomment-77271370)*

TheJJ:
This was long done!

PR #253, opened 2015-03-19
WIP - Coords testing & Tile.get_pos_on_chunk fix

jgelderloos:
Added testing for the coord system
Fixed Tile.get_pos_on_chunk to return the position of the tile on the chunk. This will be an int 0-15. It does not look like this changes anything with the game as it does not seem to be used anywhere. Let me know if you find any problems with it.


cpp/coord/tests.cpp
```diff
@@ -0,0 +1,874 @@
...
+/**
+ * Top level function for coord testing.
+ */
+void coord() {
+	int ret;
+	const char *testname;
+
+	// Some tests below need the engine to be initialized
+	util::Dir dataDir{"assets"};
+	Engine::create( &dataDir ,"testing openage");
```
TheJJ: although i can't blame you for initializing the full engine here, but we should come up with a way the coords can be fully tested without depending on the engine.

jgelderloos: Yeah I kind of figured that once I saw the build failures on travis. I'll have a look into how we can get around it.

franciscod: however, it's proposed on #159 to allow game initialization w/o video/audio support

jgelderloos:
In addition to running with no audio & video we also can't initialize the engine on travis due to not having the assets. The conversions depend on window_size, camgame_phys, camgame_window, camhud_window, and tile_halfsize for a bunch of conversions. Perhapse the best way to deal with it is to move this data out of the engine and just give the engine a pointer to it?

franciscod:
We also could set fallback values to be used for tests, or even custom
assets

mic-e:
I suggest to simply make the conversion constants static members of Engine, and set fake values during testing. In addition, that would be beneficial to performance.

At some point it would be cool to have fake assets for testing, and allow the engine to initialize even though no real video/audio is available, but that's more of a long-term plan.

TheJJ:
I think if it's impossible just testing the coordinate system without initializing the engine, then something about the design is broken (either in the engine or the test).

Imagine the engine's start will take 10s some day, it really would be painful depending on that for just testing the coordinate stuff. It must be possible to test that independently i hope.

mic-e:
@TheJJ: _somebody_ decided that it's a good idea to make information that's essential for coordinate frame translations a member variable of the Engine singleton class.

Making it a static member instead would fix that issue.

TheJJ:
Then we should somehow extract these members so that they can be used without all the other engine stuff be initialized. I'm thinking of some `cood_state` class which is a member of the engine but can also be instanciated separately for testing it.

mic-e:
Just make it an independent, struct with static members/namespace with global variables that is used/filled by the engine and used for coordinate transforms.

TheJJ:
Oh gawd no global variables please... i'd write a struct containing the information that is instanciated by the engine. The coord testing can instanciate this struct independently.

Another idea to eliminate the global state is inheritance: Every component having some coord state inherits from a coord base class. This class provides the state members (the new coord_state struct) and functions for modifying the state. Calling these will only modify the inherited state members, global state is purged.

For testing the coord system, some test class also inherits from this coord base class and simply tests the gained member methods.

TheJJ:
All coordinates instanced have to be bound to some coordinate system, where they can access the state (e.g. current camera position etc). The delta coordinates don't need that bounding i think. This way the conversion can still be done conveniently as the .to_whatever() just makes use of the initially-remembered coordinate system it was bound to.

TheJJ:
When creating new coordinates, they could be generated by a factory of the the respective coordinate base class (the one defining the coordinate space and state). This way generated coordinates are bound to that space without passing the space to the new coordinate's constructor all the time.

Jon0:
Create a camera class, take rendering code out of the GameMain and the coord stuff from Engine.

jgelderloos:
Most the coord conversions should work now without initializing the engine. Now there is just the conversions to and from the term type that are relying on some assets to be loaded.

jgelderloos:
Ok, console is now initialized without the colors needed so we can use its position data from tests. Then in the engine just an extra call to load the colors.

TheJJ:
Awesome, nice. Can you squash the commits please, otherwise the color table will be in the history anyway. If you like, you can even merge everything to one commit but ideally, each commit should be something that would make sense to revert in one batch.

jgelderloos:
Squashed it to a few sensible commits. Let me know if theres anything else we think needs to be added.

TheJJ:
nicely done!
can you change the camel case of the `coordData` to `coord_data` and `getCoordData` -> `get_coord_data` to maintain project consistency? A cool feature i discovered for that is the `git commit --fixup <hash>` in combination with `git rebase --autorebase`, it'll probably help you to make changes to the appropriate commit without rebasing interactively all the time.

TheJJ:
Ah, in your testing code you used camel case multiple times, would be cool if you could update that as well. Only class names are camel'd for us.

jgelderloos:
OK its cleaned up. That is a handy trick!

cpp/coord/tests.cpp
```diff
@@ -0,0 +1,821 @@
...
+	int stage = 0;
+	// Test camhud to window
+
+	// The center is the one place camhud and window are the same
+	window expected_w{400, 300};
+	camhud c{400, 300};
+
+	window results_w = c.to_window();
+	//log::err("%d, %d", results_w.x, results_w.y);
+	//throw;
```
TheJJ: commented leftovers

cpp/engine.cpp
```diff
@@ -62,11 +68,12 @@ Engine::Engine(util::Dir *data_dir, const char *windowtitle)
...
-	camgame_phys{10 * coord::settings::phys_per_tile, 10 * coord::settings::phys_per_tile, 0},
-	camgame_window{400, 300},
-	camhud_window{0, 600},
-	tile_halfsize{48, 24},  // TODO: get from convert script
+	//window_size{800, 600},
+	//camgame_phys{10 * coord::settings::phys_per_tile, 10 * coord::settings::phys_per_tile, 0},
+	//camgame_window{400, 300},
+	//camhud_window{0, 600},
+	//tile_halfsize{48, 24},  // TODO: get from convert script
+	engine_coord_data{this->get_coord_data()},
```
TheJJ: moar commented leftovers

jgelderloos:
OK, fixed those up. You are too good at finding all my little mistages JJ.

franciscod:
that's because jj (and many of us) made all those mistakes a lot of times
:) but yeah, and he's specially good at being an indent nazi

TheJJ:
but now you shall get the reward: let it be merged!111

PR #257, opened 2015-03-24
Fix decaying units

Jon0:
Decaying units face correct direction and disappear after a short time.

Im not sure how the original engine knows which units decay based on gamedata, so it will just check for unit names ending in "_D" for now.

### DIFF
--- a/cpp/unit/producer.cpp
+++ b/cpp/unit/producer.cpp
@@ -49,6 +49,9 @@ ObjectProducer::ObjectProducer(DataManager &dm, const gamedata::unit_object *ud)
 	default_tex{dm.get_unit_texture(ud->graphic_standing0)},
 	dead_unit_producer{dm.get_producer(ud->dead_unit_id)} {
 
+	// for now just look for type names ending with "_D"
+	this->decay = unit_data.name.substr(unit_data.name.length() - 2) == "_D";
+
 	// find suitable sounds
 	int creation_sound = this->unit_data.sound_creation0;
 	int dying_sound = this->unit_data.sound_dying;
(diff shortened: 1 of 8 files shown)


mic-e:
What do you mean by "which units decay"? Are there units that don't, or is the decay graphic not linked properly?

Jon0:
when units die the type of the unit changes (no longer has move/attack abilities), basically the gamedata says when this unit type dies it becomes another type.

the decay types (names end with "_D") use the standing graphic. the data must have some better way of indicating these units have a timed lifespan

mic-e:
Maybe the game just knows it from the fact that it just changed the unit to is "dying" type, and the info is not part of the type at all. Is there some variable that indicates the decay speed?

Jon0:
No variable I have found yet, Units are given a type by a set type function, so how can that function even know its the decay type of another, there should be a gamedata attribute on the type.

mic-e:
There _should_ be, but you might have noticed that the Genie data files are pretty horrible... my bet is that Genie simply has an internal struct for each unit, where it  where its, type, current hp, and "dead" status. Maybe it even uses the negative range of the hp value for decay state?

cpp/unit/producer.cpp
```diff
@@ -120,10 +123,8 @@ std::string ObjectProducer::producer_name() const {
 void ObjectProducer::initialise(Unit *unit, Player &player) {
 
 	// log attributes
-	unit->log(MSG(dbg) << "creating " <<
-		this->unit_data.id0 << " (" <<
-		this->unit_data.id1 << " " <<
-		this->unit_data.id2 << ") " <<
+	unit->log(MSG(dbg) << "setting unit type " <<
+		this->unit_data.id0 <<
 		this->unit_data.name);
```
mic-e: there's no space between id0 and name... is that intentional?

Jon0:
no

mic-e:
Looking good :+1:

cpp/unit/producer.cpp
```diff
@@ -196,10 +203,10 @@ TerrainObject *ObjectProducer::place(Unit *u, Terrain &terrain, coord::phys3 ini
 	// create new object with correct base shape
 	TerrainObject *obj = nullptr;
 	if (this->unit_data.selection_shape > 1) {
-		obj = new RadialObject(*u, this->unit_data.radius_size0, this->terrain_outline.get());
+		obj = new RadialObject(*u, this->unit_data.radius_size0, this->terrain_outline, !decay);
 	}
 	else {
-		obj = new SquareObject(*u, this->foundation_size, this->terrain_outline.get());
+		obj = new SquareObject(*u, this->foundation_size, this->terrain_outline, !decay);
```
TheJJ: `this->`

cpp/unit/unit.cpp
```diff
@@ -223,4 +213,21 @@ std::string Unit::logsource_name() {
 	return "Unit " + std::to_string(this->id);
 }
 
+void Unit::erase_after(std::function<bool(std::unique_ptr<UnitAction> &)> func) {
+	auto position_it = std::find_if(
+		std::begin(this->action_stack),
+		std::end(this->action_stack),
+		func);
+
+	if (position_it < std::end(this->action_stack)) {
```
TheJJ: really `<`? isn't `!=` what one wants comparing to end?

PR #259, opened 2015-03-25
More Fixes

Jon0:
Fixes:
* Ranged units no longer freeze trying to garrison
* Garrison restricted to allied buildings (and must be built)
* Siege can be placed on land not water now.
* Buildings take some time to train a unit.
* Units can ungarrison while a building is training something.
* Fix segfault when shuting down


cpp/unit/action.cpp
```diff
@@ -69,10 +69,11 @@ void UnitAction::damage_object(Unit &target, unsigned dmg) {
 	}
 }
 
-void UnitAction::move_to(Unit &target) {
+void UnitAction::move_to(Unit &target, bool use_range) {
```
TheJJ: A boolean here is good, but using an enum class would increase readability instead of just having true and false as "visible" parameters. Don't you think?

cpp/terrain/terrain.cpp
```diff
@@ -26,7 +26,11 @@ TileContent::TileContent() :
 }
 
 TileContent::~TileContent() {
-	for (auto &o : this->obj) {
+	// objects will try remove themselves from the list
+	// TODO: a better cleanup
```
mic-e: As discussed on the IRC, you'll probably want to use `weak_ptr` here

cpp/unit/ability.cpp
```diff
@@ -81,9 +88,23 @@ GarrisonAbility::GarrisonAbility(Sound *s)
 }
 
 bool GarrisonAbility::can_invoke(Unit &to_modify, const Command &cmd) {
-	return to_modify.location &&
-	       cmd.has_unit() &&
-	       cmd.unit()->has_attribute(attr_type::garrison);
```
mic-e: You could keep the indent depth lower by instead writing

    if (!cmd.has_unit()) {
        return false;
    }

cpp/unit/ability.cpp
```diff
@@ -81,9 +88,23 @@ GarrisonAbility::GarrisonAbility(Sound *s)
...
+	if (cmd.has_unit()) {
+		Unit &target = *cmd.unit();
+
+		// make sure buildings are completed
+		if (target.has_attribute(attr_type::building)) {
+			auto &build_attr = target.get_attribute<attr_type::building>();
```
mic-e: Why doesn't `has_attribute` use the same template-parameter syntax as `get_attribute`?

Jon0:
why did it fail?

goto-bus-stop:
Because of Github being semi-down from the DDOS earlier, seems like.

TheJJ:
i retriggered the build manually, should be successful now.

mic-e:
What's the status of this?

Jon0:
complete, ready to review/ pull.

Jon0:
thats all for this PR, I will make a new one once this is pulled because it will have new features not just bug fixes.

cpp/terrain/terrain_object.cpp
```diff
@@ -26,26 +26,28 @@ TerrainObject::TerrainObject(Unit &u, bool collisions)
...
-	for (auto &c : children) {
-		c->remove();
-		delete c;
-	}
 	this->unit.location = nullptr;
 }
 
+void TerrainObject::initialise() {
+
+	// remove any exisitng location
```
TheJJ: typo

cpp/terrain/terrain_object.cpp
```diff
@@ -26,26 +26,28 @@ TerrainObject::TerrainObject(Unit &u, bool collisions)
...
+void TerrainObject::initialise() {
+
+	// remove any exisitng location	
+	if (unit.location) {
+		unit.location->remove();
+	}
+
+	// shared_from_this cannot be used in the constructor,
+	// so an initalise function is required
+	unit.location = this->shared_from_this();
```
TheJJ: this is pretty dirty, but doing it sanely is probably more difficult.

mic-e:
BTW I'm really unhappy with the name `TerrainObject`; the name `PhysicalObject` would be much less ambiguous (that's not an issue with this PR, just something that I've been unhappy with for at least a year).

cpp/terrain/terrain_object.cpp
mic-e: Is there a reason you can't simply do a regular `u->location = std::make_shared<TerrainObject>();`?

cpp/terrain/terrain_object.h
```diff
@@ -97,6 +97,11 @@ class TerrainObject {
 	std::function<void()> draw;
 
 	/**
+	 * sets this as the location of the unit
+	 */
+	void initialise();
```
TheJJ: do we want murrican english? if yes, we'd want `initialize`.

cpp/unit/ability.cpp
```diff
@@ -21,6 +21,16 @@ bool has_resource(Unit &target) {
 	       target.get_attribute<attr_type::resource>().amount > 0;
 }
 
+bool is_ally(Unit &to_modify, Unit &target) {
+	if (to_modify.has_attribute(attr_type::owner) &&
+		target.has_attribute(attr_type::owner)) {
```
TheJJ: indent to the `if (`

cpp/unit/ability.cpp
```diff
@@ -21,6 +21,16 @@ bool has_resource(Unit &target) {
...
+		auto &mod_player = to_modify.get_attribute<attr_type::owner>().player;
+		auto &tar_player = target.get_attribute<attr_type::owner>().player;
+		return mod_player.is_ally(tar_player);
+	}
+	return false;
```
TheJJ: sure it's a good idea to say "it's not an ally" when one of the units doesn't have the owner attribute?

cpp/unit/ability.cpp
```diff
@@ -108,7 +132,9 @@ void UngarrisonAbility::invoke(Unit &to_modify, const Command &cmd, bool play_so
 	if (play_sound && this->sound) {
 		this->sound->play();
 	}
-	to_modify.push_action(std::make_unique<UngarrisonAction>(&to_modify, cmd.position()));
+
+	// add as secondary, so primary action is not disrupted
+	to_modify.secondary_action(std::make_unique<UngarrisonAction>(&to_modify, cmd.position()));
```
TheJJ: Do you think we should have even more than two action stacks?
It may be good in the future to let the units do any number of actions in parallel. Probably it's better extending the system once we need it.

cpp/unit/command.h
```diff
@@ -10,6 +12,34 @@
...
+namespace std {
+
+/**
+ * hasher for game command flags
+ */
+template<> 
+struct hash<openage::command_flag> {
+	typedef underlying_type<openage::command_flag>::type underlying_type;
+	typedef hash<underlying_type>::result_type result_type;
+	result_type operator()(const openage::command_flag& arg) const {
```
TheJJ: `..flag &arg`

cpp/unit/producer.cpp
```diff
@@ -41,6 +39,33 @@ std::unordered_set<terrain_t> allowed_terrains(const gamedata::ground_type &rest
 	return result;
 }
 
+std::shared_ptr<TerrainObject> UnitProducer::place_beside(Unit *u, std::shared_ptr<TerrainObject> other) const {
+	if (!other) {
```
TheJJ: can the shared ptr even be false? if yes, then i'd test for the unit as well.

cpp/unit/producer.cpp
```diff
@@ -507,7 +538,8 @@ TerrainObject *BuldingProducer::place(Unit *u, Terrain &terrain, coord::phys3 in
 		for (coord::tile check_pos : tile_list(obj->get_range(pos))) {
 			TileContent *tc = terrain.get_data(check_pos);
 			if (!tc) return false;
```
TheJJ: {} and newlines pls

cpp/unit/producer.cpp
```diff
@@ -686,11 +718,12 @@ TerrainObject *ProjectileProducer::place(Unit *u, Terrain &terrain, coord::phys3
...
 
 			// ensure no intersections with other objects
-			for (auto obj_cmp : tc->obj) {
+			for (auto item : tc->obj) {
+				auto obj_cmp = item.lock();
 				if (obj != obj_cmp && 
 					&obj_cmp->unit != launcher &&
```
TheJJ: indent!11

cpp/unit/ability.cpp
mic-e: The secondary thing is not a stack, but a set of actions that might run in parallel.

iNaKoll: I don't understand why this secondary_action is needed. Why can't we just insert the UngarrisonAction in front of the "primary stack" ?

cpp/terrain/terrain_object.cpp
Jon0: ok I will give unit a set_location(shared_ptr<TerrainObject>) function with the above logic instead

cpp/unit/ability.cpp
Jon0: because a building might be training something on the primary and should be allowed to ungarrison things at the same time

cpp/terrain/terrain_object.cpp
mic-e: `u->location` is the single owner of the TerrainObject, right?

Which cases are there where a `TerrainObject` is constructed, and then only later assigned to the location?

Jon0: On second thought, terrainObject is permanently tied to one unit, and I didnt want:
```cpp
some_other_unit->location = make_shared<TerrainObject>(some_unit);
```

Jon0:
Fixed style, will leave fixing initialise for my other branch to avoid conflicts.

PR #261, opened 2015-03-26
buildsystem: fix the cmake cmp0054 policy for using quoted if evaluation

TheJJ:
trivial change, but the warning goes away.


franciscod:
:+1: totally worth it, also we gain FOUR BYTES OF STORAGE FOR MORE SOURCE CODE!!!

janisozaur:
you can now do http://codegolf.stackexchange.com/a/47873 w00t!

PR #262, opened 2015-03-29
Abort on checkfull fail

TheJJ:
Currently, the sanity check is done before compiling, but the build doesn't abort if something is wrong.
That way, the error messages get burried before all the compilation output.

With this, Travis won't compile if checkfull fails.
Also, fancier IRC notification.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,7 +53,8 @@ notifications:
     channels:
       - "chat.freenode.net#sfttech"
     template:
-      - "%{repository}/%{branch} (%{commit} - %{author}): %{message}"
+      - "%{repository_slug}/%{branch} (%{commit} - %{author}): %{message}"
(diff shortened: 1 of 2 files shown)


TheJJ:
hm, I don't like the `&&` chaining.
using `$?` doesn't seem to work, any other ideas?

franciscod:
Maybe travis supports `set -e`, more info here: http://mywiki.wooledge.org/BashFAQ/105

mic-e:
My guess is that Travis spawns a new shell for every single command.
I think we're (ab-)using it in a way it's not meant to be used in, so either add a `travis` target to the makefile, or simply deal with it.

mic-e:
Please add a 'help' entry for runci in the makefile.

TheJJ:
done

Issue #86, opened 2014-11-02
Create asset packs and automatically locate them with the engine


mic-e:
Implement an asset loader that overlays multiple asset source directories. For an installed game, this should be `/usr/share/openage` (installed by the package) and `~/.openage` (the user-writable "userasset" folder).

You can also specify further paths to look for packs.

The userasset folder is used for storing files that are created by the user (e.g.: datapacks (converted game assets)), savegames, screenshots, etc.

The "asset locating mode" may be specified via `--assets={dev, installed}`. If not specified, it's auto-determined from the files in the folder where the executable lies (does `openage_version` exist?). `dev` => look in ./assets.

Performance optimization feature: Support packing folders in (tar) archives.

The AssetLoader takes an asset path (such as `termcolors.docx`) and an optional "on-modification" callback function (think inotify, manual "reload-assets" instruction) and returns a file-like object for reading (or raise an Exception).

All code in openage should rely on the AssetLocator instead of directly opening files.

TheJJ:
A basic data pack format will be proposed with #204 as the pack is exactly the place to store versioning information.

a-detiste:
Hi,

I'll add support for Age Of Empires II to Debian's game-data-packager, a tool that aim to support automatically packaging data for _all_ the games with a free engine and non-free assets.

You can have look here if you're curious:
http://anonscm.debian.org/cgit/pkg-games/game-data-packager.git/tree/data

Please also look for data in `/usr/share/games/openage` :smile:

mic-e:
Nice to see that debian supports something like this. We will look in `/usr/share` as well as the user's homefolder.
Note that an installed version of the game is required for performing the conversion.

a-detiste:
> debian

Feel free to adapt it to the distro of your choice if you find it usefull.

> conversion

There is a plugin system for games with specific needs and data conversion is already supported for [LGeneral][1] ; only requirement is that it can happen in a dedicated command-line tool provided by your repo.

If [this][2] is the conversion program, this could also be directly `import`ed in a to-be `openage.py` plugin:

[1]: http://anonscm.debian.org/cgit/pkg-games/game-data-packager.git/tree/game_data_packager/games/lgeneral.py

[2]: https://github.com/SFTtech/openage/tree/master/py/openage/convert
----

does this engine plan to support the demo version of the game ?

janisozaur:
You should rather try using `$XDG_DATA_HOME` instead of `~/.openage`.

[...]

Issue #79, opened 2014-11-01
Targets for media convert python module


TheJJ:
Currently, the `openage.convert` script always executes the conversion in a fixed manner, the only adjustments possible are filename matches for archive extractions.

Please implement some job-like architecture so that each conversion task can be selected separately and grouped together to common conversion targets.

The file that currently hardcodes the conversion procedure is `py/openage/convert/mediafile.py`.

TheJJ:
This is already in place with #288

PR #265, opened 2015-03-29
[WIP] Media versioning

### LINKED ISSUE #204: Media file versioning

TheJJ:
My attempt on versioning the media files. Introduces a data pack format (#86).

Basically, all converted and generated files will store some version (e.g. the gamedata specification hash). The engine can detect required/present discrepancies and trigger reconversion.

When this is merged, users won't have to reconvert everything for changes in one area only, but this will require #79 to be finished.

This implements and thus closes #204.

New features:

* [X] Hashing the exported struct member configuration
* [X] Simple mod data pack format
* [ ] Check if the compiled-in struct format hash == exported-gamedata-files-format-hash

### DIFF
--- a/py/openage/convert/dataformat/exportable.py
+++ b/py/openage/convert/dataformat/exportable.py
@@ -1,5 +1,6 @@
-# Copyright 2014-2014 the openage authors. See copying.md for legal info.
+# Copyright 2014-2015 the openage authors. See copying.md for legal info.
 
+import hashlib
 import struct
 
 from ..util import zstr
@@ -381,7 +382,7 @@ class Exportable:
             dbg(lazymsg=lambda: "%s: exporting member %s<%s>" % (repr(cls), member_name, member_type), lvl=3)
 
             if isinstance(member_type, MultisubtypeMember):
-                for subtype_name, subtype_class in member_type.class_lookup.items():
+                for subtype_name, subtype_class in sorted(member_type.class_lookup.items()):
                     if not issubclass(subtype_class, Exportable):
                         raise Exception("tried to export structs from non-exportable %s" % subtype_class)
                     ret += subtype_class.structs()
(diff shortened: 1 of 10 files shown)


py/openage/convert/dataformat/exportable.py
```diff
@@ -381,7 +382,7 @@ def structs(cls):
             dbg(lazymsg=lambda: "%s: exporting member %s<%s>" % (repr(cls), member_name, member_type), lvl=3)
 
             if isinstance(member_type, MultisubtypeMember):
-                for subtype_name, subtype_class in member_type.class_lookup.items():
+                for subtype_name, subtype_class in sorted(member_type.class_lookup.items()):
```
franciscod: why do we need to sort the items?

py/openage/convert/dataformat/exportable.py
```diff
@@ -404,11 +405,61 @@ def structs(cls):
...
+        hasher.update(cls.name_struct_file.encode())
+        hasher.update(cls.struct_description.encode())
+
+        # only hash exported struct members!
+        # non-exported values don't influence anything.
+        members = cls.get_data_format(
+            allowed_modes=(True, READ_EXPORT, NOREAD_EXPORT),
+            flatten_includes=False,
+        )
+        for is_parent, export, member_name, member_type in members:
```
franciscod: newline before this?

py/openage/pack/pack_config.py
```diff
@@ -0,0 +1,113 @@
+# Copyright 2015-2015 the openage authors. See copying.md for legal info.
+
+import configparser
```
franciscod: yay configparser!

py/openage/convert/dataformat/exportable.py
mic-e: If you don't sort the items, their order is undefined; in practice, it varies from run to run.
The iteration order translates to the order of the segments in the generated C++ code, which will be different with each run, triggering unneeded re-builds.

py/openage/pack/pack_config.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+    main config file for a mod pack.
+
+    stores the package metainformation for the mod list.
+
+    file contents:
+
+    [openage-pack]
+    name = <package name>
+    version = <package version>
```
mic-e: What precisely is the package version?
Is it the version of the used package format, or the actual revision of the package content?
You might need two different, more precisely named fields here.

py/openage/pack/pack_config.py
```diff
@@ -0,0 +1,113 @@
...
+    author = <pack author>
+    pack_type = (game|mod|...)
```
mic-e: Is there a difference between games and mods?

py/openage/convert/dataformat/exportable.py
franciscod: oh, thought I was still reading the hashing code, not the C++ struct exporting one! now it makes sense lol

py/openage/pack/pack_config.py
franciscod: maybe game is the root pack and mod sits on top of a game or a mod?

py/openage/convert/dataformat/exportable.py
TheJJ: this for-loop iterates over the members, so i thought sticking them together might point that out.

py/openage/pack/pack_config.py
TheJJ: This was intended to be the revision of the content. But yeah, the name is ambiguous.

py/openage/pack/pack_config.py
TheJJ: Hm, yes, they are probably the same. I thought of this difference because then we could separate the download lists by mods and games. The engine can run Star Wars or AOE and maybe more, and these shouldn't be in the "mod" selection list i think.

TheJJ:
This is kinda obsolete since the pyinterface was merged, I'll try updating some time and introduce mod pack config files.

TheJJ:
To revive this, the format hashing should be ported to the current version again.
It's a version indicator for the gamespec format, that should trigger a reconversion if different from what the engine currently understands.

TheJJ:
This is now upstream btw, came in with e59f5c8e178f11857dd68ae4cf51bc9357a6c9c2.

PR #267, opened 2015-03-29
Cleanup

dbrgn:
I started reading myself into the resource conversion process. These are a few cleanups I found along the way.

Reasoning for the module level imports can be found in the commit comment.

### DIFF
--- a/py/openage/codegen/codegen.py
+++ b/py/openage/codegen/codegen.py
@@ -1,4 +1,4 @@
-# Copyright 2014-2014 the openage authors. See copying.md for legal info.
+# Copyright 2014-2015 the openage authors. See copying.md for legal info.
 
 import datetime
 import os
--- a/copying.md
+++ b/copying.md
@@ -48,6 +48,7 @@ _the openage authors_ are:
 | Jonas Borchelt              | riotjones                   | jonasbr-github@bellatrix.uberspace.de |
 | Jon Gelderloos              | jgelderloos                 | jgelderloos@gmail.com                 |
 | Emmanuel Gil Peyrot         | Link Mauve                  | linkmauve@linkmauve.fr                |
+| Danilo Bargen               | dbrgn                       | mail@dbrgn.ch                         |
 
 If you're a first-time commiter, add yourself to the above list. This is not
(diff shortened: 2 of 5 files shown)


mic-e:
Looking pretty good; are those finds from pylint?

Are you planning to add any more fixes?

py/openage/codegen/codegen.py
```diff
@@ -6,7 +6,7 @@
 
 from openage.log import dbg
 
-contenttemplate = """\
+contenttemplate = """
```
mic-e: for the content template, the `\` is unfortunately necessary; otherwise the generated files will start with a newline (where a copyright line would be expected)

franciscod: how about this?
```contenttemplate = """{prefix} {copyrightline}```

mic-e: In my personal opinion, the old `"""\` is the least ugly option.

dbrgn:
> Looking pretty good; are those finds from pylint?

Flake8 and simply going through the code line by line.

> Are you planning to add any more fixes?

Probably, but this can be merged for now. I can always create new PRs.

py/openage/codegen/codegen.py
dbrgn: According to @TheJJ the backslash is not only not necessary, it actually shows up in the generated docs. Therefore I removed them all.

mic-e: That's true for docstrings, but `contenttemplate` is not a docstring :P

dbrgn: Uh, I'm sorry :) That was a mistake. I'll fix.

dbrgn: Fixed & rebased.

The alternative would be to add `.strip()` at the end. But I find the backslash cleaner.

mic-e:
:+1: (we have a four-eyes policy for non-trivial stuff, so you'll have to wait for TheJJ to review it) (but don't let that dissuade you from adding more commits)

TheJJ:
looking alright :+1:

PR #268, opened 2015-03-31
Auto-task idle units

Jon0:
* Rename Producer to UnitType
* Terrain objects are now actually cleaned up
* Annex terrain object are cleaned up
* Many more bug fixes
* Idle units find nearby units to attack
* Faster searching for dropsites
* Buildings now placed as floating outlines 
* Floating buildings do not block units
* The terrain objects can now only be constructed via units:
```c++
unit->make_location<SquareObject>(...)
```

### DIFF
--- a/cpp/datamanager.h
+++ b/cpp/datamanager.h
@@ -55,7 +55,7 @@ public:
 	void check_updates();
 
 	/**
-	 * check if loading has been completed 
+	 * check if loading has been completed
 	 */
 	bool load_complete();
 
@@ -69,7 +69,7 @@ public:
 	 */
 	index_t get_slp_graphic(index_t slp);
 
-	/** 
+	/**
 	 * get a texture by id, this specifically avoids returning the missing placeholder texture
 	 */
 	Texture *get_texture(index_t graphic_id);
@@ -95,7 +95,7 @@ public:
 	 */
 	UnitType *get_type_index(size_t type_index);
 
-	/** 
+	/**
 	 * data for a graphic
 	 */
 	const gamedata::graphic *get_graphic_data(index_t grp_id);
@@ -170,8 +170,8 @@ private:
 	void create_unit_types(const std::vector<gamedata::empiresdat> &gamedata, int your_civ_id);
 
 	/**
-	 * loads required assets to produce an entity type
-	 * adds to the producer list if the object can be created safely
+	 * loads required assets to construct a unit type
+	 * adds to the type list if the object can be created safely
 	 */
 	void load_building(const gamedata::unit_building &, unit_type_list &);
 	void load_living(const gamedata::unit_living &, unit_type_list &);
--- a/cpp/datamanager.h
+++ b/cpp/datamanager.h
@@ -86,12 +86,12 @@ public:
 	Sound *get_sound(index_t sound_id);
 
 	/**
-	 * producers by aoe unit ids -- the producer which corresponds to an aoe unit id
+	 * unit types by aoe gamedata unit ids -- the unit type which corresponds to an aoe unit id
 	 */
 	UnitType *get_type(index_t type_id);
 
 	/**
-	 * producers by list index -- a continuous array of all producers
+	 * unit types by list index -- a continuous array of all types
 	 */
 	UnitType *get_type_index(size_t type_index);
 
@@ -140,12 +140,12 @@ private:
 	std::unordered_map<index_t, std::vector<const gamedata::unit_command *>> commands;
 
 	/**
-	 * unit ids -> producer for that id
+	 * unit ids -> unit type for that id
 	 */
 	std::unordered_map<index_t, UnitType *> producers;
 
 	/**
-	 * unit ids -> producer for that id
+	 * unit ids -> unit type for that id
 	 */
 	std::unordered_map<index_t, std::shared_ptr<UnitTexture>> unit_textures;
 
--- a/cpp/datamanager.h
+++ b/cpp/datamanager.h
@@ -14,13 +14,13 @@
 namespace openage {
 
 class AssetManager;
-class UnitProducer;
+class UnitType;
 
 /**
  * the key type mapped to data objects
  */
 using index_t = int;
-using producer_list = std::vector<std::unique_ptr<UnitProducer>>;
+using unit_type_list = std::vector<std::unique_ptr<UnitType>>;
 
 /**
  * simple sound object
@@ -88,12 +88,12 @@ public:
 	/**
 	 * producers by aoe unit ids -- the producer which corresponds to an aoe unit id
 	 */
-	UnitProducer *get_producer(index_t producer_id);
+	UnitType *get_type(index_t type_id);
 
 	/**
 	 * producers by list index -- a continuous array of all producers
 	 */
-	UnitProducer *get_producer_index(size_t producer_index);
+	UnitType *get_type_index(size_t type_index);
 
 	/** 
 	 * data for a graphic
@@ -117,7 +117,7 @@ private:
 	/**
 	 * all available game objects.
 	 */
-	producer_list available_objects;
+	unit_type_list available_objects;
 
 	/**
 	 * slp to graphic reverse lookup
@@ -142,7 +142,7 @@ private:
 	/**
 	 * unit ids -> producer for that id
 	 */
-	std::unordered_map<index_t, UnitProducer *> producers;
+	std::unordered_map<index_t, UnitType *> producers;
 
 	/**
 	 * unit ids -> producer for that id
@@ -167,16 +167,16 @@ private:
 	/**
 	 * makes producers for all types in the game data
 	 */
-	void create_producers(const std::vector<gamedata::empiresdat> &gamedata, int your_civ_id);
+	void create_unit_types(const std::vector<gamedata::empiresdat> &gamedata, int your_civ_id);
 
 	/**
 	 * loads required assets to produce an entity type
 	 * adds to the producer list if the object can be created safely
 	 */
-	void load_building(const gamedata::unit_building &, producer_list &);
-	void load_living(const gamedata::unit_living &, producer_list &);
-	void load_object(const gamedata::unit_object &, producer_list &);
-	void load_projectile(const gamedata::unit_projectile &, producer_list &);
+	void load_building(const gamedata::unit_building &, unit_type_list &);
+	void load_living(const gamedata::unit_living &, unit_type_list &);
+	void load_object(const gamedata::unit_object &, unit_type_list &);
+	void load_projectile(const gamedata::unit_projectile &, unit_type_list &);
 
 	/**
 	 * has game data been load yet
--- a/cpp/game_main.cpp
+++ b/cpp/game_main.cpp
@@ -346,16 +346,27 @@ bool GameMain::on_input(SDL_Event *e) {
 			if (this->building_placement) {
 
 				// confirm building placement with left click
-				Command cmd(this->players[0], this->datamanager.get_type_index(this->editor_current_building), mousepos_phys3);
-				cmd.set_ability(ability_type::build);
-				selection.all_invoke(cmd);
+				// first create foundation using the producer
+				auto player = this->selection.owner();
+				if (player) {
+					auto container = &this->placed_units;
+					auto building_type = this->datamanager.get_type_index(this->editor_current_building);
+					auto new_building = container->new_unit(*building_type, *player, mousepos_phys3);
+
+					// task all selected villagers to build
+					if (new_building.is_valid()) {
+						Command cmd(*player, new_building.get());
+						cmd.set_ability(ability_type::build);
+						this->selection.all_invoke(cmd);
+					}	
+				}
 				this->building_placement = false;
 			}
 			else {
 
 				// begin a boxed selection
-				selection.drag_begin(mousepos_camgame);
-				dragging_active = true;
+				this->selection.drag_begin(mousepos_camgame);
+				this->dragging_active = true;
 			}
 		}
 		else if (clicking_active and e->button.button == SDL_BUTTON_LEFT and construct_mode) {
@@ -560,8 +571,11 @@ bool GameMain::on_draw() {
 	}
 
 	if (this->building_placement) {
-		auto txt = this->datamanager.get_type_index(this->editor_current_building)->default_texture();
-		txt->draw(mousepos_tile.to_phys2().to_phys3().to_camgame(), 0, 1);
+		auto building_type = this->datamanager.get_type_index(this->editor_current_building);
+		auto txt = building_type->default_texture();
+		auto size = building_type->foundation_size;
+		tile_range center = building_center(mousepos_tile.to_phys2().to_phys3(), size);
+		txt->draw(center.draw.to_camgame(), 0, 1);
 	}
 
 	return true;
--- a/cpp/game_main.cpp
+++ b/cpp/game_main.cpp
@@ -289,8 +289,8 @@ GameMain::GameMain(Engine *engine)
 	this->keybind_context.bind(keybinds::action_t::TRAIN_OBJECT, [this]() {
 		// attempt to train editor selected object
 		if ( this->datamanager.producer_count() > 0 ) {
-			UnitProducer *producer = this->datamanager.get_producer_index(this->editor_current_building);
-			Command cmd(this->players[0], producer);
+			auto type = this->datamanager.get_type_index(this->editor_current_building);
+			Command cmd(this->players[0], type);
 			this->selection.all_invoke(cmd);
 		}
 	});
@@ -346,7 +346,7 @@ bool GameMain::on_input(SDL_Event *e) {
 			if (this->building_placement) {
 
 				// confirm building placement with left click
-				Command cmd(this->players[0], this->datamanager.get_producer_index(this->editor_current_building), mousepos_phys3);
+				Command cmd(this->players[0], this->datamanager.get_type_index(this->editor_current_building), mousepos_phys3);
 				cmd.set_ability(ability_type::build);
 				selection.all_invoke(cmd);
 				this->building_placement = false;
@@ -408,7 +408,7 @@ bool GameMain::on_input(SDL_Event *e) {
 			} else if ( this->datamanager.producer_count() > 0 ) {
 				// try creating a unit
 				log::log(MSG(dbg) << "create unit with producer id " << this->editor_current_building);
-				UnitProducer &producer = *this->datamanager.get_producer_index(this->editor_current_building);
+				auto &producer = *this->datamanager.get_type_index(this->editor_current_building);
 				this->placed_units.new_unit(producer, this->players[rand() % this->players.size()], mousepos_tile.to_phys2().to_phys3());
 			}
 			break;
@@ -560,7 +560,7 @@ bool GameMain::on_draw() {
 	}
 
 	if (this->building_placement) {
-		auto txt = this->datamanager.get_producer_index(this->editor_current_building)->default_texture();
+		auto txt = this->datamanager.get_type_index(this->editor_current_building)->default_texture();
 		txt->draw(mousepos_tile.to_phys2().to_phys3().to_camgame(), 0, 1);
 	}
 
@@ -578,7 +578,7 @@ bool GameMain::on_drawhud() {
 		coord::window bpreview_pos;
 		bpreview_pos.x = e.engine_coord_data->window_size.x - 200;
 		bpreview_pos.y = 200;
-		auto txt = this->datamanager.get_producer_index(this->editor_current_building)->default_texture();
+		auto txt = this->datamanager.get_type_index(this->editor_current_building)->default_texture();
 		txt->sample(bpreview_pos.to_camhud());
 	}
 	return true;
--- a/cpp/game_main.cpp
+++ b/cpp/game_main.cpp
@@ -347,18 +347,18 @@ bool GameMain::on_input(SDL_Event *e) {
 
 				// confirm building placement with left click
 				// first create foundation using the producer
-				auto player = this->selection.owner();
+				Player *player = this->selection.owner();
 				if (player) {
-					auto container = &this->placed_units;
-					auto building_type = this->datamanager.get_type_index(this->editor_current_building);
-					auto new_building = container->new_unit(*building_type, *player, mousepos_phys3);
+					UnitContainer *container = &this->placed_units;
+					UnitType *building_type = this->datamanager.get_type_index(this->editor_current_building);
+					UnitReference new_building = container->new_unit(*building_type, *player, mousepos_phys3);
 
 					// task all selected villagers to build
 					if (new_building.is_valid()) {
 						Command cmd(*player, new_building.get());
 						cmd.set_ability(ability_type::build);
 						this->selection.all_invoke(cmd);
-					}	
+					}
 				}
 				this->building_placement = false;
 			}
@@ -419,7 +419,7 @@ bool GameMain::on_input(SDL_Event *e) {
 			} else if ( this->datamanager.producer_count() > 0 ) {
 				// try creating a unit
 				log::log(MSG(dbg) << "create unit with producer id " << this->editor_current_building);
-				auto &producer = *this->datamanager.get_type_index(this->editor_current_building);
+				UnitType &producer = *this->datamanager.get_type_index(this->editor_current_building);
 				this->placed_units.new_unit(producer, this->players[rand() % this->players.size()], mousepos_tile.to_phys2().to_phys3());
 			}
 			break;
--- a/cpp/game_main.cpp
+++ b/cpp/game_main.cpp
@@ -413,7 +413,7 @@ bool GameMain::on_input(SDL_Event *e) {
 			// delete any unit on the tile
 			if (!chunk->get_data(mousepos_tile)->obj.empty()) {
 				// get first object currently standing at the clicked position
-				auto obj = chunk->get_data(mousepos_tile)->obj[0].lock();
+				TerrainObject *obj = chunk->get_data(mousepos_tile)->obj[0];
 				log::log(MSG(dbg) << "delete unit with unit id " << obj->unit.id);
 				obj->unit.delete_unit();
 			} else if ( this->datamanager.producer_count() > 0 ) {
--- a/cpp/game_save.cpp
+++ b/cpp/game_save.cpp
@@ -5,6 +5,7 @@
 
 #include "unit/producer.h"
 #include "unit/unit.h"
+#include "unit/unit_type.h"
 #include "game_main.h"
 #include "game_save.h"
 
@@ -12,7 +13,7 @@ namespace openage {
 namespace gameio {
 
 void save_unit(std::ofstream &file, Unit *unit) {
-	file << unit->producer->producer_id() << std::endl;
+	file << unit->unit_type->id() << std::endl;
 	file << unit->get_attribute<attr_type::owner>().player.player_number << std::endl;
 	coord::tile pos = unit->location->pos.start;
 	file << pos.ne << " " << pos.se << std::endl;
@@ -27,7 +28,7 @@ void load_unit(std::ifstream &file, openage::GameMain *game) {
 	file >> ne;
 	file >> se;
 
-	auto p = game->datamanager.get_producer(pr_id);
+	auto p = game->datamanager.get_type(pr_id);
 	game->placed_units.new_unit(*p, game->players[player_no], coord::tile{ne, se}.to_phys2().to_phys3());
 }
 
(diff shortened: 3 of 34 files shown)


cpp/unit/unit.h
```diff
@@ -70,6 +71,25 @@ class Unit : public log::LogSource {
 	 */
 	graphic_set *graphics;
 
+	/**
+	 * constructs a new location for this unit
+     *
```
TheJJ: hmm

Jon0:
The discussed rename of Producer to UnitType has been completed here

cpp/unit/action.cpp
```diff
@@ -506,12 +581,19 @@ void BuildAction::update(unsigned int time) {
 			this->complete = 1.0;
 		}
 	}
-	else {
+	else if (producer) {
```
TheJJ: i thought producers were removed?^^

cpp/unit/action.cpp
```diff
@@ -681,19 +773,29 @@ void AttackAction::update(unsigned int time) {
 	else if (distance_to_target <= this->radius) {
 		strike_percent = 1.0f;
 		this->frame = 0.0f;
-		this->attack(*this->target.get());
+		this->attack(*target_ptr);
+		this->allow_move = true;
 	}
-	else {
+	else if (allow_move) {
```
TheJJ: `this->`

Jon0:
Just tested, this fixes https://github.com/SFTtech/openage/issues/274

TheJJ:
The code is looking alright, I haven't tested it yet though. Awsum werk!

TheJJ:
Kay, just tested it. The only thing i don't like about it is that all units are doing an idle task now. Can you add some key that enables the "idle actions" for selected units?

Non-activated units will do nothing but explicit commands that way, only activated units automatically start doing things.

Jon0:
Do you jst want units to not move in construct mode? maybe its better to disable unit updates in construct mode.

The only other way would have to make a SearchAbility, which is button activated - but no one  wants to manually activate that for every unit while playing?

Jon0:
I will add the stand ground stance key -- But this will still allow ranged actions when in range, just units wont move.

TheJJ:
What i wanted to change is that all units now always start doing something on their own, which is what i want to prevent.
I think engine ticks in construct mode are ok for now, someday later we probably want to disable it.

I think ideally, all units are "active" somehow to react to things, but don't start attacking or building without any user command.

cpp/datamanager.h
```diff
@@ -88,12 +88,12 @@ class DataManager {
 	/**
 	 * producers by aoe unit ids -- the producer which corresponds to an aoe unit id
```
Davidslv: is this comment still correct?

Jon0:
Should be unit types

cpp/datamanager.h
Jon0: Should be unit types

Jon0:
Done a bit more refactoring to unit actions, units now defualt to a do nothing stance.

Jon0:
Thats a strange build failure...

TheJJ:
The job manager test seems to have deadlocked, never seen this before. I restarted the build...

Jon0:
Rebase done, should be mergable, with task search turned off for now, this is purely a bug fix branch.

Jon0:
Everything is fixed now, review/pull

cpp/game_main.cpp
```diff
@@ -346,16 +346,27 @@ bool GameMain::on_input(SDL_Event *e) {
...
+					auto container = &this->placed_units;
+					auto building_type = this->datamanager.get_type_index(this->editor_current_building);
+					auto new_building = container->new_unit(*building_type, *player, mousepos_phys3);
+
+					// task all selected villagers to build
+					if (new_building.is_valid()) {
+						Command cmd(*player, new_building.get());
+						cmd.set_ability(ability_type::build);
+						this->selection.all_invoke(cmd);
+					}
```
mic-e: a trailing whitespace... the end is near!

Jon0: http://imgur.com/AN2LC7o

cpp/game_main.cpp
```diff
@@ -346,16 +346,27 @@ bool GameMain::on_input(SDL_Event *e) {
 			if (this->building_placement) {
 
 				// confirm building placement with left click
-				Command cmd(this->players[0], this->datamanager.get_producer_index(this->editor_current_building), mousepos_phys3);
-				cmd.set_ability(ability_type::build);
-				selection.all_invoke(cmd);
+				// first create foundation using the producer
+				auto player = this->selection.owner();
```
mic-e: I really don't like this exaggerated use of `auto`; it takes away the code's self-documenting properties. For example, to check whether this line doesn't actually call a copy constructor, I have to to open `game_main.h` to find out what `this->selection` is, then open `unit/selection.h` to find out what `UnitSelection.owner()` returns...

cpp/game_main.cpp
mic-e: http://pub.sft.mx/mic_e/trailingspace.png

cpp/game_main.cpp
```diff
@@ -408,7 +419,7 @@ bool GameMain::on_input(SDL_Event *e) {
 			} else if ( this->datamanager.producer_count() > 0 ) {
 				// try creating a unit
 				log::log(MSG(dbg) << "create unit with producer id " << this->editor_current_building);
-				UnitProducer &producer = *this->datamanager.get_producer_index(this->editor_current_building);
+				auto &producer = *this->datamanager.get_type_index(this->editor_current_building);
```
mic-e: pls no auto... it leads to write-only code

cpp/game_save.cpp
```diff
@@ -27,7 +28,7 @@ void load_unit(std::ifstream &file, openage::GameMain *game) {
 	file >> ne;
 	file >> se;
 
-	auto p = game->datamanager.get_producer(pr_id);
+	auto p = game->datamanager.get_type(pr_id);
```
mic-e: the same thing goes here... don't use 'auto', and rename the var.

cpp/player.cpp
```diff
@@ -24,4 +25,11 @@ bool Player::is_ally(const Player &other) const {
 	return player_number == other.player_number;
 }
```
mic-e: can there be owner-less units?

cpp/player.h
```diff
@@ -24,9 +26,21 @@ class Player {
...
 	bool is_enemy(const Player &) const;
+
+	/**
+	 * the specified player is an ally of this player
+	 */
 	bool is_ally(const Player &) const;
 
+	/**
+	 * this player owns the specified unit
+	 */
```
mic-e: even more trailing whitespace!

cpp/game_main.cpp
franciscod: http://sublimetexttips.com/show-whitespace-sublime-text/

You can control when Sublime shows white space with the draw_white_space setting. The default value is selection, but you can override that with all to always show whitespace or none to never show it.

To change your whitespace preferences, run Preferences: Settings – User in the command palette, then add the draw_white_space setting:

{
    "draw_white_space": "all"
}

detrumi: The draw_white_space setting is kinda noisy, so this is another (possibly better) option:
`"trim_trailing_white_space_on_save": true`

mic-e:
Building foundations can now overlap... is this intentional?

cpp/terrain/terrain_search.cpp
```diff
@@ -0,0 +1,80 @@
...
+namespace openage {
+
+std::shared_ptr<TerrainObject> find_near(const TerrainObject &start, 
+                                         std::function<bool(const TerrainObject &)> found) {
+	
+	auto terrain = start.get_terrain();
+	auto tile = start.pos.draw.to_tile3().to_tile();
+	TerrainSearch search(terrain, tile);
+
+	for (int i = 0; i < 500; ++i) {
```
mic-e: why 500?

cpp/terrain/terrain_search.cpp
```diff
@@ -0,0 +1,80 @@
...
+		for (auto o : terrain->get_data(tile)->obj) {
+			auto shared_o = o.lock();
```
mic-e: Does `lock()` always succeed (can we be sure the `weak_ptr` has not expired?)

If yes, write a comment for stuff like that. If no, do a check (unless `found` must do the check... but the non-existing doc for `find_near` doesn't say so)

cpp/terrain/terrain_object.h
```diff
@@ -41,6 +45,8 @@ struct tile_range {
  */
 std::vector<coord::tile> tile_list(const tile_range &rng);
```
mic-e: doc please

cpp/terrain/terrain_search.cpp
```diff
@@ -0,0 +1,80 @@
...
+		}
+		tile = search.next_tile();
+	}
+
+	return nullptr;
+}	
+
+TerrainSearch::TerrainSearch(std::shared_ptr<Terrain> t, coord::tile s)
+	:
+	TerrainSearch(t, s, .0f) {
```
mic-e: not an important issue, but I think the closing brace should be in the same line.

cpp/terrain/terrain_search.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+TerrainSearch::TerrainSearch(std::shared_ptr<Terrain> t, coord::tile s, float radius)
+	:
+	terrain{t},
+	start(s),
+	previous_radius{.0f},
+	max_radius{radius} {
+}
+
+TerrainSearch::~TerrainSearch() {}
```
mic-e: why not just set the destructor to `= default`?

cpp/terrain/terrain_search.h
```diff
@@ -0,0 +1,71 @@
...
+#include <vector>
+
+#include "../coord/tile.h"
+
+namespace openage {
+
+class Terrain;
+class TerrainObject;
+
+std::shared_ptr<TerrainObject> find_near(const TerrainObject &start,
```
mic-e: docs please

cpp/terrain/terrain_search.h
```diff
@@ -0,0 +1,71 @@
...
+class TerrainObject;
+
+std::shared_ptr<TerrainObject> find_near(const TerrainObject &start, 
+                                         std::function<bool(const TerrainObject &)> found);
+
+constexpr coord::tile_delta const neigh_tile[] = {
+	{0, 1},
+	{0,  -1},
+	{1,  0},
+	{-1,  0}
```
mic-e: if you're gonna use a completely irregular number of spaces, at least make it so that stuff is aligned.

cpp/terrain/terrain_search.h
```diff
@@ -0,0 +1,71 @@
...
+};
+
+/**
+ * searches outward from a point and returns nearby objects
```
mic-e: that sounds like the doc for a function, not a class. either this should be a function, or the doc should say something about why the class has persistent state/multiple members...

cpp/unit/ability.h
```diff
@@ -42,17 +43,14 @@ using ability_set = std::bitset<16>;
 using ability_id_t = unsigned int;
 
 /**
- * all bits set to 1
+ * set bits corresponding to abilities
  */
-const ability_set ability_all = ability_set().set();
+ability_set from_list(const std::vector<ability_type> &items);
```
mic-e: so, there's a method `::openage::from_list` which returns `std::bitset<16>`... maybe the method should have a bit more... `ability_set`-specific name. like `to_ability_set`

cpp/unit/ability.h
```diff
@@ -42,17 +43,14 @@ using ability_set = std::bitset<16>;
 using ability_id_t = unsigned int;
```
mic-e: how do you know that 16 is large enough for `ability_set`? you should add a member `ability_type::MAX`, and use that. also, ability_type should be an `enum class`. As is, we have int constants like `::openage::move` flying around in the global namespace.

cpp/unit/ability.h
```diff
@@ -78,6 +76,15 @@ class UnitAbility {
 	 * applys command to a given unit
```
mic-e: should be 'applies'
*grammar nazi*

cpp/unit/action.cpp
```diff
@@ -79,6 +96,130 @@ void UnitAction::move_to(Unit &target, bool use_range) {
 	this->entity->invoke(cmd);
 }
 
+TargetAction::TargetAction(Unit *u, graphic_type gt, UnitReference r, coord::phys_t rad)
+	:
+	UnitAction(u, gt),
+	target{r},
+	repath_attempts{10},
+	end_action{false},
+	radius{rad} {
```
mic-e: please add a newline here

cpp/unit/action.cpp
```diff
@@ -17,21 +18,37 @@ namespace openage {
...
+		auto &attack = u->get_attribute<attr_type::attack>();
+		range += attack.range;
+	}
+	return range;
 }
 
-UnitAction::UnitAction(Unit *u, graphic_type initial_gt, coord::phys_t action_radius)
+UnitAction::UnitAction(Unit *u, graphic_type initial_gt)
 	:
 	entity{u},
```
mic-e: btw, why `entity` instead of `unit`?

cpp/unit/action.cpp
```diff
@@ -79,6 +96,130 @@ void UnitAction::move_to(Unit &target, bool use_range) {
...
+	TargetAction(u, gt, r, adjacent_range(u)) {
+}
+
+void TargetAction::update(unsigned int time) {
+	auto target_ptr = update_distance();
+	if (!target_ptr) {
+		return;
+	}
+
+	if (this->entity->has_attribute(attr_type::gatherer)) {
```
mic-e: is there ever a use case that doesn't fall in the scheme `if (has_attribute) { get_attribute() }`? IMO it should simply be `get_attribute()` (returning a pointer or `nullptr`).

cpp/unit/action.cpp
```diff
@@ -79,6 +96,130 @@ void UnitAction::move_to(Unit &target, bool use_range) {
...
+	radius{rad} {
+	this->update_distance();
+}
+
+TargetAction::TargetAction(Unit *u, graphic_type gt, UnitReference r)
+	:
+	TargetAction(u, gt, r, adjacent_range(u)) {
+}
+
+void TargetAction::update(unsigned int time) {
```
mic-e: According to `UnitAction`'s docs, "each action has its own update functionality".

Please summarize the kind of update functionality that TargetAction has (either in the class doc, the doc for `TargetAction::update`, or in a comment block here in the impl).

cpp/unit/action.cpp
```diff
@@ -79,6 +96,130 @@ void UnitAction::move_to(Unit &target, bool use_range) {
...
+		this->repath_attempts -= 1;
+	}
+	else {
+
+		// unit is stuck
+		this->end_action = true;
+	}
+}
+
+void TargetAction::on_completion() {}
```
mic-e: `on_completion` seems to be empty rather often. Perhaps it would be best to make it empty by default in the base class?

cpp/terrain/terrain_object.h
```diff
@@ -81,9 +87,19 @@ class TerrainObject : public std::enable_shared_from_this<TerrainObject> {
 	Unit &unit;
 
 	/**
+	 * is the object a floating outline
+	 */
+	bool is_floating() const;
```
mic-e: when would an object be floating (more docs please, I have no idea what you're talking about here)

cpp/unit/attribute.h
```diff
@@ -67,6 +67,13 @@ enum class attr_type {
 	garrison
 };
 
+enum class attack_stance {
+	aggresive,
+	devensive,
+	stand_ground,
+	do_nothing
```
mic-e: (called 'passive' in AoC)

cpp/unit/attribute.h
```diff
@@ -84,6 +91,8 @@ class AttributeContainer {
 		type{t} {}
 
 	attr_type type;
+
+	virtual AttributeContainer *copy() const = 0;
```
mic-e: Why are we even using this complicated `AttributeContainer`/`attr_type` model instead of simply making attributes polymorphic?

cpp/unit/attribute.h
```diff
@@ -127,22 +144,28 @@ template<> class Attribute<attr_type::hitpoints>: public AttributeContainer {
...
 
+	AttributeContainer *copy() const override {
+		return new Attribute<attr_type::attack>(*this);
+	}
+
 	// TODO: can a unit have multiple attacks such as villagers hunting
 	// map target classes onto attacks
 
-	UnitProducer *pp; // projectile producer
+	UnitType *pp; // projectile type
```
mic-e: rename the var to `ptype`?

cpp/unit/producer.cpp
```diff
@@ -209,7 +207,7 @@ void ObjectProducer::initialise(Unit *unit, Player &player) {
 				), 
 				true);
 		}
-		else {
+		else if (this->graphics.count(graphic_type::dying) > 0) {
```
mic-e: you should probably use `.find()` instead of `.count() > 0`... they do the same, but `.count()` can seem a bit... wtf-inducing.

cpp/unit/producer.cpp
```diff
@@ -242,11 +239,15 @@ std::shared_ptr<TerrainObject> ObjectProducer::place(Unit *u, Terrain &terrain,
 	 * currently unit avoids water and tiles with another unit
 	 * this function should be true if pos is a valid position of the object
 	 */
-	obj->passable = [obj, &terrain, terrains](const coord::phys3 &pos) -> bool {
+	std::weak_ptr<TerrainObject> obj_ptr = u->location;
+	std::weak_ptr<Terrain> terrain_ptr = terrain;
+	u->location->passable = [obj_ptr, terrain_ptr, terrains](const coord::phys3 &pos) -> bool {
+		auto obj = obj_ptr.lock();
```
mic-e: I've noticed this in serveral places, but I'm gonna comment only here.
Everywhere where you lock a weak pointer, you should check whether the pointer was still valid, or write a commend why you're sure it can't have been expired. Everything else means segfaults are looming on the horizon.

cpp/unit/producer.cpp
```diff
@@ -519,49 +521,51 @@ void BuldingProducer::initialise(Unit *unit, Player &player) {
...
-			obj->draw_outline();
+			obj_ptr.lock()->draw_outline();
 		}
 		u->draw();
 	};
 
 	// try to place the obj, it knows best whether it will fit.
-	bool obj_placed = obj->place(&terrain, init_pos);
-	if (!obj_placed) {
+	auto state = object_state::floating; //collisions? object_state::placed : object_state::placed_no_collision;
```
mic-e: what does that comment mean?

cpp/unit/unit.cpp
```diff
@@ -106,16 +107,20 @@ void Unit::update_secondary(int64_t time_elapsed) {
 }
 
 void Unit::draw() {
+	this->draw(this->location, this->graphics);
```
mic-e: what if `this->location` is empty?

mic-e:
maybe you should explicitly delete the copy and move constructors/assignment operators for `TerrainObject`. Also, does `Unit::location` even need to be a `shared_ptr`? I think the ownership is pretty explicit. Unless you need `weak_ptr` functionality, a `unique_ptr` should do.

cpp/unit/unit.h
```diff
@@ -43,9 +43,9 @@ class Unit : public log::LogSource {
 	const id_t id;
 
 	/**
-	 * producer used to create this object
+	 * type which was used to create this object
```
mic-e: is the type really used to "create" the object? can't a unit's type actually change on the fly?

cpp/unit/unit.h
```diff
@@ -70,6 +71,41 @@ class Unit : public log::LogSource {
 	 */
 	graphic_set *graphics;
 
+	/**
+	 * constructs a new location for this unit replacing any
+	 * existing location
```
mic-e: wouldn't a name such as `set_location` be more appropriate then?

cpp/unit/unit.h
```diff
@@ -70,6 +71,41 @@ class Unit : public log::LogSource {
...
+	 *
+	 * uses same args as the location constructor
+	 * except the first which will filled automatically
+	 */
+	template<class T, typename ... Arg>
```
mic-e: is this variadic templatestuff really necessary? it seems pretty horrible/evil/making it hard to spot errors

cpp/unit/unit.h
```diff
@@ -70,6 +71,41 @@ class Unit : public log::LogSource {
...
+	/**
+	 * constructs a new annex location for this unit
+	 *
+	 * this does not replace the units location
+	 */
+	template<class T, typename ... Arg>
+	std::shared_ptr<T> make_location_annex(Arg ... args) {
+
+		// since Unit is a friend of the location
+		// make_shared will not work
```
mic-e: this pretty much sounds like a serious design flaw in C++11...

cpp/unit/unit.h
```diff
@@ -233,8 +275,10 @@ class Unit : public log::LogSource {
 	/**
 	 * erase from action specified by func to the end of the stack
 	 * all actions erased will have the on_complete function called
+	 *
+	 * popped actions on_complete function is ignored when run_completed is false
```
mic-e: please add doc for @param run_completed

cpp/unit/unit_container.cpp
```diff
@@ -101,15 +101,15 @@ UnitReference UnitContainer::new_unit(UnitProducer &producer,
...
 
-UnitReference UnitContainer::new_unit(UnitProducer &producer, 
+UnitReference UnitContainer::new_unit(UnitType &type, 
                                       Player &owner, 
                                       std::shared_ptr<TerrainObject> other) {
 	auto newobj = std::make_unique<Unit>(*this, next_new_id++);
 
 	// try placing unit
-	auto placed = producer.place_beside(newobj.get(), other);
+	auto placed = type.place_beside(newobj.get(), other);
```
mic-e: remove auto plox

cpp/unit/unit_type.h
```diff
@@ -0,0 +1,120 @@
...
+class Terrain;
+class TerrainObject;
+class Unit;
+class UnitAbility;
+class UnitContainer;
+class UnitTexture;
+
+/**
+ * Initializes a unit with the required attributes, each unit type should implement these funcrtions
+ * initialise should be called on construction of units 'new Unit(some_unit_producer)'
```
mic-e: Look, mom! I've spotted a producer!

mic-e:
Looks like I'm through; looking forward to your replies/fixes.

franciscod:
that was mad reviewing mic-e!!!

mic-e:
Yup, everybody who subscribes to openage just got flooded with 100 mails.

cpp/player.cpp
Jon0: It may happen in the future

mic-e: better play it safe, then.

cpp/unit/ability.h
Jon0: The value should be the size of of the enum class ability_type, not sure how to get the size of it here...

cpp/unit/ability.h
Jon0: The value should be the size of of the enum class ability_type, not sure how to get the size of it here...

mic-e: Usually you do it by adding a final member, `::MAX`, to the enum

enum class foo {
   val,
   bar=20,
   rofl,
   test,
   MAX
};

foo_size = foo::MAX

cpp/unit/ability.h
Jon0: the only use is:
```c++
ability_set as = from_list({ability_type::attack});
```

cpp/unit/action.cpp
Jon0: its is just a todo for now,
gathers go to idle when its resource runs out, they will have a find new resource function in the n_completion in future

mic-e: mind adding a TODO comment then?

cpp/unit/attribute.h
Jon0: the container is mostly hidden, this was the only way I could get the templated get_attribute of units to work.

cpp/unit/producer.cpp
Jon0: if location is deleted, then so is this lambda, (deleting terrain implies location is deleted), so this is safe.

cpp/unit/producer.cpp
Jon0: Its not mine, I kept it from the original placement function
It means place() does the checks as to whether placement is allowed

cpp/unit/unit.cpp
Jon0: The draw function checks it

cpp/unit/attribute.h
mic-e: this is what I had in mind:

    template<typename T>
    T *get_attribute() {
        std::type_info *map_key = &typeid(T)
    
        auto map_idx = attribute_map.find(map_key);
    
        if (map_idx == std::map::end) {
            return nullptr;
        } else {
            return reinterpret_cast<T *>(*map_idx);
        }
    }

cpp/unit/ability.h
mic-e: yeah, but `from_list` still pollutes the global namespace, and its name doesn't indicate in any way that it's specific to ability sets.

cpp/unit/producer.cpp
mic-e: I see. Please add a comment that says so, then.

cpp/unit/producer.cpp
mic-e: No matter whose fault it is, it needs fixing :P

cpp/unit/unit.h
Jon0: It just copies the params of the specific location type eg SquareLocation

cpp/unit/unit.h
Jon0: Yes, calling type.initialise(unit, player) on an already typed unit will reset the unit to the new type

cpp/unit/unit.cpp
mic-e: Uhm, doesn't the second line in `Unit::draw` say `auto &draw_pos = loc->pos.draw;`?

cpp/unit/unit.h
mic-e: Yes, but IMO that doesn't make it less evil

cpp/unit/ability.h
Jon0: I can make it static for ability

cpp/unit/unit.h
mic-e: Well I guess it's needed because `T` is templated. BTW you should add a `static_assert` that checks that `T` is a subclass of `TerrainObject`

cpp/unit/unit.cpp
Jon0: Ill add the check then, I thought it was there... It wouldnt have any affect atm as draw calls originate from terrain (to get the draw order) so a drawn unit always has a location

mic-e: if you're worried about the performance impact or anything, just `#include "util/compiler.h"` add `unlikely()` around the check condition; that should make the check pretty much zero-cost.

cpp/unit/ability.h
mic-e: yeah, that would be a good idea.

Jon0:
Fixed - the attribute thing is worth an issue as they may not be needed here at all, and then another pull request to change.

mic-e:
I agree about the need for #299.

Things are starting to look pretty mergable; however, some of my issues remain unadressed:

 - Please add doc for `building_center` in `terrain_object.h`
 - My comment "maybe you should explicitly delete the copy and move constructors and assignment operators for TerrainObject. Also, does Unit::location even need to be a shared_ptr? I think the ownership is pretty explicit. Unless you need weak_ptr functionality, a unique_ptr should do."
 - Building foundations can overlap: http://pub.sft.mx/mic_e/openage_overlapping_foundations.png

In addition, I'm getting a compiler warning in `terrain_search.cpp`:

    terrain_search.cpp:19:20: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
      for (int i = 0; i < search_limit; ++i) {

Jon0:
All should be fixed now, refactoring Unit::location to use unique_ptr implied that TileContent objects be changed from weak_ptr to raw pointers.

In AoC floating outlines (the first stage of a building site) can overlap, as 2 players can place a floating outline in the same place, however they will only see their own or allied outlines. Here we are rendering all players, so foundations are seen to overlap. Once a builder reaches the outline and construction all other floating outlines are removed (the builder who starts building first get the placement) I have fixed the building behaviour to do this now.

mic-e:
Nice. :+1: (after squashing those style fix commits)

mic-e:
Wheee!

PR #269, opened 2015-04-05
Ignore empty gamedata.drs

niklasf:
`gamedata.drs` does not exist in HD edition, but the Voobly compability patch
adds an empty file. So also check that the file is not empty.


goto-bus-stop:
:+1:

goto-bus-stop:
Oh, you should also add yourself to the [copying.md](https://github.com/SFTtech/openage/blob/master/copying.md) file for Travis to accept it :)

TheJJ:
nice, well done :+1:

PR #275, opened 2015-04-12
Documented usage

dbrgn:
Couldn't find this when trying the game :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,10 @@ Current State of the Project
 
 See [status.md](status.md).
 
+ - What can I do once I start the game?
+
+See [docs/usage.md](doc/usage.md).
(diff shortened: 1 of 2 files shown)


charlesrgould:
Thanks, I was looking for this. Didn't know about the backtick. :)

franciscod:
:+1: :D

TheJJ:
awesome. we probably can remove status.md and create a real changelog instead now that we have this guide.

dbrgn:
> Thanks, I was looking for this. Didn't know about the backtick. :)

Fun thing is, the backtick doesn't work on a Swiss keyboard, because backtick is a deadkey by default :)

PR #276, opened 2015-04-12
Add a CMake 'doc' target that explains why doc generation won't work.

charlesrgould:
#### ./configure (without Doxygen installed)

Some warnings, but you can proceed.

```
$ ./configure --mode=debug --compiler=gnu
...
-- Could NOT find Doxygen (missing:  DOXYGEN_EXECUTABLE) 
CMake Warning at buildsystem/doxygen.cmake:49 (message):
  doxygen couldn't be found, you won't be able to generate docs
  CMakeLists.txt:64 (doxygen_configure)
...
```

#### make doc

now:

```
$ make doc
make --no-print-directory -C bin doc
When you configured openage, Doxygen could not be found.
Install Doxygen and `./configure` openage again, then try to generate docs.
Built target doc
$ 
```

before:

```
$ make doc
make --no-print-directory -C bin doc
make[1]: *** No rule to make target 'doc'.  Stop.
Makefile:73: recipe for target 'doc' failed
make: *** [doc] Error 2
$ 
```

### DIFF
--- a/buildsystem/doxygen.cmake
+++ b/buildsystem/doxygen.cmake
@@ -1,4 +1,4 @@
-# Copyright 2014-2014 the openage authors. See copying.md for legal info.
+# Copyright 2014-2015 the openage authors. See copying.md for legal info.
 
 # Doxygen integration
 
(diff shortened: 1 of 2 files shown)


TheJJ:
thx for the changes!
Please update the copyright year in the `doxygen.cmake` file, the build will pass then.
Also, the email you did the commit with is not associated with your github account.

Either add that mail to github or `git commit --amend --author="Your Name <yourknown@mail.lol>"` to change the mail for the commit.

TheJJ:
maybe you could add a `COMMAND false` to the target so the invokation will fail and not display `Built target doc`.

charlesrgould:
Before merging...

I couldn't test the "COMMAND false" bit because DOXYGEN_FOUND was being set to true even when I moved the doxygen command out of my path. Any ideas?

http://www.cmake.org/cmake/help/v2.8.12/cmake.html#module:FindDoxygen

TheJJ:
hm, probably `FindDoxygen` uses pkgconfig or something, and as you still had Doxygen installed it assumed some path to be exitent. Anyway :+1: for merge.

mic-e:
For cross-platform compatibility, it's best to use the commands provided by `cmake -E` (e.g. `cmake -E echo`). Unfortunately, `cmake -E` doesn't provide a command for `false` :|

PR #278, opened 2015-04-14
Doc cleanup

detrumi:
Half of the URLs in the `doc` folder gave 404 errors because they added a relative `doc/` while already in the `doc` folder.

To keep things consistent (and to avoid this error in the future), I changed such URLs to a absolute path like `/doc/usage.md`.


mic-e:
I didn't even know about the leading-slash thing... :+1:

franciscod:
nice!

Issue #271, opened 2015-04-12
Customizable key bindings


TheJJ:
Maybe store it as key-value pairs in `~/.openage/keybindings`.

This file should have a config version on top once the format gets more sophisticated to perform/detect updates.

Ideally, the bindings are managed in it's own subsystem (`cpp/keys`?) to allow central storage and changes.
The keys should probably point to functions that can be replaced at runtime. Avoid string lookups please. 

depends on the cvar config subsystem from #486

detrumi:
I'll start working on this.

mic-e:
I guess after #284, only the "read-keybindings-from-file-or-whatever" part remains to be implemented.

Hoten:
I would like to begin contributing to this project, and I think this is a good way to start. I'll work on reading these key bindings from a configuration file.

StepS-:
You might want to cooperate with @Elnabo then as I believe he already had plans for it in the meantime.

elnabo:
The keybinds are now part of the generic config file(s). The config file reading should be done in python to be easily integrated with the cvar system (#486).

So I think you should focus on extracting any kind of couple key/value and if possible have a possibility to save modification.

elnabo:
I think that  #537 has completed this issue, isn't it ?

TheJJ:
YES! YES IT HAS!!!!


:beers:

PR #284, opened 2015-04-18
Customizable key bindings

detrumi:
I've started working on customizable key bindings, but would like some feedback on the design.
The key bindings are structured as follows: 
- There's a `action_t` enum which lists all bindable actions.
- You can bind a `std::function` object to such actions. I used lambda expressions to replace the original switch statement.
- You'll be able to point from keys to actions. This is currently hardcoded, but will be saved in a file. 

There's still some open questions:
- Currently everything is for `SDL_KEYUP` events. It would be easy to extend this to also incluse `SDL_KEYDOWN` events, but do we need this? (we'll need key modifiers like control as well)
- The meaning of the current `y`, `z`, `x` and `c` key bindings is unclear. What do they do? It would be nice to document this, or remove them if they're not needed.

Implements the basics for #271


Jon0:
I was considering making a HotkeyContext class, which maps keys to functions, depending on which ui menu you are currently looking at. That would be most useful for future development.

forgot to document some keys:
Y = uses the current building (upper right) to place a foundation when a villager is selected
Z = only move commands are accepted from right clicking, which means right click will not issue an attack / gather / ...
X = only gather commands ...
C = all commands (default)
 There could potentially be a key for every ability, this is useful for things like patrol/guard/follow which may otherwise be identical to move commands

detrumi:
Actually, couldn't we change `Keybinds` to something like `KeybindInputHandler`, which would be a subclass of `InputHandler`? We're already using the handlers as some sort of context, if I understand it correctly. I'm only not sure how this would interact with `CoreInputHandler` though.

TheJJ:
(@detrumi I think you need to rebase on the current master to fix the merge conflict in `copying.md`)
Jon's "mapping per view" is a great idea, it just extends your (also great) key -> function mapping to change dynamically.

The `CoreInputHandler` can be optimized out that way, because each mode (e.g. control, placement) would have a different handler, but we'd also need some handler to be active for both (e.g. the profiling key works in both modes).

So we need multiple dynamically activated handlers binding functions to keys, and any number of currently active ones.

detrumi:
I've finished work on the keybind contexts. When you open the ingame console, keybinds like 'space' (for toggling tile blending) stop working, while global keybindings (for example, 'escape' for exiting the game) still work. If you were to bind 'escape' again in the console, it'd take priority over the global one. You can also add keybinds while retaining old ones.

The part with writing the keybinds from and to a file is quite tricky:
- The `SDL_Keycode` can be mapped to a string, but not from one. We could just save the integers, but that hurts the readability of the file (if we care about that)
- Mapping the actions, which are implemented as an enum, to and from a string is annoying in c++. We can either write big switch statements (annoying to update) or use some [preprocessor magic](http://stackoverflow.com/questions/147267/easy-way-to-use-variables-of-enum-types-as-string-in-c#202511) to generate it.

Mapping the keybinds to a file would add quite some complexity, so it might be better to wait until we're working on the gui for changing keybinds. The separation of (key->action) and (action->function) together with the contexts might be good enough for now.

cpp/game_main.cpp
```diff
@@ -241,6 +242,74 @@ GameMain::GameMain(Engine *engine)
...
+		this->use_set_ability = false;
+	});
+	keybind_context.bind(keybinds::action_t::SET_ABILITY_MOVE, [this]() {
+		this->use_set_ability = true;
+		this->ability = ability_type::move;
+	});
+	keybind_context.bind(keybinds::action_t::SET_ABILITY_GATHER, [this]() {
+		this->use_set_ability = true;
+		this->ability = ability_type::gather;
+	});
```
TheJJ: can you add `this->keybind_context` pls? It's visible directly that it's a member variable that way.

cpp/keybinds/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+add_sources(${PROJECT_NAME}
+    action.cpp
+    keybind_context.cpp
+    key.cpp
+    keybind_manager.cpp
```
TheJJ: one tab indent pls

cpp/keybinds/action.cpp
```diff
@@ -0,0 +1,15 @@
...
+
+#include <functional>
+
+#include "action.h"
+
+namespace openage {
+namespace keybinds {
+
+int action_hash::operator()(const action_t& a) const {
+    return std::hash<int>()(static_cast<int>(a));
```
TheJJ: tab indent pls

cpp/keybinds/action.h
```diff
@@ -0,0 +1,36 @@
...
+    TRAIN_OBJECT,
+    ENABLE_BUILDING_PLACEMENT,
+    DISABLE_SET_ABILITY,
+    SET_ABILITY_MOVE,
+    SET_ABILITY_GATHER,
+    TOGGLE_CONSOLE
+};
+
+struct action_hash {
+    int operator()(const action_t& a) const;
```
TheJJ: tab indent here and in the enum class above pls

cpp/keybinds/key.cpp
```diff
@@ -0,0 +1,35 @@
...
+#include "key.h"
+
+namespace openage {
+namespace keybinds {
+
+key_t::key_t(SDL_Keycode key)
+    :
+    key(key),
+    mod(KMOD_NONE)
+{
```
TheJJ: tab indent again. For the initializer list init and more hints about the codestyle we're using (because of consistency and beautyfication reasons) please have a look at this: https://github.com/SFTtech/openage/blob/master/doc/code_style/mom.cpp#L131

all your new files are space indented :wink:

detrumi:
I've fixed the tab and style issues, and the tabs for new files should now be configured correctly for next time. I'll make sure to check, though.

cpp/console/console.h
```diff
@@ -34,6 +35,8 @@ class Console : InputHandler, TickHandler, HudHandler, ResizeHandler {
 	Buf buf;
 	Font font;
 
+	keybinds::KeybindContext keybindContext;
```
TheJJ: sorry that i noticed this that late, please name the variable `keybind_context`, but the class name is perfect :wink:

cpp/game_main.cpp
```diff
@@ -345,7 +414,7 @@ bool GameMain::on_input(SDL_Event *e) {
 			break;
 		}
 		else if (not scrolling_active
-		         and e->button.button == SDL_BUTTON_MIDDLE) {
+				 and e->button.button == SDL_BUTTON_MIDDLE) {
```
TheJJ: this was perfect before, have a look at https://github.com/SFTtech/openage/blob/master/doc/code_style/tabs_n_spaces.md for it :grinning:

cpp/keybinds/keybind_context.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+}
+
+void KeybindContext::bind(const action_t a, const std::function<void()> f) {
+	binds[a] = f;
+}
+
+bool KeybindContext::execute_if_bound(action_t a) {
+	auto f = binds.find(a);
+	if (f != binds.end()) {
```
TheJJ: `this->binds` would be nice

cpp/keybinds/keybind_manager.cpp
```diff
@@ -0,0 +1,99 @@
...
+		}
+	}
+
+	// If no local keybinds were bound, check the global keybinds
+	globalHotkeys.execute_if_bound(action);
+}
+
+
+void KeybindManager::set_key_state(SDL_Keycode k, bool is_down) {
+	key_states[k] = is_down;
```
TheJJ: `this->` for all the lines in that file pls

TheJJ:
I like the design, this is probably the right approach. Sometimes later, we might even mix this with the GUI system, or create some kind of "action" subsystem, which automatically assigns the binds and context to different active areas on the screen. This also captures the caret for text inputs and activates other keybindings then.

detrumi:
I've fixed all style issues.

I kind of wanted to generalize the context approach, but wasn't able to find a good general solution yet. But let's leave that for later, when we actually need it.

TheJJ:
Dis lookin gud nao, :+1:

cpp/game_main.cpp
```diff
@@ -400,96 +469,27 @@ bool GameMain::on_input(SDL_Event *e) {
...
-			this->ctrl_active = true;
-			break;
-		}
+	}
+
+	case SDL_KEYDOWN: {
+		this->keymod = SDL_GetModState();
 
+		SDL_Keycode sym = reinterpret_cast<SDL_KeyboardEvent *>(e)->keysym.sym;
+		engine.get_keybind_manager().set_key_state(sym, true);
```
franciscod: i'm late for this as the PR is already merged, should the game release all the keys when the window loses focus?

cpp/keybinds/keybind_manager.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+#include "keybind_manager.h"
+
+namespace openage {
+namespace keybinds {
+
+KeybindManager::KeybindManager()
+	: // TODO get this from a file instead of hardcoding it here
+	keys {{ key_t(SDLK_ESCAPE), action_t::STOP_GAME },
+		{ key_t(SDLK_F1), action_t::TOGGLE_HUD },
```
franciscod: @TheJJ did you let this one slip? 

gnazi

detrumi:
@franciscod I can't get the keys stuck when I steal focus while holding a key down, so I think SDL generates an event for the key_up when the window loses focus, or something like that.

We found an issue though where the control key stops working when num lock or caps lock is active, so I'll update this branch with some fixes later.

franciscod:
awesome about the focus one :) the hardcoding bit was more for the styling that JJ was once very strict

PR #291, opened 2015-04-26
Villager commands

detrumi:
Adding functionality to villagers. Also fixing bugs while I find them.

## Implemented features:
- [x] Place a villager with the `v` key
- [x] Kill a selected unit with `delete`
- [x] Hide some UI elements when in `Actions mode`
- [x] Villagers can build some buildings with the `q`, `w`, `e` and `r` keys (with `control` key for military buildings)
- [x] Set current player with 1-8 keys, create units with correct color
- [x] You can select multiple of your own units or buildings, but only 1 enemy at a time

## Found/fixed bugs:
- [x] Most click commands should be on mouse-up, not mouse-down
- [x] Dead units can be selected
- [x] Villagers can build another player's buildings
- [x] Building preview image is always from player 1
- [ ] Some editor building previews are always in blue, regardless of the current player
- [x] Enemy units can be controlled

## Fixed by Jon0 in another branch:
- Villagers work in place when building can't be placed (something in the way)
- When building with multiple villagers, only one of them starts building, the others work in place
- Villager building animation sometimes is choppy/slow
- Deleting a building while it's being built sometimes crashes the game (or makes the building reappear)
- Villagers sometimes invisible while in attacking animation

## Won't fix bugs:
Small bugs that probably won't be fixed any time soon
- Game freezes a bit when moving units, especially when clicking outside of the map (pathfinding just slow?)
- Buildings can be placed on water (terrain doesn't have enough information for this yet, I think)


Jon0:
Where you added has_villager property to selection, it would be safer to check the units class == CIVILIAN, rather than type 590, since there are many villager types and that is only 1.
```c++
if (it.second.is_valid() && it.second.get()->unit_class == gamedata::unit_classes::CIVILIAN) {
...
```

detrumi:
The way I did it was a quick hack, I'm aware of that. But isn't it better to check for `ability_type::build`?

Jon0:
checking for build ability works too, it might be better actually...

I have already fixed the first 5 unticked bugs on your list on my branch

detrumi:
Nice! I'll be careful not to fix the same things you're fixing then.

franciscod:
you 2 should share a branch :)

detrumi:
I've implemented most of the things I wanted to implemented, so this is more or less finished. See the topmost comment for a overview of what was added/fixed.

cpp/unit/selection.cpp
```diff
@@ -103,17 +104,75 @@ void UnitSelection::clear() {
...
+	this->units.erase(u->id);
+
+	if (this->units.empty()) {
+		this->selection_type = selection_type_t::nothing;
+	}
+}
+
+void UnitSelection::kill_unit() {
+	if (!this->units.empty()) {
+		Unit *u = this->units.begin()->second.get();
```
Jon0: There needs to be an is_valid() check, if the first unit in the vector gets killed, the unit is deleted, but the reference would remain in the selection -- this would throw an exception on get()

selections do not get updated and invalid units removed until a new selection is made.

Jon0:
The town center has id 109 if you want to add it -- it can be difficult to find in the editor list as it only get partially drawn there.

detrumi:
@Jon0: I didn't wanted to clutter the hotkeys, as we use several keys for other things already.

cpp/game_main.cpp
```diff
@@ -307,6 +307,50 @@ GameMain::GameMain(Engine *engine)
...
+				} else {
+					this->editor_current_building = building;
+				}
+			}
+		});
+	};
+	bind_building_key(keybinds::action_t::BUILDING_1, 598, 609); // House, barracks
+	bind_building_key(keybinds::action_t::BUILDING_2, 574, 558); // Mill, archery range
+	bind_building_key(keybinds::action_t::BUILDING_3, 616, 581); // Mining camp, stable
+	bind_building_key(keybinds::action_t::BUILDING_4, 611, 580); // Lumber camp, siege workshop
```
TheJJ: do you think we should add the towncenter here?

cpp/game_main.cpp
```diff
@@ -341,115 +406,97 @@ bool GameMain::on_input(SDL_Event *e) {
...
+					// delete any unit on the tile
+					if (!chunk->get_data(mousepos_tile)->obj.empty()) {
+						// get first object currently standing at the clicked position
+						TerrainObject *obj = chunk->get_data(mousepos_tile)->obj[0];
+						log::log(MSG(dbg) << "delete unit with unit id " << obj->unit.id);
+						obj->unit.delete_unit();
+					} else if ( this->datamanager.producer_count() > 0 ) {
+						// try creating a unit
+						log::log(MSG(dbg) << "create unit with producer id " << this->editor_current_building);
+						UnitType &producer = *this->datamanager.get_type_index(this->editor_current_building);
```
TheJJ: no more producers ;)

cpp/unit/selection.cpp
```diff
@@ -103,17 +104,92 @@ void UnitSelection::clear() {
...
+			u->delete_unit();
+		}
+	}
+}
+
+bool UnitSelection::contains_villagers() {
+	for (auto &it : units) {
+		if (it.second.is_valid() && it.second.get()->get_ability(ability_type::build)) {
+			return true;
+		}
```
TheJJ: sure that you want to depend "contains villavers" on the build ability? maybe should be `contains_builders`?

cpp/unit/unit_texture.h
```diff
@@ -58,7 +58,7 @@ class UnitTexture {
 	/**
 	 * a sample drawing for hud
 	 */
-	void sample(const coord::camhud &draw_pos) const;
+	void sample(const coord::camhud &draw_pos, unsigned color = 1) const;
```
TheJJ: `color=1` without spaces pls

TheJJ:
The code looks alright, many new bindings and actions, awesome! :+1: 
Somebody™ should playtest it before merging :wink:

cpp/game_main.cpp
detrumi: I was about to say the `z` key was already taken, but instead I'll just move the old `z`, `x` and `c` binds to `ctrl+z` etc., as they're just debugging-related hotkeys anyways.

Also, @Jon0 said town center has id 109, but instead I get moss with the debug message `setting unit type 335 TERRN`. So this doesn't seem to work by setting `editor_current_building` to 109.

cpp/game_main.cpp
detrumi: Did you just mean the `producer` variable on line 482? Or is something else about it wrong as well? I'm not sure, because I moved these lines without changing them.

cpp/game_main.cpp
Jon0: Oh the editor list doesnt use the type ids (to ensure the list is continuous) -- I found the tc at editor list value 568

detrumi: The id is correct this time, but the building is still broken. Does #304 fix this?

Jon0: Yes it needs the fix to work

detrumi:
Some play-testing would be nice! I'm also testing it myself of course, but it's hard to think of everything. For example, I just noticed you could control enemy units.

The town center is broken for now, but I'll leave the keybind in as it'll be fixed when #304 is merged.

Jon0:
Yea I play tested, It matches the selection behavior of AoC, just need to be sure that is what we want -- such as only being able to select 1 enemy at a time, when its really useful to select many (to see hp). This is kind of like how we dont want to restrict selection to 40 units.

I noticed the editor items get hidden for the action mode, but I normally use that view for training units with 't' and building with 'y', additionally it can be hidden with f1 anyway. Not sure what would be best here, it would eventually be better to hide since there would be menus for wat can be built/trained.

Jon0:
Made 2 cannon ships for 1 player, and another cannon ship to attack them, boxed selecting the ships while they are being attacked caused a segfault. Was able to recreate twice and this does not affect master. Maybe caused by a projectile being in the selection?

detrumi:
The way AoC handles selections is actually a bit strange. If you have multiple units selected, you can add a single enemy unit while holding control. Same with buildings.

I wanted to separate the two modes, so that you can't control+scroll while placing a building with a villager. I think it's best to only enable debug functions like `t` and `y` in construct mode.

TheJJ:
#304 is merged now, you can add the towncenter to the quick-build keys.

detrumi:
I've added the town center now (on the `z` button). The graphic is still weird when you have it selected/are placing it (only the left and right pillars are shown), but building and completing it works fine now.

Jon0:
The preview does not draw delta graphics at the moment

TheJJ:
awsum. :+1:

franciscod:
awesome!

PR #293, opened 2015-04-28
automatic conversion

asiniy:


TheJJ:
Hm, don't they do that already? I mean its their attack, and they should perform it when they have to defend themselves. The recharge is still an issue, so the continuation of converting will take a lot of time. Am I right?

detrumi:
I think he meant something like an aggressive stance for monks, so that they attack even when they're not attacked themselves.

asiniy:
> I think he meant something like an aggressive stance for monks, so that they attack even when they're not attacked themselves.
Yep! Exactly this. Sorry for my english

mic-e:
I have merged this manually (and fixed the grammar). Next time, please use a single pull request for all your ideas (PRs can simply be updated).

asiniy:
@mic-e I just think about novative things, sometimes I get interesting ideas.

mic-e:
@asinly: Yeah, thanks for the pull request, but if you already have an "idea" pull request open, just update that one instead of creating a new pull request for every single idea.

PR #298, opened 2015-05-04
Fixed keybind modifiers

detrumi:
Keybinds (#284) were influenced by keys like caps lock and num lock, and stopped working if one of those keys was active. That's fixed now.

I've also moved the modifier stuff from `GameMain` to `KeybindManager`, to keep things organized.


TheJJ:
looking good etc :+1:

PR #300, opened 2015-05-06
visual fixes for dock/ships

Jon0:
fix for docks making ground invisible, and ships becoming invisible when moiving


detrumi:
That seems to fix it :+1:

PR #304, opened 2015-05-09
Fix annexed buildings

Jon0:
Objects like the TC were not able to be built as the foundation got removed instantly (floating outlines for buildings get removed when a building begins construction, here the annex caused the TC to remove itself)

The TC has several stages wrt collisions, the floating stage causes no collisions, the construction stage allows no units on the foundation, and once completed construction parts of the tc are able to be walked over, this is now working.

Since the change to unique_ptr<TerrainObject> in Unit, the annex objects are not cleaned up (memleak), The solution is to allow TerrainObjects to also be owned by another TerrainObject, the make annex function is moved from Unit to TerrainObject. Deletion of the parent object will now cleanup all children.

### DIFF
--- a/cpp/terrain/terrain_object.cpp
+++ b/cpp/terrain/terrain_object.cpp
@@ -69,13 +69,21 @@ bool TerrainObject::place(object_state init_state) {
 		}
 
 		for (auto obj : chunk->get_data(temp_pos)->obj) {
+
+			// ignore self and annexes of self
 			if (obj != this &&
-				obj->is_floating()) {
-				to_remove.push_back(obj);
-			}
-			else if (obj != this &&
-				     obj->check_collisions()) {
-				return false;
+				obj->get_parent() != this) {
+
+				if (obj->is_floating()) {
+
+					// floating objects get removed
+					to_remove.push_back(obj);
+				}
+				else if (obj->check_collisions()) {
+
+					// solid objects obstruct placement
+					return false;
+				}
 			}
 		}
 	}
--- a/cpp/terrain/terrain_object.cpp
+++ b/cpp/terrain/terrain_object.cpp
@@ -103,8 +103,9 @@ bool TerrainObject::place(object_state init_state) {
 			}
 		}
 	}
+
+	// all obstructing objects get deleted
 	for (auto remove_obj : to_remove) {
-		this->unit.log(MSG(dbg) << "Removing...");
 		remove_obj->unit.location = nullptr;
 	}
 
--- a/cpp/terrain/terrain_object.cpp
+++ b/cpp/terrain/terrain_object.cpp
@@ -24,7 +24,8 @@ TerrainObject::TerrainObject(Unit &u)
 	passable{[](const coord::phys3 &) -> bool {return true;}},
 	draw{[]() {}},
 	state{object_state::removed},
-	occupied_chunk_count{0} {
+	occupied_chunk_count{0},
+	parent{nullptr} {
 }
 
 TerrainObject::~TerrainObject() {
@@ -103,6 +104,7 @@ bool TerrainObject::place(object_state init_state) {
 		}
 	}
 	for (auto remove_obj : to_remove) {
+		this->unit.log(MSG(dbg) << "Removing...");
 		remove_obj->unit.location = nullptr;
 	}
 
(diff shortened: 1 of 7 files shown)


cpp/terrain/terrain_object.h
```diff
@@ -108,12 +108,15 @@ class TerrainObject : public std::enable_shared_from_this<TerrainObject> {
 	bool is_floating() const;
 
 	/**
-	 * has the object been placed
+	 * returns true if this object has been placed. this indicates that the object has a position and exists
+	 * on the map, floating buildings are not considered placed as they are only an indicator
+	 * for where something can begin construction
```
mic-e: are both `is_floating` and `is_replaced` required?

cpp/unit/producer.cpp
```diff
@@ -444,7 +444,8 @@ BuldingProducer::BuldingProducer(DataManager &dm, const gamedata::unit_building
 	trainable1{dm.get_type(83)}, // 83 = m villager
 	trainable2{dm.get_type(293)}, // 293 = f villager
 	projectile{dm.get_type(this->unit_data.projectile_unit_id)},
-	foundation_terrain{ud->terrain_id} {
+	foundation_terrain{ud->terrain_id},
+	enable_collisions{this->unit_data.id0 == 109} {
```
mic-e: what's `109`? add a comment

cpp/terrain/terrain_object.cpp
```diff
@@ -184,17 +209,16 @@ void TerrainObject::set_ground(int id, int additional) {
...
 	return this->parent;
 }
 
 std::vector<TerrainObject *> TerrainObject::get_children() const {
-	return this->children;
+	std::vector<TerrainObject *> result;
+	for (auto &obj: this->children) {
+		result.push_back(obj.get());
+	}
+	return result;
```
mic-e: constructing and returning a `std::vector` is not very good for performance, obviously... in the long term we should do something about that (such as accepting a callback and passing the children to that).

detrumi: @mic-e: That's not really true. Here's a [stackoverflow](http://stackoverflow.com/questions/3134831/in-c-is-it-still-bad-practice-to-return-a-vector-from-a-function) question stating otherwise. It's kind of complex, but you can imagine that the compiler changes `return result` to `return std::move(result)`.

Jon0: I was thinking the same thing, maybe a function which accepts a lambda

mic-e: RVO might ease the pain, but there's still the unnecessary `malloc` and copying of all values into the new vector.

cpp/terrain/terrain_object.h
Jon0: you mean is_placed()? -- is_floating implies not is_placed, but not is_floating can be either removed or is_placed.

mic-e: k.
maybe you'll want to document the list of all possible states.
while you're at it, just create an enum for all possible states, and add a function that returns the current state... maybe `is_placed` and `is_floating` would both get obsoleted by that?

cpp/terrain/terrain_object.cpp
mic-e: @Jon0: That's what I was thinking about, too.

The problem is that this does not really seem to be an established pattern in C/C++; I wonder why. After all, C/C++ usually are quite nazi when it comes to performance.

cpp/terrain/terrain_object.h
Jon0: there is an enum, but there is more than one state that will make those function true.

mic-e: Ah, I see.

Jon0:
Added and fixed some docs to explain exerything

cpp/terrain/terrain_object.cpp
```diff
@@ -69,17 +85,26 @@ bool TerrainObject::place(object_state init_state) {
...
+				else if (obj->check_collisions()) {
+
+					// solid objects obstruct placement
+					return false;
+				}
 			}
 		}
 	}
 	for (auto remove_obj : to_remove) {
+		this->unit.log(MSG(dbg) << "Removing...");
```
TheJJ: maybe add a message what is being removed.

Jon0: Removed as it is longer needed -- TerrainObject cleanup will log something at this point anyway

cpp/unit/producer.cpp
```diff
@@ -13,6 +13,14 @@
 #include "unit.h"
 #include "unit_texture.h"
 
+/**
+ * Many values in this file are hardcoded, due to limited understanding of how the original
```
TheJJ: please write
```
/** @file
 * bla
 */
```
otherwise this comment will annotate the `openage` namespace :wink:

Issue #217, opened 2015-02-07
Python API


TheJJ:
To allow calls from C++ to Python and vice versa, we need an awesome interface.

@mic-e started working on a nice example here: https://github.com/mic-e/cythondemo

PR #179, opened 2014-12-17
[Orphaned] perlin noise generator

zuntrax:
C++ rewrite in progress.
* [x] C++ rewrite of Perlin class
* [ ] documentation
* [ ] test case or demo for Perlin class
* [x] delete python implementation

Should be useful for #65.

These were generated with the python implementation, the new ones will look different.
Seed: "sftworld1", 500x500px, 10x10 grid cells, polynomial interpolation:
perlin_sftworld1

Red: "sftworld0", green: "sftworld1", blue: "sftworld2", polynomial interpolation, dims as above:
multicolor

PR #305, opened 2015-05-11
[WIP] Map Engine

timohaas:
# Map Engine

moved map creation from the main loop to a new subsystem.
Created Interface for different map engines.
Implemented Engines:
* static: returns Grass map
* mandelbrot: returns Island of the shape of the mandelbrot set
* diamond square
* perline noise: true infinite random map possible


detrumi:
It would've been nice if this was communicated better, as I've merged part of this PR into [one of my branches](https://github.com/detrumi/openage/tree/random-maps) (by your request). I'm fine with you working on it or taking it over.

On the perlin noise: someone already has implemented [a prototype](https://github.com/SFTtech/openage/pull/179) for this, though I'm not sure what the status on that is, and whether it's easy to call the python code from cpp code (this is also being worked on, but that's not finished yet).

TheJJ:
Did you have a look at #179? It implements the noise generator in python, the API (#217) is close to being finished, so you can use the python implementation for the mapgen or even implement the mapgen in python itself (as it's not a time-critical operation).

I'd rather not add `libnoise` as a dependency, it doesn't seem to be widely available (at least missing in gentoo) and is very outdated (last commit 9 years ago..).

I like your modular approach, the code needs better integration though (we have an awesome logger e.g.). Keep up the good work, much potential is there, we'll come up with some nice ideas.

timohaas:
@detrumi I'm sorry, I should have said something, but @TheJJ stated (https://github.com/detrumi/openage/pull/1#issuecomment-100703842), I should use the Master Repo for Pull Requests.
How should I proceed from here? Pull request to detrumi (original author of the branch) or to master Repo?

I didn't noticed the python implementation. I will have a look into it and use it to reduce the dependency.

detrumi:
Go ahead with this PR to the master repo.

In my branch, I was just experimenting with the diamond-square algorithm, it's not finished yet. It's not flexible enough for map generation, though it may possibly be useful for generating the heightmap if we're implementing elevation.

zuntrax:
@detrumi I think #179 should work for producing perlin noise. @rkreis claims that it actually produces value noise, not perlin noise. I wanted to verify my code and discuss it with him, but I didn't get to do it yet.

rkreis:
No, I do believe it's perlin noise.  The noise in my test pictures is value noise, that's why it looks different from the in the PR.  I see nothing wrong with the noise generator the way it's in the PR.

franciscod:
mandelbrot!!!

timohaas:
Mandelbrot:
16 tiles per side
mandelbrot_16_tiles
32 tiles
mandelbrot_32_tiles
multiple 16 tiles per side chunks
mandelbrot_16_tiles_multiple_chunks
perlin: multiple 16 tiles per side chunks
perlin_16_tiles

franciscod:
that is awesome!!! :D

zuntrax:
@rkreis ah ok, then i just got your comment wrong.

TheJJ:
Before it's too late, i'm really don't like bundling libraries in our code. Using "well-known" and up2date libs shipped with distros as new dependencies is much better. If no matching lib is available, we should (e.g. like the rng) implement stuff in "our way" so we have a simple-stupid awesome implementation written and maintained by the project itself.
That means, the best solution may be to either reimplement the perlin noise py code we already have in c++ or try to come up with some api that will then be provided by the py-api.

My suggestion: write the code in c++ and I'll replace it with the python part when the api is ready.

timohaas:
You're right, this is an intermediate solution. I want a working travis ci build, so i can go on to the next working one.

mic-e:
@timohaas: It should be rather simple; just have a look at `.travis.yml`; it's rather self-explaining. You'll probably simply need to add `libnoise-dev` or something like that to the install list.

timohaas:
It's building now (Commit 478d227). I added the needed Header and Source Files directly to the Source Code. 

TheJJ  suggested, we implement our own Perlin Noise. I will look into this if needed, but would rather use the http://libnoise.sourceforge.net/ lilbrary. 
Debian has the lib in its repo (https://packages.debian.org/jessie/libnoise0).

TheJJ:
The python interface is complete now, you can use parts the noise implementation commits from #179 i hope.

zuntrax:
I'm the one working on the perlin prototype. I'll try to get this productive asap.

PR #309, opened 2015-05-28
add test for log::FileSink

kukinsula:
I managed to test log messages through a FileSink: messages are written in /tmp/openage-log-test (maybe I should change that file name because there's already /tmp/openage-log and it can get ambiguous). So to see the messages you just need to read that file (cat /tmp/openage-log-test).

I set the FileSink's loglevel to err to ignore messages with a lower loglevel and it works fine.

### DIFF
--- a/cpp/log/test.cpp
+++ b/cpp/log/test.cpp
@@ -1,6 +1,8 @@
 // Copyright 2014-2015 the openage authors. See copying.md for legal info.
 
 #include "log.h"
+#include "file_logsink.h"
+#include "level.h"
 
 #include <iostream>
 #include <iomanip>
@@ -34,8 +36,10 @@ private:
 	}
 };
 
-
-void test() {
+/**
+ * Tests log Messages through the standard output.
+ */
+void std_out_log_sink () {
 	TestLogSource logger;
 	TestLogSink sink{std::cout};
 
@@ -59,6 +63,33 @@ void test() {
 
 	t0.join();
 	t1.join();
+}
+
+/**
+ * Creates a FileSink to also write log Messages
+ * into /tmp/openage-log-test.
+ * Only Messages with a level equals or higher
+ * than level::err will be printed in that file.
+ */
+void file_log_sink () {
+	TestLogSource logger;
+	FileSink filelog("/tmp/openage-log-test", true);
+
+	filelog.loglevel = log::level::err;
+
+	// filelog should ignore these Messages
+	logger.log(MSG(dbg) << "This message should not appear in openage-log-test");
+	logger.log(MSG(info) << "This message should not appear in openage-log-test");
+	logger.log(MSG(warn) << "This message should not appear in openage-log-test");
+
+	// filelog should write these Messages
+	logger.log(MSG(err) << "This message should appear in openage-log-test");
+	logger.log(MSG(crit) << "This message should appear in openage-log-test");
+}
+
+void test() {
+	std_out_log_sink();
+	file_log_sink();
 }
 
 }}} // openage::log::tests
--- a/cpp/log/test.cpp
+++ b/cpp/log/test.cpp
@@ -9,6 +9,7 @@
 #include <thread>
 #include <sstream>
 #include <string>
+#include <algorithm>
 
 #include "../util/strings.h"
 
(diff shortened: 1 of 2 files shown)


TheJJ:
Thanks!

A few comments:
* The commit email is not associated with github. Either update the email in the commit or register it to github please.
* You didn't add yourself to `copying.md`
* You do a good job writing to that filesink, but do not verify its expected contents.
* You could do more tests that will check wether only the assigned log levels are written to that file
* Please remove the file after testing, we don't wanna clutter the filesystem :wink:

kukinsula:
Hi TheJJ !

I corrected what you told me to fix, I hope this is ok now.

To verify that only relevant messages are written in the log file I read that file (line by line) looking for "ERR" or "CRIT", is it ok or were you thinking some other way ?

I also added another test: I log messages through a FileSink with a crit log level, then read the log file which should be empty.

One last question, should I ensure that all unit tests don't return -1 to state wether it passed or failed (like other subsystems tests) ?

cpp/log/test.cpp
```diff
@@ -61,4 +68,76 @@ void test() {
...
 }
 
+/**
+ * Creates a FileSink to also write log Messages
+ * into /tmp/openage-log-test.
+ * Only Messages with a level equals or higher
+ * than level::err will be printed in that file.
+ */
+void file_log_sink () {
+	const char *filelog_name = "/tmp/openage-log-test";
```
mic-e: hard-coding the filename to `/tmp` is sub-optimal; this might not work on OSX and will not work on Win32. Use `mkstemp` instead.

cpp/log/test.cpp
```diff
@@ -61,4 +68,76 @@ void test() {
...
+	logger.log(MSG(dbg) << "This message should not appear in openage-log-test");
+	logger.log(MSG(info) << "This message should not appear in openage-log-test");
+	logger.log(MSG(warn) << "This message should not appear in openage-log-test");
+
+	// filelog should write these Messages
+	logger.log(MSG(err) << "This message should appear in openage-log-test");
+	logger.log(MSG(crit) << "This message should appear in openage-log-test");
+
+	std::string line;
+	std::ifstream infile(filelog_name);
```
mic-e: I think you should delete the FileSink object before opening the file for reading - who knows what sort of buffering might be happening in the background.

Also, I think Win32 doesn't allow you to open a file that has already been opened for writing...

cpp/log/test.cpp
```diff
@@ -61,4 +68,76 @@ void test() {
...
+/**
+ * Creates a FileSink to also write log Messages
+ * into /tmp/openage-log-test.
+ * Only Messages with a level equals or higher
+ * than level::err will be printed in that file.
+ */
+void file_log_sink () {
+	const char *filelog_name = "/tmp/openage-log-test";
+	TestLogSource logger;
+	FileSink filelog(filelog_name, true);
```
mic-e: You'll want to pass `false` here. If you pass `true`, it will append to the existing logfile, which may lead to wrong results.

cpp/log/test.cpp
kukinsula: Thank you, I'll take the time (soon) to change that.

cpp/log/test.cpp
kukinsula: Will do this too then (since I delete the file at the end of the test it doesn't lead to wrong results right now).

cpp/log/test.cpp
StepS-: @mic-e it does allow opening the file but it needs to have the "share" access. Depending on how the file was opened for writing (with or without access sharing for reading or writing), the file may or may not be accessible for reading.

cpp/log/test.cpp
kukinsula: Hey @mic-e ! I am trying to use mkstemp to create a temporary file (it has a random name /tmp/file.XXXXXX) but I can't find a way to retrieve that random name so I can give it to the FileSink. What were you exactly thinking about mkstemp ? or maybe I am missing something.

detrumi: There's an example on mkstemp usage [here](http://www.thegeekstuff.com/2012/06/c-temporary-files/).

mic-e: Sorry, looks like `mstemp` is _not_ the function you're looking for; you'll want to use `tmpnam` instead.

I guess I got irritated by the man page's warning to never use `tmpnam`, and to use `mkstemp` instead. Which is bullshit in our case.

cpp/log/test.cpp
```diff
@@ -59,6 +64,104 @@ void test() {
...
+
+	// filelog should write these Messages
+	logger.log(MSG(err) << "This message should appear in openage-log-test");
+	logger.log(MSG(crit) << "This message should appear in openage-log-test");
+
+	std::string line;
+	std::ifstream infile(filelog_name);
+
+	while (std::getline(infile, line)) {
+		if ((line.find("ERR") == std::string::npos) && (line.find("CRIT") == std::string::npos)) {
```
TheJJ: i'd recommend searching for `"should not appear"`, because this is the essential part to test, not the line prefix which may change any time.

cpp/log/test.cpp
TheJJ: If you don't want to close the file, you should at least flush the buffer to write out anything cached.

cpp/log/test.cpp
```diff
@@ -59,6 +64,104 @@ void test() {
...
+	return result;
+}
+
+
+/**
+ * Tests that the FileSink doesn't write any
+ * message into the the filelog because it
+ * has a too low log level.
+ */
+int empty_file_log_sink () {
```
TheJJ: i'm not sure whether this test does any new coverage, the do-we-output-for-some-level test is already done above.

cpp/log/test.cpp
```diff
@@ -59,6 +64,104 @@ void test() {
...
+ * Creates a FileSink to also write log Messages
+ * into /tmp/openage-log-test.
+ * Only Messages with a level equals or higher
+ * than level::err will be printed in that file.
+ */
+int file_log_sink () {
+	char filelog_name[L_tmpnam];
+	int result = 0;
+
+	tmpnam(filelog_name);
```
TheJJ: You might want to use http://linux.die.net/man/3/mkstemp.
It returns the file descriptor directly, instead of just a name, which is way easier in handling and doesn't require the `filelog_name` variable.

TheJJ: If you want/must stick to generating a filename only, you should use at least use [tempnam](http://linux.die.net/man/3/tempnam) instead.

TheJJ:
please update this test to the new project layout and testing system.

TheJJ:
If you wanna continue this, feel free to request a reopen.

PR #310, opened 2015-05-30
Made existing testcases test JobManager with atleast 10 jobs in parallel  #189

sreejithr:
Did the following things:
* `test_simple_jobs` tests the JobManager by giving it 10 jobs in parallel and checking if it succeeds.
* `test_simple_jobs_with_exception` tests by giving JobManager 5 normal jobs and 5 jobs with throw exceptions.

### DIFF
--- a/cpp/job/tests.cpp
+++ b/cpp/job/tests.cpp
@@ -3,6 +3,8 @@
 #include "job_manager.h"
 #include "../log/log.h"
 
+#include <thread>
+
 namespace openage {
 namespace job {
 namespace tests {
@@ -12,7 +14,8 @@ int test_simple_job() {
 	JobManager manager{4};
 	manager.start();
 
-	bool finished = false;
+	std::atomic<int> finish_count(0);
+	int job_count = 10;
 	bool result = -1;
 
 	auto job_function = []() -> int {
@@ -23,12 +26,14 @@ int test_simple_job() {
 		if (job_result == 42) {
 			result = 0;
 		}
-		finished = true;
-
+		finish_count++;
 	};
-	manager.enqueue<int>(job_function, job_callback);
 
-	while (not finished) {
+	for (int i = 0; i < job_count; i++) {
+		manager.enqueue<int>(job_function, job_callback);
+	}
+
+	while (finish_count.load() < job_count) {
 		manager.execute_callbacks();
 	}
 
@@ -41,30 +46,46 @@ int test_simple_job_with_exception() {
 	JobManager manager{4};
 	manager.start();
 
-	bool finished = false;
-	bool result = -1;
+	int good_jobs = 5;
+	int bad_jobs = 5;
+	std::atomic<int> finished(0);
+	std::atomic<int> errors(0);
 
-	auto job_function = []() -> int {
+	auto exception_job_function = []() -> int {
 		throw "error string";
 	};
+
+	auto job_function = []() -> int {
+		return 42;
+	};
+
 	auto job_callback = [&](result_function_t<int> get_result) {
 		try {
 			get_result();
 		} catch (const char *e) {
-			result = 0;
+			errors++;
 		}
-		finished = true;
-
+		finished++;
 	};
-	manager.enqueue<int>(job_function, job_callback);
 
-	while (not finished) {
+	for (int i = 0; i < good_jobs; i++) {
+		manager.enqueue<int>(job_function, job_callback);
+	}
+
+	for (int i = 0; i < bad_jobs; i++) {
+		manager.enqueue<int>(exception_job_function, job_callback);
+	}
+
+	while (finished.load() < good_jobs + bad_jobs) {
 		manager.execute_callbacks();
 	}
 
 	manager.stop();
-	return result;
 
+	if (errors.load() == bad_jobs) {
+		return 0;
+	}
+	return -1;
 }
 
 
--- a/copying.md
+++ b/copying.md
@@ -52,6 +52,7 @@ _the openage authors_ are:
 | Niklas Fiekas               | niklasf                     | niklas.fiekas@tu-clausthal.de         |
 | Charles Gould               | charlesrgould               | charles.r.gould@gmail.com             |
 | Wilco Kusee                 | detrumi                     | wilcokusee@gmail.com                  |
+| Sreejith R                  | sreejithr                   | sreejith.r44@gmail.com                |
 
 If you're a first-time commiter, add yourself to the above list. This is not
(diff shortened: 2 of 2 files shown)


mic-e:
Your contribution has quite a few whitespace errors; you should either disable your IDE's auto-formatting, or configure it to be more in line with openage's style guidelines.

sreejithr:
@mic-e Sorry man. My emacs started acting weird after installing helm-mode. I've fixed it :)

TheJJ:
Now all the tabs are spaces, please adhere to [our codestyle](https://github.com/SFTtech/openage/blob/7061e8bf1e54df177981415cd242f2d1dbf69ef1/doc/code_style/mom.cpp) :grin:

cpp/job/tests.cpp
```diff
@@ -3,85 +3,108 @@
...
-	return result;
+    JobManager manager{4};
+    manager.start();
+
+    std::atomic<int> finish_count(0);
+    int job_count = 10;
+    bool result = -1;
+
+    auto job_function = []() -> int {
+        std::this_thread::sleep_for(std::chrono::seconds(1));
```
TheJJ: sleeping that long is probably a bad idea, we're running the test on our ci server, and waiting for long times doesn't speed up its run.

sreejithr:
@TheJJ Fixed that. I come from the snake world (python). Spaces = indent is 2nd nature to me :) I hope my code is compliant now.

cpp/job/tests.cpp
```diff
@@ -3,33 +3,40 @@
...
 	JobManager manager{4};
 	manager.start();
 
-	bool finished = false;
+	std::atomic<int> finish_count(0);
+	int job_count = 10;
 	bool result = -1;
 
 	auto job_function = []() -> int {
+		std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
TheJJ: nice you lowered it. but thinking about it, what is the reason you need to add any delay here?

cpp/job/tests.cpp
```diff
@@ -37,38 +44,54 @@ int test_simple_job() {
...
-		finished = true;
-
+		finished++;
 	};
-	manager.enqueue<int>(job_function, job_callback);
 
-	while (not finished) {
+	for (int i = 0; i < good_jobs; i++) {
+		manager.enqueue<int>(job_function, job_callback);
+	}
```
TheJJ: why did you add the non-excepting jobs here? they should be tested in the other function?

cpp/job/tests.cpp
sreejithr: Just to simulate a long-running task.

sreejithr:
@TheJJ About mixing non-excepting jobs with excepting ones, I did that to make sure the non-excepting ones do get completed even if there's excepting ones in the queue. What do you think?

sreejithr:
I'm make a test case for abortable jobs. I spawn 20 jobs and stop the JobManager in between (at ~12jobs). Each job has a delay of 1sec. I wait for 20 sec (to make sure that if aborting didn't work, all jobs will get completed). After this, I check how many got finished. Hopefully, <20 jobs would be finished which implies aborting worked.

I know it's not elegant. But I can't find a better solution since there's no way to know whether a job is aborted or if it's still in progress (since aborting doesn't set any flags anywhere)

cpp/job/tests.cpp
```diff
@@ -3,33 +3,40 @@
...
-		if (job_result == 42) {
-			result = 0;
-		}
-		finished = true;
-
+	int job_result = get_result();
+	if (job_result == 42) {
+		result = 0;
+	}
+	finish_count++;
```
TheJJ: these lines are in `{}`, therefore need one indentation level.

cpp/job/tests.cpp
```diff
@@ -3,33 +3,40 @@
...
-	manager.enqueue<int>(job_function, job_callback);
 
-	while (not finished) {
-		manager.execute_callbacks();
+	for (int i = 0; i < job_count; i++) {
+		manager.enqueue<int>(job_function, job_callback);
+	}
+
+	while (finish_count.load() < job_count) {
+	manager.execute_callbacks();
```
TheJJ: indent this line please.

TheJJ:
Blocking the CI build should be avoided, so no timeouts please. You may want to squash your commits and meld them together so all your fixes are included in a single commit then.

Maybe you're able to add storing the exception flag to the job subsystem, then each job can be queried easily if it is in failed state (that reminds me, i think job states are already present).

sreejithr:
@TheJJ Ahh. I saw `std::exception_ptr exception` in the job state. But the `	std::shared_ptr<TypedJobStateBase<T>> state` is a private member in `Job`. So, I can't access the JobState.

TheJJ:
You can determine whether a job has failed by checking whether the job is finished and an exception is stored at the `this->state->exception`. Getting this member probably requires adding a `get_exception` function to `Job` in `job.h`, which just returns this variable.

If the job `is_finished()` and the exception ptr is non-nullptr, then it has thrown an exception.

sreejithr:
@TheJJ That'll do the trick. Will make a pull request separately for the job.h edit.

sreejithr:
@TheJJ I've squashed my commits into one.

sreejithr:
@TheJJ Jobs are aborted by throwing the `JobAbortedException`. But unlike other exceptions, this one is not logged into `this->exception` ((typed_job_state_base.h:69)[https://github.com/SFTtech/openage/blob/master/cpp/job/typed_job_state_base.h#L69]).

TheJJ:
Indeed, but you can determine whether the [default value of `T`](https://github.com/SFTtech/openage/blob/master/cpp/job/typed_job_state_base.h#L22) is different from the result value when the job has `finished = true`.

sreejithr:
@TheJJ Aborted jobs don't set `finished` flag. So, a job state still having the default value for `this->result` without the `finished` flag set, could either be an aborted job or a job in progress.

TheJJ:
Hmm, unsure what to do. I think you can skip testing aborted jobs for now. Maybe it's a bug that the job is not marked as finished when aborted, because otherwise one would assume it is running indefinitely.

sreejithr:
@TheJJ Yeah. It returns true if it aborted. But that is only accessible to the worker. Anyway, I'm skipping aborted jobs.

Is everything OK with this PR?

TheJJ:
Jup, looking alright. :+1:

sreejithr:
@TheJJ :+1:

PR #311, opened 2015-05-31
Job test

kukinsula:
Hi guys !

I made another test but this time for the job manager subsystem. My test (compare_two_job_managers) gives to two different job managers (the first one with 1 worker and the second one with 4) a significant quantity of job functions, then they separately run one after the other to measure their execution time. In the end, I just compare the two execution times expecting that the first one is slower.

### DIFF
--- a/copying.md
+++ b/copying.md
@@ -52,6 +52,7 @@ _the openage authors_ are:
 | Niklas Fiekas               | niklasf                     | niklas.fiekas@tu-clausthal.de         |
 | Charles Gould               | charlesrgould               | charles.r.gould@gmail.com             |
 | Wilco Kusee                 | detrumi                     | wilcokusee@gmail.com                  |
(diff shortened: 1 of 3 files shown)


fr3akout:
Actually this test will probably be flickering, because there is no guarantee, that the JobManager with 4 worker threads will be faster than the one with only 1 worker thread.

kukinsula:
I see, I haven't thought of that ... The test on my machine always pass but I guess it can change depending on many things.

TheJJ:
You pushed your logging-tests to this branch as well, we should keep those separate. You can just create a new branch from your current master and then cherry-pick the relevant commits to the job-test branch. Also, you should synchronize with #310.

TheJJ:
#310 is merged now, so this is either obsolete or needs to be updated.

TheJJ:
you may resubmit an updated version for the new project layout and testing system, but i think for now this patch is obsolete.

PR #316, opened 2015-06-18
show building outlines on selected buildings

Jon0:
currently the building outline does not appear on elected buildings


detrumi:
Looking good :+1:

PR #288, opened 2015-04-21
cython-based C++-/Python interface

mic-e:
- openage is now a shared library
 - moved `py/openage` -> `openage`
 - moved `py/openage/codecompliance` -> `buildsystem/codecompliance`
 - moved `cpp/` -> `libopenage/`
 - moved `doc/implementation/` -> `doc/code`
 - integrated Cython into buildsystem
 - C++ -> Py function calling
  - C++ API is specified in PXD annotations in `.h` files
  - cpp/pyinterface/pyfunc.h
   - Func<returntype, arg0type, arg1type, ...>
   - Callable objects
   - Designed to be filled with pointers to cdef functions
   - __don't use raw function pointers in the interface__
  - openage/cppinterface/setup.py
   - setup() Fills the PyFunc pointers
   - must be called before using libopenage functionality
 - Exception translation
  - cpp/pyinterface/exctranslate.h
   - translate_exc_cpp_to_py
   - translate_exc_py_to_cpp
  - py/cppinterface/exctranslate.py
   - describe_exception
   - raise_cpp_error
  - C++ functions that are declared 'except +' if they may throw
  - Python functions that are declared 'except *' if they may raise
  - GIL-safe
  - everything is translated back and forth
   - metadata
   - message
   - backtrace
 - Documentation

C++:

 - Overhauled exception code
  - now lives in `error/error.h`
  - exceptions automatically store their cause exception
  - exceptions can collect backtraces
  - see demo in `cpp/error/demo.cpp`
  - Hierarchies may now be used
  - `doc/code/exceptions`
 - Moved util/error.h to error/error
 - Added error/handlers.cpp, which provides a SIGSEGV and terminate handler
 - refactored LZXD code
  - removed many C-isms and micro-optimizations
  - simplified codeflow, split methods into smaller ones
  - file split into multiple smaller ones (`cpp/util/compress`)
  - `decompress` replaced by `decompress_frame`, which decompresses just
    32768 bytes at a time instead of everything at once.
  - exposed to Python using the PyInterface
 - Simplified C++ logging
  - ExternalStringStream: like stringstream, but appends to a custom buffer
 - Modernized C++ testing
  - now uses TestError (e.g. via the `TESTFAIL` macro) instead of return values.
 - C++ exceptions can collect backtraces
 - `SIGSEGV` and `std::terminate` handlers
 - Enum class that allows adding new members at runtime, and associated data.
  - used by `log::lvl`
[...]

PR #317, opened 2015-06-23
Multithreaded convert

janisozaur:
Use '-j' option to 'media' command to specify amount of jobs to use.
Unfortunately, make cannot pass its own '-j' value.

Down to roughly 2m53s from 10m31s on 4-core (+HT) i7, with HDD.
Verified sha1sums for singlethreaded and multithreaded conversion the same.
Log on console can be out of order now.

### DIFF
--- a/py/openage/convert/mediafile.py
+++ b/py/openage/convert/mediafile.py
@@ -9,7 +9,8 @@ import os
 import os.path
 import pickle
 import subprocess
-import multiprocessing as mp
+from multiprocessing import Pool
+import traceback
 
 from .. import util
 from .colortable import ColorTable, PlayerColorTable
@@ -136,9 +137,11 @@ def media_convert(args):
         args.extract.append('*:*.*')
 
     extraction_rules = [ExtractionRule(e) for e in args.extract]
-    if args.jobs == 0:
-        args.jobs = os.cpu_count()
-    pool = mp.Pool(processes=args.jobs)
+    if args.jobs is None:
+        threads = os.cpu_count()
+    else:
+        threads = args.jobs
+    pool = Pool(processes=threads)
     media_out = []
 
     dbg("age2 input directory: %s" % (args.srcdir,), 1)
@@ -299,8 +302,7 @@ def media_convert(args):
 
     file_list = defaultdict(lambda: list())
     media_files_extracted = 0
-    dbg("Converting data using %i threads, this may take a while" % args.jobs,
-        1)
+    dbg("Converting data using %i threads, this may take a while" % threads, 1)
 
     # iterate over all available files in the drs, check whether they should
     # be extracted
@@ -349,15 +351,20 @@ def media_convert(args):
             media_files_extracted += 1
 
     pool.close()
-    for i in media_out:
+    exc_count = 0
+    for entry in media_out:
         try:
-            i.wait()
+            entry.wait()
             # if anything was raised in child, it will be re-raised here
-            i.get()
+            entry.get()
         except:
-            dbg("got exception", 1)
+            exc_count += 1
+            traceback.print_exc()
     pool.join()
     dbg("Conversion done", 1)
+    if exc_count > 0:
+        raise Exception("There were %i exceptions during media conversion. "
+                        "Please check the logs." % exc_count)
 
     if write_enabled:
         dbg("media files extracted: %d" % (media_files_extracted), 0)
--- a/py/openage/convert/mediafile.py
+++ b/py/openage/convert/mediafile.py
@@ -9,6 +9,7 @@ import os
 import os.path
 import pickle
 import subprocess
+import multiprocessing as mp
 
 from .. import util
 from .colortable import ColorTable, PlayerColorTable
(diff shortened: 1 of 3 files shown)


franciscod:
nice!

janisozaur:
Hopefully should come in handy just in time for patch 4.0 conversion, whenever that happens.

detrumi:
Neat! Maybe you could add a note to `building.md` so people will know about this.

TheJJ:
Cool stuff! Do i see this right that this will fill up the queue with all jobs in the beginning and then process them?

The job order of course doesn't matter, it was randomized due to py3's nonteterministic sets anyway.

janisozaur:
`apply_async()` will start processing right away, doing load balancing (by queueing jobs into all workers) automatically, you just need to wait until all tasks return.

To prepare a queue and then batch-process it, you would use one of `Pool.*map()` functions (see https://docs.python.org/3/library/multiprocessing.html#module-multiprocessing.pool)

You could possibly squeeze more performance out of the code by `mmap`ing the DRS file, so the kernel can load the file into memory perhaps more efficiently, but not sure if it's worth it.

I also check for possible exceptions by calling `async_result.get()`

py/openage/convert/mediafile.py
```diff
@@ -77,6 +136,10 @@ def media_convert(args):
         args.extract.append('*:*.*')
 
     extraction_rules = [ExtractionRule(e) for e in args.extract]
+    if args.jobs == 0:
+        args.jobs = os.cpu_count()
```
TheJJ: You could set `default=os.cpu_count()` in the argparse definition. Setting it to 0 as default, dunno :)

py/openage/convert/mediafile.py
```diff
@@ -327,6 +348,17 @@ def media_convert(args):
...
             media_files_extracted += 1
 
+    pool.close()
+    for i in media_out:
+        try:
+            i.wait()
+            # if anything was raised in child, it will be re-raised here
+            i.get()
+        except:
+            dbg("got exception", 1)
```
TheJJ: Can you dump the rethrown exception here? Just `"got exception"` won't help debugging. Also we probably don't want to catch all exceptions here, not sure though.

py/openage/convert/mediafile.py
janisozaur: Could do, but I didn't want to add the `import os` in `__main__.py`. Also, in the first version, in this file the code was
```
if args.jobs == 0:
    pool = mp.Pool()
else:
    pool = mp.Pool(processes=args.jobs)
```
As `Pool`'s `processes` argument defaults to `os.cpu_count`, but it's not exposed later on and I wanted to let user know how many threads we use.

TheJJ: kay, then just leave it for now, although i don't see the problem with `import os` :)

py/openage/convert/mediafile.py
janisozaur: Should I just call `i.get()` without `try … except` block then?

mic-e:
Heh, this will be fun to integrate into my main pyinterface branch :)

py/openage/convert/mediafile.py
mic-e: Dump the exception with `traceback.print_exc()`.

py/openage/convert/mediafile.py
mic-e: Ideally, you'd do `default=None`, then `if args.jobs is None`.

py/openage/convert/mediafile.py
```diff
@@ -9,6 +9,7 @@
 import os.path
 import pickle
 import subprocess
+import multiprocessing as mp
```
mic-e: Better do `from multiprocessing import Pool`

py/openage/convert/mediafile.py
janisozaur: can I `default=None` with `type=int`? That's a cool trick.

janisozaur: If there was a way to retrieve thread count from Pool, I'd love to do that. That's not possible and I want to let user know we're using X threads for conversion, so I decided to use `threads` for that.

TheJJ:
i think this is good to go, we should integrate this into other pullrequests separately, the benefits do outweight the merge problems of others i think :)
:+1:

mic-e:
I disagree. Literally every single line of your pull request will conflict with #288, and the basic idea (*fork*-based multiprocessing) is unfeasible in programs that have multiple (non-python) threads and/or a GUI.

Thank you for your contribution; I will manually apply your contribution (preserving your authorship) on top of #288 once that is done *real soon now*™

franciscod:
GOOD COP VS BAD COP EVERYBODY

mic-e:
I have done the integration; see [this commit](https://github.com/mic-e/openage/commit/9315cdfbe1f31e15f0a43c4725b6173e4b93f125).

I have attributed you as author, but unfortunately I wasn't able to salvage a single line of your code. If you're unhappy with the fact that you're named as author, or think that some code should be changed, just tell me.

Issue #285, opened 2015-04-18
Create logo


TheJJ:
We really need some artwork.

* [x] Nice big logo/banner for the `README.md`, can be pixel or vector graphics (#319)
* [x] Square desktop icon, should be vector graphics (#319)
* [ ] Some fanart or whatever

franciscod:
https://www.reddit.com/r/linux_gaming/comments/2xy9tf/openage_devs_no_binary_blobs_community_svg_gaben/

JonasBorchelt:
😂👍

TheJJ:
some [inspiration](http://trebuxet.deviantart.com/gallery/52719403/Age-of-Empires) from http://trebuxet.deviantart.com/:
![medieval battle](https://cloud.githubusercontent.com/assets/673543/7361432/7d9e0238-ed58-11e4-96f1-ee9da03cea6f.jpg)
![how do you turn this on](https://cloud.githubusercontent.com/assets/673543/7361433/82595854-ed58-11e4-83ac-8136159fa71d.jpg)

franciscod:
Those are awesome! Let's ask him! :D

detrumi:
As for the icon/logo: some of the AOE2 icons are a helmet. How about an open helmet for openage?

franciscod:
the open helmet idea is awesome! ...what about gaben.svg with an open helmet?

detrumi:
frickler01 has made [some sketches](https://github.com/frickler01/openage-design) for us! Let's give him some feedback. You can find a overview of his ideas [here](https://github.com/frickler01/openage-design/blob/master/sketches/sketches.md).

For the helmets, I liked # 6 the most recognizable AOE-like helmet.
Most of the escutcheon images are a bit too busy, I'd prefer something simple like # 14 (though # 18 is awesome too!)
The omega/A idea is awesome! Maybe use the omega as a helmet? # 20 looks nice, but having 'age' in it instead of just the 'A' would make it more recognizable.

detrumi:
It would be nice to have a recognizable icon for openage. How do people feel about [this one](https://github.com/frickler01/openage-design/blob/master/svg/crown.svg)? It's already in `svg` format.

linkmauve:
If you decide to go for this one, I wiped the unneeded attributes and the embedded PNG here, for a nice 46KiB → 955B ratio, and that’s before compression, where gzip can achieve a 491B file: http://linkmauve.fr/files/crown.svg

[...]

PR #319, opened 2015-06-24
fancy new logo!

TheJJ:
yay! finally some logo!

@frickler01 please provide me your real name and some actual email so i can update the commit author to you.

btw i have no idea how the readme will display the logo, but we'll see that in a minute.


franciscod:
https://github.com/SFTtech/openage/issues/285#issuecomment-114790988

@TheJJ you might have wanted to mention @linkmauve instead of @frickler0

detrumi:
@franciscod: We mentioned @frickler01 because he's the one who made the logo. @linkmauve only changed a small bit of it.

Though I'm not sure it's a good thing to set the commit's name to someone else's.

franciscod:
contributors.md needs him too!

mic-e:
What about this?

boombatower:
logo++

TheJJ:
The problem is that github doesn't support rendering repo-relative svgs, why ever.

>

> Hi JJ,
> Thanks for getting in touch! There's nothing amiss with your markup -- It's not currently possible to render relative SVG images in Markdown on GitHub. I believe it's set up this way to block a potential security vulnerability, but I'll certainly let the team know you'd like to see relative SVG image rendering enabled in the future if possible. 

>

charlesrgould:
Try [RawGit](http://rawgit.com) to serve the banner?

Example:
https://cdn.rawgit.com/TheJJ/openage/98abfcda1b1c3d17a096886804bd96215bf74261/assets/logo/banner.svg

TheJJ:
two problems with that:
0. it's not repo-relative then. if you update the image, it should be updated in your readme as well
1. it doesn't support the `xlink` in there. to avoid redundancy, the banner includes the logo image.

TheJJ:
>> Can you please add support for placing svg images in readme files?
>
>This is currently disabled for security reasons. We hope to enable it again soon, but I don't have a date that I can give you, unfortunately. I'll pass your feedback onto the team though.

TheJJ:
>> What security reasons? You already render the svgs in the image viewer,
>> why don't apply the security reasons there?
>
>As far as I'm aware, directly embedding from within the repo was disabled due to the fact that SVGs can contain JS scripting that would then have access to our fileserver. To prevent this, we serve all files with that capability as plain-text when accessed directly from the repository.
>
>Embedding from an external site such as rawgit.com means that if there was scripting, it would not have direct access to our server. Same thing for serving them in the image viewer, which uses a different domain that's responsible for rendering more complicated files.
>
>I know this isn't ideal, and like I said before, we're working on finding more secure ways to render these files relatively. I'll pass your concerns on to the team working on it. Thanks for taking the time to write in!

TheJJ:
What do you think? https://github.com/TheJJ/openage/blob/logo/README.md

charlesrgould:
Looks good on desktop and mobile.

PR #321, opened 2015-06-27
Gamemain refactoring

Jon0:
Gamemain split across several classes
Seperating existing map rendering code into GameRenderer
The input handling and ui rendering now done in GameControl

DataManager renamed to GameSpec. this is supposed to be a class covering the available tile sets and unit types available (which may change with game data mods)

The file loading for Terrain now done by the GameSpec so it only needs loading once, Terrain objects now constructed just using meta data, which means constructing a new Terrain is easier. In future this makes exiting one game returning to the main menu and then creating another possible.

The GameControl splits the existing input code into EditorMode and ActionMode, to be handled seperatly, and a new mode CreateMode for constructing a game will eventually have gui elements to select settings for creating a new game.

I am considering refactoring the KeybindManager to handle mouse actions as well, maybe renamed to InputManager -- this could avoid a bunch of duplicate mouse handling code

This pr slightly overlaps the existing gui and renderer pull requests but should be easy to intergrate with them


detrumi:
Good idea on also handling mouse actions in the KeybindManager. And renaming it to InputManager seems logical then.

Jon0:
I have come up with a slight redesign of the input handling, Each input context (different menus/panels which may receive focus) should be able to map event -> action in any context, previously this was done event -> binding -> action, but this would lead to creating too many bindings which are not really needed (clicking a button wont need a binding it should just be mapped by the click event) 

The bindings will be still be passed to the event receiver in an event arg struct, this will tell the reciever any global binds on an event, the arg struct will also contain extra information like mouse position and the source event.

Jon0:
No longer WIP, please review / test.

cpp/game_renderer.cpp
```diff
@@ -0,0 +1,253 @@
...
+	glColor3f(0.0, 0.0, 0.0);
+	glBegin(GL_LINES); {
+
+		for (int i = -k; i < k; i++) {
+				glVertex3f(i * tilesize_x + x0, y1, 0);
+				glVertex3f(i * tilesize_x + x1, y0, 0);
+
+				glVertex3f(i * tilesize_x + x0, y0 - 1, 0);
+				glVertex3f(i * tilesize_x + x1, y1 - 1, 0);
+		}
```
TheJJ: double-indent fail ;)

mic-e: /me is utterly horrified

cpp/game_spec.cpp
```diff
@@ -269,7 +308,62 @@ void DataManager::load_projectile(const gamedata::unit_projectile &proj, unit_ty
 	}
 }
 
-void DataManager::create_abilities(const std::vector<gamedata::empiresdat> &gamedata) {
+void GameSpec::load_terrain(AssetManager *am) {
```
TheJJ: use `const AssetManager &` to avoid checking for `nullptr`?

cpp/game_spec.h
```diff
@@ -155,6 +196,11 @@ class DataManager {
 	std::unordered_map<index_t, Sound> available_sounds;
 
 	/**
+	 * begin the main loading job
+	 */
+	void initialize(AssetManager *am);
```
TheJJ: `const &`?

cpp/game_spec.h
```diff
@@ -179,10 +230,15 @@ class DataManager {
 	void load_projectile(const gamedata::unit_projectile &, unit_type_list &);
 
 	/**
+	 * fill in the terrain_data attribute of this
+	 */
+	void load_terrain(AssetManager *am);
```
TheJJ: `const &`?

cpp/game_renderer.cpp
TheJJ: yes, but please blame me for not having a better renderer finished already. gnargh.

cpp/generator.cpp
```diff
@@ -0,0 +1,330 @@
...
+		sheep.object_id = 594;
+
+		player_regions.push_back(player);
+		player_regions.push_back(obj_space);
+		player_regions.push_back(trees1);
+		player_regions.push_back(trees2);
+		player_regions.push_back(stone);
+		player_regions.push_back(gold);
+		player_regions.push_back(forage);
+		player_regions.push_back(sheep);
```
TheJJ: Btw we wanted to have the available ressources as a dynamic set (namely defined by the gamespec/nyan), so the mods can easily add new ressources someday. Probably too early, but we should at least think about it.

cpp/generator.cpp
```diff
@@ -0,0 +1,330 @@
...
+				m.placed_units.new_unit(*otype, p, tile.to_tile3().to_phys3());
+			}
+		}
+
+		// A space for starting town center and villagers
+		else if (r.owner) {
+			Player &p = m.players[r.owner];
+			auto tctype = this->spec->get_type(109); // town center
+			auto mvtype = this->spec->get_type(83);  // villager
+			auto fvtype = this->spec->get_type(293); // villager
```
TheJJ: both are villagers? male/female?

cpp/generator.cpp
```diff
@@ -0,0 +1,330 @@
...
+	e.start_game(*this);
+}
+
+
+void lol() {
+	// TODO: move these somewhere else
+	//this->assetmanager.check_updates();
+}
+
+} // namespace openage
```
TheJJ: missing `\n` and some leftover `lol()` function :D

cpp/input/action.h
```diff
@@ -0,0 +1,110 @@
...
+
+// TODO: use action_hint_t = std::pair<action_t, int>
+// for example a building command
+// std::make_pair(action_t::BUILD, 123)
+using action_id_t = action_t; //std::pair<action_t, int>;
+
+
+struct action_arg_t {
+public:
+	action_arg_t(Event e, coord::window m, coord::window_delta mm);
```
TheJJ: struct already public by default. do i see it right that the constructor is just for ommitting the hints member? when reading "set globally" down there i get the feeling this can be solved in a better way...

cpp/input/event.cpp
```diff
@@ -0,0 +1,152 @@
...
+}
+
+
+bool Event::operator ==(const Event &other) const {
+	return this->cc == other.cc && this->mod == other.mod;
+}
+
+
+int event_hash::operator()(const Event &e) const {
+	return class_code_hash()(e.cc); // ^ std::hash<modset_t>()(e.mod) * 3664657;
```
TheJJ: we might want to define a template instance for `std::hash` for a `const Event &`.

cpp/input/event.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+
+int modifier_hash::operator()(const modifier &m) const {
+	return std::hash<int>()(static_cast<int>(m));
+}
+
+
+int class_code_hash::operator()(const class_code_t &e) const {
+	return event_class_hash()(e.eclass) ^
+	       std::hash<int>()(e.code) * 3664657;
```
TheJJ: `std::hash` template instance for `class_code_t` as well? also, magic numbers?

cpp/input/action.h
mic-e: constructors are generally not a very great idea for structs (there are exceptions ofc).

cpp/input/event.cpp
mic-e: Why do we need `Event` to be hashable at all? At first glance, this seems like something that should be unhashable by design.

cpp/input/event.h
```diff
@@ -0,0 +1,149 @@
...
+using modset_t = std::unordered_set<modifier, modifier_hash>;
+
+
+/**
+ * a set sdl event types
+ * TODO to char function
+ */
+struct class_code_t {
+public:
+	class_code_t(event_class cl, code_t code);
```
TheJJ: with constructor and member functions, this is most likely a class, and especially not a `_t` type right?

cpp/input/input_manager.cpp
```diff
@@ -0,0 +1,255 @@
...
+	      { sdl_key(SDLK_5), action_t::SWITCH_TO_PLAYER_5},
+	      { sdl_key(SDLK_6), action_t::SWITCH_TO_PLAYER_6},
+	      { sdl_key(SDLK_7), action_t::SWITCH_TO_PLAYER_7},
+	      { sdl_key(SDLK_8), action_t::SWITCH_TO_PLAYER_8},
+	      { sdl_key(SDLK_UP), action_t::UP_ARROW},
+	      { sdl_key(SDLK_DOWN), action_t::DOWN_ARROW},
+	      { sdl_key(SDLK_LEFT), action_t::LEFT_ARROW},
+	      { sdl_key(SDLK_RIGHT), action_t::RIGHT_ARROW},
+	      { sdl_mouse(1), action_t::PAINT_TERRAIN},
+	      { sdl_mouse(1), action_t::SELECT},
```
TheJJ: double define of `mouse 1`?

cpp/input/input_manager.cpp
```diff
@@ -0,0 +1,255 @@
...
+	return this->global_hotkeys;
+}
+
+
+void InputManager::override_context(InputContext *context) {
+	// Create a new context list, unless no context lists exist
+	if (!this->contexts.empty()) {
+		this->contexts.push(std::vector<InputContext *>());
+	}
+	this->contexts.top().push_back(context);
```
TheJJ: `const context &` and then placint this in the vector is better i think.

cpp/input/event.h
mic-e: generally, `_t` should only be used for types that are actually just a single integer, or, in extreme cases, a collection of similar integers (like `phys3_t`).

cpp/generator.cpp
Jon0: What is your plan for terrain generation? Should a script file be used to decide regions and contents? I thought this would be replaced by some python functions in future so I left out some customisation options atm

TheJJ: yeah, all the terrain and game generation will be py someday, so this is good for now. Just wanted to say that the available ressource types will also be dynamic then.

cpp/options.cpp
```diff
@@ -0,0 +1,308 @@
...
+	for (auto &f : actions) {
+		result.push_back(inner_indent + "func " + f.first + "()");
+	}
+	for (auto &p : panels) {
+		for (auto &line : p.second->list_options(recurse, inner_indent)) {
+			result.push_back(line);
+		}
+	}
+	result.push_back(indent + "}");
+	return result;
```
TheJJ: for all these stack-constructed vectors we should use `std::move` for returning them.

cpp/input/action.h
Jon0: It was mainly for setting the structs const members

cpp/options.h
```diff
@@ -0,0 +1,235 @@
...
+	 */
+	const std::string name;
+
+protected:
+
+	/**
+	 * deprecated
+	 */
+	void add_bool(const std::string &vname, bool *value);
+	void add_int(const std::string &vname, int *value);
```
TheJJ: why do you add them even though they're already deprecated :D

cpp/input/event.cpp
Jon0: Events (a tuple of: class [keyboard, mouse], code, modifiers) are mapped to actions using std::unordered_map

cpp/player.h
```diff
@@ -14,8 +14,11 @@ class Unit;
 
 class Player {
 public:
-	Player(unsigned int number);
+	Player(unsigned int number, std::string name);
 
+	/**
+	 * numbers 0-7
+	 */
 	const unsigned int player_number;
```
TheJJ: nah, not just 0-7, it's uint128_t of course :grin:
but srsly, we should remove the 8 player restriction as soon as possible.

cpp/input/input_manager.cpp
Jon0: this is a multimap of hints, mouse click 1 produces an event_arg struct with hints PAINT_TERRAIN and SELECT, it is up to the context it gets passed to, to decide which one gets used.

cpp/player.h
Jon0: the comment was meant to imply a range 0..n-1 rather than the colour value which is 1..n

cpp/texture.cpp
```diff
@@ -203,6 +208,7 @@ void Texture::draw(coord::pixel_t x, coord::pixel_t y,
                    unsigned int mode, bool mirrored,
                    int subid, unsigned player,
                    Texture *alpha_texture, int alpha_subid) const {
+	main_thread_load();
```
TheJJ: `this->`

cpp/texture.cpp
```diff
@@ -385,7 +391,8 @@ void Texture::get_subtexture_size(int subid, int *w, int *h) const {
 
 
 GLuint Texture::get_texture_id() const {
-	return this->id;
+	main_thread_load();
```
TheJJ: `this->`

cpp/unit/producer.cpp
```diff
@@ -605,7 +607,7 @@ TerrainObject *BuildingProducer::make_annex(Unit &u, std::shared_ptr<Terrain> t,
 	u.log(MSG(dbg) << "adding annex " << annex_id);
 
 	// find annex foundation size
-	auto b = datamanager.get_building_data(annex_id);
+	auto b = dataspec.get_building_data(annex_id);
```
TheJJ: `this->`

cpp/unit/producer.cpp
```diff
@@ -622,7 +624,7 @@ TerrainObject *BuildingProducer::make_annex(Unit &u, std::shared_ptr<Terrain> t,
 	annex_loc->place(t, start_tile, state);
 
 	// weak ptr for use in lambda drawing functions
-	auto annex_type = datamanager.get_type(annex_id);
+	auto annex_type = dataspec.get_type(annex_id);
```
TheJJ: `this->`

cpp/util/variable.h
```diff
@@ -0,0 +1,67 @@
...
+ * A typed value holder
+ */
+class VariableBase {
+public:
+	virtual ~VariableBase() {}
+
+	/**
+	 * sets the type and value
+	 */
+    template<class T, class V> void set(const V &value);
```
TheJJ: indent forkup!

cpp/util/variable.h
```diff
@@ -0,0 +1,67 @@
...
+		value(initial_value) {}
+
+
+    const T &get() const {
+    	return this->value;
+    }
+
+    void set(const T &v) {
+    	this->value = v;
+    }
```
TheJJ: moar indent forkups

cpp/input/input_manager.cpp
TheJJ: i see, nice.

Jon0:
The rebase is done, some testing would be nice.

libopenage/options.h
```diff
@@ -0,0 +1,291 @@
...
+ * option node allowing reflection, while also
+ * being directly accessable as a typed member
+ */
+template<class T>
+class Var : public util::Variable<T> {
+public:
+	Var(OptionNode *owner, const std::string &name, const T &init)
+		:
+		util::Variable<T>{init} {
+		owner->add(name, &this->value);
```
mic-e: warning: address of 'this->value' will always evaluate to 'true' [-Wpointer-bool-conversion]

libopenage/game_control.h
```diff
@@ -0,0 +1,155 @@
...
+	// currently selected terrain id
+	openage::terrain_t editor_current_terrain;
+	unsigned int editor_current_type;
+	unsigned int editor_category;
+	UnitType *selected_type;
+	std::string category;
+
+	// TODO assign placement to player object
+	bool paint_terrain;
+	bool building_placement;
```
mic-e: warning: private field 'building_placement' is not used [-Wunused-private-field]

libopenage/options.cpp
```diff
@@ -0,0 +1,346 @@
...
+	for (auto &f : actions) {
+		result.push_back(inner_indent + "func " + f.first + "()");
+	}
+	for (auto &p : children) {
+		for (auto &line : p.second->list_options(recurse, inner_indent)) {
+			result.push_back(line);
+		}
+	}
+	result.push_back(indent + "}");
+	return std::move(result);
```
mic-e: warning: moving a local object in a return statement prevents copy elision [-Wpessimizing-move]

(the `std::move` makes stuff slower by preventing optimization)

libopenage/input/input_manager.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+
+modset_t InputManager::get_mod() const {
+	SDL_Keymod mod = SDL_GetModState();
+	return sdl_mod(mod);
+}
+
+
+bool InputManager::on_input(SDL_Event *e) {
+	Engine &engine = Engine::get();
```
mic-e: unused variable

mic-e:
This doesn't run; the 50500 palette file is missing.

```
mic@mic ~/git/openage $ ./run
INFO launching engine with data directory './assets'
INFO Using audio device: default [freq=48000, format=32784, channels=, samples=2048]
INFO Font queried: DejaVu Serif:style=Book
INFO Font file: /usr/share/fonts/truetype/dejavu/DejaVuSerif.ttf
INFO Font queried: DejaVu Serif:style=Book
INFO Font file: /usr/share/fonts/truetype/dejavu/DejaVuSerif.ttf
INFO Font file: /usr/share/fonts/truetype/dejavu/DejaVuSansMono.ttf
INFO Loading time [engine]: 0.123 s
  File "run.py", line 15, in init run (/home/mic/git/openage/run.cpp:829)
    main()
  File "/home/mic/git/openage/openage/__main__.py", line 87, in main
    return args.entrypoint(args, cli.error)
  File "/home/mic/git/openage/openage/game/main.py", line 39, in main
    return run_game(args, assets)
  File "openage/game/main_cpp.pyx", line 20, in openage.game.main_cpp.run_game (/home/mic/git/openage/openage/game/main_cpp.cpp:855)
    result = run_game_cpp(args_cpp)
  File "libopenage/util/file.h", line 74, in void openage::util::read_csv_file(const std::string &, std::vector<lineformat> &) [lineformat = gamedata::palette_color]
    throw Error(MSG(err) <<
openage.cppinterface.exctranslate.CPPException: Failed to read CSV file ./assets/converted/player_palette_50500.docx: Empty or missing.
```

mic-e:
You'll of course want to squash the menu selection bug/save file improvement commits to prevent those 14000+/- from showing up in the git history.

TheJJ:
Whee, are we mergable now?

Jon0:
All issues have been fixed. Should be mergable now.

TheJJ:
Bug: i somehow managed to be in the selection menu where you navigate with arrow keys, but the arrow keys also scroll the map.

TheJJ:
It's done! Thanks for the great new features jon!

TheJJ:
Ah, we might wanna update `doc/usage.md` or even delete it before we tag a new release.

PR #322, opened 2015-06-29
[WIP] Fixed point numbers

detrumi:
Fixed point numbers, to replace the int_64 and related calculations on the `phys_t` type. This also removes the ugly `settings` namespace.

It compiles, but still has quite some bugs I'm currently fixing (I'll also be moving it to the `util` namespace).

### DIFF
--- a/cpp/coord/fixed_point.h
+++ b/cpp/coord/fixed_point.h
@@ -1,0 +1,286 @@
+// Copyright 2013-2015 the openage authors. See copying.md for legal info.
+
+#ifndef OPENAGE_COORD_FIXED_POINT_H_
+#define OPENAGE_COORD_FIXED_POINT_H_
+
+#include <cmath>
+#include <functional>
+
+#include "../util/misc.h"
+
+namespace openage {
+namespace coord {
+
+template<typename int_type, unsigned fractional_bits>
+struct FixedPoint {
+
+	int_type data;
+
+	static const int_type one = int_type(1) << fractional_bits;
+
+	static const int_type fraction_bitmask = one - 1;
+
+
+	// constructors
+	constexpr FixedPoint() : data(0) {
+	}
+
+	constexpr FixedPoint(int_type n) : data(n << fractional_bits) {
+	}
+
+	constexpr FixedPoint(int n) : data(n << fractional_bits) {
+	}
+
+	constexpr FixedPoint(unsigned int n) : data(n << fractional_bits) {
+	}
+
+	// Constructing FixedPoint numbers with values like 1/2 and 1/8 won't lose precision
+	constexpr FixedPoint(float n) : data(static_cast<int_type>(n * one)) {
+	}
+
+	constexpr FixedPoint(double n) : data(static_cast<int_type>(n * one)) {
+	}
+
+	static FixedPoint from_raw(int_type data) {
+		FixedPoint result;
+		result.data = data;
+		return result;
+	}
+
+	constexpr FixedPoint(const FixedPoint &o) : data(o.data) {
+	}
+
+	FixedPoint &operator=(const FixedPoint &o) {
+		data = o.data;
+		return *this;
+	}
+
+
+	// Comparison operators
+	bool operator==(const FixedPoint &o) const {
+		return data == o.data;
+	}
+
+	bool operator!=(const FixedPoint &o) const {
+		return data != o.data;
+	}
+
+	bool operator<(const FixedPoint &o) const {
+		return data < o.data;
+	}
+
+	bool operator>(const FixedPoint &o) const {
+		return data > o.data;
+	}
+
+	bool operator<=(const FixedPoint &o) const {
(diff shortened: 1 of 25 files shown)


cpp/unit/producer.cpp
```diff
@@ -678,7 +678,7 @@ void ProjectileProducer::initialise(Unit *unit, Player &) {
 	unit->graphics = &this->graphics;
 
 	// projectile speed
-	coord::phys_t sp = this->unit_data.speed * coord::settings::phys_per_tile / 666;
+	coord::phys_t sp = coord::phys_t{this->unit_data.speed} / 666;
```
mic-e: projectile speed of the beast

cpp/unit/unit_texture.cpp
```diff
@@ -145,7 +145,7 @@ unsigned int dir_group(coord::phys3_delta dir, unsigned int angles) {
 	unsigned int first_angle = 5 * angles / 8;
 
 	// normalise direction vector
-	double len = std::hypot(dir.ne, dir.se);
+	double len = std::hypot(dir.ne, dir.se).to_double();
```
mic-e: this should be `phys3_t len = std::hypot(dir.ne, dir.se);`

mic-e:
This looks pretty nice, so far. A lot of the ugliness that resulted from my earlier hacks has been removed.

cpp/unit/unit_texture.cpp
detrumi: I was thinking of changing it to this:
```
	coord::phys_t len = std::hypot(dir.ne, dir.se);
	double dir_ne = static_cast<double>(dir.ne / len);
	double dir_se = static_cast<double>(dir.se / len);
```

Or maybe use the to_double function instead of the static_cast:
`double dir_ne = (dir.ne / len).to_double();`

I think having the operators return FixedPoint is nicer, because it's nice to be able to divide by two or add 0.5f, we use that a lot.

mic-e: But _which_ FixedPoint should we return? If you e.g. divide a FixedPoint<int, 256> by a FixedPoint<int, 256>, should we return a FixedPoint<int, 256>?

detrumi: I think having different precisions as result type would complicate this a lot. Isn't it good enough to keep everything a `phys_t`?

mic-e: Users will indeed expect `operator /(A a, B b)` to return `A`, you're right.

While `a * 0.5` should work as expected, I think we'd best not support `0.5 * a`.

__However__, in this particular use case, we don't want that behavior of the division operator. We know that the result will be in the interval `[-1; 1]`, and there `phys_t` gives us only 16 bits, significantly less than `float` or something like`FixedPoint<int64_t, 32>`. It's not that bad for `phys_t`, but other types such as `tile_t` have lower resolution (0 bits).
So I suggest adding a templated member function, `div`, where the caller can specify their own return type (`float`, `FixedPoint<...>`, ...).

cpp/coord/fixed_point.h
```diff
@@ -0,0 +1,286 @@
...
+	static const int_type one = int_type(1) << fractional_bits;
+
+	static const int_type fraction_bitmask = one - 1;
+
+
+	// constructors
+	constexpr FixedPoint() : data(0) {
+	}
+
+	constexpr FixedPoint(int_type n) : data(n << fractional_bits) {
```
mic-e: We probably want those single-argument constructors to be `explicit` to prevent some implicit conversions.

cpp/coord/fixed_point.h
```diff
@@ -0,0 +1,286 @@
...
+	int_type data;
+
+	static const int_type one = int_type(1) << fractional_bits;
+
+	static const int_type fraction_bitmask = one - 1;
+
+
+	// constructors
+	constexpr FixedPoint() : data(0) {
+	}
```
mic-e: Our current style says

    constexpr FixedPoint()
        :
        data{0} {}

cpp/coord/fixed_point.h
```diff
@@ -0,0 +1,286 @@
...
+
+	FixedPoint &operator=(const FixedPoint &o) {
+		data = o.data;
+		return *this;
+	}
+
+
+	// Comparison operators
+	bool operator==(const FixedPoint &o) const {
```
mic-e: We don't want references here; `FixedPoint` contains just a single int-like value, which may even take less memory on the stack than a reference to it.

TheJJ:
hows this going?

cpp/coord/fixed_point.h
TheJJ: But then the compiler can optimize it away anyway as it's a `const &`. I'd say this is good the `&`-way.

TheJJ:
superseeded by #662

TheJJ:
went upstream in #975

PR #328, opened 2015-07-24
Profiler

Miguellissimo:
I started with a simple profiler to measure several processing times happening in a frame. At the moment the profiler has categories for EVENT_PROCESSING, RENDERING and IDLE_TIME, but the idea is to keep it general to add more at later time.

### DIFF
--- a/libopenage/util/profiler.cpp
+++ b/libopenage/util/profiler.cpp
@@ -1,0 +1,177 @@
+// Copyright 2015-2015 the openage authors. See copying.md for legal info.
+
+#include "profiler.h"
+#include "../engine.h"
+#include "misc.h"
+
+#include <chrono>
+#include <epoxy/gl.h>
+#include <iostream>
+
+namespace openage {
+namespace util {
+
+Profiler::~Profiler() {
+	this->unregister_all();
+}
+
+void Profiler::register_component(std::string com, color component_color) {
+	if (this->registered(com)) {
+		return;
+	}
+
+	component_time_data cdt;
+	cdt.display_name = com;
+	cdt.drawing_color = component_color;
+
+	for (auto &val : cdt.history) {
+		val = 0;
+	}
+
+	this->components[com] = cdt;
+}
+
+void Profiler::unregister_component(std::string com) {
+	if (not this->registered(com)) {
+		return;
+	}
+
+	this->components.erase(com);
+}
+
+void Profiler::unregister_all() {
+	std::vector<std::string> registered_components = this->registered_components();
+
+	for (auto com : registered_components) {
+		this->unregister_component(com);
+	}
+}
+
+std::vector<std::string> Profiler::registered_components() {
+	std::vector<std::string> registered_components;
+	for (auto &pair : this->components) {
+		registered_components.push_back(pair.first);
+	}
+
+	return registered_components;
+}
+
+void Profiler::start_measure(std::string com, color component_color) {
+	if (not this->registered(com)) {
+		this->register_component(com, component_color);
+	}
+
+	this->components[com].start = std::chrono::high_resolution_clock::now();
+}
+
+void Profiler::end_measure(std::string com) {
+	if (this->registered(com)) {
+		std::chrono::high_resolution_clock::time_point end = std::chrono::high_resolution_clock::now();
+		this->components[com].duration = end - this->components[com].start;
+	}
+}
+
+void Profiler::draw_component_performance(std::string com) {
+	color rgb = this->components[com].drawing_color;
+	glColor4f(rgb.r, rgb.g, rgb.b, 1.0);
+
+	glLineWidth(1.0);
+	glBegin(GL_LINE_STRIP);
+	float x_offset = 0.0;
+	float offset_factor = (float)PROFILER_CANVAS_WIDTH / (float)MAX_DURATION_HISTORY;
+	float percentage_factor = (float)PROFILER_CANVAS_HEIGHT / 100.0;
+
+	for (auto i = this->insert_pos; mod(i, MAX_DURATION_HISTORY) != mod(this->insert_pos-1, MAX_DURATION_HISTORY); ++i) {
+		i = mod(i, MAX_DURATION_HISTORY);
+
+		auto percentage = this->components[com].history.at(i);
+		glVertex3f(PROFILER_CANVAS_POSITION_X + x_offset, PROFILER_CANVAS_POSITION_Y + percentage * percentage_factor, 0.0);
+		x_offset += offset_factor;
+	}
+	glEnd();
+
+	// reset color
+	glColor4f(1.0, 1.0, 1.0, 1.0);
+}
+
+void Profiler::show(bool debug_mode) {
+	if (debug_mode) {
+		this->show();
+	}
+}
+
+void Profiler::show() {
+
+	this->draw_canvas();
+	this->draw_legend();
+
+	for (auto com : this->components) {
+		this->draw_component_performance(com.first);
+	}
+}
+
+bool Profiler::registered(std::string com) const {
+	return this->components.find(com) != this->components.end();
+}
+
+unsigned Profiler::size() const {
+	return this->components.size();
+}
+
+void Profiler::start_frame_measure() {
+	this->frame_start = std::chrono::high_resolution_clock::now();
+}
+
+void Profiler::end_frame_measure() {
+	auto frame_end = std::chrono::high_resolution_clock::now();
+	this->frame_duration = frame_end - this->frame_start;
+
+	for (auto com : this->registered_components()) {
+		double percentage = this->duration_to_percentage(this->components[com].duration);
+		this->append_to_history(com, percentage);
+	}
+
+	this->insert_pos++;
+}
+
+void Profiler::draw_canvas() {
+	glColor4f(0.2, 0.2, 0.2, PROFILER_CANVAS_ALPHA);
+	glRecti(PROFILER_CANVAS_POSITION_X,
+	        PROFILER_CANVAS_POSITION_Y,
+	        PROFILER_CANVAS_POSITION_X + PROFILER_CANVAS_WIDTH,
+	        PROFILER_CANVAS_POSITION_Y + PROFILER_CANVAS_HEIGHT);
+}
+
+void Profiler::draw_legend() {
+	int offset = 0;
+	for (auto com : this->components) {
+		glColor4f(com.second.drawing_color.r, com.second.drawing_color.g, com.second.drawing_color.b, 1.0);
+		int box_x = PROFILER_CANVAS_POSITION_X + 2;
+		int box_y = PROFILER_CANVAS_POSITION_Y - PROFILER_COM_BOX_HEIGHT - 2 - offset;
+		glRecti(box_x, box_y, box_x + PROFILER_COM_BOX_WIDTH, box_y + PROFILER_COM_BOX_HEIGHT);
+
+		glColor4f(0.2, 0.2, 0.2, 1);
+		coord::window position = coord::window();
+		position.x = box_x + PROFILER_COM_BOX_WIDTH + 2;
+		position.y = box_y + 2;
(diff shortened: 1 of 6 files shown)


cpp/engine.cpp
```diff
@@ -170,9 +170,16 @@ Engine::Engine(util::Dir *data_dir, const char *windowtitle)
 	if (devices.empty()) {
 		throw util::Error{MSG(err) << "No audio devices found"};
 	}
+
+	// initialize profiler
+	this->profiler.register_category(util::Profiler::Categorie::EVENT_PROCESSING);
+	this->profiler.register_category(util::Profiler::Categorie::RENDERING);
+	this->profiler.register_category(util::Profiler::Categorie::IDLE_TIME);
```
mic-e: It would be best if we could avoid havoing to manually call this for every member of the enum (what happens if I forget to do it? Undefined behavior?)

Miguellissimo: Yeah, I was thinking about moving the register part into the start_measure method and register the categorie only if it is not already registered

cpp/util/profiler.h
```diff
@@ -0,0 +1,53 @@
...
+	void end_measure(Categorie cat);
+	long last_duration(Categorie cat);
+	void show(Categorie cat);
+	void show(bool debug_mode);
+	void show();
+	bool registered(Categorie cat) const;
+	unsigned size() const;
+
+private:
+	std::map<Categorie, std::shared_ptr<CategorieTimeData>> categories;
```
mic-e: `shared_ptr` would only be needed if the object had multiple owners. Usually, and here, you'd want to use `unique_ptr`.

But here, you shouldn't use a pointer at all. Just make it a `std::unordered_map<Categorie, CategorieTimeData>`.

cpp/util/profiler.h
```diff
@@ -0,0 +1,53 @@
...
+#ifndef OPENAGE_UTIL_PROFILER_H_
+#define OPENAGE_UTIL_PROFILER_H_
+
+#include <map>
+#include <vector>
+#include <chrono>
+#include <memory>
+#include <list>
+
+#define MAX_DURATION_HISTORY 100
```
mic-e: Please use `constexpr` here.

cpp/util/profiler.h
```diff
@@ -0,0 +1,53 @@
...
+
+namespace openage {
+namespace util {
+
+struct CategorieTimeData {
+	std::chrono::high_resolution_clock::time_point start;
+	std::chrono::high_resolution_clock::duration duration;
+	std::list<std::chrono::high_resolution_clock::duration> history;
```
mic-e: Isn't `std::list` rather memory-heavy? Not Landau Notation-wise, but I imagine that it involves a lot of slow small heap allocations 'n stuff.
`std::vector` would be preferable if heap allocation is required (you'd have to manually manage the cyclic access thingie, but that's just one extra variable and a few lines of code).

However, since we know the history size as a compile-time constant, you can simply use `std::array` instead of `std::vector`, which uses no heap allocation at all.

cpp/util/profiler.cpp
```diff
@@ -0,0 +1,93 @@
+// Copyright 2015-2015 the openage authors. See copying.md for legal info.
+
+#include "profiler.h"
+#include <chrono>
```
mic-e: Note that we have `cpp/util/timer.h`, which provides `util::Timer`.

cpp/util/profiler.h
```diff
@@ -0,0 +1,52 @@
...
+#include <vector>
+#include <chrono>
+#include <list>
+
+constexpr int MAX_DURATION_HISTORY = 100;
+
+namespace openage {
+namespace util {
+
+struct CategorieTimeData {
```
TheJJ: struct names are lowercased and underscored please (see the `doc/codestyle/` examples)

cpp/util/profiler.h
```diff
@@ -0,0 +1,52 @@
...
+	std::chrono::high_resolution_clock::time_point start;
+	std::chrono::high_resolution_clock::duration duration;
+	std::list<std::chrono::high_resolution_clock::duration> history;
+};
+
+class Profiler {
+public:
+	enum class Categorie {
```
TheJJ: enum class names are lowercased again, also the correct word is `category`, but you might want to name it `component` or `subsystem`. which brings me to the idea to have some general subsystem registry which we can then use for the profiler.

cpp/util/profiler.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+namespace openage {
+namespace util {
+
+void Profiler::register_category(Categorie cat) {
```
TheJJ: the category registering could be avoided entirely as they all can be just used as default. Using a "activate" and "deactivate" would come in handy, though.

cpp/util/profiler.h
```diff
@@ -0,0 +1,52 @@
...
+	void end_measure(Categorie cat);
+	long last_duration(Categorie cat);
+	void show(Categorie cat);
+	void show(bool debug_mode);
+	void show();
+	bool registered(Categorie cat) const;
+	unsigned size() const;
+
+private:
+	std::map<Categorie, CategorieTimeData> categories;
```
TheJJ: you want to use `std::unordered_map` here to have a hashmap and not a searchtree.

TheJJ:
You can create a test or demo to run your code independently of the actual game for now. You can then start it with `openage --test profiler` or something like that.

mic-e:
@TheJJ this is based on the openage master, not pyinterface.

cpp/engine.cpp
```diff
@@ -274,6 +278,10 @@ void Engine::loop() {
 				}
 			}
 		}
+		this->profiler.end_measure(util::Profiler::component::EVENT_PROCESSING);
+
+		// TODO remove me later
+		std::cout << "Measured time: " << this->profiler.last_duration(util::Profiler::component::EVENT_PROCESSING) << std::endl;
```
TheJJ: you want to use the logger subsystem, even now, for testing.

cpp/util/profiler.h
```diff
@@ -0,0 +1,52 @@
...
+	void end_measure(component cat);
+	long last_duration(component cat);
+	void show(component cat);
+	void show(bool debug_mode);
+	void show();
+	bool registered(component cat) const;
+	unsigned size() const;
+
+private:
+	std::map<component, component_time_data> components;
```
TheJJ: why do you keep using the `map` instead of `unordered_map`?

cpp/engine.cpp
```diff
@@ -170,9 +170,12 @@ Engine::Engine(util::Dir *data_dir, const char *windowtitle)
 	if (devices.empty()) {
 		throw util::Error{MSG(err) << "No audio devices found"};
 	}
+
 }
 
 Engine::~Engine() {
+	this->profiler.unregister_all();
```
mic-e: shouldn't this happen in the profiler's constructor instead?

TheJJ:
was your rebase successful? If not, we can help you :)

Miguellissimo:
I removed the WIP tag. Review, Feedback please

libopenage/util/profiler.cpp
```diff
@@ -0,0 +1,182 @@
...
+void Profiler::register_component(std::string com, color component_color) {
+	if (this->registered(com)) {
+		return;
+	}
+
+	component_time_data cdt = component_time_data();
+	cdt.display_name = com;
+	cdt.drawing_color = component_color;
+
+	for (auto it = cdt.history.begin(); it != cdt.history.end(); ++it) {
```
TheJJ: `(auto &it : cdt.history)`

libopenage/util/profiler.cpp
```diff
@@ -0,0 +1,182 @@
...
+Profiler::~Profiler() {
+	this->unregister_all();
+}
+
+void Profiler::register_component(std::string com, color component_color) {
+	if (this->registered(com)) {
+		return;
+	}
+
+	component_time_data cdt = component_time_data();
```
TheJJ: `this->component_time_data`

libopenage/util/profiler.cpp
```diff
@@ -0,0 +1,182 @@
...
+	std::vector<std::string> registered_components = this->registered_components();
+
+	for (auto com : registered_components) {
+		this->unregister_component(com);
+	}
+}
+
+std::vector<std::string> Profiler::registered_components() {
+	std::vector<std::string> registered_components;
+	for (auto it = this->components.begin(); it != this->components.end(); ++it) {
```
TheJJ: `auto &it : this->components`

libopenage/util/profiler.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+	this->insert_pos++;
+}
+
+void Profiler::draw_canvas() {
+	glColor4f(0.2, 0.2, 0.2, PROFILER_CANVAS_ALPHA);
+	glRecti(PROFILER_CANVAS_POSITION_X,
+			PROFILER_CANVAS_POSITION_Y,
+			PROFILER_CANVAS_POSITION_X + PROFILER_CANVAS_WIDTH,
+			PROFILER_CANVAS_POSITION_Y + PROFILER_CANVAS_HEIGHT);
```
TheJJ: indent:
tabs to reach position of `glRecti`, spaces to move `PROFILER...` after the `(`

libopenage/util/profiler.cpp
```diff
@@ -0,0 +1,182 @@
...
+}
+
+void Profiler::append_to_history(std::string com, double percentage) {
+	if (this->insert_pos == MAX_DURATION_HISTORY) {
+		this->insert_pos = 0;
+	}
+	this->components[com].history[this->insert_pos] = percentage;
+}
+
+int Profiler::keep_in_duration_bound(int value) {
```
TheJJ: you might move a general version of this function to `util/misc`.

libopenage/util/profiler.cpp
```diff
@@ -0,0 +1,182 @@
...
+	int result = value % MAX_DURATION_HISTORY;
+	return result < 0 ? result + MAX_DURATION_HISTORY : result;
+}
+
+} //namespace util
+} //namespace openage
```
TheJJ: we recently switched the namespace closing style:
`}} // openage::util`

libopenage/util/profiler.h
```diff
@@ -0,0 +1,126 @@
...
+	 *returns a vector of registered component names
+	 */
+	std::vector<std::string> registered_components();
+
+	/*
+	 * starts a measurement for the component com. If com is not yet
+	 * registered, its getting registered and the profiler uses the color
+	 * information given by component_color. The default value is white.
+	 */
+	void start_measure(std::string com, color component_color = {1.0, 1.0, 1.0});
```
TheJJ: no spaces around a defaultvalue `=`

libopenage/util/profiler.h
```diff
@@ -0,0 +1,126 @@
...
+	void draw_canvas();
+	void draw_legend();
+	int keep_in_duration_bound(int value);
+	void draw_component_performance(std::string com);
+	double duration_to_percentage(std::chrono::high_resolution_clock::duration duration);
+	void append_to_history(std::string com, double percentage);
+};
+
+} //namespace util
+} //namespace openage
```
TheJJ: namespace stype pls

TheJJ:
looks pretty nice, the api is clean and simple ;)

libopenage/util/profiler.cpp
Miguellissimo: there is actually already one ;)

libopenage/util/profiler.cpp
Miguellissimo: component_time_data is not a class member. Therefore this is not possible.

TheJJ: whoops, didn't check that. I think the default constructor for `component_time_data` is already run, so you can just `component_time_data cdt;`

TheJJ:
continued in #348

PR #331, opened 2015-07-29
Update openSUSE build instructions

boombatower:


zuntrax:
Do we still need glew? I think it was replaced by libepoxy.

TheJJ:
Jup, glew isn't needed any more. And you might try to simplify the disclaimer. It's "a look" [and not the brother of alot](http://hyperboleandahalf.blogspot.de/2010/04/alot-is-better-than-you-at-everything.html) btw :grin:

TheJJ:
@boombatower can you reword the disclaimer to make it more readable and remove `glew` from the list please?

boombatower:
I ended up getting rid of the disclaimer since it is only relevant for package building. Not sure why I didn't realize before that this is for manually building.

TheJJ:
Thanks!

Issue #333, opened 2015-07-29
Possible deadlock in the job manager code


mic-e:
About one in 100 times or so the job manager test freezes with a deadlock.

This is the stack trace info I managed to collect:

```
thread 2:
#0  pthread_cond_wait@@GLIBC_2.3.2 ()
    at ../nptl/sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
#1  0x00007f9097596f3c in std::condition_variable::wait(std::unique_lock<std::mutex>&) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x00007f9097af7403 in openage::job::Worker::process (this=0x2ace400)
    at /home/mic/git/openage/cpp/job/worker.cpp:70
#3  0x00007f909759bdf0 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007f909975a0a4 in start_thread (arg=0x7f9083af9700)
    at pthread_create.c:309
#5  0x00007f909873b04d in clone ()
    at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111
thread 1:
#0  0x00007f909975b4db in pthread_join (threadid=140258661340928, 
    thread_return=0x0) at pthread_join.c:92
#1  0x00007f909759bd37 in std::thread::join() ()
   from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x00007f9097af2538 in openage::job::JobManager::stop (this=0x7ffce85e9848)
    at /home/mic/git/openage/cpp/job/job_manager.cpp:49
#3  0x00007f9097af63cb in openage::job::tests::test_simple_job ()
    at /home/mic/git/openage/cpp/job/tests.cpp:43
#4  0x00007f9097af68d6 in openage::job::tests::test_job_manager ()
    at /home/mic/git/openage/cpp/job/tests.cpp:94
```

PR #336, opened 2015-07-30
Remove deadlock in JobManager/Worker

fr3akout:
This pull request removes a deadlock when stopping the worker-threads of the JobManager.


mic-e:
fixes #333

TheJJ:
nice!

fr3akout:
Actually there is still a little chance of a deadlock, but it will be fixed with the next commit...

franciscod:
osom!

Issue #318, opened 2015-06-23
convert: HD 4.0 patch compatibility


goto-bus-stop:
The HD Edition 4.0 patch changed a lot in data file storage. For example, DRS files are a thing of the past. The good part is that the new storage format is like, a million times better. The bad part is that the openage conversion script no longer works for up-to-date HD installations, and needs more special cases, probably.

I'll have time to work on this in a bit over a week, but if somebody else wants to pick it up in the mean time, feel free :smile:

franciscod:
Are we still supporting the old format?

goto-bus-stop:
I'd say yes, at least for a while. The 4.0 patch is very buggy so a lot of people are still on 3.9.

janisozaur:
and for original release of the game as well?

detrumi:
It's probably best to keep supporting the original release, as we don't want to force people to buy the HD version if they already have the old version.

We can drop 3.9 support after a while, for example when 4.1 or something comes out, as most people will update their game. But that depends on how much has changed between versions. If it's just small changes, we can keep old versions in.

goto-bus-stop:
Yeah, absolutely, the original releases should be supported. In fact, the original releases might still be more common than HD at this point among people who want to install openage :stuck_out_tongue_winking_eye:  4.0 support would be in addition to all other releases that are currently supported, just like how HD support itself was an extra thing.

TheJJ:
We have #61 for this to support all kinds of input configs and select them automatically. #288 will bring a filesystem abstraction comparable to [physFS](http://icculus.org/physfs/), so we will be able to use the version detection to mount all the archives (including drs) transparently for the then-following conversion.
We can support the `.cab` files from the iso the same way as the new folder structure that way `\o/`!

mic-e:
If anyone wants to work on this, __please__ base your work on the [current pyinterface branch](https://github.com/mic-e/openage/tree/pyinterface).

There have been __so many changes__ and simplifications that merging your patch would pretty much result in a re-write.

TheJJ:
should be fixed by #337 now

PR #338, opened 2015-08-01
[WIP] Bugfix of the convert script in the AoE HD edition

Miguellissimo:
The convert script is not working any more with the assets from the HD edition of Age of Empires (without the new update). The conversion terminates in an exception that the listdirs function is undefined. 

This error should be fixed with this first commit, but the asset conversion is still not running. 
```bash
...
INFO [py] [3128/3133] terrain/15030.slp
INFO [py] [3129/3133] terrain/15031.slp
INFO [py] [3130/3133] terrain/blends
  File "run.py", line 15, in init run (/home/miguel/git/openage/run.cpp:810)
    main()
...
File "/home/miguel/git/openage/openage/util/fslike/union.py", line 70, in open_r
    raise FileNotFoundError(b'/'.join(parts))
FileNotFoundError: b'terrain/blends'
```

### DIFF
--- a/openage/convert/driver.py
+++ b/openage/convert/driver.py
@@ -36,10 +36,15 @@ def get_string_resources(srcdir):
     else:
         count = 0
         from .hdlanguagefile import read_hd_language_file
-        for lang in srcdir.listdirs("bin"):
+        
+        bin_path = srcdir.joinpath("bin") 
+        for lang in bin_path.list():
+            if bin_path.joinpath(lang).is_file():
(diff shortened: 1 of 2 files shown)


zuntrax:
See #318, they changed the file structure.

jfeo:
My fix is to make sure no directories are added to the `files_to_convert`-list. Makes sure `terrain/blends` is not added. The game runs, so I guess it is not really to be converted?

PR #343, opened 2015-08-12
Changed FAQ bullets to subheadings

frickler01:
Explanation of a bullet point was in a paragraph, now it's the block below a subheading - makes more sense IMHO.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,15 +52,15 @@ Our goals specifically *exclude*:
 Current State of the Project
 ----------------------------
 
- - What features are currently implemented?
+###What features are currently implemented?
 
 See [doc/status.md](/doc/status.md).
 
- - What can I do once I start the game?
+###What can I do once I start the game?
 
 See [docs/usage.md](/doc/usage.md).
 
- - What's the plan?
+###What's the plan?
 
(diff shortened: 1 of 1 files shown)


TheJJ:
Thanks. I think it'll look better with level 4 or more, what do you think?. Also please leave a space after the `#`s.

TheJJ:
Alternatively, because i think subheadings won't look as cool as bullets, the answers to these questions could also be a `*`, but with two spaces indent so they're a subitem to the question.

frickler01:
Since the current `h3` font size is only one possible representation, I'd prefer to mark it up semantically as the next subdivision of a given topic. When you introduce a new layer between these, just move the questions down one `#`.
Also having kind of real sentences as answers (instead of bullets) seems more considerate of the reader.

TheJJ:
These texts are way too big then in my opinion.

About real sentences: I think full sentences are the one of the main reasons why people don't read readme text. The information needs to be trimmed out, nobody wants to read text blocks in man pages or readmes.

TheJJ:
How do you like https://github.com/TheJJ/openage/blob/logo/README.md ? Including your logo :grinning:

frickler01:
Thant works as well.

Issue #11, opened 2014-10-27
Battlefield style battlescreen


boombatower:
Basically a large second screen minimap with more detail. Perhaps showing unit icons instead of just colored pixels.

The Battlefield version is done using a web interface (not sure if connected to local game or central server). For openage it could be done in any variety of options.

I bring this up since I've been doing some basic research into this before I discovered openage. My plan was to either intercept network traffic (like voobly does for observer proxy) or read the record game file as it is being written (assuming it dumps fairly often).

There is some code out there to parse the starting point of recorded games and stats and what not, which voobly uses on their site, but not a lot of info on the meat of the file. Does openage support recorded game files? As such I imagine I could look there.

I would be interested in creating this if it sounds feasible, although my plan was do it as a local webservice. If there is information on recorded game format perhaps it would be better to create it standalone so it works for AoC, HD, and openage. Thoughts?

boombatower:
Just an updated set of thoughts. I assume this could be done as a enlarged and enhanced version of minimap #344. Where it could represent group of units by their icon...like a group of knights by the icon shown a stables to build them. I have a lot more ideas like trying to show map control heatmaps, battles, and alerts, but it really boils down to a large minimap at the most basic level.

I confirmed that game record files are written as the game progresses (quite frequently) and there are a number of projects that document the spec and provide basic parsing ability. Assuming one wrote a parser and used something like inotify to detect updates and parse added bytes it could be conceivable to issue those commands in openage to provide a synced game state if slightly delayed. Similar to an observer, but without the network/sync/lag strain nor having people agree to that. It would be an alternative to set it up as network connected if openage ever supported interoperability with original age of empires games. The main benefit of building this in openage as opposed to a separate project being that openage will have all the features built in and it can share the openage game simulation code rather than having to build that on top of parser.

I understand the comment about not need to read recorded game file if built into openage, but the reason for that was to be able to also use this with the original game since I expect to have to keep using it for quite sometime.

Anyway, once the game state is synced the battle screen provided by openage could be used in conjunction with the original age of empires client as a supplemental display. Ideally the games would be interoperable, but I feel like someone stated that would be unlikely.

[...]

PR #344, opened 2015-08-12
[WIP] Minimap

### LINKED ISSUE #111: Implement minimap
Should we limit the world terrain size for this? What if the world grows? The minimap scale changes accordingly to accomodate the whole world?

attached screenshot for reference
minimap
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/5766131-implement-minimap?utm_campaign=plugin&utm_content=tracker%2F371436&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F371436&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jfeo:
Here's my progress on a minimap.

There's probably a lot of style issues, etc. and I need to rebase some commits...

Right now, I think it's an issue that the Minimap registers itself with the engine for drawing, but then is deleted when the game is ended (dangling pointer?)

A good way to draw resources is also needed. Where are the colors/palette indices for them?

[Video](https://vid.me/Xx6B) for those who just want to see what it looks like.
implements and fixes #111


Jon0:
The game_main class should only include pure game state data, so its not the best owner for the minimap, I would suggest looking at main.cpp, line 42-44 the where the are other components, and similarly minimap can be made on the stack (no 'new' constructor), 

You would need to change the constructor parameters to `minimap(Engine *engine, ... size, ... hudpos)` The minimap wont need to have pointers to the unit container and terrain since these can be retrieved in the `minimap::on_drawhud()` call using `GameMain *game = engine.get_game()` note: that will return nullptr when there is no active game, in which case no minimap should be drawn.

if the result is not null you will just use `game->terrain`, and `game->placed_units`, as before.

franciscod:
oh the video looks awesome! nice work :D 

I almost forgot about gaben, and chuckled when he popped in the background :smile:

TheJJ:
If you perform your rebase, you might want to use `git rebase -m -i upstream/master`.

jfeo:
Is there any reason for me to keep the commit history of my own commits, or would it be okay to just squash it all into 1 commit? I think it would be easier for me

TheJJ:
Sure, you can squash them into one if you like. Most of the time my work builds up incrementally so it makes sense keeping the history (e.g. `"implemented ..."`, then `"engine: make use of ..."`, then `"bla: enable ... globally"`) so each step could be reverted easily some day.

jfeo:
Yes, I think I need to use a more structured approach in the future, instead of haphazardly adding everything and commiting :-P I'll just squash everything then.

libopenage/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ add_sources(libopenage
 	player.cpp
 	screenshot.cpp
 	texture.cpp
-
+  minimap.cpp
```
franciscod: whtitespace

libopenage/game_spec.cpp
```diff
@@ -345,6 +348,9 @@ void GameSpec::load_terrain(AssetManager &am) {
 		// TODO: terrain double-define check?
 		terrain_data.terrain_id_priority_map[terrain_id]  = line->blend_priority;
 		terrain_data.terrain_id_blendmode_map[terrain_id] = line->blend_mode;
+    terrain_data.terrain_id_map_color_hi_map[terrain_id] = line->map_color_hi;
+    terrain_data.terrain_id_map_color_med_map[terrain_id] = line->map_color_med;
+    terrain_data.terrain_id_map_color_low_map[terrain_id] = line->map_color_low;
```
franciscod: whitespace

assets/shaders/minimap.vert.glsl
```diff
@@ -0,0 +1,13 @@
...
+attribute vec4 vertex_position;
+attribute vec3 color;
+
+varying vec4 out_vertex_position;
+varying vec3 out_color;
+
+void main() {
+  out_vertex_position = vertex_position;
+  out_color = color;
+  gl_Position = gl_ModelViewProjectionMatrix * vertex_position;
```
TheJJ: hihi whitespace, tab indent pls

assets/shaders/minimap.frag.glsl
```diff
@@ -0,0 +1,20 @@
...
+  ivec2 center = ivec2(minimap_size.x/2, minimap_size.y/2);
+  ivec2 vpos = minimap_orig + center - ivec2(out_vertex_position.x, out_vertex_position.y);
+  vec3 color;
+
+  if (!(abs(vpos.x) * center.y + abs(vpos.y) * center.x <= center.x * center.y))
+    discard;
+  else
+    color = out_color;
+
+  gl_FragColor = vec4(color, 1.0);
```
TheJJ: tab indent pls.

libopenage/terrain/terrain.h
```diff
@@ -307,6 +311,22 @@ class Terrain {
...
+
+  /**
+   * get the red color component for a given terrain id.
+   */
+  uint8_t map_color_med(terrain_t terrain_id);
+
+  /**
+   * get the red color component for a given terrain id.
+   */
+  uint8_t map_color_low(terrain_t terrain_id);
```
TheJJ: tab indent pls

boombatower:
Does this predate the python interface and need to be updated for that approach or is this ready to be built on? Debating using this in referencing issue.

jfeo:
Hey @boombatower. I'm a bit unsure of what you mean by the python interface. I just pulled from master locally, and it works, so there's that :-) In regards to #11, I suppose it would make sense to perhaps use this as a base (perhaps make it more generic?).
Arbitrary sizes are already possible, like this:
2015-09-10-193505_800x600_scrot

TheJJ:
Used in #559 and integrated in the qt ui.

PR #345, opened 2015-08-12
convert: fixed checking for hd version availability

r00tler:
Fixed a bug where checking for the AOE2 HD version would always return true, leading to failed conversion of the original AOE2 game files.


TheJJ:
wheee typooooooooo

jfeo:
woopsie

PR #346, opened 2015-08-14
Fixes for compilation on Mac OS X

jprashanth:
libopenage no longer compiles with AppleClang. Something to do with thread_local storage not being supported.

Compilation with gcc 4.9 is successful.


TheJJ:
> In Mac OS X, the undefined symbols should not yield a linker error. These symbols will be resolved by the interpreter when the module is loaded.

They really should give an linker error. Otherwise if you forget to implement any function in `libopenage`, it will silently be ignored as it could be provided by some other library. The error will show up when launching, which is too late.

jprashanth:
That fixes all the compilation and linker issues. Now, the game runs.

jprashanth:
Rebase done.

libopenage/coord/tests.cpp
```diff
@@ -557,6 +557,10 @@ void camhud_0() {
 	// and top right of console as a toptight that is the camhud type
 	// is negative in its y and thus would be off the screen
 
+	// Currently, the constructor of Console invoke's ftgl font advance method.
+	// This can be an issue if OpenGL was not initialized.
+	// TODO Fix this issue before uncommenting the following block of code.
+	/*
 	console::Console con;
```
TheJJ: hehe, nope, we won't comment out code in the upstream repo.

This needs to be fixed properly.
If the console depends on initialized opengl, this is not correct and needs to be corrected.

jprashanth:
Removed the last commit. The coord tests concerning the Console are producing SIGSEGV on MacOSX. Due to this signal, make test is failing. Otherwise, this pull request is ready for merging.

Will open a new issue for console.

jprashanth:
#354

TheJJ:
Hmm, these 2 commits somehow overlap, probably not intended^^

jprashanth:
No, it was not intended. I made a mistake. I will squash the two commits together if needed.

buildsystem/python.cmake
```diff
@@ -169,7 +183,6 @@ function(add_cython_modules)
 				set(STANDALONE_NEXT FALSE)
 			else()
 				set_target_properties("${TARGETNAME}" PROPERTIES LINK_DEPENDS_NO_SHARED 1)
-				target_link_libraries("${TARGETNAME}" "${PYEXT_LINK_LIBRARY}")
```
TheJJ: can't you add

```cmake
# Since this module is not embedded with python interpreter,
# Mac OS X requires a link flag to resolve undefined symbols
if(APPLE and NOT EMBED_NEXT)
	set_target_properties("${TARGETNAME}" PROPERTIES LINK_FLAGS "-undefined dynamic_lookup" )
endif()
```

below and leave this line? should have the same effect.

jprashanth: even if EMBED_NEXT is TRUE, it would be set to FALSE by this line.

TheJJ: then set some other variable in the line where embed_next is still active, which then triggers the `dynamic_lookup` addition here. I just noticed that the `set_target_properties` might override already existing flags, we might want to prevent this.

jprashanth:
Can you check the changes now.

TheJJ:
This might work. Even flag combination is intact. Can you squash the commits please?

jprashanth:
Done.

TheJJ:
I'll call for testing in the IRC, I don't have a mac installation :)

jprashanth:
:+1:

TheJJ:
Nice work!

PR #349, opened 2015-08-15
'log-tail' in-game console command

tomolt:
It's an in-game console command that shows you the 8 latest log messages.
BTW, The log sink I created for this (TailSink) doesn't store any older messages than these 8 since openage's logs get big really fast and it would just litter the user's RAM.


TheJJ:
Thanks! If you aren't annoyed by this, please make minor adjustments to your [code style](https://github.com/SFTtech/openage/tree/master/doc/code_style) (your brace style).

For future planning, can you change the global tailsink so we can have any number of those sinks? We want to have any amount of per-unit log that shows up as a little box above a unit. And we also want to have multiple console windows.

TheJJ:
continued in #351

PR #351, opened 2015-08-16
'log-tail' in-game console command

tomolt:
It's an in-game console command that shows you the 8 latest log messages.
BTW, The log sink I created for this (TailSink) doesn't store any older messages than these 8 since openage's logs get big really fast and it would just litter the user's RAM.

P.S.:I've reuploaded this pull request because I accidentally pushed my master branch in the first request. Sorry!

### DIFF
--- a/libopenage/console/console.cpp
+++ b/libopenage/console/console.cpp
@@ -42,7 +42,7 @@ Console::Console()
 	topright.y = charsize.y * buf.dims.y;
 }
 
-Console::~Console () {}
+Console::~Console() {}
 
 void Console::load_colors(std::vector<gamedata::palette_color> &colortable) {
 	for (auto &c : colortable) {
@@ -59,6 +59,9 @@ void Console::register_to_engine(Engine *engine) {
 	engine->register_tick_action(this);
 	engine->register_drawhud_action(this);
 	engine->register_resize_action(this);
+
+	// create & register the TailSink
+	this->tailsink = std::make_unique<log::TailSink>();
 
 	// TODO bind any needed input to InputContext
 
@@ -112,6 +115,10 @@ void Console::write(const char *text) {
 void Console::interpret(const std::string &command) {
 	if (command == "exit") {
 		this->set_visible(false);
+	}
+	else if (command == "log-tail") {
+		for (auto &line : tailsink->tail)
+			this->buf.write(line.c_str());
 	}
 	else if (command == "list") {
 		Engine &e = Engine::get();
--- a/libopenage/console/console.h
+++ b/libopenage/console/console.h
@@ -14,6 +14,7 @@
 #include "../util/color.h"
 #include "../font.h"
 #include "../gamedata/color.gen.h"
+#include "../log/tail_logsink.h"
 
 namespace openage {
 
@@ -26,7 +27,6 @@ class Console : InputHandler, TickHandler, HudHandler, ResizeHandler {
 
 public:
 	Console();
-	~Console();
 
 	coord::camhud bottomleft;
 	coord::camhud topright;
@@ -43,6 +43,11 @@ public:
 
 	// the command state
 	std::string command;
+
+	/**
+	 * the log sink used for the 'log-tail' command.
+	 */
+	std::unique_ptr<log::TailSink> tailsink;
 
 	/**
 	 * load the consoles color table
--- a/libopenage/log/logsource.cpp
+++ b/libopenage/log/logsource.cpp
@@ -6,6 +6,7 @@
 
 #include "logsink.h"
 #include "stdout_logsink.h"
+#include "tail_logsink.h"
 
 namespace openage {
 namespace log {
--- a/libopenage/log/tail_logsink.cpp
+++ b/libopenage/log/tail_logsink.cpp
@@ -1,0 +1,35 @@
+// Copyright 2015-2015 the openage authors. See copying.md for legal info.
+
+#include "tail_logsink.h"
+
+#include <sstream>
+#include <cstring>
+#include <iostream>
+
+#include "message.h"
+#include "logsource.h"
+
+namespace openage {
+namespace log {
+
+
+void TailSink::output_log_message(const message &msg, LogSource *source) {
+	// write log to str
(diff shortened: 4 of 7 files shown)


tomolt:
From the first upload:

 TheJJ commented 13 hours ago

Thanks! If you aren't annoyed by this, please make minor adjustments to your code style (your brace style).

For future planning, can you change the global tailsink so we can have any number of those sinks? We want to have any amount of per-unit log that shows up as a little box above a unit. And we also want to have multiple console windows.

libopenage/log/logsource.cpp
```diff
@@ -21,6 +22,9 @@ void LogSource::log(const message &msg) {
 	// (and thus at least one sink exists).
 	global_stdoutsink();
 
+	// HACK tailsink should probably be initialized somewhere else.
+	global_tailsink();
```
TheJJ: The tailsink should be an independent class which is a member of the console. the console then registers itself at the logger to provide the tailsink. You can avoid global state by that way.

libopenage/log/tail_logsink.h
```diff
@@ -0,0 +1,31 @@
...
+#include "logsink.h"
+
+namespace openage {
+namespace log {
+
+
+class TailSink : public LogSink {
+public:
+	const static unsigned int tail_size = 8;
+	char *tail[tail_size] = {};
```
TheJJ: can you use stl-types for that. some queue might be suitable, e.g. http://en.cppreference.com/w/cpp/container/queue

TheJJ:
You can use rebase to [meld the bug-fixing commits into other ones](https://github.com/SFTtech/openage/blob/master/doc/contributing.md#rebasing---what-the-hell-is-it-and-why-do-i-need-it) btw.

tomolt:
Okay, now TailSink is a member of console and uses std::list (which is easier to iterate through than std::queue). :)

libopenage/console/console.cpp
```diff
@@ -60,6 +63,9 @@ void Console::register_to_engine(Engine *engine) {
 	engine->register_drawhud_action(this);
 	engine->register_resize_action(this);
 
+	// create & register the TailSink
+	tailsink = new log::TailSink();
```
TheJJ: to simplify this, you can make the class member a `std::unique_ptr<log::TailSink>` and initialize it with `this->tailsink = std::make_unique<log::TailSink>();`
Then you don't need the delete in the `~`.

libopenage/log/tail_logsink.cpp
```diff
@@ -0,0 +1,34 @@
...
+	ss << msg.lvl->name << "|";
+	ss << source->logsource_name() << "|";
+	ss << msg.filename << ":" << msg.lineno << "|";
+	ss << "T" << msg.thread_id << "|";
+	ss << std::setprecision(7) << std::fixed << msg.timestamp / 1e9 << "|";
+	ss << msg.text << std::endl;
+	std::string str = ss.str();
+
+	if(tail.size() > tail_size)
+		tail.pop_front();
```
TheJJ:
```cpp
	  v space             brace  v 
	if (tail.size() > tail_size) {
		tail.pop_front();
	}
```

libopenage/log/tail_logsink.h
```diff
@@ -0,0 +1,27 @@
...
+namespace log {
+
+
+class TailSink : public LogSink {
+public:
+	const static unsigned int tail_size = 8;
+	std::list<std::string> tail;
+
+private:
+	virtual void output_log_message(const message &msg, LogSource *source) override;
```
TheJJ: as you already specify the override keyword, the `virtual` is unnecessary.

libopenage/console/console.cpp
```diff
@@ -113,6 +119,10 @@ void Console::interpret(const std::string &command) {
 	if (command == "exit") {
 		this->set_visible(false);
 	}
+	else if (command == "log-tail") {
+		for (auto it = tailsink->tail.begin(); it != tailsink->tail.end(); it++)
+			this->buf.write((*it).c_str());
```
TheJJ: braces here as well pls, and you can use `auto &it : this->tailsink`. and `it->c_str()`.

tomolt:
Okay, working on it

TheJJ:
So, now for this one. First i'd propose you rebase this branch against the master where we just merged the other pull. `for (auto &line : tailsink->tail)` wants some braces, otherwise it looks pretty ok for now. But i'll do a more "in-depth" review when you rebased.

libopenage/console/console.h
```diff
@@ -26,7 +27,6 @@ class Console : InputHandler, TickHandler, HudHandler, ResizeHandler {
 
 public:
 	Console();
-	~Console();
```
TheJJ: We should keep the console destructor and implement it as an empty function.

libopenage/log/tail_logsink.cpp
```diff
@@ -0,0 +1,35 @@
...
+#include "message.h"
+#include "logsource.h"
+
+namespace openage {
+namespace log {
+
+
+void TailSink::output_log_message(const message &msg, LogSource *source) {
+	// write log to str
+	// (mainly copied from StdOutSink).
```
TheJJ: If it's just copied, can we unify both somehow?

tomolt:
Well, no, because we might want different output formats in the end...

TheJJ:
How's it going? Can we assist you somehow?

TheJJ:
Are you still interested in bringing this upstream?

tomolt:
Oh, sorry for completely forgetting that :/
If you want to I could still bring this upstream, but I dont know if this really fits with the direction you want to bring the in-game console in; From what I recall, you are more interested in either having a fully-fledged terminal emulator or an interactive python console than your own basic command interpreter ...

TheJJ:
Yes, we want a real terminal emulator. Not sure if we need a full terminal for the debug text boxes above unit heads, but it needs to be rendered in any case. Because of that I think you are right and we should not duplicate that scrollback. Except if we implemented something like `less` to do independent paging. If you wanna go for it, would be awesome!

PR #352, opened 2015-08-16
in-game console UTF-8 backspace support

tomolt:
Supports UTF-8 characters and line-wrapping just fine.

### DIFF
--- a/libopenage/console/buf.cpp
+++ b/libopenage/console/buf.cpp
@@ -385,6 +385,29 @@ void Buf::write(char c) {
 	}
 }
 
+void Buf::pop_last_char() {
+	if (this->cursorpos.x > 0) {
+		if (this->cursor_special_lastcol) {
+			this->cursor_special_lastcol = false;
+		}
+		else {
+			this->cursorpos.x -= 1;
+		}
+
+		buf_char *ptr = this->chrdataptr(this->cursorpos);
+		*ptr = this->current_char_fmt;
+		ptr->cp = ' ';
+
+		if (this->cursorpos.x == 0 && this->linedataptr(this->cursorpos.y - 1)->type == LINE_WRAPPED) {
+			this->linedataptr(this->cursorpos.y)->type = LINE_EMPTY;
+			this->cursorpos.y--;
+			this->linedataptr(this->cursorpos.y)->type = LINE_REGULAR;
+			this->cursorpos.x = this->dims.x - 1;
+			this->cursor_special_lastcol = true;
+		}
+	}
+}
+
 void Buf::process_codepoint(int cp) {
 	//if the terminal is currently in escaped state, tread the codepoint as
 	//part of the current escape sequence.
--- a/libopenage/console/console.cpp
+++ b/libopenage/console/console.cpp
@@ -5,6 +5,7 @@
 #include "../log/log.h"
 #include "../error/error.h"
 #include "../util/strings.h"
+#include "../util/unicode.h"
 
 #include "draw.h"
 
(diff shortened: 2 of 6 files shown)


libopenage/console/console.cpp
```diff
@@ -81,13 +81,26 @@ void Console::register_to_engine(Engine *engine) {
...
-			this->command = "";
-			return true;
-		}
-		return false;
+		switch (arg.e.as_char()) {
+			case 8: // remove a single UTF-8 character
+				if (command.size() > 0) {
+					// remove all the extra UTF-8 bytes
+					while((command.back() & 0x80) && !(command.back() & 0x40))
+						command.pop_back();
```
TheJJ: this is kinda redundant with the buf utf8 char dropping. can you unify that? maybe even put a function in `util/unicode.h` so any unicode string could benefit from this deletion extension?

libopenage/console/buf.cpp
```diff
@@ -385,6 +385,28 @@ void Buf::write(char c) {
...
+			ptr->cp = ' ';
+			this->cursorpos.x--;
+			ptr = this->chrdataptr(this->cursorpos);
+		}
+		*ptr = this->current_char_fmt;
+		ptr->cp = ' ';
+	} else if (this->linedataptr(this->cursorpos.y-1)->type == LINE_WRAPPED) {
+		this->cursorpos.y--;
+		this->cursorpos.x = this->dims.x;
+		pop_last_char();
```
TheJJ: `this->pop_last_char()`

TheJJ:
I don't know if you know and how this behaves now, but when you fill a whole line in urxvt or xterm, then the last char will keep the cursor on the same line even though its full now. We might wanna test what happens if we now delete something (then the cursorpos stays the same i think but the char is deleted). If we're in the next line then, and delete, what happens then?

@mic-e might know this already.

tomolt:
*"this is kinda redundant with the buf utf8 char dropping. can you unify that? maybe even put a function in util/unicode.h so any unicode string could benefit from this deletion extension?"*

Okay, I'm trying to factor it into a utility function. The only problem is that (as far as i know) the console buffer doesn't store the characters one after another but always as a pair of a character and some "options" like terminal color etc.

*"I don't know if you know and how this behaves now, but when you fill a whole line in urxvt or xterm, then the last char will keep the cursor on the same line even though its full now"*

I didn't know our console does this too, so my code doesn't check for it, which means that if you press backspace at the EOL, it will remove the *second last* character and move the cursor back by 1, leaving the last character at the end of the line. I'll change it so that the cursor will stay and the last char will be deleted if the cursor is at the EOL.

*"If we're in the next line then, and delete, what happens then?"*

Then the cursor will jump back to the end of the previous line.

TheJJ:
The goal is to implement a full urxvt/xterm compatible term here, so yea it should behave the same i guess. The console code so far is by @mic-e, i just remember that weird newline stuff. You might wanna check if the behavior is what you expect.

Vim partially relies on this feature i remember barely.
You can use the non-opengl-renderer (the `console::tests::interactive` demo), it just prints out the calculated buffer to your already-functional terminal.

tomolt:
Okay, I just discovered something: On my system (Arch Linux), xterm + bash doesn't even have this feature! If the line is full, the cursor will directly jump to the next line.

TheJJ:
bash prevents it. run `cat` and then fill up the line. you will notice what happens at line's end.

tomolt:
Okay, you're right. Too bad. I already thought openage's in-game console would one day be more advanced than GNU's bash that already exists for almost 3 decades! :)

TheJJ:
I'd really wish if we could e.g. establish better escape codes or something. But we kinda have to keep the damn legacy shit so that vim continues to work :grin:

tomolt:
Okay, I've completely reworked the whole feature;
1. I've refactored the whole UTF-8 character size handling into the two functions *unicode::utf8_last_char_size* and *unicode::utf8_pop_back*.
2. it now supports this EOL special cursor state.
3. I've removed some dead code I accidentally wrote ...

libopenage/util/unicode.cpp
```diff
@@ -149,4 +151,20 @@ size_t utf8_encode(int cp, char *outbuf) {
...
 
+size_t utf8_last_char_size(char *str) {
+	int r = 0;
+	int i = strlen(str) - 1;
+	while ((i >= 0) && (str[i] & 0x80) && !(str[i] & 0x40)) {
+		i--;
+		r++;
+	}
+	if(i >= 0)
+		r++;
```
TheJJ: `if (bla) {` please :)

TheJJ:
Nice. This looks much cleaner and less redundant now, well done.

tomolt:
*"(btw the bash has nothing to do with the terminal. bash is a program being displayed by the terminal, and we will implement just the terminal. What runs in there, the ingame python console or vim or whatever, sends special codes so that the buffer is arranged accordingly)"*
Yes I know, I just derped there ...

TheJJ:
kay^^ if you now merged all of the commits into one (because all of them are fixups only), we cann pull it.

tomolt:
Okay ...

tomolt:
Okay, rebased it! :)

TheJJ:
Whee awesome!

PR #287, opened 2015-04-21
[WIP] New renderer

TheJJ:
This is basically a totally new graphics engine, designed for our needs.
- [X] Abstraction of render backend
- [X] OpenGL 3.3 requirement
- [X] Preparations for Vulkan (we don't know much..)
- [X] File watching subsystem
- [X] `constexpr` key-value map
- [X] Screenshot support
  - [ ] optimizations (maybe [this](http://www.songho.ca/opengl/gl_pbo.html#pack) or [this](http://www.opengl-tutorial.org/intermediate-tutorials/tutorial-14-render-to-texture/))
- [ ] Render pipeline abstraction
  - [X] Textures
  - [X] Shaders
  - [x] Compiletime shader variable access
    - [x] Uniforms
    - [x] Vertex attributes
    - [x] Dynamic buffer uploading
  - [ ] Samplers
  - [ ] Materials
  - [ ] Geometries
- [x] Quaternions
- [ ] Texture position transformations without reuploading vertex buffer
- [ ] Texture packing
  - [ ] Efficient binpacking
  - [ ] 3D Textures
- [ ] Separate renderer thread and other event loop improvements
- [ ] Efficient render task queuing (minimize shader/texture changes)
  - [ ] Viewport clipping
- [ ] Smooth zooming
- [ ] Enable Z-Buffering
- [ ] Geometry drawing
  - [ ] squares
  - [ ] circles
  - [ ] smooth paths
  - [ ] boxes (for the gui)
  - [ ] text rendering (use freetype / harfbuzz / pango)
- [ ] Perspective translation (e.g. the unit-selected-circle)
- [ ] Terrain rendering
  - [ ] Terrain texture merging to single square tile
  - [ ] Perspective rendering
  - [ ] Blending cache
  - [ ] Terrain blending in the shader (in perspective to easily split tiles)
- [ ] Outline rendering with obstruction dual-pass (z-flip)
- [ ] [Occlusion](https://a248.e.akamai.net/f/248/10/10/http.developer.nvidia.com/GPUGems/gpugems_ch29.html) [queries](http://vertostudio.com/gamedev/?p=177)
- [ ] Pixel-perfect unit hitbox for unit selection and damage areas

implements #34, #286 and ideas from #153 
closes #264 (file watching subsystem).
will do #149 (terrain blending in shader), #158 /  #154 (blending cache), #141 (terrain perspective adjustments), #359 (outlines), #368 (unit selection box), #374 (texture clipping incorrect).

[...]

PR #357, opened 2015-08-21
Text renderer with support for colors

jprashanth:


libopenage/engine.cpp
```diff
@@ -222,6 +222,8 @@ Engine::Engine(util::Dir *data_dir, const char *windowtitle)
 	bind_player_switch(input::action_t::SWITCH_TO_PLAYER_6, 6);
 	bind_player_switch(input::action_t::SWITCH_TO_PLAYER_7, 7);
 	bind_player_switch(input::action_t::SWITCH_TO_PLAYER_8, 8);
+
+	this->text_renderer.reset(new renderer::TextRenderer());
```
TheJJ: `this->text_renderer = std::make_unique<renderer::TextRenderer>()`

TheJJ:
Nice work! We can merge it soon, i'll integrate it into #287 then but it is functionaly standalone.

libopenage/renderer/text_renderer.h
```diff
@@ -0,0 +1,56 @@
...
+	void render();
+
+private:
+	struct TextRenderBatchPass {
+		int x;
+		int y;
+		std::string text;
+	};
+
+	struct TextRenderBatch {
```
TheJJ: One small thing left: struct names are lowercase-underscored. `text_render_batch` and `text_render_batch_pass`.

Would also be nice if you added some `/** documentation */` for the members of this class.

libopenage/renderer/color.h
```diff
@@ -0,0 +1,33 @@
...
+
+	uint8_t r;
+	uint8_t g;
+	uint8_t b;
+	uint8_t a;
+
+};
+
+bool operator==(const Color &left, const Color &right);
+bool operator!=(const Color &left, const Color &right);
```
TheJJ: These should also be members ov the Color class:
`class Color { bool operator==(const Color &other); };`

libopenage/engine.cpp
```diff
@@ -278,22 +280,28 @@ void Engine::end_game() {
...
 	);
 
+	this->text_renderer->set_color(renderer::Color{0, 255, 0, 255});
 	// Draw version string in the lower left corner
 	this->render_text(
 		{5, 35}, 20,
 		"openage %s", config::version
 	);
+
+	this->text_renderer->set_color(renderer::Color{0, 0, 255, 255});
```
TheJJ: I'm not sure if we should color the on-screen texts all differently :smile:, leaving them white might be a bit better for readability.

libopenage/renderer/text_renderer.cpp
```diff
@@ -0,0 +1,114 @@
+// Copyright 2015-2015 the openage authors. See copying.md for legal info.
+
+#include "text_renderer.h"
+
+#include <algorithm>
+#include <epoxy/gl.h>
```
TheJJ: there's no opengl call in that file, so no need for the epoxy header.

jprashanth: glColor4f is being used in the render method.

libopenage/engine.cpp
jprashanth: Yes, just demonstrating the usage.

libopenage/renderer/text_renderer.cpp
TheJJ: you're right, sorry.

jprashanth:
Completed the required changes.

TheJJ:
Thanks! Cool stuff.

Issue #356, opened 2015-08-20
No configure check for pygments


LordAro:
Not having it installed results in a make error, rather than a configure error
This is different to pil & numpy

I should think this is very easy to fix :D

jprashanth:
this issue can be closed now.

PR #372, opened 2015-08-27
Attempt at cleaning up buildsystem

jprashanth:
Buildsystem restructuring and cleanup.


buildsystem/modules/HandleOpenageCXXOptions.cmake
```diff
@@ -0,0 +1,151 @@
...
+		include(ProcessorCount)
+		ProcessorCount(N)
+		if(NOT N EQUAL 0)
+			set_cxx_flags("-flto=${N}")
+			set_linker_flags("-flto=${N}")
+		endif()
+	endif()
+endif()
+
+if(NOT OPENAGE_SANITIZE_MODE)
```
TheJJ: should be `${oa san mode}` i think as you're evaluating it?

TheJJ:
I like the changes, well done!

I think you may even remove the openage-specificness from the `HandleOpenageCXXOptions.cmake` and name it `HandleCXXOptions.cmake` then.

buildsystem/modules/HandleOpenageCXXOptions.cmake
jprashanth: That line is checking if `OPENAGE_SANITIZE_MODE` is defined. If it is not defined, i am defaulting it to `none`.

jprashanth:
Yea, i will remove the openage-specificness.

buildsystem/modules/HandleOpenageCXXOptions.cmake
TheJJ: ok then^^

jprashanth:
30c3cfeef08a2693d362e846977507f0529db7dd fixes #356

TheJJ:
Your buildsystem cleanup is really nice, keep up the good work!

jprashanth:
Thanks @TheJJ . I have made some changes regarding the CMake Modules idea pointed out before.

TheJJ:
Please notify me when i should have another look at this and then merge it.

jprashanth:
Hi @TheJJ , I think i am done with the refactoring for now. You can go ahead and review the commits.

buildsystem/Doxygen.cmake
```diff
@@ -0,0 +1,66 @@
...
+set(DOXYFILE_TEMPLATE "${BUILDSYSTEM_DIR}/templates/Doxyfile.in")
+set(DOCS_ROOT_DIRECTORY "${CMAKE_BINARY_DIR}/docs")
+
+function(add_doxygen_target TARGET TARGET_SRC_DIR)
+	if(NOT DOXYGEN_FOUND)
+		return()
+	endif()
+
+	set(DOXYFILE "${CMAKE_BINARY_DIR}/${TARGET}_Doxyfile")
+	set(OUTPUT_DIRECTORY "${DOCS_ROOT_DIRECTORY}/${TARGET}")
```
TheJJ: How does that turn out to be? Are the html and latex docs stored to this folder then?
If yes we need to update the doc readme file and make sure the folder clutter is not too big and make clean works as expected.

buildsystem/Doxygen.cmake
```diff
@@ -0,0 +1,66 @@
...
+		VERBATIM
+	)
+	add_dependencies(doc ${TARGET}_docs)
+endfunction()
+
+# If disabled, don't even bother going forward
+if(NOT GENERATE_DOCS)
+	add_custom_target(doc
+		COMMAND echo "openage was configured with GENERATE_DOCS option set to OFF."
+		COMMAND echo "Reconfigure the project with this option set as ON to be able to generate docs."
```
TheJJ: the `echo` is not wrong, but may lead to problems and difficulties in porting. Is there any `cmake -E touch` like variant for printing messages?

TheJJ: Ah, yes it is: `cmake -E echo` :)

jprashanth:
The cpp files generated from the pyx files depend on the `pxdgen` target. Therefore, these targets are only built after all the pxd files were generated.
 
You are right about having hundreds of cmake targets in the future. I implemented this functionality hoping mic-e's addition to cmake -  https://github.com/SFTtech/openage/blob/master/CMakeLists.txt#L94 will be merged soon.

buildsystem/Doxygen.cmake
jprashanth: Yes, you are right. I will change this to `${CMAKE_COMMAND} -E echo`

buildsystem/Doxygen.cmake
jprashanth: This implementation is almost same as before. The output folders will now become `bin/docs/libopenage` and `bin/docs/openage` instead of `bin/doc`. If this is ok, i can modify the doc readme file.

TheJJ:
Yea, it seems to be merged now:
http://www.cmake.org/pipermail/cmake-developers/2015-July/025676.html
http://www.cmake.org/gitweb?p=cmake.git;a=commitdiff;h=1d398478

buildsystem/Doxygen.cmake
TheJJ: Ok. Please update `doc/README.md`, i think that's the only one referencing to the location.
Don't you think we should merge both folders into one as it was before? That way python and c++ docs will be put into the website and latex files.

jprashanth: In my opinion, libopenage and openage should have different doc folders. But if you think otherwise, the only change required in this implementation would be to change https://github.com/jprashanth/openage/blob/buildsystem/CMakeLists.txt#L75-L76 to `add_doxygen_target("openage" libopenage/ openage/)`. This would create both the python and c++ docs together in the folder `bin/docs/openage`

TheJJ: It's not that i want them in a single folder, I want them in the same doxygen output.
Maybe (I'm not sure) it is better to have totally separate doxygen invocations, but when we combine them it may lead to name clashes between c++ and python names.

What do you think? Should the same webpage include docs for both py and c++?

jprashanth: Maybe it is better to have two different doxygen invocations and have a webpage link them both.
```
bin/
    docs/
        libopenage/
        openage/
        index.html
```
`index.html` will just contains links to both the documentation.

TheJJ:
So `__pycache__` is no longer installed. That's bad, we want to ship the compiled files to the filesystem.

jprashanth:
Ok, i will revert those changes or split this pull request. I am busy for a few days. I can make the changes after that.

jprashanth:
@TheJJ I have moved the `debatable`, `controversial`, and `in progress` commits to a new branch.  I will open a separate pull request for these commits.

You can go ahead and review this pull request.

TheJJ:
Haha nice, thanks! I will look into it soon.

TheJJ:
After reading through it again, I think it's ready to merge, any other opinions? :+1:

jprashanth:
I think it is complete with no more additions in this pull request. The
other pull request I have in plan will need some looking into. I will open
that pull request in a week. Busy for two days.

Issue #379, opened 2015-09-14
Python 3.4 library not found on Mac OS X.


darrenstrash:
**System setup:**
Mac OS X Yosemite 10.10.4 (14E46)
Python 3.4 installed with `brew install python3`
Using g++ version 5.2.0

**The symptom:** When building the "run" executable, the linker fails with error `ld: library not found for -lpython3.4`

My investigation reveals the following:

**The superficial cause:** Python libraries are not copied to system library directories in Mac installs. And no `-Lpath` compiler option is used for compiling "run".

**The root cause:** The distutils python package is currently being erroneously used to grab the C/C++ library information. In fact, distutils is currently returning the location of libraries to *import* into python scripts. It's a coincidence that the real library name has the same name `python3.4`. Therefore, since it wasn't designed for it, distutils can't be used to the find the library's location...

**Recommendation:** We use the python configuration tool to tell us the exact linker arguments instead: (e.g., with `python$(VERSION)-config --ldflags`). While we are at it, we could grab the cflags with `--cflags` as well. This way we are always using the appropriate flags that the Python package builders meant for us to use when linking against Python. 

**Drawbacks:** This solution works with the Homebrew Python 3.4 installation on Mac OS X. However, I have observed the ldflags can sometimes contain extra -lfile libs that are not currently included in our build system, which can then lead to a similar error. This was observed with the MacPorts installation.

franciscod:
excellent investigation report!

TheJJ:
Would you mind doing the changes and submitting a pull request? :stuck_out_tongue: 

Your explanation sounds reasonable, but mainly the ldpath has to be extended, right? (I mean that the `-lpython3.4m` or -lpython3.4` is correct, but that our `-L"/somewhere"` is not sufficient at the moment)

darrenstrash:
Of course! I'll make the changes. Please assign it to me :-)

TheJJ:
Go for it. Do you wanna be "really" assigned or is it ok to tell you you're now assigned? :wink:

darrenstrash:
Either way is fine. Making it official might be the best way to avoid duplicate work.

PR #384, opened 2015-09-16
Include -L<dir> flag for linking against python

darrenstrash:
This is a proposed fix for issue #379, in which the python library could not be found by the linker. Now we are grabbing the correct library directory using distutils, and also generating the library name the same way as the python-config tool. These are then given to the linker with -L and -l, respectively.

In the future, we should use the python-config tool to grab all ldflags and cflags. I currently don't use the python-config tool because it can include many flags (e.g, extra libraries) that we simply aren't ready to incorporate yet, and therefore would require much filtering. Since this is not currently a problem, it's a feature, not a bug.

I tested the fix on:

Mac OS X Yosemite 10.10.4 (14E46)
Python 3.4 installed with brew install python3
Using g++ version 5.2.0

and

Ubuntu 14.0.4-3
Python 3.4 installed with apt-get
Using both g++ version 4.9.2, and clang++ version 3.6.0

### DIFF
--- a/buildsystem/modules/FindPython.cmake
+++ b/buildsystem/modules/FindPython.cmake
@@ -6,10 +6,10 @@
 #
 # This file defines the following variables:
 #
-# PYTHON_FOUND       - True when python was found.
-# PYTHON             - The full path to the Python interpreter.
-# PYTHON_INCLUDE_DIR - Include path for Python extensions.
-# PYTHON_LIBRARY     - Linker flags for Python extensions.
+# PYTHON_FOUND        - True when python was found.
+# PYTHON              - The full path to the Python interpreter.
+# PYTHON_INCLUDE_DIR  - Include path for Python extensions.
+# PYTHON_LIBRARY      - Library and Linker options for Python extensions.
 #
 # Also defines py_exec and py_get_config_var.
 #
@@ -32,7 +32,6 @@ function(py_exec STATEMENTS RESULTVAR)
 	set("${RESULTVAR}" "${PY_OUTPUT}" PARENT_SCOPE)
 endfunction()
 
-
 function(py_get_config_var VAR RESULTVAR)
 	# uses py_exec to determine a config var as in distutils.sysconfig.get_config_var().
 	py_exec(
@@ -41,6 +40,21 @@ function(py_get_config_var VAR RESULTVAR)
 	)
 
 	set("${RESULTVAR}" "${RESULT}" PARENT_SCOPE)
+endfunction()
+
+function(py_get_lib_name RESULTVAR)
+	# uses py_exec to compute Python's C/C++ library name, just like python-config does.
+	py_get_config_var(VERSION PYTHON_VERSION)
+	if(NOT "${PYTHON_VERSION}" VERSION_LESS "3.2")
+		py_exec(
+			"import sys; print(sys.abiflags)"
+			ABIFLAGS
+		)
+	else()
+		set(ABIFLAGS, "")
+	endif()
+
+	set("${RESULTVAR}" "python${PYTHON_VERSION}${ABIFLAGS}" PARENT_SCOPE)
 endfunction()
 
 function(find_python_interpreter_builtin)
@@ -112,13 +126,9 @@ endforeach()
 foreach(PYTHON ${PYTHON_INTERPRETERS})
 	# ask the interpreter for the essential extension-building flags
 	py_get_config_var(INCLUDEPY PYTHON_INCLUDE_DIR)
-	if(APPLE)
-		py_get_config_var(DESTLIB PYTHON_LIBRARY_PATH)
-		get_filename_component(PYTHON_LIBRARY ${PYTHON_LIBRARY_PATH} NAME)
-		get_filename_component(PYTHON_LINK_DIRECTORIES ${PYTHON_LIBRARY_PATH} DIRECTORY)
-	else()
-		py_get_config_var(BLDLIBRARY PYTHON_LIBRARY)
-	endif()
+	py_get_config_var(LIBPL PYTHON_LIBRARY_DIR)
+	py_get_lib_name(PYTHON_LIBRARY_NAME)
+	set(PYTHON_LIBRARY "-L${PYTHON_LIBRARY_DIR} -l${PYTHON_LIBRARY_NAME}")
 
 	# there's a static_assert that tests the Python version.
 	try_compile(PYTHON_TEST_RESULT
@@ -127,7 +137,6 @@ foreach(PYTHON ${PYTHON_INTERPRETERS})
 		LINK_LIBRARIES "${PYTHON_LIBRARY}"
 		CMAKE_FLAGS
 			"-DINCLUDE_DIRECTORIES=${PYTHON_INCLUDE_DIR}"
-			"-DLINK_DIRECTORIES=${PYTHON_LINK_DIRECTORIES}"
 		OUTPUT_VARIABLE PYTHON_TEST_OUTPUT
 	)
 
--- a/buildsystem/python.cmake
+++ b/buildsystem/python.cmake
@@ -64,8 +64,8 @@ function(python_init)
 	set(CYTHON_VERSION "${CYTHON_VERSION}" PARENT_SCOPE)
 
 	set(PYEXT_CXXFLAGS "${PYEXT_CXXFLAGS}" PARENT_SCOPE)
-	set(PYEXT_PYLIB "${PYTHON_LIBRARY}" PARENT_SCOPE)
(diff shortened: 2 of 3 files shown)


buildsystem/modules/FindPython.cmake
```diff
@@ -43,6 +43,29 @@ function(py_get_config_var VAR RESULTVAR)
...
+function(py_get_lib_name RESULTVAR)
+	# uses py_exec to compute Python's C/C++ library name, just like python-config does.
+	py_get_config_var(VERSION PYTHON_VERSION)
+	if(NOT "${PYTHON_VERSION}" VERSION_LESS "3.2")
+		set(USE_ABIFLAGS "use")
+	else()
+		unset(USE_ABIFLAGS)
+	endif()
+
+	set(BASE_STATEMENT "import sys; from distutils.sysconfig import get_config_var; print(\"python\" + get_config_var(\"VERSION\")")
```
TheJJ: Hmm isn't this what `py_get_config_var` can do?
If not, you can extend that function to purge the redundant usage of `distutils.sysconfig`.

buildsystem/modules/FindPython.cmake
```diff
@@ -43,6 +43,29 @@ function(py_get_config_var VAR RESULTVAR)
 	set("${RESULTVAR}" "${RESULT}" PARENT_SCOPE)
 endfunction()
 
+function(py_get_lib_name RESULTVAR)
+	# uses py_exec to compute Python's C/C++ library name, just like python-config does.
+	py_get_config_var(VERSION PYTHON_VERSION)
+	if(NOT "${PYTHON_VERSION}" VERSION_LESS "3.2")
+		set(USE_ABIFLAGS "use")
```
TheJJ: we don't support `python<3.3`, so you can remove that test.

darrenstrash: True that we don't support it. However, if we invoke this on an older version of python (which will likely be called, since we are searching through all interpreters) then python fails because it can't find sys.abiflags. This behavior changed in version 3.2

buildsystem/modules/FindPython.cmake
darrenstrash: Yes, I can clean it up, and just make a call to collect sys.abiflags.

buildsystem/modules/FindPython.cmake
TheJJ: Why should it ever be called? we explicitly search for `py>=3.4` which will already fail if not found.

darrenstrash: I should note that I observed this failure in my testing, which is why I put the check in.

darrenstrash: The code searches through all python interpreters, regardless of version. For each one, the FindPython_test.cpp test file is compiled and the version number is extracted from that. This can probably be cleaned up too.

Therefore, it's not known a priori if it is the correct version.

jprashanth: @TheJJ , currently we are creating a list of all the python interpreters and try to compile a python extension using each interpreter. The version check for python is in the form of a static assert in `FindPython_test.cpp`.

TheJJ: Ah I see, you're right. Mind that #372 does a lot of fixing for the buildsystem. Can you test @jprashanth's branch if the error is maybe even fixed in there already?

jprashanth: Also, CMake has an in-built module called `FindPythonLibs`. We can investigate this for our purpose.

darrenstrash: @TheJJ As far as I can tell, @jprashanth's branch has the same error. 

@jprashanth FindPythonLibs sounds like a good idea. I can look into that.

darrenstrash: Seems that there is a long history of FindPythonLibs not working with Homebrew. https://github.com/Homebrew/homebrew/issues/25118

Still might be worth a try though...

darrenstrash: cmake has an open issue for this: http://www.itk.org/Bug/view.php?id=14809

Let's stick to the current solution for now.

darrenstrash:
I would recommend that we push the FindPythonLibs cmake work until later (as a refactoring of the build system), and stick with this solution for now.

Has anyone tested this solution on their system?

TheJJ:
So you mean merge this pull and then adapt the build system refactor pull request?
This pull seems to work on travis at least, so i guess it will work well for Linux, but I will test that before merging.

darrenstrash:
Yes, I would recommend merging this for now, since the Mac OS X build is currently broken. I can investigate FindPythonLibs in the meantime, unless @jprashanth would like to.

mic-e:
I wonder why the previous way of determining the Python linker flags didn't work properly; after all, that's the way that all those `distutils` `setup.py` scripts are using.
I took that directly from the distutils code in order to be as cross-platform as possible.

I wonder why it has become necessary to return both `LIBRARY_DIR` and `LIBRARY_NAME` instead of pre-cooked linker flags, as seems to be common practice with cmake modules.

darrenstrash:
This solution still uses distutils, but grabs different configuration variables. We need the library directory since Mac doesn't install into the system's lib directory. 

I've not seen the setup.py solutions, but my understanding is that the variable we are currently grabbing (**DESTLIB**) is for the location of python libraries that can be imported into scripts. Perhaps that's what the setup.py scripts need?

mic-e:
Maybe I've been accidentally grabbing the wrong config variable (buying a Mac just to test openage builds isn't in my budget right now, and no OSX user was around...).

The thing about the separate variables is that I'd prefer seeing them condensed into one single variable because that "seems to be how it's done". I might be wrong there too, but most other modules seem to pass the path - if applicable - as a "-L" flag inside the LIBRARY variable.

buildsystem/python.cmake
```diff
@@ -137,7 +138,8 @@ function(add_cython_modules)
 				set_property(GLOBAL APPEND PROPERTY SFT_CYTHON_MODULES_EMBED "${source}")
 				add_executable("${TARGETNAME}" "${CPPNAME}")
 
-				target_link_libraries("${TARGETNAME}" "${PYEXT_PYLIB}")
+				# TODO: use full ldflags and cflags provided by python${VERSION}-config
+				target_link_libraries("${TARGETNAME}" "-L${PYEXT_LIBRARY_DIR} -l${PYEXT_LIBRARY_NAME}")
```
mic-e: Why don't you assemble this string into a traditional `PYTHON_LIBRARY` variable back in `FindPython.cmake`?

darrenstrash: I can do that. In that case though, I would prefer to call it PYTHON_LDFLAGS, since it is no longer a library name.

mic-e: In the other modules, the variables have names like

    FTGL_LIBRARIES
    BACKTRACE_LIBRARY
    OPUS_LIBRARIES
    SDL2_LIBRARIES

Even though they clearly contain linker flags, not (file)names. I agree that it doesn't really make sense, so... choose however you wish, but wisely.

darrenstrash: I agree, thanks for all of your feedback. To be consistent, I'll use `*_LIBRARY` variable name. It's also unclear why `target_link_libraries` should even take flags...given the name.

mic-e:
Looking good now :+1:

darrenstrash:
Thanks!

buildsystem/modules/FindPython.cmake
```diff
@@ -112,13 +126,9 @@ endforeach()
...
-	if(APPLE)
-		py_get_config_var(DESTLIB PYTHON_LIBRARY_PATH)
-		get_filename_component(PYTHON_LIBRARY ${PYTHON_LIBRARY_PATH} NAME)
-		get_filename_component(PYTHON_LINK_DIRECTORIES ${PYTHON_LIBRARY_PATH} DIRECTORY)
-	else()
-		py_get_config_var(BLDLIBRARY PYTHON_LIBRARY)
-	endif()
+	py_get_config_var(LIBPL PYTHON_LIBRARY_DIR)
+	py_get_lib_name(PYTHON_LIBRARY_NAME)
+    set(PYTHON_LIBRARY "-L${PYTHON_LIBRARY_DIR} -l${PYTHON_LIBRARY_NAME}")
```
TheJJ: indent fail, needs to be a tab

TheJJ:
Looking good. You may squash the commits into one by `git rebase -i upstream/master`.

buildsystem/modules/FindPython.cmake
darrenstrash: D'oh! Good catch. Old habits die hard... I'll check in the change with the squash.

darrenstrash:
Squashing done. Thanks again for all of your feedback.

TheJJ:
Thanks for the fix!

Issue #382, opened 2015-09-16
Compiler errors with apple clang


darrenstrash:
OSX Yosemite 10.10.4 (14E46)
clang: Apple LLVM version 6.1.0 (clang-602.0.49) (based on LLVM 3.6.0svn)
Target: x86_64-apple-darwin14.4.0
Thread model: posix

The clang build currently fails. Here is a rundown:

1. **Easy:** function call with invalid signature in game_control.cpp
2. **Easy:** missing #include \<string\> in header files.
3. **Hard:** thread_local storage is unsupported for types with "non-trivial destructors"

TheJJ:
Just saying, under Linux with clang version 3.7.0 it builds perfectly fine. We'd never push a version to the master that fails building on our machines, but I acknowledge that it fails for you.

But this is also the reason I can hardly test if I fixed it, as it builds for me anyway.

So you either need to fix it yourself or tell me exactly what is wrong so I can fix things without experiencing the problems.

darrenstrash:
Interesting. I will take a look and see what the major difference is. I see on the clang website that:

"thread_local support currently requires the C++ runtime library from g++-4.8 or later." 

http://clang.llvm.org/cxx_status.html

darrenstrash:
After some investigation, I found that Apple's clang uses *old* gcc 4.2 headers, which have very limited support for C++11. I'll investigate what it takes to use updated headers / library with clang on Mac OS X. Then we should update building.md.

darrenstrash:
This is fixed by:

1. installing clang 3.6 with Homebrew: brew install homebrew/versions/llvm36, 
2. modifying the game_control.cpp file to fix an argument mistmatch, and
3. removing the thread_local redeclaration from compiler.h (it's for old/unsupported compilers anyway).

I will make the code changes, modify building.md, and submit a pull request.

TheJJ:
Thanks for the investigation. Why on earth are the apple headers outdated more than a year.... Jeez.

darrenstrash:
The brewed clang needs to be version 3.5 or higher, version 3.4 doesn't have the necessary thread_local or initializer functionality. I will change building.md to reflect this.

TheJJ:
should be fixed now, right?

darrenstrash:
Should be. Thanks!

PR #388, opened 2015-09-18
Change code (and documentation) to support clang on Mac OS X.

darrenstrash:
This is a proposed fix for issue #382. In order to build openage on Mac, clang 3.5 or higher should be installed through Homebrew. I have updated building.md (increase clang version, add extra Homebrew and Mac OS X instructions) and made minor modifications to source files to support clang on Mac.

### DIFF
--- a/doc/building.md
+++ b/doc/building.md
@@ -26,10 +26,10 @@ Dependencies are needed for:
 
 Dependency list:
 
-    C     gcc >=4.9 or clang >=3.5
-    CRA   python >=3.4
+    C     gcc >= 4.9 or clang >= 3.4 (clang >= 3.5 for Mac OS X)
+    CRA   python >= 3.4
     C     cython >= 0.22
-    C     cmake >=3.0.0
+    C     cmake >= 3.0.0
       A   numpy
       A   python imaging library (PIL) -> pillow
     CR    opengl >=2.1
@@ -93,7 +93,7 @@ if all packages can be installed.
  - `brew tap caskroom/fonts`
  - `brew tap homebrew/version`
  - `brew tap homebrew/python`
- - `brew update`
+ - `brew update` (yes, again)
  - `brew cask install font-dejavu-sans`
  - `brew install python3 libepoxy ftgl freetype fontconfig cmake sdl2 sdl2_image opus opus-tools opusfile`
  - `brew install numpy --with-python3`
@@ -128,7 +128,7 @@ You can install both compilers and select the one to be used by `./configure`.
  - For *nix: `./configure --compiler=clang` will prepare building
   - You could also use `./configure --mode=release --compiler=gcc` here
  - For Mac OS X: use instead:  `./configure --cpp-compiler=clang++-3.5`
-  -  Or: `./configure --mode=release --compiler=g++-4.9`
+  -  Or: `./configure --mode=release --cpp-compiler=g++-4.9`
  - `make` does code generation, builds all Cython modules, and libopenage.
  - `make run`, `./run`, `./run.py` or `python3 -m openage` runs the game.
     try `./run --help` and `./run game --help`.
--- a/doc/building.md
+++ b/doc/building.md
@@ -26,7 +26,7 @@ Dependencies are needed for:
 
 Dependency list:
 
-    C     gcc >=4.9 or clang >=3.4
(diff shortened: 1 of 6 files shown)


doc/building.md
```diff
@@ -87,13 +87,23 @@ if all packages can be installed.
...
 
- - Install [homebrew](http://brew.sh). Other package managers and building from source might also work, but aren't supported as of now. If you figure out fixes for those other options, you're welcome to open pullrequests, as long as the homebrew build doesn't break.
- - Install the needed fonts (e.g. using [homebrew-cask](https://github.com/caskroom/homebrew-cask)).
+ - Install [Homebrew](http://brew.sh). Other package managers and building from source might also work, but aren't supported as of now. If you figure out fixes for those other options, you're welcome to open pullrequests, as long as the homebrew build doesn't break.
+ - `brew update`
+ - `brew install caskroom/cask/brew-cask`
+ - `brew tap caskroom/fonts`
+ - `brew tap homebrew/version`
  - `brew tap homebrew/python`
  - `brew update`
```
TheJJ: double `brew update`

darrenstrash: True, but shouldn't it be updated before installing `cask`, and then after the `taps`?

darrenstrash:
Failed since I wasn't in copying.md. I'll update after I receive more feedback

doc/building.md
TheJJ: I have no idea how brew works :grin:

doc/building.md
```diff
@@ -115,8 +125,10 @@ You can install both compilers and select the one to be used by `./configure`.
 
  - (obviously) clone this repo or acquire the sources some other way
  - make sure you have everything from the [dependency list](#dependencies)
- - `./configure --compiler=clang` will prepare building
+ - For *nix: `./configure --compiler=clang` will prepare building
   - You could also use `./configure --mode=release --compiler=gcc` here
+ - For Mac OS X: use instead:  `./configure --cpp-compiler=clang++-3.5`
+  -  Or: `./configure --mode=release --compiler=g++-4.9`
```
TheJJ: What happens for mac users if you chose `cpp-compiler=clang` or `clang++`? Why is the explicit version required?

darrenstrash: Brew  installs these compilers with the version number as a part of the
executable name, so that they do not not interfere with already-installed
Apple tools. Brew doesn't create symlinks without the version number
(unlike Linux).

If someone uses `--compiler=clang` then it would run the Apple-installed
clang and fail. The same would happen with gcc.

--

Karlsruhe Institute of Technology (KIT)
Institute of Theoretical Informatics

Dr. Darren Strash
Postdoctoral Researcher

Am Fasanengarten 5
Building 50.34, Room 206
76131 Karlsruhe

Telephone: +49 721 608-4-6602
E-Mail: strash@kit.edu

KIT - Universität des Landes Baden-Württemberg und nationales
Großforschungszentrum in der Helmholtz-Gemeinschaft

darrenstrash: Brew  installs these compilers with the version number as a part of the executable name, so that they do not not interfere with already-installed Apple tools. Brew doesn't create symlinks without the version number (unlike Linux). 

If someone uses `--compiler=clang` then it would run the Apple-installed clang and fail. The same would happen with gcc.

doc/building.md
```diff
@@ -26,10 +26,10 @@ Dependencies are needed for:
...
-    C     gcc >=4.9 or clang >=3.4
-    CRA   python >=3.4
+    C     gcc >= 4.9 or clang >= 3.4 (clang >= 3.5 for Mac OS X)
+    CRA   python >= 3.4
     C     cython >= 0.22
-    C     cmake >=3.0.0
+    C     cmake >= 3.0.0
       A   numpy
       A   python imaging library (PIL) -> pillow
     CR    opengl >=2.1
```
TheJJ: if you're adding all the spaces after the `>=` then please do it everywhere :)
(the `>=0.0.0` without space syntax is from gentoo version numbering btw, that's why I actually wanna keep it)

darrenstrash: Ok, no problem. One had a space and the others didn't. I'll take them all out to make it consistent. How does it look otherwise?

TheJJ: Thanks man!

Issue #377, opened 2015-08-27
Villagers bring their resources to nearest building


Miguellissimo:
regardless of the building type. It can be houses, towers, ...

Normal behaviour would be that they bring them to the town center, lumberjack building, mining building or mill. I should be dependant of the type of resource

TheJJ:
Fixed in #390

PR #390, opened 2015-09-19
Villagers bring their resources to the correct buildings

Miguellissimo:
I am working on the issue #377. 

- [x] villagers bring their resources not to random buildings any more.
- [x] according to the resource type, gather the information from the unit instead of casting to living producer to get the dropsite ids
- [x] cleanup

### DIFF
--- a/libopenage/unit/action.cpp
+++ b/libopenage/unit/action.cpp
@@ -862,12 +862,21 @@ UnitReference GatherAction::nearest_dropsite() {
 	// find nearest dropsite from the targeted resource
 	auto ds = find_near(*this->target.get()->location,
 		[=](const TerrainObject &obj) {
-			return &obj.unit != this->entity &&
-			       &obj.unit != this->target.get() &&
-			       obj.unit.has_attribute(attr_type::building) &&
-			       obj.unit.get_attribute<attr_type::building>().completed >= 1.0f &&
+
+			if (not obj.unit.has_attribute(attr_type::building) or &obj.unit == this->entity or &obj.unit == this->target.get()) {
+				return false;
+			}
+
+			auto living_unit = static_cast<LivingProducer*>(this->entity->unit_type);
+			auto building_unit = static_cast<BuildingProducer*>(obj.unit.unit_type);
+
+			auto dropsite0_id = living_unit->unit_data.drop_site0;
+			auto dropsite1_id = living_unit->unit_data.drop_site1;
+
+			return obj.unit.get_attribute<attr_type::building>().completed >= 1.0f &&
 			       obj.unit.has_attribute(attr_type::owner) &&
-			       obj.unit.get_attribute<attr_type::owner>().player.owns(*this->entity);
+			       obj.unit.get_attribute<attr_type::owner>().player.owns(*this->entity) &&
+			       (building_unit->id() == dropsite0_id or building_unit->id() == dropsite1_id);
 	});
 
 	if (ds) {
--- a/libopenage/unit/action.cpp
+++ b/libopenage/unit/action.cpp
@@ -772,14 +772,13 @@ GatherAction::GatherAction(Unit *e, UnitReference tar)
 	target_resource{true},
 	target{tar} {
 
-		Unit *target = this->target.get();
-		if (target->has_attribute(attr_type::resource)) {
-			auto &resource_attr = target->get_attribute<attr_type::resource>();
-			this->resource_type = resource_attr.resource_type;
-
-			//TODO toggle docks
-			this->docks = false;
-		}
+	Unit *target = this->target.get();
+	if (target->has_attribute(attr_type::resource)) {
+		auto &resource_attr = target->get_attribute<attr_type::resource>();
+		this->resource_type = resource_attr.resource_type;
+	} else {
+		throw std::invalid_argument("Unit reference has no resource attribute");
+	}
 }
 
 GatherAction::~GatherAction() {}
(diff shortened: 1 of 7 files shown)


Jon0:
Edit unit/attribute.h, there you can add values to the building attribute, resource_type is currently unused. It should probably be changed to a set since a town center accepts all 4 resources.  Consider docks and fishing boats, there may need to be a required unit class for which units can drop at a particular building

Creating a new drop site attribute might work better, since `has_attribute(attr_type::dropsite)` could rule out a lot of non drop site buildings quickly.

Miguellissimo:
Review and feedback please :+1:

libopenage/unit/attribute.h
```diff
@@ -287,6 +289,34 @@ template<> class Attribute<attr_type::building>: public AttributeContainer {
...
+public:
+
+	Attribute(std::vector<game_resource> types)
+			:
+			AttributeContainer{attr_type::dropsite},
+			resource_types{types} {}
+
+		bool shared() const override {
+			return false;
+		}
```
TheJJ: lines 296 to 302 have to be indented with one tab less please :)

libopenage/unit/attribute.h
```diff
@@ -287,6 +289,34 @@ template<> class Attribute<attr_type::building>: public AttributeContainer {
...
+
+	std::shared_ptr<AttributeContainer> copy() const override {
+			return std::make_shared<Attribute<attr_type::dropsite>>(*this);
```
TheJJ: this one is 2-tab indented as well

libopenage/unit/producer.cpp
```diff
@@ -534,12 +535,53 @@ void BuildingProducer::initialise(Unit *unit, Player &player) {
 		unit->give_ability(std::make_shared<AttackAbility>());
 	}
 
+	// dropsite attribute
+	std::vector<game_resource> accepted_resources = get_accepted_resources();
```
TheJJ: `this->get_accepted_resources()`

libopenage/unit/producer.cpp
```diff
@@ -534,12 +535,53 @@ void BuildingProducer::initialise(Unit *unit, Player &player) {
...
 	unit->give_ability(std::make_shared<SetPointAbility>());
 	unit->give_ability(std::make_shared<TrainAbility>());
 	unit->give_ability(std::make_shared<UngarrisonAbility>());
 }
 
+std::vector<game_resource> BuildingProducer::get_accepted_resources() {
+	//TODO use a more general approach instead of hard coded ids
+
+	auto id_in = [id = this->id()](std::initializer_list<int> ids){
+		return std::any_of(ids.begin(), ids.end(), [=](int n){ return n == id; });
```
TheJJ: you can use `this->id()` in the body and capture by value `=` in the lambda, then the `id = this->id()` is no longer necessary.

libopenage/unit/producer.cpp
```diff
@@ -534,12 +535,53 @@ void BuildingProducer::initialise(Unit *unit, Player &player) {
...
+		return std::vector<game_resource>{game_resource::food};
+	}
+
+	if (id_in({562, 563, 564, 565})) { //Lumberjack camp
+		return std::vector<game_resource>{game_resource::wood};
+	}
+
+	if (id_in({45, 47, 51})) { //Docks
+		return std::vector<game_resource>{game_resource::fish};
+	}
```
TheJJ: you can probably use `else if` for all of those, but it's more a workaround for not having the accepted resources directly available i guess. I think you might want to use https://github.com/SFTtech/openage/blob/master/openage/convert/gamedata/unit.py#L792

This needs some preprocessing though, as there is a villager type for each resource hunter in the original game and we only have one villager with several behavior types, you might want to map the above `drop_site0` and `1` to a list of possible drop sites for one villager type. But this avoids the current hardcoding I think.

TheJJ:
Cool stuff, nice improvements. If we can avoid the hardcoding someday, even better.

libopenage/unit/resource.h
```diff
@@ -14,7 +14,8 @@ enum class game_resource {
 	wood,
 	food,
 	gold,
-	stone
+	stone,
+	fish
```
Jon0: This is supposed to be the resource stockpile types, so it shouldn't include fish (fishing should be added to the food stockpile)

libopenage/unit/producer.cpp
Jon0: building a multimap of: gather id -> dropsite id in `GameSpec` before constructing the `UnitType` classes, would allow an inverse lookup in the building type constructors

Miguellissimo:
I've dropped the docks as a dropsite for now and concentrated on the four fundamental resources, because it is like a special case for food. In the future, we should possibly think about adding relics as a resource type as well. Basically they can be gathered and get only accepted by a special dropsite id (monastery).

libopenage/unit/action.cpp
```diff
@@ -771,6 +771,15 @@ GatherAction::GatherAction(Unit *e, UnitReference tar)
...
 	target{tar} {
+
+		Unit *target = this->target.get();
+		if (target->has_attribute(attr_type::resource)) {
+			auto &resource_attr = target->get_attribute<attr_type::resource>();
+			this->resource_type = resource_attr.resource_type;
+
+			//TODO toggle docks
+			this->docks = false;
+		}
```
TheJJ: These lines are one indent level too much, they need to be at the same level as `target{tar}`.
What is `resource_type` initialized to when the target doesn't have the resource attribute? You probably wanna add it to the initializer list above or else but i think you need to handle it.

TheJJ:
Thanks, seems to be good for merge! :+1:

PR #393, opened 2015-09-29
Added missing dependency under Arch Linux

paddatrapper:
Cython and Pygments are required, but not listed under the list of dependencies to install for Arch Linux.


TheJJ:
Thanks a lot!

PR #394, opened 2015-10-06
Villagers stop gathering resources

### LINKED ISSUE #386: Villagers continue collecting resources when told to move away.
The situation:

1. Villager is collecting resources
2. I right-click on terrain in a different location
3. Villager **correctly** moves to that location
4. Then, villager **incorrectly** goes right back to collecting the same resource.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26779041-villagers-continue-collecting-resources-when-told-to-move-away?utm_campaign=plugin&utm_content=tracker%2F371436&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F371436&utm_medium=issues&utm_source=github).
</bountysource-plugin>

paddatrapper:
Fixes #386  - when the user moves a villager away from gathering a resource the villager will not go back to collecting it.

### DIFF
--- a/copying.md
+++ b/copying.md
@@ -60,6 +60,7 @@ _the openage authors_ are:
 | Thomas Oltmann              | tomolt                      | thomas.oltmann.hhg@gmail.com          |
 | Miguel Kasparick            | miguellissimo               | miguellissimo@gmail.com               |
 | Darren Strash               | darrenstrash                | darren.strash@gmail.com               |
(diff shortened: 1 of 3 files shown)


TheJJ:
thanks very much!

Issue #181, opened 2014-12-18
Custom font rendering engine


TheJJ:
We could think about removing `ftgl` in favor of a customized rendering system.

For efficient rendering of the ingame console, something handcrafted renderer is needed anyway, so we can use this system for all font rendering.

Some routines for loading fonts to draw surfaces can probably still be used from `ftgl`.

TheJJ:
#395 is trying to achieve this.

PR #395, opened 2015-10-20
text rendering using HarfBuzz and vbo's

### LINKED ISSUE #383: coord test uses nonexistant opengl context on Mac OS X
**System setup:**
Mac OS X Yosemite 10.10.4 (14E46)
Using g++ version 5.2.0

[ 8/19] cpp coord::tests::coord
INFO Font file: /Users/strash/Library/Fonts/DejaVuSansMono.ttf

**SIGSEGV**

**FATAL:** terminate has been called

current stack

  File ?, in openage::error::StackAnalyzer::analyze()+0x77 [0x1099a153b]

standard terminate handler

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26724694-coord-test-uses-nonexistant-opengl-context-on-mac-os-x?utm_campaign=plugin&utm_content=tracker%2F371436&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F371436&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jprashanth:
- [X] Replaces FTGL's immediate mode rendering with HarfBuzz and vbo based rendering (previously 30fps and now VSYNC on my machine).
- [X] Console now uses this text rendering mechanism (much better performance - VSYNC)
This is a step towards #287
- [X] update prerequisites in `doc/building.md`
- [X] Remove `FTGL` completely, if required

To satisfy travis, i had to add an external ppa providing `libharfbuzz-dev` for ubuntu precise.

Fixes #383


libopenage/renderer/font.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+bool font_description::operator!=(const font_description &other) const {
+	return !(*this == other);
+}
+
+size_t font_description::hash() const {
+	std::size_t h1 = std::hash<std::string>()(this->font_file);
+	std::size_t h2 = std::hash<unsigned int>()(this->size);
+	// Taken from http://www.boost.org/doc/libs/1_55_0/doc/html/hash/reference.html#boost.hash_combine
+	return h1 ^ (h2 + 0x9e3779b9 + ((h1 << 6) + (h1 >> 2)));
```
TheJJ: i think you can just create a `std::pair` with those two and then hash the pair.

libopenage/renderer/font_manager.h
```diff
@@ -0,0 +1,86 @@
...
+	FT_Library ft_library;
+
+	freetype_library() {
+		if (FT_Init_FreeType(&this->ft_library) != 0) {
+			throw Error{MSG(err) << "Failed to initialize freetype library."};
+		}
+	}
+
+	~freetype_library() {
+		FT_Done_FreeType(this->ft_library);
```
TheJJ: You probably want to explicitly delete the move and copy constructors/assignments for this struct. Also, this is more likely to be a class than a struct as it does way more then just packing data.

libopenage/renderer/font_manager.h
```diff
@@ -0,0 +1,86 @@
...
+	 */
+	Font *get_font(const char* font_file, unsigned int size);
+
+private:
+	// Font cache. the hash of font's description is used as the key
+	std::unordered_map<size_t, std::unique_ptr<Font>> fonts;
+
+private:
+	// The freetype library instance
+	static freetype_library library;
```
TheJJ: Why is the freetype library required to be static? Does it only allow one instance? Otherwise, each font manager could have it's own freetype instanciation.

libopenage/renderer/glyph_atlas.cpp
```diff
@@ -0,0 +1,153 @@
...
+	memset(this->buffer.get(), 0, this->width * this->height);
+
+	glGenTextures(1, &this->texture_id);
+	glBindTexture(GL_TEXTURE_2D, this->texture_id);
+	glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_CLAMP);
+	glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_CLAMP);
+	glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
+	glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR);
+
+	glTexImage2D(GL_TEXTURE_2D, 0, GL_RED, this->width, this->height, 0, GL_RED, GL_UNSIGNED_BYTE, this->buffer.get());
```
TheJJ: this is gonna be fun porting to the backend-independent renderer then, but sure, for now it's perfectly ok as the renderer isn't merged :D

libopenage/renderer/glyph_atlas.cpp
```diff
@@ -0,0 +1,153 @@
...
+	Glyph glyph;
+	std::unique_ptr<unsigned char> image = font->load_glyph(codepoint, glyph);
+	return this->set(key, glyph, image.get());
+}
+
+size_t GlyphAtlas::get_cache_key(Font *font, codepoint_t codepoint) const {
+	size_t fd_hash = font->description.hash();
+	size_t codepoint_hash = std::hash<codepoint_t>()(codepoint);
+	// Taken from http://www.boost.org/doc/libs/1_55_0/doc/html/hash/reference.html#boost.hash_combine
+	return fd_hash ^ (codepoint_hash + 0x9e3779b9 + ((fd_hash << 6) + (fd_hash >> 2)));
```
TheJJ: again, using a `std::pair` here may obsolete the manual hash combination.

TheJJ:
cool stuff!

libopenage/renderer/font.cpp
jprashanth: I don't see a standard specialization of `std::hash` for `std::pair` in [cppreference](http://en.cppreference.com/w/cpp/utility/hash)

A note in the same page says
> Note: additional specializations for std::pair and the standard container types, as well as utility functions to compose hashes are available in boost.hash

libopenage/renderer/font_manager.h
jprashanth: FreeType provides the option to create multiple instance of `FT_Library`. Typically, you would one library instance to manage all the font faces loaded in this library.

The API reference also adds additional notes with respect to multithreading
> A handle to a FreeType library instance. Each ‘library’ is completely independent from the others; it is the ‘root’ of a set of objects like fonts, faces, sizes, etc.

> In multi-threaded applications it is easiest to use one ‘FT_Library’ object per thread. In case this is too cumbersome, a single ‘FT_Library’ object across threads is possible also (since FreeType version 2.5.6), as long as a mutex lock is used around FT_New_Face and FT_Done_Face.

libopenage/renderer/glyph_atlas.cpp
jprashanth: Not necessarily. `GlyphAtlas` itself can be made independent of backend renderer. I can remove the `gl` calls inside the atlas and provide an api (let's say `get_buffer`). `TextRenderer` (or any other manager) can use this api to flush the buffer to texture.

This can be done once renderer is merged :D

libopenage/renderer/font_manager.h
TheJJ: Then i'd say we use the ft library as non-static member here, we won't use more than one font manager anyway i hope.

libopenage/renderer/font.cpp
TheJJ: Hm you seem to be right. I'd suggest adding a function in `util` to allow combining hashes, so you start with some seed and can update it again and again with any another hash (a `size_t` number). Then i'd add `std::hash<font_description>` that uses that hash combination function. Better than duplicating the combination algorithm all over the codebase.
The existing combinations in some other parts of the code should be updated as well then.

libopenage/renderer/font_manager.h
jprashanth: Yes, that would be the ideal case. I will make some changes regarding this.

TheJJ:
So this actually does fix #181, nice.

libopenage/renderer/font_manager.cpp
```diff
@@ -0,0 +1,88 @@
...
+namespace openage {
+namespace renderer {
+
+std::string FontManager::get_font_filename(const char *family, const char *style) {
+	// Initialize fontconfig
+	if (!FcInit()) {
+		throw Error{MSG(err) << "Failed to initialize fontconfig."};
+	}
+
+	//FcPattern *font_pattern = FcNameParse((const unsigned char *)"DejaVu Serif:style=Book");
```
TheJJ: leftover debug stuff

libopenage/renderer/font_manager.cpp
```diff
@@ -0,0 +1,88 @@
...
+	//tell fontconfig to find the best match
+	FcResult font_match_result;
+	FcPattern *font_match = FcFontMatch(nullptr, font_pattern, &font_match_result);
+
+	/*
+	//debug output: display matching font pattern as parsable string
+	FcChar8 *match_string = FcNameUnparse(font_match);
+	log::dbg2("resulting font: %s", match_string);
+	free(match_string);
+	*/
```
TheJJ: more here

TheJJ:
We can leave the harfbuzz package in the other ppa as travis will soon be ditched anyway.
Can you add the dependency info to the `building.md` please, then we can merge.

We might think about moving the font rendering stuff to `renderer/text/`, or what do you think?

jprashanth:
I will update the docs with `harfbuzz` dependency. Should I leave the `FTGL` stuff as it is?

Yes, I think it is a good idea to move font stuff to `renderer/text/`. I can go ahead with moving the files to this directory, if there are no other objections.

TheJJ:
If you are able to remove `ftgl` completely, would be even better. The "high-level" text renderer api should stay in `renderer/text.h` or something, this will be the class that will then create render tasks someday later.

jprashanth:
This pull request completely removed the usage of `FTGL`. So, yes we can cleanly remove it. Just to be on the safe side, I will keep this removal in a separate commit.

How about we keep the high-level api `renderer/text_renderer.h` as it is, and move font-related stuff to `renderer/font`.

TheJJ:
Jup, good idea. Go for it.

jprashanth:
Done. :+1:

Issue #391, opened 2015-09-20
Fix devmode detection by not installing __pycache__/devmode.pyc


TheJJ:
To detect if we're running in the repo or as installed program in `/usr` or whatever,
we test if the devmode python module can be imported.

The `devmode.py` file is not installed to the installation directory. That way, when running as installed program, the module can't be imported.

The problem now is that the buildsystem converts all python code to compiled `.pyc` files.
The `devmode.py` is not installed, but the `devmode.pyc` file is.

Python does import the `pyc` file even though the `.py` file is missing.

=> Prevent the buildsystem to install corresponding `pyc` files to not-installed `py` files.

jprashanth:
#372 removes `__pycache__` installation altogether.

TheJJ:
Hmm, didn't see that, and it's not good. Weodp want to install the compiled py objs in the pycache folder, setup.py etc does this as well.

TheJJ:
#372 is merged now, but devmode.pyc is still installed :)
If anyone has time and motication to tinker with cmake, go for it.

PR #399, opened 2015-10-23
Out of Source Builds

### LINKED ISSUE #398: Better out-of-source build python support
We should consider boosting the buildsystem with ideas from:

https://stackoverflow.com/questions/13959434/cmake-out-of-source-build-python-files

Basically:
- Don't rely on the source folder at all when running
- Achieve this by symlinking/copying into build directory
- This makes `make install` easier to implement and understand as the runimage already is present in the binary dir (although symlinks have to be fixed then)

jprashanth:
- python sources in openage folder are replicated in build dir.

- cpp, html, and cython modules are now generated in build dir
  and therefore keeps the source directory clean.

- pxd files are generated in the build dir.

- removed os-dependent functionality like `symlinks` and `find`.

- replaced add_cython_modules with better api `add_cython_module`.

- with these changes, individual cython modules depend on only their
  source files. Therefore, changes to cython module source files
  will trigger build of only dependent targets. Also, removed the
  need to explicitly specify pxd files through `add_pxds`.

- in general, simplified the build system by removing the custom targets
  compilepy, inplacemodules, cleanpxdgen, cythonize, cleancython.

- As CMake is now responsible for managing timestamps of pxd files
  and corresponding source files, pxdgen is now simpler.

- fixed devmode detection.

Fixes #398 , #391 

This completes the changes required for "fully" out-of-source builds. The only files generated in source tree are `codegen` files. These files are not dependent on any configuration and therefore acceptable (for now).

Also removed the generation of configuration dependent files like `inplace_module_list` in configuration independent dir `bin/py/`. Therefore, the project can now be built using multi config cmake generators (Tested with Xcode).


jprashanth:
Now, able to debug :+1: 

openage_debugging

franciscod:
holy cow, that's a nice debugger! is it only on osx ?

jprashanth:
Xcode

franciscod:
:heart_eyes:

jprashanth:
Hahaha

TheJJ:
Nope, we won't turn back having a `python` folder again, unless there's a very good reason not to stick to the current folder layout. We changed it from `py` -> `openage` once and it was a huge mess for contributors.

Also, the centralized sourcefile registry was something we decided to purge to keep the system more modular and resistant for renames, reverting that back to one file to list sources should also really be avoided.

jprashanth:
The problem with using `add_subdirectory` is that it creates a separate build context for each added directory. It was provided for the purpose of building different units. I think we are using it to list the files of a directory. I haven't tested this, but theoretically, the build complexity and time also increases with the increase in the number of directories added through `add_subdirectory`.

jprashanth:
Reverted back to the current folder layout.

jprashanth:
@TheJJ This pull request is now complete. You can review and suggest any required modifications.

TheJJ:
Apart from smaller bugs like touching a python file which then is not copied to the build folder after that, the only thing that annoys me is that there now is `Debug` and `Release` within the build folder.

This is precisely what we're currently avoiding with the `./configure` script and the separate build folders.
Are those folders needed for xcode or is it sufficient to have the current bin folder created by regular cmake?

@mic-e has done a more deeper review and will hopefully share his findings soon here.

jprashanth:
It's been a while since I have looked at this branch. I will have to read through the code to see which python file is being touched. I will get back to you regarding this.

For single configuration CMake build systems (`make`, `ninja`, e.t.c), there is no need for a special directory like `Debug` and `Release`. However, multi config generators of CMake like Xcode, Visual Studio e.t.c create different output files for different build configuration. When using these generators (Xcode in my case), cmake will fail with an error.

TheJJ:
But is it possible to just keep the single configuration? I mean it's not likely that you will do any release build anyway. Also, can't Xcode create these debug and release folders itself?

My goal is to make the folder layout as beautiful as possible, and having `bin/openage` is way better than `bin/Debug/openage`, especially because the `Debug` and `Release` switch is done by `configure` at the moment.

Which brings me to the next idea: How about we try to unify both approaches?
Currently, `./configure` creates different cmake bin dirs in `.bin/$folder_cxxflagencoding`.
Can we rename `Debug` and `Release` to the above but keep them in the same cmake bin dir?
Then, IDEs should be happy and we can still do convenient switching with the `configure` script.

If we don't unify it, we have two chained switches, one by `configure` (to set all the flags) and one by the Debug and Release folder.

In short: `configure` should do the same switch as Xcode.

TheJJ:
#500 fixed some things you tried to address here. If you can update your code, would be cool if we had true out of source builds some day :)

TheJJ:
I guess we'll close this in favor of #560.

PR #401, opened 2015-10-25
General grammar and styling fixes

VOTProductions:
Consistency in code is important, and I've tried to make small changes in order to maintain it.

Please don't be too shouty at how terrible I am at Git, it's my first public commit.


configure
```diff
@@ -7,7 +7,7 @@
 
 import sys
 if not sys.version_info >= (3, 4):
-    print("openage requires Python 3.4 or higher")
+    print("Openage requires Python 3.4 or higher.")
```
TheJJ: heh :) `openage` is actually meant to be a lowercase-word, so I'm not sure if we should capitalize it even on sentence beginnings.

libopenage/engine.cpp
```diff
@@ -47,13 +47,13 @@ void Engine::create(util::Dir *data_dir, const char *windowtitle) {
...
 	} else {
-		throw Error{MSG(err) << "You tried to create another singleton engine instance!!111"};
+		throw Error{MSG(err) << "You tried to create another singleton engine instance!"};
 	}
 }
 
 void Engine::destroy() {
 	if (Engine::instance == nullptr) {
-		throw Error{MSG(err) << "You tried to destroy a nonexistant engine."};
+		throw Error{MSG(err) << "You tried to destroy a none-xistent engine."};
```
TheJJ: uh

configure
VOTProductions: Things like iPad are the exception, but you would capitalize Firefox and Steam. Stack Overflow is being inconsistent, as they are using "stackoverflow" for their logo, and yet their site is called "Stack Overflow". It's all over the place. :P

TheJJ:
Thanks so far. I don't approve removing "funny" stuff, but otherwise the improvements look good. Can you now please [sqash those commits together](https://github.com/SFTtech/openage/blob/master/doc/contributing.md#rebasing---what-the-hell-is-it-and-why-do-i-need-it)?

TheJJ:
It looks like you dropped one commit in the rebase, as the dash-typo is in there again.

TheJJ:
Thanks. But before merging, you need to rebase in order to eliminate the merge commit and to fix the conflicts. You may also squash the commits into one.

Sorry I didn't notice earlier, but I'm not notified if somebody pushes new commits to a pull request.

configure
franciscod: remember to turn this back to lowercase :)

copying.md
```diff
@@ -61,6 +61,7 @@ _the openage authors_ are:
 | Miguel Kasparick            | miguellissimo               | miguellissimo@gmail.com               |
 | Darren Strash               | darrenstrash                | darren.strash@gmail.com               |
 | Kyle Robbertze              | paddatrapper                | paddatrapper@gmail.com                |
+| Robert Chomicz              | Votato, vot                 | votproductions@gmail.com              |\
```
franciscod: whoops, `\` at the end

franciscod:
I'm not sure if I like bullet lists with periods:

- I don't like this.
- The bullet list already has items and don't need extra punctuation.
- You already see when one starts and another ends very clearly.

- So this is better!
- No periods
- Cleaner
- **you even save bytes**

TheJJ:
Indeed, it looks better without the periods I'd say.

PR #411, opened 2015-12-02
convert: Check for game version in only one place

azrdev:
Introduce GameVersion enum, and store an instance once determined.
Everybody else can switch version-specific code upon it, instead of
reinventing the wheel

I'm using [enum](https://docs.python.org/3.4/library/enum.html), which is available as of python 3.4 - didn't find a note how backwards compatible you want to be.

### DIFF
--- a/copying.md
+++ b/copying.md
@@ -61,6 +61,7 @@ _the openage authors_ are:
 | Miguel Kasparick            | miguellissimo               | miguellissimo@gmail.com               |
 | Darren Strash               | darrenstrash                | darren.strash@gmail.com               |
 | Kyle Robbertze              | paddatrapper                | paddatrapper@gmail.com                |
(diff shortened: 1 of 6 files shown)


TheJJ:
Thanks! This is looking very good.

Unfortunately, some update to pylint now complains about all sorts of errors you aren't responsible for.
We'll fix them as soon as possible, and then ask you to rebase to include those changes.

No worries, i think your code looks fine and does good things :)

TheJJ:
Please now rebase on the current master, I fixed the pylint problems in our upstream code :)

azrdev:
@TheJJ rebased & improved my work. Once13014b9 survives CI, please merge!

TheJJ:
Awsum! Thanks! Wheee!

PR #417, opened 2015-12-16
Improvments of the team colors shader

spjoe:
+ Made the number of possible player colors variable (assets/converted/player_palette.docx)
+ Simplified equals with epsilon checks
+ Introduces a way to share common functions between shaders
 + Shader constructor takes a list of shader code snippets instead of on code char array
 + removed shader header from shaders.
 + add equalsEpsilon.glsl as first shared shader code.
+ game_renderer derives number of player colors from player_color_lines

### DIFF
--- a/assets/shaders/alphamask.frag.glsl
+++ b/assets/shaders/alphamask.frag.glsl
@@ -1,5 +1,3 @@
-#version 120
-
 // alpha masking shader
 //
 // applies an alpha mask texture to a base texture,
--- a/assets/shaders/teamcolors.frag.glsl
+++ b/assets/shaders/teamcolors.frag.glsl
@@ -1,5 +1,3 @@
-#version 120
-
 //team color replacement shader
 //
 //looks for an alpha value specified by 'alpha_marker'
@@ -16,13 +14,13 @@ uniform int player_number;
 uniform float alpha_marker;
 
 //color entries for all players and their subcolors
-uniform vec4 player_color[64];
+uniform vec4 player_color[NUM_OF_PLAYER_COLORS];
 
 //interpolated texture coordinates sent from vertex shader
 varying vec2 tex_position;
 
 //create epsilon environment for float comparison
-const float epsilon = 0.001;
+const float EPSILON = 0.001;
 
 //do the lookup in the player color table
 //for a playernumber (red, blue, etc)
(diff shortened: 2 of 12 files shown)


assets/shaders/alphamask.frag.glsl
```diff
@@ -1,5 +1,3 @@
-#version 120
```
TheJJ: Hmm, isn't the version specifier required? we shouldn't abandon them.

assets/shaders/teamcolors.frag.glsl
```diff
@@ -31,49 +29,26 @@ vec4 get_color(int playernum, int subcolor) {
...
+				gl_FragColor = get_color(player_number, i);
+				return;
 			}
 		}
-	}
-	//else the texel had no marker so we can just draw it without player coloring
 
+	 //unknown base color gets pink muhahaha
+	 pixel = vec4(255.0/255.0, 20.0/255.0, 147.0/255.0, 1.0);
+ }
```
TheJJ: the indentation seems to be wrong here.

TheJJ:
Doesn't look bad and I'll merge it after I tested it and you fixed the indentation fail.
Are you aware of #287 ? All that stuff will be migrated over there some day.

assets/shaders/alphamask.frag.glsl
spjoe: yes it is, but it also has to be in the first not comment line and it is not allowed to be defined twice. What I did is adding a possibility to add common functions before the real shader code. To be able to do it, the version header can not be in the shader file.

instead it is added in the code. see https://github.com/spjoe/openage/commit/c12aff570d96f42a1cf5011df0147f9b05c2c41f#diff-76e3d56a04902eefda9a5140e3af8bbeR69

spjoe:
@TheJJ. Sorry I was not aware of #287. I will merge your renderer branch and see if it is working

TheJJ:
Don't merge my renderer please, I'm actively continuing to work on it until it's sufficiently good :dancers:  I'll merge your improvements but they'll be obsolete sometime later when i migrate all that stuff into the new renderer, but it's a good short-term improvement.

You can do a rebase to get rid of the merge commit by the way. Look at `contributing.md` or as me if you need assistance :)

spjoe:
Just have done a rebase, but build now fails, from the logs a see it is a race condition, but it seems unrelated to my changes. it happend in the job_manager test.

TheJJ:
Test successful, mergemergemergemerge!

Issue #56, opened 2014-10-29
preprocessor-free coordinate system


TheJJ:
The preprocessor hackery has to be replaced by the new `openage.codegen` python generation.

This should create the necessary headers and should end the pain of the current C preprocessing implementation.

The implementation is located in `libopenage/coord`.

franciscod:
could someone explain the pain of the current implementation, and how would the paradisiac version be like?

TheJJ:
By "someone" you probably mean @mic-e as he wrote the `coord` system in the first place.

mic-e:
The pain with the current version is that it's pretty unmaintainable and extremely hard to read because it uses preprocessor `#defines` and `#includes` in an unconventional way to produce the required `.cpp` and `.h` files for the absolute and delta classes for each coordinate system, including all of their overloaded operators.

Writing all code by hand would be an option, though it would create immense redundancy. It also would make adding a new coordinate system pretty hard.

An ideal solution would have instruction files of some kind == in the `coord` folder, and have a python code generator module use them to create the full `.h` and `.cpp` files from them.

detrumi:
I'm planning to implement this. There's not much that changes between the different types, so the following information would be enough:

	namespace = coord;
	scalar_type = phys2;
	members = ne, se;
	body = {
		phys3 to_phys3(phys_t up = 0) const;
		tile to_tile() const;
		phys2_delta get_fraction() const;
	};
	delta_body = {
		phys3_delta to_phys3(phys_t up = 0) const;

		//allowing conversions of phys deltas to tile deltas isn't such a great
		//idea. you never know whether a phys change of 0.9 will _actually_
		//change the tile.
	};
This example would be something like `phys2.scalar`, which would then generate `phys2.h` with the structs and all operations for the `phys2` and `phys2_delta` types.

Some loose ends:
- It would be nice to be able to define these files for example in the `cpp/console` folder, with the correct namespace generated. At the moment `cpp/coords` also contains some types for the console, so it would be nice to separate them.
- Some of the types have a hash function defined, but `phys3` only uses the first two values in the hash. Can we generalize this, or otherwise just specify which members are to be used in the hash?

TheJJ:
A few thoughts that I had about the coord system: (for the record, i just extracted the following from a 2000 lines-xmpp-conversation):

* Constraint: conversions should be very fast.
* Eliminate the global state (e.g. scroll position) by inheritance:
  * Every component having some coord state inherits from a coord base class.
[...]

PR #419, opened 2015-12-18
[WIP]coords with c++ templates

spjoe:
Instead of using macros and out of place `#include` I introduced c++ templates to create easily new coord types.

Cons
* construtor is not a constexpr constructor anymore :-1: 
* lot of code due to limitation on partial template specialization on methods :-1:

Pros
* no wired `#include` and `#defines` :+1:

Actual I am not really convinced anymore if this was the right call to do it with c++ templates. maybe code generation is better :). Or if somebody knows todo this c++ templates better (I wanted to use the  [get&lt;N&gt;](https://github.com/SFTtech/openage/compare/master...spjoe:feature/coord-template?expand=1#diff-19f485c287b0e7d5dede1d170aabae03R17) method on xyz_coord but the compiler did not let me, due to limitations partial template specialization on methods). 

e.g.
```cpp
template <typename AT, typename RT, typename ST, unsigned int N>
RT absolute_xyz_coord<AT, RT, ST, N>::as_relative() const {
        RT retval;
        for (unsigned int i = 0; i < N; i++) {
                retval.get<i>() = this->get<i>();
        }
        return retval;
}
```
Compiler should be able to do unrolling at compile time (N is known).


TheJJ:
Hmm, this is _much_ more code for the same functionality as before. I'd say we should either use code generation or somehow clarify the `#define` hell that, yes, should be avoided.

The main thing that has to be improved is to no longer have global state (e.g. the camera position).
For those ideas, see (and discuss in) #56 please.

TheJJ:
Do you plan on continuing this? Otherwise, I'll close this and somebody else might give it a shot.

spjoe:
Sorry no time at the moment and I will not continue with it.

TheJJ:
Ok. I'll close this now, but we can always dig it out again. Thanks for the attempt, though.

Issue #397, opened 2015-10-21
Migrate to kevin-ci


TheJJ:
Migrate our CI system to https://github.com/SFTtech/kevin

This is currently work-in-progress and will replace travis for us.

TheJJ:
Only thing left is adding project support to kevin, then we can do the transition. Stay tuned!

TheJJ:
It's done!

Issue #329, opened 2015-07-24
Purge engine singleton


TheJJ:
In our big mission to remove all global state, we can remove the engine singleton the following way:
each subsystem, upon initialization, gets passed the engine pointer.
Member functions of this subsystem can then use `this->engine->` to access the engine instead of `auto engine& = Engine::get(); engine.bla()`.

I recommend creating a `class SubSystem { public: Engine *engine };`, and all subsystem classes that currently use the global `Engine::get` method should inherit from that one. When the subsystem gets initialized, the engine pointer is stored.

The only funny thing left to solve is the coordinate subsystem. #322 is a big improvement in that direction.
##

amit2103:
@JJ can I work on this? Also can you or anyone point out how to identify the submodules which might need be refactored this way. Any help will be appreciated.

Disclaimer:- I am highly profiecient in JAVA and Scala. But have recently started learning C++ and looking to contribute to open source C++ games like openage.

zuntrax:
@amit2103 Go for it. Also hello.

amit2103:
@zuntrax  okay..Will try to catch you guys in Freenode.

spjoe:
I suggestion using a Dependency Injection framework. With the proposed solution the Engine Object has knowledge of too much(see anti pattern God Object).

TheJJ:
@spjoe what is a dependency injection framework? It sounds kinda intrusive :)
If you wanna strip information out of the engine (namely making some subsystems unavailable), how do we reach them then?

spjoe:
google for "dependency injection framework" and for "inversion of control", you will find a tone of information on this subject. The idea is that you do not need to "reach" your dependencies(in our case subsystems) you get them injected during construction time.

A C++ impl. of a DI framework https://github.com/krzysztof-jusiak/di

amit2103:
Okay..I had actually started to code this...I come from a JAVA background where DI is almost a de facto standard for wiring dependencies...I can try this with the C++ DI framework.

spjoe:
@amit2103 it was just a suggestion, you do not have to abandon your work ;-), I have experience with big code bases and IoC helps in my opinion a lot, but it is maybe not needed here :-)

amit2103:
Well ...it might be worth while to do it. It might be better to do now..than to do it fully and again do it :)

TheJJ:
@amit2103 how's it going? How much work is left to do to close this issue?

TheJJ:
[...]

PR #422, opened 2015-12-20
QtQuick GUI

ChipmunkV:
I've started the integration of the QtQuick as an ingame GUI. Current state is that it can render the game an the QtQuick scene at the same time.

Simple hardcode version to be able to understand the workings is on **different** branch: https://github.com/ChipmunkV/openage/tree/qt-gui-minimal

There is some simple stuff left to do: resize, input and actual QML code of the skin. (I won't start from that)
Locale troubles too: https://github.com/SFTtech/openage/issues/423

I'll start from the more important question: the connection to the game logic. For now, I can't see how that can be done without at least the separation of the main engine from the SDL initialization code.


TheJJ:
is https://stackoverflow.com/questions/10348098/embedding-qt-gui-into-existing-opengl-program no longer a problem?

ChipmunkV:
Not a problem. Qt doesn't need to create a different context. And it doesn't take control of the input event loop. Events can be injected via Qt postEvent/sendEvent functions.

TheJJ:
Good, good :) Have you looked at the gui implementation of [Stellarium](https://bazaar.launchpad.net/%7Estellarium/stellarium/trunk/files/head:/src/gui/)? Maybe they came up with some cool concepts you can also benefrit from.

But all in all i'd say using qt for the gui stuff is probably the right thing to do instead of developing some homebrew stuff again.

ChipmunkV:
Stellarium GUI is based on the raster graphics, it's not the same approach. Honestly, I don't understand that code. It's too different.

ChipmunkV:
Figured out a method of passing data. Now will look how it works for the real classes of the game.

TheJJ:
Nice! Btw, please try adhering to our `doc/code_style/`, we wanna keep it consistent.

ChipmunkV:
Added mouse and resize (no keyboard for the moment, figuring out how to simulate the focus).

Now it's possible to launch world generator from the GUI.

Run with LC_ALL=C.

TheJJ:
on a side node: `dev-qt/qtquickcontrols` is needed under gentoo.

```
jj@jjpad ~/devel/openage [openage/ChipmunkV-qt-gui]% LC_ALL=C make -j4 run
./run game
INFO launching engine with data directory './assets'
INFO Using audio device: default [freq=48000, format=32784, channels=, samples=2048]
INFO Font queried: DejaVu Serif:style=Book
INFO Font file: /usr/share/fonts/dejavu/DejaVuSerif.ttf
INFO Font queried: DejaVu Serif:style=Book
INFO Font file: /usr/share/fonts/dejavu/DejaVuSerif.ttf
WARN [T1] WARNING: QApplication was not created in the main() thread.
INFO Font file: /usr/share/fonts/dejavu/DejaVuSansMono.ttf

SIGSEGV
run: /tmp/portage/media-libs/fontconfig-2.11.94/work/fontconfig-2.11.94/src/fccache.c:533: FcCacheFini: Assertion `fcCacheChains[i] == ((void *)0)' failed.

SIGSEGV

FATAL: terminate has been called

current stack

  File ?, in ? [0x7f4cdf54bc7b]
  File ?, in QV4::Runtime::getQmlQObjectProperty(QV4::ExecutionEngine*, QV4::Value const&, int, bool)+0x11b [0x7f4d09cbe13b]
  File ?, in QV4::QObjectWrapper::getProperty(QObject*, QV4::ExecutionContext*, QQmlPropertyData*, bool)+0x765 [0x7f4d09ca37c5]
  File ?, in ? [0x7f4d09ca2862]
  File ?, in ? [0x7f4d0a1d1c8b]
  File ?, in ? [0x7f4d0a1d1aa7]
  File ?, in QFontMetricsF::averageCharWidth() const+0x11 [0x7f4d096e66e1]
  File ?, in QFontPrivate::engineForScript(int) const+0x75 [0x7f4d096c8cf5]
  File ?, in QFontDatabase::load(QFontPrivate const*, int)+0x346 [0x7f4d096f24e6]
  File ?, in QFontDatabase::findFont(QFontDef const&, int)+0x1bd [0x7f4d096f1b2d]
  File ?, in ? [0x7f4d096ecba1]
  File ?, in ? [0x7f4cf809dc19]
  File ?, in FcFontList+0x4f [0x7f4d0be1133f]
  File ?, in FcFontSetList+0x1a1 [0x7f4d0be10b01]
  File ?, in ? [0x7f4d0be15953]
  File ?, in ? [0x7f4d081e25c0]
  File ?, in ? [0x7f4d0c3a630f]
  File ?, in ? [0x7f4d08ae59c1]
  File ?, in ? [0x7f4d08ae5976]
  File ?, in openage::error::terminate_handler()+0xf4 [0x7f4d0c3a6191]
  File ?, in openage::error::StackAnalyzer::analyze()+0x87 [0x7f4d0c3a6795]

standard terminate handler

terminate called without an active exception

FATAL: terminate has been called
FATAL: terminate has been called
Makefile:22: recipe for target 'run' failed
```

ChipmunkV:
Try disabling Console. There is no synchronization around the fontconfig at all right now. Btw, that backtrace doesn't look like from a debug build.

ChipmunkV:
Got some keyboard input working. Also changed to the single-thread mode. That should fix the fontconfig-related crashes too.

TheJJ:
Awesome :dancers:

ChipmunkV:
On the way to implementing the world editor: now GUI can show game textures.

TheJJ:
Holy fucking shit i can press buttons and enter text!!!! YEEEAA!

ChipmunkV:
Also add buttons and change stuff without restart.

franciscod:
so libopenage will be a  qt project with qmake or the build process stays
the same?

TheJJ:
It's just another dependency, the buildsystem won't be changed other than that.
```
qt >=5
dev-qt/qtgui >=5
qtquick >=5
```

franciscod:
probably a noob question but anyway: why there's a lot of these?

```c++
class guiItemLink;
```

ChipmunkV:
All 10 lines of changes to the build files (except for adding source files) are in the first commit "qt build deps".

The question about the Link classes is actually the central question to this thing.

In order to get some data from a game class (for example, suppose we have class ResourceAmount in `libopenage/economy`), there must be a QObject-derived class ResourceAmountLink in `libopenage/gui` that adapts it to what QmlEngine uses. GuiItemLink pointer in ResourceAmount points to ResourceAmountLink to be able to know who is who when looking at ResourceAmount object.

ChipmunkV:
GUI integration is done. There is almost a feature parity (maybe some button was forgotten; and the Console wasn't touched at all).

For the resource preparation I'll start another PR.

franciscod:
niiiiiiice

TheJJ:
Cool stuff!

You either have to take and fight travis (probably with http://askubuntu.com/a/279422/371397)... or wait until #397 is in place and we can simply install qt5 ourselves.

I'll review the code, overall it looks good so far!

TheJJ:
Now that we finally have kevin, please tell me the exact new dependencies so I can install them in the VM. Then, just amend a commit and push again so the build is triggered.

ChipmunkV:
qtdeclarative5-dev package pulls everything that is needed, if on Debian. The version has to be at least 5.4.

ChipmunkV:
Removed a non-atlas image provider, fixed tiling support, split CMakeLists, moved around some stuff in the EditorMode because of the changes in the GameSpec/Civilisation.

TheJJ:
Under the arch vm, i installed `qt5-quickcontrols`, in debian `qtdeclarative5-dev`. Should be building there now :)
Please click the details links then and see what the builder complains about. If you view that link with `curl -N` you can see the output live on your terminal.

ChipmunkV:
Added install. Anything code-related left?

CMakeLists.txt
```diff
@@ -64,6 +64,13 @@ endif()
 set(BUILDSYSTEM_DIR "${CMAKE_SOURCE_DIR}/buildsystem")
 set(CMAKE_MODULE_PATH "${BUILDSYSTEM_DIR}" "${BUILDSYSTEM_DIR}/modules/")
 
+# TODO: why this breaks when moved to the libopenage/CMakeLists.txt?
+set(QT_VERSION_REQ "5.4")
+find_package(Qt5Core ${QT_VERSION_REQ} REQUIRED)
+find_package(Qt5Quick ${QT_VERSION_REQ} REQUIRED)
+# TODO: do not automoc everything
+set(CMAKE_AUTOMOC ON)
+
```
TheJJ: we should move that to the correct location :)

ChipmunkV: Try it :)

ChipmunkV: There is a conflict with this line: buildsystem/python.cmake:126:

    target_link_libraries("${TARGETNAME}" "${PYEXT_LINK_LIBRARY}")

ChipmunkV:
Explicitly stating symbol visibility for the libraries that libopenage uses.

Cython things when linking to the libopenage link to these libraries too. Had to declare Qt "PRIVATE" to prevent that.

It's needed to be prevented because the definition of the targets Qt5::Core and Qt5::Quick must come before the `target_link_libraries` of the cython module in this case.

Maybe someone knows CMake better to fix it differently.

doc/code/gui.md
```diff
@@ -0,0 +1,180 @@
...
+
+    	template<>
+    	struct Wrap<ResourceAmount> {
+    		typedef ResourceAmountLink Type;
+    	};
+
+    	template<>
+    	struct Unwrap<ResourceAmountLink> {
+    		typedef ResourceAmount Type;
+    	};
```
TheJJ: to increase visibility, you can add ```cpp to the code block so syntax highlighting is done correctly.

libopenage/assetmanager.cpp
```diff
@@ -17,10 +17,11 @@
 
 namespace openage {
 
-AssetManager::AssetManager(util::Dir *root)
+AssetManager::AssetManager(gui::GuiItemLink *gui_link)
 	:
-	root{root},
-	missing_tex{nullptr} {
+	root{std::string()},
```
TheJJ: You can leave out this init. An error should occur if the data dir wasn't set but is fetched, otherwise funny bugs may occur?

libopenage/assetmanager.cpp
```diff
@@ -32,15 +33,27 @@ AssetManager::AssetManager(util::Dir *root)
 }
 
 util::Dir *AssetManager::get_data_dir() {
-	return this->root;
+	return &this->root;
+}
+
+std::string AssetManager::get_data_dir_string() const
+{
+	return this->root.basedir;
```
TheJJ: bracket on the same line as the declaration pls :)

libopenage/engine.cpp
```diff
@@ -548,4 +542,12 @@ void Engine::move_phys_camera(float x, float y, float amount) {
 	this->engine_coord_data->camgame_phys += cam_delta.to_phys3();
 }
 
+void Engine::start_game(std::unique_ptr<GameMain> game) {
+	// TODO: maybe implement a proper 1-to-1 connection
+	assert(game);
```
TheJJ: We didn't use any asserts so far, please use our exception system instead (awesome backtraces!).

libopenage/assetmanager.cpp
ChipmunkV: Assumes current directory. Then gui figures out what to set. For now, it's taken from the command line arguments.

libopenage/engine.cpp
ChipmunkV: Asserts are indicating broken code, exceptions inform about wrong user-supplied data. Can't mix them.

ChipmunkV:
https://github.com/SFTtech/openage/pull/504 allowed to partially show a background of the HUD.

libopenage/engine.cpp
TheJJ: So far we have `util::Error` is handled exactly like an assert, it quits the program, but allows backtrace collection and error messages. We don't have an exception hierarchy yet, so all exceptions are fatal and we only used them for that. So I'd say there's no need for any `assert` and we can use `util::Error`.

TheJJ:
On my gentoo machine I get:

```
CMake Error at libopenage/CMakeLists.txt:152 (target_link_libraries):
  The plain signature for target_link_libraries has already been used with
  the target "libopenage".  All uses of target_link_libraries with a target
  must be either all-keyword or all-plain.

  The uses of the plain signature are here:

   * libopenage/CMakeLists.txt:110 (target_link_libraries)
```

This is my patch to make it work, as all `target_link_libraries` calls need to have `PUBLIC`/`PRIVATE` specifiers if they're used once.

``` patch
diff --git a/libopenage/CMakeLists.txt b/libopenage/CMakeLists.txt
index b5a29e5..f6c66de 100644
--- a/libopenage/CMakeLists.txt
+++ b/libopenage/CMakeLists.txt
@@ -94,7 +94,7 @@ endif()
 
 if(GCCBacktrace_FOUND)
        include_directories(${GCCBacktrace_INCLUDE_DIRS})
-       target_link_libraries(libopenage "${GCCBacktrace_LIBRARIES}")
+       target_link_libraries(libopenage PUBLIC "${GCCBacktrace_LIBRARIES}")
        have_config_option(backtrace BACKTRACE true)
 else()
        have_config_option(backtrace BACKTRACE false)
@@ -107,7 +107,7 @@ endif()
 if(WANT_GPERFTOOLS_PROFILER AND GPERFTOOLS_FOUND)
        have_config_option(gperftools-profiler GPERFTOOLS_PROFILER true)
        include_directories(${GPERFTOOLS_INCLUDE_DIR})
-       target_link_libraries(libopenage ${GPERFTOOLS_PROFILER})
+       target_link_libraries(libopenage PUBLIC ${GPERFTOOLS_PROFILER})
 else()
        have_config_option(gperftools-profiler GPERFTOOLS_PROFILER false)
 endif()
@@ -115,7 +115,7 @@ endif()
 if(WITH_GPERFTOOLS_TCMALLOC AND GPERFTOOLS_FOUND)
        have_config_option(gperftools-tcmalloc GPERFTOOLS_TCMALLOC true)
        include_directories(${GPERFTOOLS_INCLUDE_DIR})
-       target_link_libraries(libopenage ${GPERFTOOLS_TCMALLOC})
+       target_link_libraries(libopenage PUBLIC ${GPERFTOOLS_TCMALLOC})
 else()
        have_config_option(gperftools-tcmalloc GPERFTOOLS_TCMALLOC false)
 endif()
```

ChipmunkV:
Ok, there is no need to link anything else to qt. So, link them PRIVATEly until qt build scripts change declaration of the IMPORTED libraries to GLOBAL.

libopenage/engine.cpp
ChipmunkV: Asserts are removed by the preprocessor in release, so there is a need for additional macros that is mimicking or using NDEBUG, if we go that way.

Something like:
```cpp
#ifdef EXCEPTION_ASSERT
	#ifndef NDEBUG
		#include <exception>
		#define Assert(expr) \
			((expr) \
				? static_cast<void> (0) \
				: throw std::exception(#expr##__FILE__##__LINE__##__func__)))
	#else
		#define Assert(expr) (void)0
	#endif
#else
	#include <cassert>
	#define Assert(expr) assert(expr)
#endif
```

TheJJ: ok, but why don't we just leave it in? or do you wanna optimize the branch out? I mean even in the "release" build those errors should be detected and reportable I think.

ChipmunkV: There are also some crazy checks that can't be in release. So, I didn't invent any macro for them and used the assert from the library.

I don't see exceptions as fatal errors, so haven't considered them to be the right tool. If there is a explicitly defined fatal exception that no one would try to start handling later in the project, then it'll be usable.

Another concern is the `libopenage/gui/guisys`, which is actually a future qt-sdl library. It should stay as Qt: exception-neutral.

TheJJ: Yes, the `error::Error` is fatal by design. If one inherits from that and uses it for code flow (which will happen someday for user input, mod packs, ...) then the backtrace collection can be disabled upon construction, this already works.

If parts of the code should stay exception neutral, then i'm ok with that part, but whenever a backtrace would make sense elsewhere, our `Error` is better.

ChipmunkV: Ok, we have that:

there are errors, caused by the bug and errors, caused by the user
the former should propagate to the main and stop the program
the latter should be handled and print error dialogs or whatever
and for both we can use our current exception system

What about the syntax? You're saying that it's "`if() throw Error()`"? Inability to use them liberally will discourage checking and documenting.

TheJJ: What do you mean with using them liberally? The only important part is to use `Error(MSG(err) << )`, how and when you throw it is up to you.

ChipmunkV: I mean, it's too much text to type.

TheJJ: Hmm, I tend to disagree: Explicit is better than implicit, and the `MSG` macro does all the magic to track the originating line (btw what will happen if we have a generated cpp file via `moc`? is the line numbering converted as well?) and the exception is thrown through the native c++ mechanism. I hardly see a way of making it more minimal except having another magic macro as an alias.

ChipmunkV: moc doesn't throw and it doesn't copy-paste user-written code (so it doesn't throw).

Explicit is better than implicit, but implicit is better then none. And you get none if the syntax is more complex than an assert. It's like that for a reason. You normally move and change dozens of asserts while developing.

TheJJ: Hm, if you think it's good and necessary, then feel free to add a macro that throws the `Error`. We already have very similar code for the testing subsystem here: `libopenage/testing/testing.h`. It should still be visible instantly that this isn't the same as the standard `assert`, so maybe a `ENSURE(cond, message)` is a good keyword.

libopenage/engine.cpp
```diff
@@ -168,6 +168,19 @@ Engine::Engine(util::Dir *data_dir, const char *windowtitle)
 
 	glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
 
+	// qml sources will be installed to the asset dir
+	// otherwise assume that launched from the source dir
+	using namespace std::string_literals;
+	auto qml_search_paths = {this->data_dir->basedir, "libopenage/gui"s};
```
TheJJ: isn't it automatically constructed? http://en.cppreference.com/w/cpp/string/basic_string/basic_string method (5)

(maybe not cause it's an rvalue)

libopenage/engine.h
```diff
@@ -219,6 +233,8 @@ class Engine : public ResizeHandler, public options::OptionNode {
 	*/
 	input::InputManager &get_input_manager();
 
+	void announce_global_binds();
```
TheJJ: what does this do? documentation please :)

libopenage/game_control.cpp
```diff
@@ -458,58 +321,22 @@ bool EditorMode::available() const {
...
 
+void EditorMode::set_current_type_id(int current_type_id) {
+	this->current_type_id = current_type_id;
+}
 
-bool EditorMode::on_mouse_wheel(int direction, coord::window) {
-	Engine &engine = Engine::get();
-	Player *player = engine.player_focus();
-	GameSpec *spec = engine.get_game()->get_spec();
+void EditorMode::set_current_terrain_id(openage::terrain_t current_terrain_id) {
```
TheJJ: we're in `namespace openage` already

libopenage/game_control.cpp
```diff
@@ -555,83 +382,209 @@ bool EditorMode::on_single_click(int del, coord::window point) {
...
 	return false;
 }
 
-GameControl::GameControl(openage::Engine *engine)
-	:
-	engine{engine},
-	active_mode_index{0} {
+void EditorMode::announce_categories() {
+	if (this->game_control)
+		if (auto player = this->game_control->get_current_player())
```
TheJJ: You check if the current player is not `nullptr` here, is it ever `nullptr`?

libopenage/game_control.cpp
```diff
@@ -555,83 +382,209 @@ bool EditorMode::on_single_click(int del, coord::window point) {
...
+}
 
-	// modes list
-	this->modes.push_back(std::make_unique<CreateMode>());
-	this->modes.push_back(std::make_unique<EditorMode>());
-	this->modes.push_back(std::make_unique<ActionMode>());
+void EditorMode::announce_category_content(const std::string &category_name) {
+	if (this->game_control)
+		if (auto player = this->game_control->get_current_player()) {
+			auto inds = player->civ->get_category(category_name);
```
TheJJ: but you omit the check here, which is why I'm not sure it's necessary?

libopenage/game_control.cpp
```diff
@@ -555,83 +382,209 @@ bool EditorMode::on_single_click(int del, coord::window point) {
...
+	active_mode{},
+	active_mode_index{-1},
+	current_player{1},
+	gui_signals{this},
+	gui_link{gui_link} {
+}
+
+void GameControl::set_engine(Engine *engine) {
+	// TODO: decide to either go for a full Engine QML-singleton or for a regular object
+	assert(!this->engine || this->engine == engine);
```
TheJJ: I'd go for a regular object, but it's kinda hard accessing it from everywhere. See #329.

libopenage/game_control.cpp
```diff
@@ -555,83 +382,209 @@ bool EditorMode::on_single_click(int del, coord::window point) {
...
+GameControl::GameControl(gui::GuiItemLink *gui_link)
+	:
+	engine{},
+	game{},
+	active_mode{},
+	active_mode_index{-1},
+	current_player{1},
+	gui_signals{this},
+	gui_link{gui_link} {
+}
```
TheJJ: Why can't we initialize all those members when constructing? I feel bad when we have to set all those after constructing by calling setter functions.

libopenage/game_control.h
```diff
@@ -103,6 +135,32 @@ class ActionMode : public OutputMode {
 
 	// used for random type creation
 	rng::RNG rng;
+
+	std::tuple<int, int, int, int> resources;
```
TheJJ: We decided to use an arbitrary amount of resources and resource types which are defined via the gamespec (later `nyan`). For now it's ok, but we should keep that flexibility in mind.

libopenage/game_singletons_info.h
```diff
@@ -0,0 +1,30 @@
...
+	GameSingletonsInfo(Engine *engine, const std::string &data_dir)
+		:
+		GuiSingletonItemsInfo{},
+		engine{engine},
+		data_dir{data_dir} {
+	}
+
+	Engine *engine;
+	std::string data_dir;
+};
```
TheJJ: That's a good idea. This way we can make the engine a regular object, and for the "current" qml and nyan context, those values are directly accessible.

libopenage/gamestate/game_spec.cpp
```diff
@@ -377,4 +379,74 @@ void Sound::play() const {
...
+	this->active = active;
+
+	start_loading_if_needed();
+}
+
+void GameSpecHandle::set_asset_manager(AssetManager *asset_manager) {
+	if (this->asset_manager != asset_manager) {
+		this->asset_manager = asset_manager;
+
+		start_loading_if_needed();
```
TheJJ: `this->` please for all the `start_loading_if_needed()` calls.

libopenage/gamestate/game_spec.cpp
```diff
@@ -377,4 +379,74 @@ void Sound::play() const {
...
+
+std::shared_ptr<GameSpec> GameSpecHandle::get_spec() {
+	return this->spec;
+}
+
+void GameSpecHandle::start_loading_if_needed() {
+	if (this->active && this->asset_manager && !this->spec) {
+		this->spec = std::make_shared<GameSpec>(*this->asset_manager);
+
+		start_load_job();
```
TheJJ: `this->`

libopenage/gui/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+	integration/private/gui_standalone_subtexture.cpp
+	integration/private/gui_texture.cpp
+	integration/private/gui_texture_factory.cpp
+	integration/private/gui_texture_handle.cpp
+)
+
+add_subdirectory("guisys")
+
+add_sources(libopenage
+	${QT_SDL_SOURCES}
```
TheJJ: Is there any reason the `add_sources` isn't called in the `CMakeLists.txt` of each folder? Currently all our source directories add the sources in that directory and recurse to other directories then.

libopenage/gui/assetmanager_link.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include "assetmanager_link.h"
+
+#include <QtQml>
+
+namespace openage {
+namespace gui {
+
+namespace {
+const int registration = qmlRegisterType<AssetManagerLink>("com.openage", 1, 0, "AssetManager");
```
TheJJ: Why `com.openage`? Can't we use something less corporate like `lol.openage`? :grinning:

libopenage/gui/category_contents_list_model.cpp
```diff
@@ -0,0 +1,100 @@
...
+			QObject::connect(this->editor_mode, &EditorModeLink::categories_content_changed, this, &CategoryContentsListModel::on_categories_content_changed);
+			QObject::connect(this->editor_mode, &EditorModeLink::category_content_changed, this, &CategoryContentsListModel::on_category_content_changed);
+		}
+
+		this->on_categories_content_changed();
+	}
+}
+
+void CategoryContentsListModel::on_category_content_changed(const std::string &category_name, std::vector<std::tuple<index_t, uint16_t>> type_and_texture) {
+	if (this->name == QString::fromStdString(category_name)) {
```
TheJJ: hrrrg qt...

libopenage/gui/engine_link.cpp
```diff
@@ -0,0 +1,62 @@
...
+	return this->global_binds;
+}
+
+void EngineLink::on_global_binds_changed(const std::vector<std::string>& global_binds) {
+	QStringList new_global_binds;
+	std::transform(std::begin(global_binds), std::end(global_binds), std::back_inserter(new_global_binds), [] (const std::string &s) {return QString::fromStdString(s);});
+
+	if (this->global_binds != new_global_binds) {
+		this->global_binds = new_global_binds;
+		emit global_binds_changed();
```
TheJJ: `this->`?

libopenage/gui/game_control_link.cpp
```diff
@@ -0,0 +1,334 @@
...
+CreateModeLink::~CreateModeLink() {
+}
+
+ActionModeLink::ActionModeLink(QObject *parent)
+	:
+	Inherits{parent},
+	food{},
+	wood{},
+	gold{},
+	stone{} {
```
TheJJ: Again, we shouldn't fix the resource types :) Think of awesome mods adding whatever you can imagine. Oil, silicon, ...

libopenage/gui/game_control_link.h
```diff
@@ -0,0 +1,259 @@
...
+class OutputModeLink;
+
+template<>
+struct Wrap<OutputMode> {
+	typedef OutputModeLink Type;
+};
+
+template<>
+struct Unwrap<OutputModeLink> {
+	typedef OutputMode Type;
```
TheJJ: In other parts of the code we used `using` instead of `typedef`.

libopenage/gui/game_control_link.h
```diff
@@ -0,0 +1,259 @@
...
+	typedef EditorMode Type;
+};
+
+class EditorModeLink : public Inherits<OutputModeLink, EditorModeLink> {
+Q_OBJECT
+
+Q_PROPERTY(int currentTypeId READ get_current_type_id WRITE set_current_type_id)
+Q_PROPERTY(int currentTerrainId READ get_current_terrain_id WRITE set_current_terrain_id)
+Q_PROPERTY(bool paintTerrain READ get_paint_terrain WRITE set_paint_terrain)
+Q_PROPERTY(QStringList categories READ get_categories NOTIFY categories_changed)
```
TheJJ: these should be indented, right?

libopenage/gui/game_control_link.h
```diff
@@ -0,0 +1,259 @@
...
+	void on_resources_changed(std::tuple<int, int, int, int> resources);
+	void on_ability_changed(const std::string &ability);
+
+private:
+	virtual void on_core_adopted() override;
+
+	int food;
+	int wood;
+	int gold;
+	int stone;
```
TheJJ: the engine should especially be capable running star wars galactic battlegrounds some day as it's the exact same engine just with another game loaded.

libopenage/game_control.cpp
ChipmunkV: That's related to the QML fundamental lack of constructors (because the entire declaration of the object in the script is something like a constructor).

The script engine creates C++ objects in no particular order and with default values. Then it calls setters on the objects according to the values that are defined in the script for their members.

libopenage/gui/game_control_link.h
ChipmunkV: If something isn't clearly in the codestyle doc, then it's up to my auto-formatter. :)

Ok, I'll indent that everywhere.

libopenage/gui/assetmanager_link.cpp
ChipmunkV: Actually, can do. Because have to add 'sfttech' there anyways.

libopenage/engine.cpp
ChipmunkV: Ok, I've looked into it and with `vector<string> qml_search_paths = {...}`, the implicit char* to string conversion works. But with `auto qml_search_paths = {...}`, the behaviour is more like with `template<typename T>`.

libopenage/gui/CMakeLists.txt
ChipmunkV: I'll make separate `CMakeLists.txt` for everything except `libopenage/gui/guisys`. That last thing should become a lib some day.

ChipmunkV:
Separated a qtsdl namespace. Added some `CMakeLists.txt`, fixed asserts, comment, redundant namespace specifiers, missing 'this', using/typedef, qobject indent, qml namespaces, resource types.

libopenage/game_control.cpp
```diff
@@ -555,83 +387,209 @@ bool EditorMode::on_single_click(int del, coord::window point) {
...
 
+Player* GameControl::get_current_player() const {
+	if (auto game = this->engine->get_game()) {
+		unsigned int number = game->players.size();
+		return &game->players[this->current_player % number];
+	}
+	return nullptr;
+}
+
+void GameControl::do_set_mode(int mode_index) {
```
TheJJ: Why do we have the `set_mode` detour?

libopenage/game_singletons_info.h
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Accessible during creation of any singleton QML item.
+ */
+struct GameSingletonsInfo : qtsdl::GuiSingletonItemsInfo {
+	GameSingletonsInfo(Engine *engine, const std::string &data_dir)
+		:
+		GuiSingletonItemsInfo{},
+		engine{engine},
+		data_dir{data_dir} {
+	}
```
TheJJ: We could move the constructor implementation to the cpp file, or is it any advantage having in the header?

ChipmunkV: Advantage is that there is no cpp file to add to CMakeLists.txt. I'll make a cpp file.

libopenage/game_control.cpp
ChipmunkV: Maybe because there are some internal uses. Make it one function?

janisozaur:
I came here looking for screenshots and got none. pls fix. kthxbai.

ChipmunkV:
Well, it looks exactly like Age of Empires, no surprises (if you use AoE assets).

libopenage/game_control.cpp
TheJJ: Couldn't spot any, i think you can meld them, yes.

goto-bus-stop:
Here's one!

@janisozaur :eyes:

ChipmunkV:
Ok, I'll recap here what I remember about this pull request:

* locale issue (probably with languages that use decimal commas): when some locale is set, the game loads the gamedata with errors, so all units spawn dead (https://github.com/SFTtech/openage/issues/423). Qt sets locale, everything dies.
Need to fix loading, I think.

* Mac OS X: I have a [branch](https://github.com/ChipmunkV/openage/tree/qt-gui-cocoa) with an implementation of a platform-specific bit for that. I may include it, but I don't have a Mac for testing. (the windows version I've tested separately from openage),

ChipmunkV:
Ok, I've fixed locale according to the [Qt docs](http://doc.qt.io/qt-5/qcoreapplication.html#locale-settings) and added the Mac code.

TheJJ:
Awesome! I'll go and merge it, if we encounter minor issues, we can fix them afterwards.

janisozaur:
So nice! Great job @ChipmunkV

PR #426, opened 2015-12-25
Fix: Civilian holding ressources when idled

tim2000:
Civilian where taking theirs idle sprite even when holding resources.
Since the carrying sprite is walking, we have to stop the frame update when it happened.

It's my first contribution, and I have some difficulties with git, so I've made just a tiny commit for the first time.


TheJJ:
Cool, works and is better than before. Thanks :)

Only issue remaining (that can be a separate fix of course):
The idle units with gather type and amount >0 no longer have an idle _animation_ and may even stand frozen mid-step.

PR #453, opened 2016-01-04
Fix a typo in ideas/gameplay

steelbrain:
Just like the title says

### DIFF
--- a/doc/ideas/gameplay.md
+++ b/doc/ideas/gameplay.md
@@ -173,7 +173,7 @@ Resources
 
 ### Stone balancing
 
-Stone cost for all buildings, rebalce stone amount per pile.
+Stone cost for all buildings, rebalance stone amount per pile.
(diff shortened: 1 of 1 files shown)


franciscod:
I think the `!!!111` is intentional, see this: http://urbandictionary.com/define.php?term=!!!111oneone

steelbrain:
Well, you **think** that, I would change it back on a confirmation

franciscod:
¯\_(ツ)_/¯

TheJJ:
Jup, it's intended of course, thanks for the link @franciscod :D

steelbrain:
@TheJJ Done

TheJJ:
Might it be possible that you meld those two commits so that only the typo fix remains? Also, please remove the emoji in the commit message as it will only be displayed on github in a graphical way.

steelbrain:
@TheJJ Done

PR #455, opened 2016-01-05
hacking vs hecking???

kbohinski:


steelbrain:
I think it's intentional

kbohinski:
that was fast. feel free to close then...

steelbrain:
I am not a collaborator, close at will

TheJJ:
`Hecke` is german for hedge, and yes it's an intentional pun :)

PR #464, opened 2016-01-10
Optimize texture frame merging

niklasf:
Directly use numpy arrays instead of building and looping through
Python lists.

### DIFF
--- a/openage/convert/texture.py
+++ b/openage/convert/texture.py
@@ -1,4 +1,4 @@
-# Copyright 2014-2015 the openage authors. See copying.md for legal info.
+# Copyright 2014-2016 the openage authors. See copying.md for legal info.
 
 """ Routines for texture generation etc """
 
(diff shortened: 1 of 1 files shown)


TheJJ:
Looks like an improvement to me :)
However, I played around a lot with the image writer back then and the current solution was the fastest one.

Of course I may not have tried `atlas_data[pos_y:pos_y+sub_h, pos_x:pos_x+sub_w] = sub_frame.data`, but still it would be nice if you could benchmark it if you got time. You can use the `singlefile.py` tool for an easy invocation. I'm pretty sure though that this is faster than before.

niklasf:
Command: time python3 -m openage convert-file Age2HD/Data/graphics.drs 877.slp $PWD/test

Total times for master (70d2bf99421a4100822): `0.50 0.61 0.48 0.51 0.48 0.50`
Total times for optimize-frame-merging: `0.47 0.55 0.37 0.42 0.44 0.36 0.38`

Looks like a decent improvement, not only in style.

Also bumped the copyright year.

PR #469, opened 2016-01-12
Expanded the critical section to include init()

lyle-nel:
Won't this init() interleave when multiple threads enter the function?
The potential result would be the wrong thread_id paired with the wrong timestamp.
Moving the guard up would mitigate it.


TheJJ:
Indeed, `this->thread_id = util::get_current_thread_id();` is racy otherwise.

TheJJ:
And now the copyright year in `libopenage/log/message.cpp` is 2015, but we have 2016 :)

lyle-nel:
Sorry, I haven't used travis before.

TheJJ:
Thanks man, nice catch :)

PR #475, opened 2016-01-15
generator: use math::PI instead of M_PI

franciscod:
gentlemen let's honor #218 pls


TheJJ:
can you quickly squash it then? :stuck_out_tongue:

franciscod:
sure!

PR #476, opened 2016-01-15
Some tweaks for configuring via cmake on MSYS2

franciscod:
These are mostly random fixes that I needed to do in order to run cmake on MSYS2.

I hope they don't break anything!

### DIFF
--- a/libopenage/CMakeLists.txt
+++ b/libopenage/CMakeLists.txt
@@ -164,4 +164,4 @@ if(NOT APPLE)
 	)
 endif()
 
-install(TARGETS libopenage LIBRARY DESTINATION lib)
+install(TARGETS libopenage DESTINATION lib LIBRARY DESTINATION lib)
--- a/buildsystem/modules/FindPython.cmake
+++ b/buildsystem/modules/FindPython.cmake
@@ -20,7 +20,7 @@ function(py_exec STATEMENTS RESULTVAR)
 	# aborts with a fatal error on error.
 	# no single quotes are allowed in STATEMENTS.
 	exec_program(
-		"${PYTHON}" ARGS -c "'${STATEMENTS}'"
+		"${PYTHON}" ARGS -c "\"${STATEMENTS}\""
 		OUTPUT_VARIABLE PY_OUTPUT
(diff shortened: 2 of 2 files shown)


libopenage/CMakeLists.txt
```diff
@@ -164,4 +164,4 @@ if(NOT APPLE)
 	)
 endif()
 
-install(TARGETS libopenage LIBRARY DESTINATION lib)
+install(TARGETS libopenage DESTINATION lib LIBRARY DESTINATION lib)
```
TheJJ: uh what? but destination lib was already in there?

franciscod: yeah i dont even understand... but added that and it worked! and doesnt break on linux

TheJJ: according to https://cmake.org/cmake/help/v3.4/command/install.html#installing-targets it shouldn't even work specifying destination before the type.

franciscod: One or more groups of properties may be specified in a single call to the TARGETS form of this command. A target may be installed more than once to different locations. Consider hypothetical targets myExe, mySharedLib, and myStaticLib. The code:
```
install(TARGETS myExe mySharedLib myStaticLib
        RUNTIME DESTINATION bin
        LIBRARY DESTINATION lib
        ARCHIVE DESTINATION lib/static)
install(TARGETS mySharedLib DESTINATION /some/full/path)
```
will install myExe to <prefix>/bin and myStaticLib to <prefix>/lib/static. On non-DLL platforms mySharedLib will be installed to <prefix>/lib and /some/full/path. On DLL platforms the mySharedLib DLL will be installed to <prefix>/bin and /some/full/path and its import library will be installed to <prefix>/lib/static and /some/full/path.

---

what i've commited looks like a shorthand version of that example

TheJJ: according to https://cmake.org/cmake/help/v3.4/command/install.html#installing-targets it shouldn't even work specifying destination before the type.

franciscod: the "On DLL / On non-DLL" bit explains this

PR #477, opened 2016-01-15
crossplatform/timing: use C++11 chrono

franciscod:
BAM, C++11!!

http://en.cppreference.com/w/cpp/chrono


franciscod:
should we move this outside crosplatform?

TheJJ:
Probably, yes. just `util/timing.{cpp,h}` or something. The crossplattform folder is kinda weird, we should keep it but avoid using it as a measure to the not-knowing how to use it properly^^

Issue #364, opened 2015-08-25
Villager carry resource types 'switch'


zesterer:
I had a villager who was carry a lot of berries. When I told him to go collect gold instead, his berries miraculously turned into gold (it wasn't just an animation either - when he went back to the Town Centre, it registered it as gold resource). Either this is a bug, or you've added an alchemy feature.

TheJJ:
Didn't you say in #363 that they don't carry back to the town centre?

zesterer:
My villager had a lot of berries. Then, I told him to go mine gold (I never let him reach the gold mine). But his animation turned to him carrying gold, and when I finally got him to hand his resources over to the town centre, it registered what he was carrying as gold.

mirelon:
I think the cause is here: https://github.com/SFTtech/openage/blob/master/libopenage/unit/action.cpp#L793
Am I right? Can I make a pull request with added condition that if the resource type is different, then the amount of harvested resource will be set to zero? (As my first contribution)

mirelon:
It is fixed only for a moving villager. When a villager becomes idle, the amount magically reappears. Is the amount stored also somewhere else?

mirelon:
I think it is because of the wrong graphic used. When moving, there is a conditional graphic: https://github.com/SFTtech/openage/blob/master/libopenage/unit/action.cpp#L464. But when initializing IdleAction, there is no such condition: https://github.com/SFTtech/openage/blob/master/libopenage/unit/action.cpp#L368

PR #482, opened 2016-01-15
Correct graphic for idle gathering villager carrying zero amount

mirelon:
Relates to the follow-up in https://github.com/SFTtech/openage/issues/364


TheJJ:
Your commit has the wrong email address :)

mirelon:
I see, it was not a good idea to run this script:) https://help.github.com/articles/changing-author-info/

PR #492, opened 2016-01-24
Added support for importing string data from HD 2.0

jamesmcm:
Added support for HD Edition 2.0, where string data is in DLLs in PATH/Data/LANG/*.dll

This might be better done by detecting that version explicitly, rather than checking for DLLs in the languages folder, but this works for now.

Prior to this fix, the importer would try to read the .dlls as text files, resulting in the following error:

      File "run.py", line 15, in <module>
        main()
      File "/home/archie/openage/openage/__main__.py", line 94, in main
        return args.entrypoint(args, cli.error)
      File "/home/archie/openage/openage/game/main.py", line 33, in main
        if not convert_assets(assets, args):
      File "/home/archie/openage/openage/convert/main.py", line 125, in convert_assets
        for current_item in convert(args):
      File "/home/archie/openage/openage/convert/driver.py", line 126, in convert
        yield from convert_metadata(args)
      File "/home/archie/openage/openage/convert/driver.py", line 179, in convert_metadata
        stringres = get_string_resources(args)
      File "/home/archie/openage/openage/convert/driver.py", line 69, in get_string_resources
        stringres.fill_from(read_hd_language_file(langfile, lang, enc='iso-8859-1'))
      File "/home/archie/openage/openage/convert/hdlanguagefile.py", line 25, in read_hd_language_file
        num, string = line.split(None, 1)
    ValueError: not enough values to unpack (expected 2, got 1)

### DIFF
--- a/copying.md
+++ b/copying.md
@@ -71,6 +71,7 @@ _the openage authors_ are:
 | Michael Kilby               | kilbyjmichael               | kilbyjmichael@gmail.com               |
 | Michal Kováč                | mirelon                     | miso@github.ksp.sk                    |
 | Patrik Stutz                | VanCoding                   | patrik.stutz@gmail.com                |
+| James McMurray              | jamesmcm                    | jamesmcm03@gmail.com                  |
 
 If you're a first-time commiter, add yourself to the above list. This is not
 just for legal reasons, but also to keep an overview of all those nicknames.
--- a/openage/convert/driver.py
+++ b/openage/convert/driver.py
@@ -1,5 +1,5 @@
 # Copyright 2015-2016 the openage authors. See copying.md for legal info.
-
+# pylint: disable=R0912
 """
 Receives cleaned-up srcdir and targetdir objects from .main, and drives the
 actual conversion process.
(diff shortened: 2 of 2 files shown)


TheJJ:
Your change looks good, but pylint complains about it (see the CI log):
```
************* Module openage.convert.driver
R: 24, 0: Too many branches (14/12) (too-many-branches)
WARNING linting issue: python code is noncompliant
```

I suggest you either define a pylint exception (semi-good) or actually fix the problem by improving the structure by creating new functions :)

jamesmcm:
Yeah, I was looking at this earlier, I think it'd be best to pull out the DLL reading code in to its own function. However, this means we need a function to take a list of directory names and recurse through the filesystem dictionary.

This is somewhat tackled here: http://stackoverflow.com/questions/14692690/access-python-nested-dictionary-items-via-a-list-of-keys

But I'm not sure if using reduce and all that, is actually simpler than the additional if statements and longer function at the moment.

jamesmcm:
So for now I recommend we merge in this code with the pylint exception since it's quite important for new users, and then we can work on a separate code improvement pull request this week or so.

TheJJ:
Can you quickly squash those commits by interactively rebasing? (see `contributing.md` or ask me on irc if you got problems)

jamesmcm:
I keep getting:

"Cannot 'squash' without a previous commit"

When trying to squash it, but I haven't used it before. How do I specify the commands exactly, as git rebase -i just opens a vim window.

iNaKoll:
Hi, 
To squash those 4 commit you need to re base from your first commit ( HEAD~3 ) and squash your last 3 commits
`git rebase -i HEAD~3`
Then in your editor mark as 'squash' your last 3 commits. Then save and exit.

jamesmcm:
Thanks! I think that did it, I'll have to set my editor to emacs instead of vi as well :smile:

TheJJ:
If you need configuration hints for your emacs, [see my .emacs](https://github.com/TheJJ/conffiles/blob/master/homedir/.emacs) :grinning:

TheJJ:
For the future, I recommend following the workflow in `contributing.md` so you have feature branches and don't do a force push to your master branch.

Thanks for the fix, good job!

Issue #425, opened 2015-12-21
Support shell expansions in original game path


TheJJ:
When entering the path to the original game installation, `~` and `$HOME` can't be used.

A simple replace with the environment variable value is probably sufficient.

franciscod:
yessss this annoys us since day 0

niklasf:
This has been fixed by @lukky513: fcc4d8fa69d49f18fafbe6e1

TheJJ:
Wheeeeeee

TheJJ:
Wheeeeeee

PR #493, opened 2016-01-25
applying asset path expansion in both AGE2DIR var and input path

lukky513:
Acording to @TheJJ and **revoof** suggestions, moved the expansion a bit further, possibly enabling #425 close.


TheJJ:
Thanks :)

Issue #389, opened 2015-09-19
Some converted interface images have distorted colors


jfeo:
Quite a few of the converted files look similar to this:
![distorted_colors](https://cloud.githubusercontent.com/assets/915204/9975669/0e506536-5ec5-11e5-9100-8df204f103a6.png)

While other files look mostly fine, with just small 'issues':
![50131 slp](https://cloud.githubusercontent.com/assets/915204/9975685/7162117e-5ec5-11e5-8167-2f45b24af91c.png)

TheJJ:
The reason for this is some slp files require a different color table than `50500 + x`, where `x=0`. If you can figure out what `x` is needed so that these files are generated correctly it would help us. Also, if you can find a link between `x` and the slp file (how it says i wanna have color table `x`) even better.

franciscod:
somehow related to the windows 7 bug which was solved by killing explorer.exe while aoe was running?

TheJJ:
Maybe, but I doubt there exists a relation. The issue there was just that some colors were wrong because of whatever memory error, and in our case all colors for specific slps are wrong because of the wrong color table.

ChipmunkV:
`palette_offset` that is "apparently never used"?

What is the command to get a raw .slp to look into?

TheJJ:
The `openage.convert` python module provides it all:

```python
import openage.convert as conv
slp = conv.slp.SLP(conv.drs.DRS(open("/home/jj/games/aoe2/aoe2/Data/graphics.drs", "rb")).root[b"5311.slp"].open("rb").read())
```

TheJJ:
See #432 for working code :)

goto-bus-stop:
The appropriate palettes for interface files are stored in the .bin files in `interfac.drs`.

Default rendering of 50100.slp: (palette 50500)

![Default](http://i.imgur.com/cn30ur0.png)

Finding the correct palette:

![CLI majik](http://i.imgur.com/Uxt2GmD.png)

Correct rendering:

![Correct](http://i.imgur.com/d8qH9V8.png)

goto-bus-stop:
So the convert script could read the necessary .bin files first (actually `.sin`, apparently, but stored as `.bin` in the DRS tables) and build a mapping of SLP Ids → Palette Ids before rendering SLPs.

TheJJ:
Oh nice! That palette assignments were a missing key point! Cooool :smile: !

franciscod:
great!!

[...]

PR #504, opened 2016-02-07
GUI ingame background assets adaptation (AOC/HD)

ChipmunkV:
There are 1280x1024 backgrounds currently extractable from the AOC game data (if interface.dat is enabled in the conversion script). No subtextures defined.

1. For the ingame HUD: I'm going to manually specify the coordinates of the bottom rectangles that are in the corners.

2. For the top stripe of the ingame HUDs it's possible to detect a repeating pattern and get coordinates of a good piece. I've ported a `visgrep` utility from xautomation to python. It does the pattern search without having something like OpenCV or GIMP as a dependency.

Middle-bottom part of the HUD - that should be probably generated at runtime because the aspect ratio is dynamic. So, it's not a part of this PR.

For the first push there is only the `visgrep.py` utility.

If someone can confirm that we can't extract more metadata from the game resources, I'll hardcode the points 1 and 2. Especially useful info would be the list of filenames that correspond to these 1280x1024 backgrounds, otherwise that will have to be hardcoded too.

### DIFF
--- a/openage/convert/interface/interfacecutter.py
+++ b/openage/convert/interface/interfacecutter.py
@@ -1,0 +1,57 @@
+# Copyright 2016-2016 the openage authors. See copying.md for legal info.
+
+""" Cutting some user interface assets into subtextures """
+
+from ..texture import TextureImage
+
+from .interfacehardcoded import (is_ingame_hud_background,
+                                 TOP_STRIP_PATTERN_CORNERS,
+                                 KNOWN_SUBTEX_CORNER_COORDS)
+
+from .visgrep import visgrep
+
+
+class InterfaceCutter(object):
+    """
+    Callable object that cuts interface textures.
+    """
+    # pylint: disable=too-few-public-methods
+
+    def __init__(self, filename):
+        self.filename = filename
+
+    def __call__(self, image):
+        if is_ingame_hud_background(self.filename):
+            pil_image = image.get_pil_image()
+
+            yield self.__cut_top_strip(pil_image)
+
+            for coords in KNOWN_SUBTEX_CORNER_COORDS:
+                yield TextureImage(pil_image.crop(coords))
+        else:
+            yield image
+
+    def __cut_top_strip(self, pil_image):
+        """
+        Finds a horisontally tilable piece of the strip that's at the top of the HUD.
+
+        ||----///////////-------------///////////-------------///////////-------------///////////||
+                  ^TOP_STRIP_PATTERN_CORNERS ^                    ^  where it is found last  ^
+                  ^           this piece is tileable              ^
+
+        so, cut out a subtexture:
+                  ///////-------------///////////-------------////
+        """
+
+        pattern = pil_image.crop(TOP_STRIP_PATTERN_CORNERS)
+        matches = visgrep(pil_image, pattern, 100000, TOP_STRIP_PATTERN_CORNERS[3])
+
+        if len(matches) < 2 or matches[-1].point[1] != TOP_STRIP_PATTERN_CORNERS[1]:
+            raise Exception("visgrep failed to find repeating pattern (%s)\n" % self.filename)
+
+        return TextureImage(pil_image.crop((
+            TOP_STRIP_PATTERN_CORNERS[0],
+            TOP_STRIP_PATTERN_CORNERS[1],
+            matches[-1].point[0],
+            TOP_STRIP_PATTERN_CORNERS[3]
+        )))
--- a/openage/convert/interface/interfacerename.py
+++ b/openage/convert/interface/interfacerename.py
@@ -1,0 +1,15 @@
+# Copyright 2016-2016 the openage authors. See copying.md for legal info.
+
+""" Renaming interface assets and splitting into direvtories """
+
+from .interfacehardcoded import ingame_hud_background_index
+
+
+def interface_rename(filename):
+    """
+    Returns a human-usable name according to the original and hardcoded metadata.
+    """
+    try:
+        return "interface/hud" + str(ingame_hud_background_index(filename)).zfill(4) + ".slp"
+    except ValueError:
+        return filename
--- a/openage/convert/interface/visgrep.py
+++ b/openage/convert/interface/visgrep.py
@@ -1,0 +1,394 @@
+# Copyright 2016-2016 the openage authors. See copying.md for legal info.
+
+""" Python version of the visgrep utility """
+
+import argparse
+import logging
+from collections import namedtuple
+from PIL import Image
+
+TOOL_DESCRIPTION = """Python translation of the visgrep v1.09
+visual grep, greps for images in another image
+Author of the original C version: Steve Slaven - http://hoopajoo.net"""
+
+EPILOG = """The image.png is
+scanned for detect.png starting from X,Y specified above. When detect.png
+is found, then all the match.png files are scanned at an offset of x,y as
+specified above.  If a match is found, then visgrep prints the x,y and
+index of the item.
+
+For example, image.png is a screenshot and match1.png .. to match5.png are
+images of letters a to e.  Each of these letters is enclosed in a blue box,
+so detect.png is an image of the upper left corner of the box.  This box is
+not included in the match*.png files, so they are actually offset 5 pixels
+down and 4 pixels to the left.  You might run it like this then:
+
+  visgrep -b -t50 -x-4 -y5 image.png match_corner.png match_a.png match_b.png ...
+
+Etc, with all matches listed.  Now suppose the screen showed 'ace' so
+visgrep might output:
+
+0 10,10 0
+12 50,10 2
+7 90,10 4
+
+Showing that match_a.png (index 0) is at 10,10 on the screen.  If no match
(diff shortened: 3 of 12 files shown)


TheJJ:
I think it's possible to figure out all the metadata about the hud graphics, just like @goto-bus-stop posted [here](https://github.com/SFTtech/openage/issues/389#issuecomment-167990267).

I doubt that the positions where it can be clicked are hardcoded in the original.
But for slicing up the HUD images, we probably need some predefined cutting points.

openage/convert/texture.py
```diff
@@ -99,9 +104,8 @@ def __init__(self, input_data, palette=None):
                 # (at least a 10x improvement, 50x would be better).
                 # ideas: remove PIL and use libpng via CPPInterface,
                 #        cythonize parts of SLP.py
-                TextureImage(frame.get_picture_data(palette, self.player_id),
-                             hotspot=frame.info.hotspot)
-                for frame in input_data.frames
+                custom_frame for frame in input_data.frames for
+                custom_frame in (custom_cutter or (lambda x: [x]))(TextureImage(frame.get_picture_data(palette, self.player_id), hotspot=frame.info.hotspot))
```
TheJJ: you might wanna create explicit loops here, it's pretty unreadable otherwise :)

openage/convert/interface/interfacecutter.py
```diff
@@ -0,0 +1,123 @@
...
+                os.path.join(os.path.dirname(os.path.realpath(__file__)), 'visgrep.py'),
+                '-J', str(TOP_STRIP_PATTERN_XYWH[1] + TOP_STRIP_PATTERN_XYWH[3]),
+                '-u', str(TOP_STRIP_PATTERN_XYWH[0]),
+                '-v', str(TOP_STRIP_PATTERN_XYWH[1]),
+                '-M', str(TOP_STRIP_PATTERN_XYWH[2]),
+                '-N', str(TOP_STRIP_PATTERN_XYWH[3]),
+                '-W', str(TOP_STRIP_PATTERN_XYWH[2]),
+                '-H', str(TOP_STRIP_PATTERN_XYWH[3]),
+                '-t', '100000',
+                uncut.name, uncut.name, uncut.name], stdout=PIPE)
```
TheJJ: after the proof of concept we should of course directly call the python code instead of calling a new process.

ChipmunkV: I disagree. The tool has a well-defined command-line interface that existed before. There is no need to make the project be a monolithic python pile of code.

If we make a python interface, that thing will become an integral part of the openage (random NIH implementation of pattern matching), rather than just a "dependency that we had to rewrite, but kept the old interface".

If there is a need for the python interface, I'll try something else with python interface, like OpenCV.

TheJJ: No, there's no need for a py interface. I mean it's still possible implementing in a way it can either be used via commandline or by importing and using it, which would even improve the "dependency we had to rewrite" by extending it with a nicer API interface.

Pulling OpenCV instead of that might be a bit heavy.

ChipmunkV:
It's functionally complete. So, the 1 and 2 points can be shown ingame and change according to the civilisation. (no middle paperish HUD part, no buttons)

Now cleaning mess in the code.

TheJJ:
What advantage does the grepping bring? Isn't it enough saving the positions for the detected pattern as well in `KNOWN_SUBTEX_CORNER_COORDS`? Not that I don't want the `visgrep`, I'm just curious.

ChipmunkV:
KNOWN_SUBTEX_CORNER_COORDS is the same for all the civilisations: it's the coordinates of the bottom left, bottom right rectangles and another five squares at the top (wood, stone, gold, food, people).

Hardcoding the detected position will be per HUD.

I don't know if there are different HUDs in other game versions, or maybe the some GUI mods exist (end we have to be able to import them too). So, I assumed that per-HUD coordinates won't work.

ChipmunkV:
And I'm not emotionally attached to that code. At the beginning I thought that there is something in the python image processing libraries to do the pattern matching.

TheJJ:
I see, ok. But what does the `visgrep` grep for? How do you determine the pattern? Or does the tool detect the repetition and then return the pattern that is repeated? Still, we'd have to supply the "general area" where the search is performed, and this is again dependent on knowlege of the hud image.  I just want to optimize the process as good as possible so we stay flexible in the future especially for mods, as you already said :sunglasses:

ChipmunkV:
```
||----///////////-------------///////////-------------///////////-------------///////////||
          ^ arbitrary search pattern ^                    ^  where it is found last  ^
          ^           this piece is tileable              ^

so, cut it into a subtexture:
          ///////-------------///////////-------------////
```

The coordinates of the "arbitrary search pattern" are set to (400, 0, 64, 32) for no particular reason (well, before 400 there are resource indicators).

ChipmunkV:
And it's cleaned up.

openage/convert/interface/interfacerename.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def interface_rename(filename):
+    """
+    Returns a human-usable name according to the original and hardcoded metadata.
+    """
+    try:
+        return "interface/hud" + str(ingame_hud_background_index(filename)).zfill(4) + ".slp"
+    except ValueError:
+        return filename
```
TheJJ: ok, but that's rather hacky :stuck_out_tongue_winking_eye: 

how about something like?

``` python
if filename in ingame_hud_background_index:
    return "interface/hud/%s.slp" % filename.zfill(4)
else:
    return filename
```

ChipmunkV: ingame_hud_background_index is the civ index.

TheJJ: The string formatting thingy doesn't matter much, I was just trying to get rid of the exception handling. But it you think your current way is the best one, then it's ok.

openage/convert/interface/visgrep.py
```diff
@@ -0,0 +1,389 @@
...
+            mpt = Point(sptx + where.x, spty + where.y)
+
+            logging.debug("Pixel cmp @ main: %d,%d, subimage: %d,%d", mpt.x, mpt.y, sptx, spty)
+
+            # Grab pels and see if they match
+            mpx = img_pixel_get(master, mpt)
+            spx = img_pixel_get(subimage, Point(sptx, spty))
+
+            badness += abs(img_pixel_cmp(mpx, spx))
+            # DMSG( "Badness {}\n", badness );
```
TheJJ: small leftover

openage/convert/interface/visgrep.py
```diff
@@ -0,0 +1,389 @@
...
+        logging.debug("Subimage would not fit here")
+        return 1000
+
+    badness = 0
+    for sptx in range(subimage.size[0]):
+        for spty in range(subimage.size[1]):
+            # Map U/V to X/Y
+            mpt = Point(sptx + where.x, spty + where.y)
+
+            logging.debug("Pixel cmp @ main: %d,%d, subimage: %d,%d", mpt.x, mpt.y, sptx, spty)
```
TheJJ: maybe you could think about using our internal logger, but then the script would depend on libopenage and can't work standalone, so leaving it like this is probably best.

openage/convert/interface/visgrep.py
```diff
@@ -0,0 +1,389 @@
...
+                        all images at every pixel instead of just at detection points. Also, in
+                        this mode the detection image is ignored, there will be no matches
+                        for tile -1""", action="store_true")
+    PARSER.add_argument("-t", dest="tolerance", default=0, metavar="TOLERANCE", type=int,
+                        help="Set tolerance for 'fuzzy' matches, higher numbers are more tolerant")
+    PARSER.add_argument("-b", dest="show_badness",
+                        help="""Display 'badness' value, higher numbers mean match is less accurate,
+                        a badness value of 0 means the match is pixel-perfect""",
+                        action="store_true")
+    PARSER.add_argument("-d", metavar="DEBUGLEVEL", type=int, help="Print debug messages")
```
TheJJ: Here you can use type=count, then `-dddd` will be level 4. I'd recommend using `-v` though except if you don't want to break the c-visgrep compat.

openage/convert/interface/interfacerename.py
ChipmunkV: Exception is necessary to catch the case when the file shouldn't be renamed (because it's not a HUD at all).

openage/convert/interface/visgrep.py
ChipmunkV: It's about commented out DMSG? Wanted to keep to original. :)

Ok, don't need that.

openage/convert/interface/visgrep.py
ChipmunkV: type=count breaks compatibility too.

TheJJ:
Thanks, we can of course further improve it, but it's very good for now i'd say :+1:

ChipmunkV:
First time using python as a programming language.

TheJJ:
:grinning: I hope it wasn't that bad

PR #509, opened 2016-02-21
[WIP] Changes to cmake to generate Visual Studio 2015 solution and project files

synth2014:
Some changes were needed in the way Python was detected. 

py_get_config_var(LIBDIR PYTHON_LIBRARY_DIR) did not get the python lib directory correctly on Windows, so the lib directory is created after reading the python config var ''BINDIR". 

Changes for windows paths using '\' which are converted into posix paths before any comparison.


TheJJ:
Before we can merge this, you need to rebase and meld those commits. As there's no logical separation of the changes you did, it's enough to have one commit.

Content-wise, I already told you some findings in IRC, but if you like i can add comments directly to the source code lines here.

TheJJ:
This is outdated for quite a while and conflicts need to be fixed.

Issue #511, opened 2016-02-21
'make doc' fails, if a parent directory has a white space


castilma:
'make doc' didn't create the documentation, only a sceleton index.html. This was repeatedly written on the shell:
`Warning: tag INPUT: input source '/home/castilma/somedir/otherdir' does not exist`
`Warning: tag INPUT: input source 'withwhitespace/openage/libopenage/' does not exist"`

Where 'otherdir withwhitespace' has a space in it's name.

As a workaround I renamed the 'otherdir withwhitespace' removing the space and rerun ./configure  and 'make doc'. 

I have no experience with cmake, but I imagine there's just somewhere one line like 
.... $OPENAGE_DIR ... 
which should be
... "$OPENAGE_DIR" ...
.

TheJJ:
Ah, it is it because you cloned openage to some folder where a parent folder has whitespace in it, right?
The place where doxygen gets multiple of the input dirs is https://github.com/SFTtech/openage/blob/7f68dc4e0c526c9dbe17bf3d39667fdb3920dc83/buildsystem/templates/Doxyfile.in#L657.

CMake creates that variable here: https://github.com/SFTtech/openage/blob/7ad1e94e8b409fb8f621b6610c82690cedd76892/buildsystem/doxygen.cmake#L37

So maybe it's enough to write `'${CMAKE_CURRENT_SOURCE_DIR}/${folder}'` there. Can you try that please?

(Btw I wanted to answer you in IRC but both times you went offline already :)

castilma:
That did not work. I guess >'< does not work inside " "?
The warning output included the >'<.
```
Warning: tag INPUT: input source `'/home/mcd/somedir/otherdir' does not exist
Warning: tag INPUT: input source `withwhitespace/openage/libopenage/'' does not exist
```

I'm not sure if I understood correctly, so here is the git diff.
```
diff --git a/buildsystem/doxygen.cmake b/buildsystem/doxygen.cmake
index 6f42bdf..39a4393 100644
--- a/buildsystem/doxygen.cmake
+++ b/buildsystem/doxygen.cmake
@@ -34,7 +34,7 @@ function(doxygen_configure)
 
                # create doc folder name list
                foreach(folder ${ARGN})
-                       set(DOXYGEN_SCAN_FOLDERS "${DOXYGEN_SCAN_FOLDERS} ${CMAKE_CURRENT_SOURCE_DIR}/${folder}")
[...]

PR #512, opened 2016-02-23
fixes documentation issue #511

castilma:
If a parent directory of openage contains a space, 'make doc' (or better, doxygen)
splits the absolute path from the INPUT tag in bin/Doxygen in two directorys.
That way, the necessary files cannot be found and no documentation is created.

This commit adds quotation marks to the responsible cmake file.

	geändert:               buildsystem/doxygen.cmake


TheJJ:
Please add yourself to the `copying.md` author list :)

TheJJ:
Nice, thanks!

PR #516, opened 2016-02-27
Fixed the attack enemy sheep bug

### LINKED ISSUE #369: Archers can't kill sheep
No more words are needed.

Kvetoant:
The `command_flag::attack_res` flag was not being added, and therefore the `AttackAbility::can_invoke` method was returning false. Please review my changes.

Fixes #369.

### DIFF
--- a/copying.md
+++ b/copying.md
@@ -74,6 +74,7 @@ _the openage authors_ are:
 | James McMurray              | jamesmcm                    | jamesmcm03@gmail.com                  |
 | Łukasz Raszka               | lukky513                    | lukky513@gmail.com                    |
 | Martin Castillo             | castilma                    | castilma@uni-bremen.de                |
+| Antonín Květoň              | kvetoant                    | kvetoant@fel.cvut.cz                  |
 
 If you're a first-time commiter, add yourself to the above list. This is not
 just for legal reasons, but also to keep an overview of all those nicknames.
--- a/libopenage/unit/command.cpp
+++ b/libopenage/unit/command.cpp
@@ -30,9 +30,11 @@ Command::Command(const Player &p, Unit *unit, coord::phys3 position)
 	this->pos = position;
 
     //allow units to attack enemy sheep
-    if(unit->unit_type->unit_class == gamedata::unit_classes::SHEEP &&
-       unit->unit_type->owner.is_enemy(p))
-       this->add_flag(command_flag::attack_res);
+    if ((unit->unit_type->unit_class == gamedata::unit_classes::SHEEP)
+        && unit->unit_type->owner.is_enemy(p)) {
+
+        this->add_flag(command_flag::attack_res);
+    }
 }
 
 Command::Command(const Player &p, UnitType *t)
--- a/libopenage/unit/command.cpp
+++ b/libopenage/unit/command.cpp
@@ -28,6 +28,11 @@ Command::Command(const Player &p, Unit *unit, coord::phys3 position)
 	:
 	Command{p, unit, true, nullptr} {
 	this->pos = position;
+
+    //allow units to attack enemy sheep
+    if(unit->unit_type->unit_class == gamedata::unit_classes::SHEEP &&
+       unit->unit_type->owner.is_enemy(p))
+       this->add_flag(command_flag::attack_res);
 }
 
 Command::Command(const Player &p, UnitType *t)


TheJJ:
1. please add yourself to `copying.md`
2. please adhere to the [code and indentation style](https://github.com/SFTtech/openage/tree/master/doc/code_style)
3. I think we should try a more generalized approach and not only add sheep as a "special exception". I'm not sure how the `command_flag::attack_res` flag is set normally (I'd have do dig a bit first), but I suspect that we should place the sheep fix somewhere else, e.g. [when creating the data pack](https://github.com/SFTtech/openage/blob/master/openage/convert/fix_data.py) if necessary)

Jon0:
The `command_flag::attack_res` was made to prevent attacking being prioritised over gathering, allowing the gather action to placed on stack first and set correct graphics. Recent changes (villagers now change unit types fixing graphic issues, and global priority of abilities in `ability.h` which shows gathering is highest) might mean this flag can be removed, and the logic in `AttackAbility::can_invoke` could be refactored.

Gathering must be a higher priority, otherwise the target is killed and the villager returns to an idle state. When gather is used, it automatically stacks the attack on top, so when the target is killed, the villager begins gathering.

Kvetoant:
I have fixed 1.-3.  (hopefully 2. is correct), sorry for the mess.

As for 4., I had the same feeling as you do. When I was skimming through the code, I found that the only other place in the code this flag is being set is in `GatherAction::update_in_range`, which means this flag is normally used with respect to villagers - attacking trees, sheep and game. This was interfering with the logic in `AttackAbility::can_invoke`. I cannot think of a better fix at this time as I am new to the project however. I will continue in searching for a better solution.

Jon0:
Looking at `GatherAction::update_in_range` since `set_ability(ability_type::attack)` restricts the actions to attacking anyway, the `add_flag(command_flag::attack_res)` should be safe to remove. In this case delete that `attack_res` flag completely as it has no other use. Refactoring `AttackAbility::can_invoke` would be the best fix.

Simply removing `(cmd.has_flag(command_flag::attack_res) == target_is_resource)` in `AttackAbility::can_invoke` could work.

Kvetoant:
Removing `(cmd.has_flag(command_flag::attack_res)` == target_is_resource)` has the unfortunate consequence that military units are able to attack friendly sheep (which they were not able to in the original game, IIRC). I keep converging towards adding an exception for sheeps (because the sheep really is an exception - a resource which can be owned by a player) - maybe this could be synchronized with farms somehow?

Jon0:
You're right, you should check the attacker has attribute `attr_type::gatherer`

TheJJ:
Then, aren't other resources like turkeys an exception as well?

(btw, please use tab indentation, otherwise it looks good now code-style wise :)

zuntrax:
Also see #369.
I'm still a bit confused about the whole unit type thing, is it possible that turkeys also have unit type `SHEEP`? Also, there seems to be a resource trait `DOMINANT_ANIMAL_DISCOVERY` that is related to sheep and turkeys.

Kvetoant:
My IDE was messing the indentation up, should be ok from the next commit on. 

Upon further examination, `command_flag::attack_res` might be needed in some form after all. Everything depends on the final implementation of Gaia (whether it will be hostile to the players or not). Because without this, military units could attack trees, etc if Gaia was to be hostile. On the other hand, military units could not attack wolves if Gaia was to be friendly.

Therefore, further classification of units might be needed (is a military attackable resource), a rundown:

Tree - is enemy - is a resource - is not a military attackable resource (but what about the trebuchet?)
Boar - is enemy - is a resource - is a military attackable resource
Sheep - is enemy/friendly - is a resource - is a military attackable resource
Wolf - is enemy - is not a resource

A proposal of the new part of the logic would then look something like this in pseudocode:

IF(is enemy AND is not resource) OR (is resource AND selected unit is gatherer) OR (is enemy AND is resource AND is military attackable resource) THEN can attack

This is under assumption gaia stays hostile.

Unsure how to handle the trebuchet/onager - tree thing. Perhaps yet another classifier will be needed.

zuntrax:
I'm pretty sure the onager-tree-interaction was by setting an area to attack. The trees get killed by splash damage. There might be another special case to consider, the most basic one of the onager-style siege weapons was not able to kill trees.
However, the whole gaia thing will probably a lot easier once #239 is finished.

Kvetoant:
Can't remember the original game, but the trees are directly targetable by onagers and trebuchets in the HD version, at least (without setting an area to attack).

Jon0:
The extracted game data contains an `attack_mode` value for each unit type. Currently openage ignores this, but it might indicate which types are a military attackable resource.

`producer.cpp` is where the gamedata gets converted into openage unit types. A new attribute for units could be added.

TheJJ:
Timeout detected, but ready for a revive.

PR #519, opened 2016-03-08
ENSURE macro for the assertions

ChipmunkV:
Adding a macro that throws the Error in a form of `ENSURE(cond, message)`.


TheJJ:
It's finished now, right?

PR #522, opened 2016-03-11
convert: Propose some paths for source_dir, including wine. Fixes #347

### LINKED ISSUE #347: Use wine to determine original game path
The convert script could use wine to read the registry key where the original game is installed. This should work for both HD and AoC.

This is of course an optional step, we don't want to depend on wine.

When media conversion is required, the user could be prompted to "use wine?" or enter the path directly.

Currently favored idea:

A simple readline-based interface:
```
…Media needs to be converted…
Select an Age of Kings installation directory [or a setup folder]:
(0) Use previously used path "..."
(1) Insert manually
(2) $WINEPREFIX/drive_c/Program Files (x86)/Microsoft Games/Age of Empires II/
(3) $HOME/.wine/drive_c/Program Files (x86)/Microsoft Games/Age of Empires II/
[default: 0] > _
```
The list could be populated programmatically:
* If `$WINEPREFIX` is set, offer something like (1)
* If `wine` is in PATH, offer wine (I'd suggest offering it always, instead)
* Once `reg query` is implemented, use the value it returns, if any
* Once we can import from installation media, add it in the query & offer common CD drive locations and loop-mount locations (or maybe XDG has a means for getting a CD mount path?)
* ...

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25765638-use-wine-to-determine-original-game-path?utm_campaign=plugin&utm_content=tracker%2F371436&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F371436&utm_medium=issues&utm_source=github).
</bountysource-plugin>

azrdev:


openage/convert/main.py
```diff
@@ -159,25 +171,45 @@ def acquire_conversion_source_dir():
 
     if 'AGE2DIR' in os.environ:
         sourcedir = os.environ['AGE2DIR']
-        print("using AGE2DIR: " + sourcedir)
+        print("found environment variable AGE2DIR")
```
TheJJ: Dound? I'd rather keep the "using".

openage/convert/main.py
```diff
@@ -159,25 +171,45 @@ def acquire_conversion_source_dir():
...
+                sourcedir = proposals[int(user_selection)]
+            else:
+                sourcedir = user_selection
         except KeyboardInterrupt:
-            print("")
+            print("Interrupted, aborting")
             exit(0)
         except EOFError:
-            print("")
+            print("EOF, aborting")
```
TheJJ: The keyboard interrupt is triggered on ctrl-c, and the eof on ctrl-d, so not sure if we wanna print a message then as the user is actively triggering the abort.

openage/convert/main.py
```diff
@@ -159,25 +171,45 @@ def acquire_conversion_source_dir():
...
 
 
+def _get_source_dir_proposals():
+    """Yield a list of directory names where an installation might be found"""
+    if "WINEPREFIX" in os.environ:
+        yield "$WINEPREFIX/" + _STANDARD_PATH_IN_32BIT_WINEPREFIX
+        yield "$WINEPREFIX/" + _STANDARD_PATH_IN_64BIT_WINEPREFIX
+    yield "~/.wine/" + _STANDARD_PATH_IN_32BIT_WINEPREFIX
+    yield "~/.wine/" + _STANDARD_PATH_IN_64BIT_WINEPREFIX
+    # TODO: wine reg query "HKLM\Software\Microsoft\Microsoft Games\Age of Empires\2.0"
```
TheJJ: cool :) can we even detect the bittness of a wineprefix and also check for existence? then we can remove the non-matching entries.

azrdev: line 178 is checking for existence. I found it easier to dump all possible paths here, and filter over there.
Standard way to differentiate WINEARCH is checking if `drive_c/Program Files (x86)` exists, AFAIK.

openage/convert/main.py
azrdev: It seemed useful to me when trying it out. And it tells more than an empty print. If you don't like the message, I suggest deleting the print altogether

openage/convert/main.py
azrdev: "Using" is printed below on line 198, for every path (after conversion from relative to absolute paths & expansion of variables)

openage/convert/main.py
TheJJ: The prints are there to emit a newline after the `^C` so the new message doesn't look like `^Clolmessage` but instead
```
^C
lolmessage
```

but it's up to you, if you think the eof info is good, then we keep it.

openage/convert/main.py
TheJJ: my bad, very well then :)

openage/convert/main.py
```diff
@@ -16,12 +16,18 @@
 from ..util.fslike.directory import CaseIgnoringDirectory
 from ..util.strings import format_progress
 
+_STANDARD_PATH_IN_32BIT_WINEPREFIX =\
+    "drive_c/Program Files/Microsoft Games/Age of Empires II/"
+_STANDARD_PATH_IN_64BIT_WINEPREFIX =\
+    "drive_c/Program Files/Microsoft Games (x86)/Age of Empires II/"
```
TheJJ: I don't think there wlll ever be 64-bit microsoft games :grin: 
It should be `Program Files (x86)` instead.

azrdev: If it's a 32bit wineprefix, there will only be `Program Files/`, and in it age.
If it's a 64bit wineprefix, there will be both directories, and age be in the `(x86)` variant.

TheJJ: nah:
`drive_c/Program Files/Microsoft Games (x86)/Age of Empires II/`
vs
`drive_c/Program Files (x86)/Microsoft Games/Age of Empires II/`

TheJJ:
thx :)

PR #523, opened 2016-03-12
Documentation for damage calculations

TheJJ:
How does the damage calculation work?
How can we prevent more people from collaborating?

All this, in one magnificent pull request!
Incredible !!!


franciscod:
great! it might be a good opportunity to go on a bit more explaining that you shouldn't do push forces or rebases in the wild, or everyone will hate you

TheJJ:
done, added a sentence about it :)

Issue #486, opened 2016-01-17
Create cvar configuration subsystem


TheJJ:
In the style of [Half-Life 2 cvars](https://developer.valvesoftware.com/wiki/List_of_HL2_Cvars), we should implement a configuration system that uses simple key-value assigning.

Once a key is defined, the values has to be of a type: `str`, `int`, `float`, ...
When parsing a config file, the type conversion is attempted and errors can be detected.

This configuration is then used by the engine to control all sorts of things.
Ideally, the definition of known cvars should be in Python or Cython, with an interface to the C++ code (just follow the [python interface docs](https://github.com/SFTtech/openage/blob/master/doc/code/pyinterface.md) for that).

As this code is not performance-critical, it should be in Python, which simplifies modding later.

A settings that could be implemented as a proof-of-concept is e.g. the initial window resolution.
Then, it lays the foundation for #271.

franciscod:
do cvars have types or are their values always strings?

TheJJ:
Good question. I'd say they have types. For now `int` and `string` should be enough. Having types, we can directly fail setting them if e.g. the number is mismatched.

franciscod:
Also `float` is a good one for in-game constants like various speed coefficients (movement speed, arrow shooting speed, building or tech development speed)

mic-e:
Also, for customizable key bindings, there should probably be a `key` type.

elnabo:
After a discussion with @mic-e in the chat yesterday.

A proposition for the cvar system is to use a python dictionary mapping each var name to its getter/setter. This way the variable is only stored once, where it is used.

Eg : 

Registering a variable to the cvar system would be something like : 
`cvar[varname] = (get,set); set(varname, configvalue[varname])`

Updating the value 
`cvar[varname].set(varname, value)`


This approach could easily allow different treatment of different type of variable (float, double, key, string, dimension).

TheJJ:
@elnabo this is also in place now, right?
Wanna add more things to configure :stuck_out_tongue:?

elnabo:
Yeah it should be working. And I don't see what could be added in that file.

PR #526, opened 2016-03-13
Dynamic key bindings

elnabo:
Started working on the usage of dynamic key bindings #271.

At the moment, it is only a preparsed (no trailing/leading space) String to Event parser and a dynamic modification of the event/action mapping.

### DIFF
--- a/libopenage/engine.cpp
+++ b/libopenage/engine.cpp
@@ -49,6 +49,7 @@ void Engine::create(util::Dir *data_dir, const char *windowtitle) {
 	if (Engine::instance == nullptr) {
 		// reset the pointer to the new engine
 		Engine::instance = new Engine(data_dir, windowtitle);
+		Engine::instance->get_input_manager().initialize(); // until non hardcoded keybind
 	} else {
 		throw Error{MSG(err) << "You tried to create another singleton engine instance!!111"};
 	}
@@ -183,19 +184,20 @@ Engine::Engine(util::Dir *data_dir, const char *windowtitle)
 
 	// initialize engine related global keybinds
 	auto &global_input_context = this->get_input_manager().get_global_context();
-	global_input_context.bind(input::actions::STOP_GAME, [this](const input::action_arg_t &) {
+	input::ActionManager &action = this->get_action_manager();
+	global_input_context.bind(action.get("STOP_GAME"), [this](const input::action_arg_t &) {
 		this->stop();
 	});
-	global_input_context.bind(input::actions::TOGGLE_HUD, [this](const input::action_arg_t &) {
+	global_input_context.bind(action.get("TOGGLE_HUD"), [this](const input::action_arg_t &) {
 		this->drawing_huds.value = !this->drawing_huds.value;
 	});
-	global_input_context.bind(input::actions::SCREENSHOT, [this](const input::action_arg_t &) {
+	global_input_context.bind(action.get("SCREENSHOT"), [this](const input::action_arg_t &) {
 		this->get_screenshot_manager().save_screenshot();
 	});
-	global_input_context.bind(input::actions::TOGGLE_DEBUG_OVERLAY, [this](const input::action_arg_t &) {
+	global_input_context.bind(action.get("TOGGLE_DEBUG_OVERLAY"), [this](const input::action_arg_t &) {
 		this->drawing_debug_overlay.value = !this->drawing_debug_overlay.value;
 	});
-	global_input_context.bind(input::actions::TOGGLE_PROFILER, [this](const input::action_arg_t &) {
+	global_input_context.bind(action.get("TOGGLE_PROFILER"), [this](const input::action_arg_t &) {
 		if (this->external_profiler.currently_profiling) {
 			this->external_profiler.stop();
 			this->external_profiler.show_results();
@@ -220,14 +222,14 @@ Engine::Engine(util::Dir *data_dir, const char *windowtitle)
 
 	};
 
-	bind_player_switch(input::actions::SWITCH_TO_PLAYER_1, 1);
-	bind_player_switch(input::actions::SWITCH_TO_PLAYER_2, 2);
-	bind_player_switch(input::actions::SWITCH_TO_PLAYER_3, 3);
-	bind_player_switch(input::actions::SWITCH_TO_PLAYER_4, 4);
-	bind_player_switch(input::actions::SWITCH_TO_PLAYER_5, 5);
-	bind_player_switch(input::actions::SWITCH_TO_PLAYER_6, 6);
-	bind_player_switch(input::actions::SWITCH_TO_PLAYER_7, 7);
-	bind_player_switch(input::actions::SWITCH_TO_PLAYER_8, 8);
+	bind_player_switch(action.get("SWITCH_TO_PLAYER_1"), 1);
+	bind_player_switch(action.get("SWITCH_TO_PLAYER_2"), 2);
+	bind_player_switch(action.get("SWITCH_TO_PLAYER_3"), 3);
+	bind_player_switch(action.get("SWITCH_TO_PLAYER_4"), 4);
+	bind_player_switch(action.get("SWITCH_TO_PLAYER_5"), 5);
+	bind_player_switch(action.get("SWITCH_TO_PLAYER_6"), 6);
+	bind_player_switch(action.get("SWITCH_TO_PLAYER_7"), 7);
+	bind_player_switch(action.get("SWITCH_TO_PLAYER_8"), 8);
 
 	this->text_renderer = std::make_unique<renderer::TextRenderer>();
 }
@@ -500,6 +502,10 @@ audio::AudioManager &Engine::get_audio_manager() {
 
 ScreenshotManager &Engine::get_screenshot_manager() {
 	return this->screenshot_manager;
+}
+
+input::ActionManager &Engine::get_action_manager() {
+	return this->action_manager;
 }
 
 input::InputManager &Engine::get_input_manager() {
--- a/libopenage/game_control.cpp
+++ b/libopenage/game_control.cpp
@@ -11,8 +11,9 @@ CreateMode::CreateMode()
 	selected{0},
 	setting_value{false} {
 
+	auto &action = Engine::get().get_action_manager();
 	// action bindings
-	this->bind(input::actions::START_GAME, [this](const input::action_arg_t &) {
+	this->bind(action.get("START_GAME"), [this](const input::action_arg_t &) {
 		Engine &engine = Engine::get();
 		options::OptionNode *node = engine.get_child("Generator");
 		if (!node) {
@@ -69,12 +70,12 @@ CreateMode::CreateMode()
 		}
 
 	});
-	this->bind(input::actions::UP_ARROW, [this](const input::action_arg_t &) {
+	this->bind(action.get("UP_ARROW"), [this](const input::action_arg_t &) {
 		if (!this->setting_value) {
 			this->selected -= 1;
 		}
 	});
-	this->bind(input::actions::DOWN_ARROW, [this](const input::action_arg_t &) {
+	this->bind(action.get("DOWN_ARROW"), [this](const input::action_arg_t &) {
 		if (!this->setting_value) {
 			this->selected += 1;
 		}
@@ -154,7 +155,8 @@ ActionMode::ActionMode()
 	type_focus{nullptr},
 	rng{0} {
 
-	this->bind(input::actions::TRAIN_OBJECT, [this](const input::action_arg_t &) {
+	auto &action = Engine::get().get_action_manager();
+	this->bind(action.get("TRAIN_OBJECT"), [this](const input::action_arg_t &) {
 
 		// attempt to train editor selected object
 		Engine &engine = Engine::get();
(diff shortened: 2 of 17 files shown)


elnabo:
I am working on a Py interface to allow the modification of the binding through python (for #486) and was wondering if there was already some python code called after the engine initialization, for a quick & dirty test of the interface.

TheJJ:
No, there's no callback to python yet after the engine initializes. Be sure to read the `doc/code/pyinterface.md` to learn about how you can achieve this. I can try to assist you in IRC as well.

elnabo:
Alright should be working. I think I'm gonna try to make a cvar system in an other branch to initialize the binding (see my comments on #486).

libopenage/input/action.h
```diff
@@ -17,57 +18,67 @@ struct action_t {
...
+	{"RIGHT_ARROW", action_t{41, "RIGHT_ARROW"}},
+	{"SELECT", action_t{42, "SELECT"}},
+	{"DESELECT", action_t{43, "DESELECT"}},
+	{"FORWARD", action_t{44, "FORWARD"}},
+	{"BACK", action_t{45, "BACK"}},
+	{"PAINT_TERRAIN", action_t{46, "PAINT_TERRAIN"}},
+	{"BUILDING_5", action_t{47, "BUILDING_5"}},
+	{"BUILDING_6", action_t{48, "BUILDING_6"}},
+	{"BUILDING_7", action_t{49, "BUILDING_7"}},
+	{"BUILDING_8", action_t{50, "BUILDING_8"}}
```
TheJJ: I know that would mean more work, but having redundancy is bad in almost all cases.
Can't we create a list of "register" those `action_t` objects and register them? This registration could then also happen from around the engine, whereever a new keybinding is created for the appropriate context.

libopenage/input/input_manager.cpp
```diff
@@ -12,57 +12,139 @@ namespace input {
...
+	      { sdl_mouse(1), actions::ACTION_MAP.at("PAINT_TERRAIN") },
+	      { sdl_mouse(1), actions::ACTION_MAP.at("SELECT") },
+	      { sdl_mouse(3), actions::ACTION_MAP.at("DESELECT") },
+	      { sdl_wheel(1), actions::ACTION_MAP.at("FORWARD") },
+	      { sdl_wheel(-1), actions::ACTION_MAP.at("BACK") }},
 	relative_mode{false} {
 }
 
+std::string InputManager::get_key(const std::string action_str) {
+	for (auto it=keys.begin(); it!=keys.end(); ++it) {
```
TheJJ: you can just use `for (auto &it : this->keys) { it.first.stuff; it.second.bla(); }`

libopenage/input/input_manager.cpp
```diff
@@ -12,57 +12,139 @@ namespace input {
...
+			return false;
+		keys.emplace(ev, it->second);
+		return true;
+	}
+	catch (int error) {
+		return false;
+	}
+}
+
+Event InputManager::textToEvent(const char* key_char)
```
TheJJ: For our code, we don't use camel case method names :)
And our ptr style is `char *key_char`. If you like, please have a look at `doc/code_style`.

libopenage/input/input_manager.cpp
```diff
@@ -12,57 +12,139 @@ namespace input {
...
+	do {
+		next = key_str.find(" ", prev);
+		key_mod |= parseMod(key_str.substr(prev,next-prev));
+		prev = next+1;
+	} while (next != last);
+
+	return sdl_key(key_code, static_cast<SDL_Keymod>(key_mod));
+}
+
+int InputManager::parseMod(std::string mod) {
```
TheJJ: As you don't modify the string, use `const std::string &mod` to allow the compiler to optimize. The `parse_mod` function can be marked `const` as it doesn't modify `this`.

libopenage/input/input_manager.h
```diff
@@ -150,8 +180,25 @@ class InputManager : public openage::InputHandler {
 	 */
 	coord::window mouse_position;
 	coord::window_delta mouse_motion;
-
 	friend InputContext;
+
+	std::map<std::string, int> stringToMod = {
```
TheJJ: The `std::unordered_map` is hashed and way faster than the map which just is a search tree. Usage is the same, except you need `std::hash<key_type>` for the hashmap, the map only needs `operator <`

libopenage/input/input_manager.cpp
```diff
@@ -12,57 +12,139 @@ namespace input {
...
+			throw 0;
+		}
+		return sdl_key(key_code);
+	}
+
+	SDL_Keycode key_code = SDL_GetKeyFromName(key_str.substr(last+1).c_str());
+	if (key_code == SDLK_UNKNOWN) {
+			throw 0;
+	}
+	int key_mod = 0;
```
TheJJ: How can this happen? Is this the splitup at `" "` to allow more than one key? If yes, it can be implemented shorter I think, without the copy-paste of `GetKeyFromName` and the check.

elnabo:
@TheJJ 

> Is this the splitup at " " to allow more than one key?

This is to allow key modifier eg "Ctrl Alt i"

> Can't we create a list of those action_t objects and "register" them?

Do you mean something like `for each action do ACTION_MAP.emplace(action.info, action)`. And for the registration around the engine, do you mean something like 

`register(key) { `
`    if (! key in ACTION_MAP) { ACTION_MAP.emplace(action_t(size+1, key))}`
`    return ACTION_MAP.at(key);`
`}`

TheJJ:
I see, but still i think we can place this into a while loop and do the parsing for each part separated by `" "`.

The registration could happen that way, exactly. Although we should do it for each context separately so the same key (and combo) does different things depending on the active context (e.g. chat window or ingame hotkey). Additionally, but that's for the gui, depending on the focus.

A getter function should avoid to modify the state. Did you do this to avoid the exception throwing?

elnabo:
Oh now I see what you meant for the key parsing.

As for the getter function. It is moslty the name that is incorrect, I meant this function for registering the key only once. And yes action should be defined for a (multiple ?) context.

elnabo:
Ok so I reduced the redundancy to in the action declaration, more separation to the storage of the action and the rest of the code and it is now possible to register new action.

libopenage/input/action.h
```diff
@@ -18,65 +18,73 @@ struct action_t {
...
-	{"FORWARD", action_t{44, "FORWARD"}},
-	{"BACK", action_t{45, "BACK"}},
-	{"PAINT_TERRAIN", action_t{46, "PAINT_TERRAIN"}},
-	{"BUILDING_5", action_t{47, "BUILDING_5"}},
-	{"BUILDING_6", action_t{48, "BUILDING_6"}},
-	{"BUILDING_7", action_t{49, "BUILDING_7"}},
-	{"BUILDING_8", action_t{50, "BUILDING_8"}}
+/** Return true if the action has been succesfully registered */
+bool register_action(const std::string type);
+action_t get_action(const std::string type);
```
TheJJ: this is already in the `action` namespace, so you can remove further redundancy and just name them `register` and `get`, as the access is `actions::get()` anyway then.

libopenage/input/action.h
```diff
@@ -18,65 +18,73 @@ struct action_t {
...
+void initialize_actions();
+
+static std::vector<std::string> default_action = {
```
ChipmunkV: Is default_action const?

libopenage/engine.cpp
```diff
@@ -47,6 +47,8 @@ Engine *Engine::instance = nullptr;
 void Engine::create(util::Dir *data_dir, const char *windowtitle) {
 	// only create the singleton instance if it was not created before..
 	if (Engine::instance == nullptr) {
+		// initializa the default actions
+		input::actions::initialize_actions();
```
ChipmunkV: If we do input::actions::initialize_actions() per engine, maybe include it in engine?

If we do global, maybe initialize when static globals are initialized?

elnabo: The current idea is to make an ActionManager class included in the engine. But it has to be initialized before anything that requires the actions.

libopenage/engine.cpp
```diff
@@ -47,10 +47,9 @@ Engine *Engine::instance = nullptr;
 void Engine::create(util::Dir *data_dir, const char *windowtitle) {
 	// only create the singleton instance if it was not created before..
 	if (Engine::instance == nullptr) {
-		// initialize the default actions
-		input::actions::initialize_actions();
 		// reset the pointer to the new engine
 		Engine::instance = new Engine(data_dir, windowtitle);
+		Engine::instance->get_input_manager().initialize(); // until non hardcoded keybind
```
elnabo: @TheJJ @ChipmunkV  The segfault problem was because the InputManager initialization was made before the Engine instance was finalized. Thanks for the help.

elnabo:
I think I am done. Do you see more features to add for the keybinding management ?

libopenage/input/action.cpp
```diff
@@ -3,18 +3,44 @@
...
+		if (it.second == action) {
+			return it.first;
+		}
+	}
+	return "UNDEFINED";
 }
 
+ActionManager::ActionManager() {
+	for (auto &type : default_action) {
+		ActionManager::create(type);
```
TheJJ: why `ActionManager::` and not `this->`?

libopenage/input/action.cpp
```diff
@@ -3,18 +3,44 @@
...
+	}
+	return false;
+}
+
+action_t ActionManager::get(const std::string type) {
+	auto it = actions.find(type);
+	if ( it != actions.end()) {
+		return it->second;
+	}
+	return actions.at("UNDEFINED");
```
TheJJ: can you please add `this->` for all the member accesses? (there's a space too much in front of `it` btw)

libopenage/input/input_manager.cpp
```diff
@@ -11,58 +11,149 @@ namespace openage {
...
+	set_key("Left", "LEFT_ARROW");
+	set_key("Right", "RIGHT_ARROW");
+
+	ActionManager &action_manager = Engine::get().get_action_manager();
+	keys.emplace(utf8("`"), action_manager.get("TOGGLE_CONSOLE"));
+	keys.emplace(sdl_mouse(1), action_manager.get("PAINT_TERRAIN"));
+	keys.emplace(sdl_mouse(1), action_manager.get("SELECT"));
+	keys.emplace(sdl_mouse(3), action_manager.get("DESELECT"));
+	keys.emplace(sdl_wheel(1), action_manager.get("FORWARD"));
+	keys.emplace(sdl_wheel(-1), action_manager.get("BACK"));
```
TheJJ: `this->` please :)

libopenage/input/input_manager.cpp
```diff
@@ -11,58 +11,149 @@ namespace openage {
...
+}
+
+std::string InputManager::get_key(const std::string action_str) {
```
TheJJ: you can use a `const std::string &action_str` here, no need to copy.

libopenage/input/input_manager.cpp
```diff
@@ -11,58 +11,149 @@ namespace openage {
...
+			return false;
+		}
+
+		for(auto &it : keys) {
+			if (it.second == action) {
+				keys.erase(it.first);
+			}
+		}
+		Event ev = text_to_event(key_char);
+		keys.emplace(ev, action);
```
TheJJ: `this->keys` here and above pls

libopenage/input/input_manager.cpp
```diff
@@ -11,58 +11,149 @@ namespace openage {
...
+	int key_mod = 0;
+	SDL_Keycode key_code = SDLK_UNKNOWN;
+
+	std::size_t start = 0;
+	std::size_t end;
+	do {
+
+		end = key_str.find(" ", start+1);
+		std::string substring = key_str.substr(start,end-start);
+		int mod = parse_mod(substring);
```
TheJJ: `this->parse_mod` pls

elnabo:
Ok so I've corrected the lack of this-> and also added support for text to mouse button/wheel support. So now every hardcoded shortcut can be hardcoded with set_bind.

libopenage/input/input_manager.cpp
```diff
@@ -16,84 +17,72 @@ InputManager::InputManager()
...
-			auto end = it.first.mod.end();
-			if (it.first.mod.find(modifier::ALT) != end) {
-				key_str = "Alt "+key_str;
+			switch (it.first.cc.eclass) {
+				case event_class::MOUSE_BUTTON :
+					return this->mouse_bind_to_string(it.first);
+				case event_class::MOUSE_WHEEL :
+					return this->wheel_bind_to_string(it.first);
+				default:
+					return this->key_bind_to_string(it.first);
```
TheJJ: Please dedent this region by one tab. The case labels are at the same indent level as the switch, but the code is then indented one tab.

libopenage/input/input_manager.cpp
```diff
@@ -117,39 +106,110 @@ bool InputManager::set_key(const char *key_char, const std::string action_str) {
...
+std::string InputManager::wheel_bind_to_string(const Event &ev) {
+	std::string base = "WHEEL ";
+	switch (ev.cc.code) {
+		case 1:
+			return base + "UP";
+		case -1:
+			return base + "DOWN";
+		default:
+			return "";
+	}
```
TheJJ: here as well :)

TheJJ:
It looks very good now. Can you please sqash the commits and perform a rebase to elimintate the merge commits? I'll test it and then we can merge.

elnabo:
Are you ok with that kind of hardcoding removal (using action for everything but mouse motion).

libopenage/game_control.cpp
```diff
@@ -220,6 +223,50 @@ ActionMode::ActionMode()
...
+			}
+		});
+	};
+
+	bind_build(action.get("BUILD"), false);
+	bind_build(action.get("KEEP_BUILDING"), true);
+
+	auto bind_select = [this](input::action_t action, const bool increase) {
+		this->bind(action, [this, increase](const input::action_arg_t &arg) {
+			log::log(MSG(dbg) << "select");
```
TheJJ: this debug message will probably spam a bit :)

TheJJ:
We are required to read the settings from a file as strings, but internally it could still be possible to handle them as enums for speed reasons. Then we'd need a mapping of enum<->string to display and set them via file later. If you are motivated, adding this performance optimization would be great.

Otherwise, it looks good and improving the hardcoding by less or better hardcoding is the way to go :)

elnabo:
I didn't find much more removable hardcoding stuff, so I think I've nothing more to add.

As for the optimization, I don't see the point, so I let this for someone else.

PR #527, opened 2016-03-13
GUI middle part of the HUD

ChipmunkV:
Rendering some paper.


ChipmunkV:
~~Solved~~ Hacked around the transparency issue. So, the paper class has enough flexibility for the edges and scales properly.

assets/textures/paper.docx
```diff
@@ -0,0 +1,8 @@
+# struct subtexture
+# one sprite, as part of a texture atlas.
+# 
+# this struct stores information about positions and sizes
+# of sprites included in the 'big texture'.
```
TheJJ: dunno if this comment is helpful :)

assets/textures/torn_paper_edge.docx
```diff
@@ -0,0 +1,8 @@
+# struct subtexture
+# one sprite, as part of a texture atlas.
+# 
+# this struct stores information about positions and sizes
+# of sprites included in the 'big texture'.
```
TheJJ: same here

TheJJ:
We also need the graphicaleffects package in the dependency matrix (in that `core, quick, quickcontrols ...` listing).

I couldn't try it yet, but why do we need the paper texture? Isn't something suitable in the original data?

Nevertheless, we should soon merge your awesome gui stuff, the diff is getting bigger and bigger :)

assets/textures/paper.docx
ChipmunkV: Copy-pasted from other generated docx. Probably, only the 'x,y,w,h,cx,cy' comment line is useful.

ChipmunkV:
In the original data there is a piece of paper of a fixed size. Not tileable. The HD version uses solid color for paper. Maybe there was a suitable paper texture from some other thing, but I looked only at the HUD ones.

TheJJ:
rebase pls :)

ChipmunkV:
Done, you might want to review the recent changes to the convert scripts here.

openage/convert/driver.py
```diff
@@ -222,6 +223,39 @@ def convert_metadata(args):
...
+        filename = b'/'.join(filepath.parts).decode()
+        if filename.endswith(suffix):
+            try:
+                for line in filepath.open():
+                    match = re.match(r"\w+_files\s+(\w+)\s+\w+\s+(\w+)", line)
+                    if match:
+                        groups = match.group(2, 1)
+                        names_map[groups[0]] = groups[1].lower()
+            except UnicodeDecodeError:
+                continue
```
TheJJ: why ignore the errors? i think we can avoid them alltogether if we `open(..)` them in `ISO-8859-1` or whatever microsoft used for the encoding.

ChipmunkV: I don't think that these are all text files.

TheJJ:
Looks good so far.
I'm a bit confused by the shaders in qml, which will pin us to OpenGL probably forever, but so be it.

It seems that the file renames require a reconversion for interface graphics, right? You need to add that to `changelog.py`.

ChipmunkV:
Updated the `changelog.py`.

openage/convert/driver.py
```diff
@@ -222,18 +223,55 @@ def convert_metadata(args):
...
+        ignored.update([('graphics', '.slp'), ('terrain', '.slp')])
+    if args.flag("no_interface"):
+        ignored.add(('interface', '.slp'))
 
     files_to_convert = []
     for dirname in ['sounds', 'graphics', 'terrain', 'interface']:
         for filepath in args.srcdir[dirname].iterdir():
-            if filepath.suffix in ignored_suffixes:
+            if (((None, filepath.suffix) in ignored) or
+                    ((filepath.parts[0].decode(), filepath.suffix) in ignored)):
```
TheJJ: why do you decode here? isn't it a `str` already?

ChipmunkV: No idea. I've taken that from the `path.py`:
``` 
   def name(self):
        """ The name of the topmost component (str). """
        return self.parts[-1].decode()
```

ChipmunkV:
Forgot to comment: removed that texture file,

TheJJ:
Kay :) How did you replace it, i thought it was necessary?

openage/convert/driver.py
TheJJ: seems alright then as you access internal components.
but the `parts` thingy is wrong itself i think.
``` python
>>> from pathlib import Path
>>> a = Path("/usr/local/bin/rofl/stuff")
>>> a.parts[0]
'/'
>>> a.parts
('/', 'usr', 'local', 'bin', 'rofl', 'stuff')
```

libopenage/gui/qml/Paper.qml
```diff
@@ -0,0 +1,215 @@
...
+
+import QtQuick 2.4
+
+Item {
+	id: root
+
+	/*
+	 * A paper texture.
+	 */
+	property url paperTextureSource: "image://by-filename/converted/interface/scr5b.slp.png.0"
```
ChipmunkV: That's a texture.

ChipmunkV:
Made the paper edge mask black.

About the shaders: the opacity shader won't be needed in Qt 5.7, the other one is the precise kind of burn effect that I wanted.

openage/convert/interface/interfacecutter.py
```diff
@@ -24,34 +27,42 @@ def __call__(self, image):
...
+                                   MID_STRIP_PATTERN_CORNERS,
+                                   MID_STRIP_PATTERN_SEARCH_AREA_CORNERS)
 
             for coords in KNOWN_SUBTEX_CORNER_COORDS:
                 yield TextureImage(pil_image.crop(coords))
         else:
             yield image
 
-    def __cut_top_strip(self, pil_image):
+    def __cut_strip(self, pil_image, pattern_corners, search_area_corners):
```
TheJJ: hm i think we can remove the `__`, and if you really wanna keep it, a single `_` should be enough.

TheJJ:
Looks really cool!
openage_2016-06-01_15-19-51_00

openage/convert/interface/interfacecutter.py
ChipmunkV: Ok, and I'll fix the comment, so it's not specifically about the top strip.

TheJJ:
Awesome! Woooh!

PR #528, opened 2016-03-19
convert: Propose install. directory read from wine registry

### LINKED ISSUE #347: Use wine to determine original game path

azrdev:
Fixes #347.

### DIFF
--- a/openage/convert/main.py
+++ b/openage/convert/main.py
@@ -223,10 +223,10 @@ def _get_source_dir_proposals():
             for suffix in _REGISTRY_SUFFIX_AOK, _REGISTRY_SUFFIX_TC:
                 reg_key = _REGISTRY_KEY + suffix
                 if reg_key in reg_parser:
-                    if 'InstallationDirectory' in reg_parser[reg_key]:
-                        yield reg_parser[reg_key]['InstallationDirectory']
-                    if 'EXE Path' in reg_parser[reg_key]:
-                        yield reg_parser[reg_key]['EXE Path']
+                    if '"InstallationDirectory"' in reg_parser[reg_key]:
+                        yield reg_parser[reg_key]['"InstallationDirectory"']
+                    if '"EXE Path"' in reg_parser[reg_key]:
+                        yield reg_parser[reg_key]['"EXE Path"']
             os.remove(tmp_reg_file)
     except OSError as e:
         dbg("wine registry extraction failed: " + str(e))
--- a/openage/convert/main.py
+++ b/openage/convert/main.py
@@ -5,6 +5,7 @@ import os
 # importing readline enables the raw_input calls to have history etc.
 import readline  # pylint: disable=unused-import
 import subprocess
+from subprocess import Popen, PIPE
 from configparser import ConfigParser
 
 from .game_versions import GameVersion, get_game_versions
@@ -206,6 +207,19 @@ def acquire_conversion_source_dir():
 
     return CaseIgnoringDirectory(sourcedir).root
 
+def _wine_to_real_path(path):
+    """
+    Turn a Wine file path (C:\\xyz) into a local filesystem path (~/.wine/xyz)
+    """
+    winepath = Popen(('winepath', path), stdout=PIPE)
+    (real_path, err) = winepath.communicate()
+    if winepath.returncode != 0:
+        raise Exception('Could not convert Wine path into real path')
+    return real_path.strip().decode()
+
+def unescape_winreg(value):
+    """Remove quotes and escapes from a Wine registry value"""
+    return value.strip('"').replace(r'\\\\', '\\')
 
 def _get_source_dir_proposals():
     """Yield a list of directory names where an installation might be found"""
@@ -232,9 +246,11 @@ def _get_source_dir_proposals():
                 reg_key = _REGISTRY_KEY + suffix
                 if reg_key in reg_parser:
                     if '"InstallationDirectory"' in reg_parser[reg_key]:
-                        yield reg_parser[reg_key]['"InstallationDirectory"']
+                        yield _wine_to_real_path(unescape_winreg(
+                            reg_parser[reg_key]['"InstallationDirectory"']))
                     if '"EXE Path"' in reg_parser[reg_key]:
-                        yield reg_parser[reg_key]['"EXE Path"']
+                        yield _wine_to_real_path(unescape_winreg(
+                            reg_parser[reg_key]['"EXE Path"']))
             os.remove(tmp_reg_file)
     except OSError as e:
         dbg("wine registry extraction failed: " + str(e))
--- a/openage/convert/main.py
+++ b/openage/convert/main.py
@@ -8,6 +8,7 @@ import readline  # pylint: disable=unused-import
 import subprocess
 from configparser import ConfigParser
 from pathlib import Path
+from tempfile import NamedTemporaryFile
 
 from .game_versions import GameVersion, get_game_versions
 from . import changelog
@@ -241,25 +242,24 @@ def _get_source_dir_proposals():
     # (e.g. wine-devel from wine upstream debian repos)
     try:
         # get wine registry key of the age installation
-        tmp_reg_file = 'aoe_temp.reg'
-        if not subprocess.call(('wine', 'regedit', '/E', tmp_reg_file,
-                                REGISTRY_KEY)) and Path(tmp_reg_file).is_file():
-            # strip the REGEDIT4 header, so it becomes a valid INI
-            lines = open(tmp_reg_file, 'r').readlines()
-            del lines[0:2]
+        with NamedTemporaryFile(mode='r') as reg_file:
+            if not subprocess.call(('wine', 'regedit', '/E', reg_file.name,
+                                    REGISTRY_KEY)):
+                # strip the REGEDIT4 header, so it becomes a valid INI
+                lines = reg_file.readlines()
+                del lines[0:2]
 
-            reg_parser = ConfigParser()
-            reg_parser.read_string(''.join(lines))
-            for suffix in REGISTRY_SUFFIX_AOK, REGISTRY_SUFFIX_TC:
-                reg_key = REGISTRY_KEY + suffix
-                if reg_key in reg_parser:
-                    if '"InstallationDirectory"' in reg_parser[reg_key]:
-                        yield _wine_to_real_path(unescape_winereg(
-                            reg_parser[reg_key]['"InstallationDirectory"']))
-                    if '"EXE Path"' in reg_parser[reg_key]:
-                        yield _wine_to_real_path(unescape_winereg(
-                            reg_parser[reg_key]['"EXE Path"']))
-            os.remove(tmp_reg_file)
+                reg_parser = ConfigParser()
+                reg_parser.read_string(''.join(lines))
+                for suffix in REGISTRY_SUFFIX_AOK, REGISTRY_SUFFIX_TC:
+                    reg_key = REGISTRY_KEY + suffix
+                    if reg_key in reg_parser:
+                        if '"InstallationDirectory"' in reg_parser[reg_key]:
+                            yield _wine_to_real_path(unescape_winereg(
+                                reg_parser[reg_key]['"InstallationDirectory"']))
+                        if '"EXE Path"' in reg_parser[reg_key]:
+                            yield _wine_to_real_path(unescape_winereg(
+                                reg_parser[reg_key]['"EXE Path"']))
     except OSError as error:
         dbg("wine registry extraction failed: " + str(error))
 
--- a/openage/convert/main.py
+++ b/openage/convert/main.py
@@ -4,6 +4,8 @@
 import os
 # importing readline enables the raw_input calls to have history etc.
 import readline  # pylint: disable=unused-import
+import subprocess
+from configparser import ConfigParser
 
 from .game_versions import GameVersion, get_game_versions
 from . import changelog
@@ -20,6 +22,10 @@ _STANDARD_PATH_IN_32BIT_WINEPREFIX =\
     "drive_c/Program Files/Microsoft Games/Age of Empires II/"
 _STANDARD_PATH_IN_64BIT_WINEPREFIX =\
     "drive_c/Program Files (x86)/Microsoft Games/Age of Empires II/"
+_REGISTRY_KEY = \
+    "HKEY_LOCAL_MACHINE\\Software\\Microsoft\\Microsoft Games\\"
+_REGISTRY_SUFFIX_AOK = "Age of Empires\\2.0"
+_REGISTRY_SUFFIX_TC = "Age of Empires II: The Conquerors Expansion\\1.0"
 
 
 class DirectoryCreator(FSLikeObjWrapper):
@@ -174,10 +180,11 @@ def acquire_conversion_source_dir():
         print("found environment variable AGE2DIR")
     else:
         # TODO: use some sort of GUI for this (GTK, QtQuick, zenity?)
-        proposals = [proposal for proposal in _get_source_dir_proposals()
-                     if os.path.exists(_expand_relative_path(proposal))]
+        proposals = set(proposal for proposal in _get_source_dir_proposals()
+                        if os.path.exists(_expand_relative_path(proposal)))
         print("Select an Age of Kings installation directory. "
               "Insert the index of one of the proposals, or any path:")
+        proposals = sorted(proposals)
         for index, proposal in enumerate(proposals):
             print("({}) {}".format(index, proposal))
 
@@ -207,7 +214,22 @@ def _get_source_dir_proposals():
         yield "$WINEPREFIX/" + _STANDARD_PATH_IN_64BIT_WINEPREFIX
     yield "~/.wine/" + _STANDARD_PATH_IN_32BIT_WINEPREFIX
     yield "~/.wine/" + _STANDARD_PATH_IN_64BIT_WINEPREFIX
-    # TODO: wine reg query "HKLM\Software\Microsoft\Microsoft Games\Age of Empires\2.0"
+    try:
+        # get wine registry key of the age installation
+        tmp_reg_file = 'aoe_temp.reg'
+        if subprocess.call(['wine', 'regedit', '/E', tmp_reg_file,
+                        _REGISTRY_KEY]) and os.path.exists(tmp_reg_file):
+            reg_parser = ConfigParser(tmp_reg_file)
+            for suffix in _REGISTRY_SUFFIX_AOK, _REGISTRY_SUFFIX_TC:
+                reg_key = _REGISTRY_KEY + suffix
+                if reg_key in reg_parser:
+                    if 'InstallationDirectory' in reg_parser[reg_key]:
+                        yield reg_parser[reg_key]['InstallationDirectory']
+                    if 'EXE Path' in reg_parser[reg_key]:
+                        yield reg_parser[reg_key]['EXE Path']
+            os.remove(tmp_reg_file)
+    except OSError as e:
+        dbg("wine registry extraction failed: " + str(e))
 
 
 def conversion_required(asset_dir, args):
--- a/openage/convert/main.py
+++ b/openage/convert/main.py
@@ -1,11 +1,11 @@
 # Copyright 2015-2016 the openage authors. See copying.md for legal info.
 
 """ Entry point for all of the asset conversion. """
+
 import os
 # importing readline enables the raw_input calls to have history etc.
 import readline  # pylint: disable=unused-import
 import subprocess
-from subprocess import Popen, PIPE
 from configparser import ConfigParser
 
 from .game_versions import GameVersion, get_game_versions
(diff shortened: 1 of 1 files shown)


goto-bus-stop:
```
➜  openage git:(convert-wine-reg) ✗ ./run convert                         
INFO [py] No converted assets have been found
media conversion is required.
Select an Age of Kings installation directory. Insert the index of one of the proposals, or any path:
> Interrupted, aborting
➜  openage git:(convert-wine-reg) ✗ grep -A1 -B4 "EXE Path" aoe_temp.reg

[HKEY_LOCAL_MACHINE\Software\Microsoft\Microsoft Games\Age of Empires\2.0]
"CDPath"="C:\\Program Files (x86)\\aoc\\"
"Custom Mouse"=dword:00000000
"EXE Path"="C:\\Program Files (x86)\\aoc\\"
"InstallationDirectory"="C:\\Program Files (x86)\\aoc\\"
--
[HKEY_LOCAL_MACHINE\Software\Microsoft\Microsoft Games\Age of Empires II: The Conquerors Expansion\1.0]
"Adjust Terrains"=dword:00000000
"CDPath"="C:\\Program Files (x86)\\aoc\\"
"Custom Mouse"=dword:00000000
"EXE Path"="C:\\Program Files (x86)\\aoc\\"
"Extend Population"=dword:00000001
```

Hmm, it doesn't appear to detect the suggestions for me, although it does find the correct registry keys.

azrdev:
@goto-bus-stop okay, so `_get_source_dir_proposals` runs through, but `aoe_temp.reg` still exists… I guess this means `subprocess.call()` returns `False`, even though it did what we wanted

goto-bus-stop:
Yep, I added a `print(subprocess.call(…))` before the if check, and it prints `0` (which almost looks like an exit code?)

azrdev:
@goto-bus-stop It is the exit code. Seems like I got the boolean logic wrong:
~~~
>>> import subprocess
>>> subprocess.call('false')
1
>>> bool(subprocess.call('false'))
True
>>> bool(subprocess.call('true'))
False
~~~

If you want to try, just invert the condition in my `if subprocess.call`, i.e. insert `not`.

goto-bus-stop:
Yep. I got another error in the `ConfigParser`[0]. Applying this:

```diff
-reg_parser = ConfigParser(tmp_reg_file)
+reg_parser = ConfigParser()
+reg_parser.read(tmp_reg_file)
```

fixes that error, but introduces _another_:

```
configparser.MissingSectionHeaderError: File contains no section headers.
```

…which appears to be because `regedit` doesn't _quite_ dump a valid ini file:

```
➜  openage git:(convert-wine-reg) ✗ head -3 aoe_temp.reg
REGEDIT4

[HKEY_LOCAL_MACHINE\Software\Microsoft\Microsoft Games]
```

So that first line `REGEDIT4` should be stripped away somewhere.

[0]: because ConfigParser takes default values in the first param

    File "/usr/lib/python3.4/configparser.py", line 590, in __init__
        for key, value in defaults.items():
    AttributeError: 'str' object has no attribute 'items'

goto-bus-stop:
Pushed a few changes that make it work for me to [goto-bus-stop/sfttech-openage#convert-wine-reg](https://github.com/goto-bus-stop/sfttech-openage/tree/convert-wine-reg).

Compare view with this PR:

https://github.com/azrdev/openage/compare/convert-proposals...goto-bus-stop:convert-wine-reg

```
➜  openage git:(convert-wine-reg) ./run convert
INFO [py] No converted assets have been found
media conversion is required.
Select an Age of Kings installation directory. Insert the index of one of the proposals, or any path:
(0) /home/user/.wine/dosdevices/c:/Program Files (x86)/aoc/
> 
```

openage/convert/main.py
```diff
@@ -207,7 +214,22 @@ def _get_source_dir_proposals():
         yield "$WINEPREFIX/" + _STANDARD_PATH_IN_64BIT_WINEPREFIX
     yield "~/.wine/" + _STANDARD_PATH_IN_32BIT_WINEPREFIX
     yield "~/.wine/" + _STANDARD_PATH_IN_64BIT_WINEPREFIX
-    # TODO: wine reg query "HKLM\Software\Microsoft\Microsoft Games\Age of Empires\2.0"
+    try:
+        # get wine registry key of the age installation
+        tmp_reg_file = 'aoe_temp.reg'
+        if subprocess.call(['wine', 'regedit', '/E', tmp_reg_file,
+                        _REGISTRY_KEY]) and os.path.exists(tmp_reg_file):
```
TheJJ: you can use `with NamedTempFile() as regfile:` from [tempfile](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile), the it gets autoremoved once the `with` block is left.

TheJJ:
I think `ConfigParser` supports a `DEFAULT` section for things not written under some `[rofl]`, so I think it should work with that `REGEDIT4` thingy.

openage/convert/main.py
```diff
@@ -20,6 +22,10 @@
     "drive_c/Program Files/Microsoft Games/Age of Empires II/"
 _STANDARD_PATH_IN_64BIT_WINEPREFIX =\
     "drive_c/Program Files (x86)/Microsoft Games/Age of Empires II/"
+_REGISTRY_KEY = \
+    "HKEY_LOCAL_MACHINE\\Software\\Microsoft\\Microsoft Games\\"
+_REGISTRY_SUFFIX_AOK = "Age of Empires\\2.0"
+_REGISTRY_SUFFIX_TC = "Age of Empires II: The Conquerors Expansion\\1.0"
```
TheJJ: You should name all of those without a leading `_`, looks better then.

openage/convert/main.py
```diff
@@ -174,10 +180,11 @@ def acquire_conversion_source_dir():
         print("found environment variable AGE2DIR")
     else:
         # TODO: use some sort of GUI for this (GTK, QtQuick, zenity?)
-        proposals = [proposal for proposal in _get_source_dir_proposals()
-                     if os.path.exists(_expand_relative_path(proposal))]
+        proposals = set(proposal for proposal in _get_source_dir_proposals()
+                        if os.path.exists(_expand_relative_path(proposal)))
         print("Select an Age of Kings installation directory. "
               "Insert the index of one of the proposals, or any path:")
+        proposals = sorted(proposals)
```
TheJJ: you should check if the entered userselection is a valid proposal, currently, it will raise an index error.

openage/convert/main.py
```diff
@@ -207,7 +214,22 @@ def _get_source_dir_proposals():
...
     yield "~/.wine/" + _STANDARD_PATH_IN_32BIT_WINEPREFIX
     yield "~/.wine/" + _STANDARD_PATH_IN_64BIT_WINEPREFIX
-    # TODO: wine reg query "HKLM\Software\Microsoft\Microsoft Games\Age of Empires\2.0"
+    try:
+        # get wine registry key of the age installation
+        tmp_reg_file = 'aoe_temp.reg'
+        if subprocess.call(['wine', 'regedit', '/E', tmp_reg_file,
+                        _REGISTRY_KEY]) and os.path.exists(tmp_reg_file):
+            reg_parser = ConfigParser(tmp_reg_file)
+            for suffix in _REGISTRY_SUFFIX_AOK, _REGISTRY_SUFFIX_TC:
```
TheJJ: not sure if we should even check for AOK, since we require TC anyway currently.

goto-bus-stop:
> I think ConfigParser supports a DEFAULT section for things not written under some [rofl], so I think it should work with that REGEDIT4 thingy.

I can't get that to work:

```shell
➜  cat test.reg 
REGEDIT4

key=value

➜  python3
Python 3.4.3+ (default, Oct 14 2015, 16:03:50) 
[GCC 5.2.1 20151010] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import configparser
>>> c = configparser.ConfigParser()
>>> c.read('test.reg')
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.4/configparser.py", line 671, in read
    self._read(fp, filename)
  File "/usr/lib/python3.4/configparser.py", line 1057, in _read
    raise MissingSectionHeaderError(fpname, lineno, line)
configparser.MissingSectionHeaderError: File contains no section headers.
file: 'test.reg', line: 1
'REGEDIT4\n'
>>> 
```

I thought it only accepts stray key/value pairs, perhaps, but:

```shell
➜  cat test2.reg 
key=value

➜  python3      
Python 3.4.3+ (default, Oct 14 2015, 16:03:50) 
[GCC 5.2.1 20151010] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import configparser
>>> c = configparser.ConfigParser()
>>> c.read('test.reg')
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.4/configparser.py", line 671, in read
    self._read(fp, filename)
  File "/usr/lib/python3.4/configparser.py", line 1057, in _read
    raise MissingSectionHeaderError(fpname, lineno, line)
configparser.MissingSectionHeaderError: File contains no section headers.
file: 'test.reg', line: 1
'key=value\n'
>>> 
```

TheJJ:
I think you are right: https://docs.python.org/3/library/configparser.html#supported-ini-file-structure

>  The only bit of magic involves the DEFAULT section which provides default values for all other sections

-> seems like we need to ignore the first line then.

openage/convert/main.py
azrdev: as I read tempfile docs, they always construct *and open* a temp file for you, and remove it once you close it. Since we want another process (wine regedit) to write to it, we cannot use tempfile, because python would hold a write fd locking the file

openage/convert/main.py
azrdev: will be done. I just thought of these as implementation details, thus made them "protected"

openage/convert/main.py
azrdev: A loop around everything sounds useful, will do. Everything else… do it if you wish, for me it seemed most convenient to "try if it's a valid index, otherwise try as path".
However, I just note I didn't quite implement that ;) will do

openage/convert/main.py
azrdev: Well, we talked about partial converts, once we've got that it becomes useful. Until then `#TODO`

azrdev:
> -> seems like we need to ignore the first line then.
I tried unsuccessfully with something like
~~~
with open('aoe.reg') as tmpfile:
  parser.read_file(next(tmpfile))
~~~
which *should* skip the first line, but still gives me errors. Now I'm following up to what @goto-bus-stop did, by just `readlines()`ing the whole file into memory, deleting the first to lines, and then using `configparser.read_string()`
commits incoming

openage/convert/main.py
TheJJ: See here, I did the same thing:
https://github.com/SFTtech/kevin/blob/f978fbd9ee3f1bae8c68b9b1f5266642ff21367c/kevin/util.py#L110
This case is the other way round, python writes it and another process (`ssh`) reads it. Note the context managers of this class.

The complete code would be:
``` python
with NamedTempFile(mode="r") as regfile:
    ret = subprocess.call(['wine', 'regedit', '/E', regfile.name, _REGISTRY_KEY])
    if ret != 0:
        raise whatever
```

openage/convert/main.py
TheJJ: my patch for it:
``` patch

diff --git a/openage/convert/main.py b/openage/convert/main.py
index 30d573c..53e7f73 100644
--- a/openage/convert/main.py
+++ b/openage/convert/main.py
@@ -5,14 +5,14 @@ import os
 # importing readline enables the raw_input calls to have history etc.
 import readline  # pylint: disable=unused-import
 
-from .game_versions import GameVersion, get_game_versions
+from pathlib import Path
+
 from . import changelog
+from .game_versions import GameVersion, get_game_versions
 
 from ..log import warn, info, dbg
-from ..util.fslike.wrapper import (
-    Wrapper as FSLikeObjWrapper,
-    Synchronizer as FSLikeObjSynchronizer
-)
+from ..util.fslike.wrapper import (Wrapper as FSLikeObjWrapper,
+                                   Synchronizer as FSLikeObjSynchronizer)
 from ..util.fslike.directory import CaseIgnoringDirectory
 from ..util.strings import format_progress
 
@@ -92,6 +92,10 @@ def convert_assets(assets, args, srcdir=None):
     This method prepares srcdir and targetdir to allow a pleasant, unified
     conversion experience, then passes them to .driver.convert().
     """
+
+    # import here so codegen.py doesn't depend on it.
+    from .driver import convert
+
     # acquire conversion source directory
     if srcdir is None:
         srcdir = acquire_conversion_source_dir()
@@ -103,6 +107,7 @@ def convert_assets(assets, args, srcdir=None):
     versions = "; ".join(str(version) for version in args.game_versions)
     if not any(version.openage_supported for version in args.game_versions):
         warn("None supported of the Game version(s) {}".format(versions))
+        warn("You need \x1b[34mAge of Empires 2: The Conquerors\x1b[m")
         return False
     info("Game version(s) detected: {}".format(versions))
 
@@ -125,7 +130,6 @@ def convert_assets(assets, args, srcdir=None):
 
     args.flag = flag
 
-    from .driver import convert
     converted_count = 0
     total_count = None
     for current_item in convert(args):
@@ -154,7 +158,7 @@ def convert_assets(assets, args, srcdir=None):
     return True
 
 
-def _expand_relative_path(path):
+def expand_relative_path(path):
     """Expand relative path to an absolute one, including abbreviations like
     ~ and environment variables"""
     return os.path.realpath(os.path.expandvars(os.path.expanduser(path)))
@@ -167,40 +171,54 @@ def acquire_conversion_source_dir():
     Returns a file system-like object that holds all the required files.
     """
     # ask for the conversion source
-    print("media conversion is required.")
+    print("\x1b[33mmedia conversion is required.\x1b[m")
 
     if 'AGE2DIR' in os.environ:
         sourcedir = os.environ['AGE2DIR']
-        print("found environment variable AGE2DIR")
+        print("found environment variable 'AGE2DIR'")
     else:
         # TODO: use some sort of GUI for this (GTK, QtQuick, zenity?)
-        proposals = [proposal for proposal in _get_source_dir_proposals()
-                     if os.path.exists(_expand_relative_path(proposal))]
-        print("Select an Age of Kings installation directory. "
-              "Insert the index of one of the proposals, or any path:")
-        for index, proposal in enumerate(proposals):
-            print("({}) {}".format(index, proposal))
-
-        try:
+        proposals = [proposal for proposal in source_dir_proposals()
+                     if os.path.exists(expand_relative_path(proposal))]
+        print("Select an \x1b[34m'Age of Empires 2: The Conquerors'\x1b[m "
+              "installation directory.")
+
+        if proposals:
+              print("Insert the index of one of the proposals")
+              for index, proposal in enumerate(proposals):
+                  print("({}) {}".format(index, proposal))
+
+        while True:
             user_selection = input("> ")
+            if not user_selection:
+                continue
+
             if user_selection.isdecimal():
-                sourcedir = proposals[int(user_selection)]
+                proposal = int(user_selection)
+
+                if proposal in proposals:
+                    sourcedir = proposals[proposal]
+                else:
+                    print("invalid selection: '%d', try again." % (proposal))
+                    continue
+
             else:
                 sourcedir = user_selection
-        except KeyboardInterrupt:
-            print("Interrupted, aborting")
-            exit(0)
-        except EOFError:
-            print("EOF, aborting")
-            exit(0)
 
-    sourcedir = _expand_relative_path(sourcedir)
-    print("converting from " + sourcedir)
+            sourcedir = expand_relative_path(sourcedir)
+
+            if not Path(sourcedir).is_dir():
+                print("'%s' not a valid directory." % sourcedir)
+                continue
+            else:
+                break
+
+    print("converting from '%s'" % sourcedir)
 
     return CaseIgnoringDirectory(sourcedir).root
 
 
-def _get_source_dir_proposals():
+def source_dir_proposals():
     """Yield a list of directory names where an installation might be found"""
     if "WINEPREFIX" in os.environ:
         yield "$WINEPREFIX/" + _STANDARD_PATH_IN_32BIT_WINEPREFIX
@@ -216,24 +234,35 @@ def conversion_required(asset_dir, args):
 
     Sets options in args according to what sorts of conversion are required.
     """
+
     try:
-        version_path = asset_dir['converted', changelog.ASSET_VERSION_FILENAME]
+        version_path = asset_dir['converted',
+                                 changelog.ASSET_VERSION_FILENAME]
+
+        spec_path = asset_dir['converted',
+                              changelog.GAMESPEC_VERSION_FILENAME]
+
         with version_path.open() as fileobj:
             asset_version = fileobj.read().strip()
+
+        with spec_path.open() as fileobj:
+            spec_version = fileobj.read().strip()
+
+        asset_version = int(asset_version)
+
     except FileNotFoundError:
         # assets have not been converted yet
         info("No converted assets have been found")
         return True
 
-    # TODO: datapack parsing
-
-    try:
-        asset_version = int(asset_version)
     except ValueError:
-        info("Converted assets have improper format; expected integer version")
+        info("Converted assets have improper format; "
+             "expected integer version")
         return True
 
-    changes = changelog.changes(asset_version)
+    # TODO: datapack parsing
+
+    changes = changelog.changes(asset_version, spec_version)
 
     if not changes:
         dbg("Converted assets are up to date")
@@ -247,6 +276,7 @@ def conversion_required(asset_dir, args):
                 # don't reconvert this component:
                 setattr(args, "no_{}".format(component), True)
 
+        # TODO
         if "metadata" in changes:
             args.no_pickle_cache = True
 
```

TheJJ: (but ignore the spec version stuff, i created it for the version hashing in my `convert` branch)

openage/convert/main.py
TheJJ: some whistleblower will leak them anyway, we have nothing to protect :P

azrdev:
@TheJJ adapted your proposal. Hope I didn't oversee anything in that big patch :P 
@goto-bus-stop you wanna test?

openage/convert/main.py
```diff
@@ -227,11 +234,13 @@ def _get_source_dir_proposals():
...
     yield "~/.wine/" + STANDARD_PATH_IN_32BIT_WINEPREFIX
     yield "~/.wine/" + STANDARD_PATH_IN_64BIT_WINEPREFIX
+    #TODO: a switch to never call wine binaries (which might accidentally modify a wineprefix)
+    #TODO: a possibility to call different wine binaries (e.g. wine-devel from wine upstream debian repos)
     try:
         # get wine registry key of the age installation
         tmp_reg_file = 'aoe_temp.reg'
         if not subprocess.call(('wine', 'regedit', '/E', tmp_reg_file,
-                                REGISTRY_KEY)) and os.path.exists(tmp_reg_file):
+                                REGISTRY_KEY)) and Path(tmp_reg_file).is_dir():
```
goto-bus-stop: → `Path(tmp_reg_file).is_file()`

goto-bus-stop:
It works if you change the `Path(tmp_reg_file).is_dir()` check to check for `is_file` instead ;)

goto-bus-stop:
Screenshot

:+1:

openage/convert/main.py
```diff
@@ -174,40 +182,86 @@ def acquire_conversion_source_dir():
...
+    # TODO: a switch to never call wine binaries
+    # (which might accidentally modify a wineprefix)
+
+    # TODO: a possibility to call different wine binaries
+    # (e.g. wine-devel from wine upstream debian repos)
+    try:
+        # get wine registry key of the age installation
+        tmp_reg_file = 'aoe_temp.reg'
+        if not subprocess.call(('wine', 'regedit', '/E', tmp_reg_file,
+                                REGISTRY_KEY)) and Path(tmp_reg_file).is_file():
```
TheJJ: did you try my tempfile suggestion yet?

azrdev:
@goto-bus-stop can you try 9093972 if @TheJJ's `NamedTempFile` approach still works?
Either way, I'd like to have this on master soon, nothing TBD here IMHO.

goto-bus-stop:
looks good :smile:

TheJJ:
Cool :trophy:

elnabo:
When you use wine to detect the installation and are unable to enter a valid prefix you can't leave the loop. Or at least I didn't find how.

`This does not appear to be a valid WINEPREFIX.

Enter a valid one, or leave it empty to skip.

WINEPREFIX=

This does not appear to be a valid WINEPREFIX.

Enter a valid one, or leave it empty to skip.
`

TheJJ:
Isn't it possible to just press enter as noted in "or leave it empty to skip."?
Or, to fully abort it, `Ctrl-c` as always?

elnabo:
I tried to cancel it with enter but with no avail so I shut it down with `Ctrl-c`

azrdev:
@elnabo  please file a separate issue for such things. Especially since this PR did not yet contain the possibility to specify a wineprefix interactively (this came with bb02b09431ccd019019c4449c13ce10f2833fd8a)

PR #529, opened 2016-03-19
[WIP] new savefile format with working triggers

timohaas:
hi,

I tried to specify a new savefile format with triggers.
for information look at doc/map-format.md

with this generic trigger approach it is possible to implement nearly everything (example: relics gold,technologies)
relics:
condition: player-has relic => 0.2Gold/s

but before further implemantation, i would like your thoughts on this.


libopenage/gamestate/game_save.cpp
```diff
@@ -12,93 +14,242 @@
 #include "game_main.h"
 #include "game_save.h"
 #include "game_spec.h"
+#include "picojson.h"
```
TheJJ: i think we can add it as an external dependency easily instead of bundling it. also avoids the licensing overhead and adding stuff to our copying.md

doc/map-format.md
```diff
@@ -0,0 +1,188 @@
+# Map Format
+
+**What is this?** This document descrcibes the file format for save files of openage.
```
TheJJ: So you propose a scenario file is exactly the same as a saved game state? Could be done, like a "snapshot".
Much of the gamestate is lost in this simple format, we need a full recording format as well.

I was thinking of using the ideas from https://blog.forrestthewoods.com/the-tech-of-planetary-annihilation-chronocam-292e3d6b169a :)

But still, this is waaay better than our current save format.

TheJJ:
uh why did you close it?

timohaas:
I did some mistakes with git bevore pushing it. I should do a pull request when its working.

TheJJ:
no problem, you can always force-push to the same branch after rewriting history, then this pullrequest can be updated.

libopenage/gamestate/game_save.cpp
timohaas: yeah, we are now at a point where it is possible to change the json lib we choose. So we should choose wisely.

doc/map-format.md
timohaas: If I understand the article correctly, we need to add to the units additionally fields like:
vector, velocity etc (and if we want a recorded file everything as array with timestamps, this can be done later).
What we need now: unique IDs(uint64) for units. (maybe gold/forest too if we have triggers who can create and destroy them)

PR #537, opened 2016-03-28
Adding CVar subsystem

elnabo:
See #486

Please overlook the crappy cython method didn't want to rewrite all the hardcoded keybind yet.
Keybinding set from python are working.


TheJJ:
You can create a `cvar` subsystem folder in `libopenage` as well, there will be more files soon :)

elnabo:
So as of now, it is possible to create a cvar, access it from python and the ingame "console" with `set name value` and `get name` (no more than one space).

I removed the size cvar, since my brain didn't want to do it (or SDL didn't want me to do it)

libopenage/console/console.cpp
```diff
@@ -131,6 +132,23 @@ void Console::interpret(const std::string &command) {
...
+		}
+	}
+	else if (command.substr(0,3) == "get") {
+		std::size_t first_space = command.find(" ");
+		if (first_space != std::string::npos) {
+			std::string name = command.substr(first_space+1, std::string::npos);
+			std::string value = Engine::get().get_cvar_manager().get(name);
+			this->write(value.c_str());
+		}
+	}
```
TheJJ: it works for now, but we really should implement the interpreter in python. that way, arbitrary commands can be entered and the game controlled later. The commands for cvars within the python environment should be something like `openage.cvar.set("name", value)`, which can be abbrevated in the console by doing `import openage.cvar as cvar` in the `code.interact()`, if we wanna have the regular interactive python as default console.

Otherwise we can start some own interpreter parsing commands like you did here, which can then invoke the `code.interact()` by some command like `python`.

libopenage/cvar/cvar.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include "cvar.h"
+#include "../engine.h"
+
+namespace openage {
+namespace cvar {
+
+bool CVarManager::create(const std::string &name,  const std::pair<get_func, set_func> accessors) {
+	auto it = this->store.find(name);
+	if (it == this->store.end()) {
+		this->store.insert(std::make_pair(name, accessors));
```
TheJJ: isn't that `this->store[name] = accessors`?

elnabo:
The reason I added the 2 crappy console commands with my commit is that they are really usefull for debugging purpose until a true console is added. This is why I added access to the cvar from python.
But I think, the console should be made in python, as python string handling looks better :)

As for the insertion, I like the verbose syntax, but if you prefer i'll use the op[] instead.

elnabo:
Keybinds are not hardcoded anymore :)

They are loaded from a config file, that can load other config file (without infinte loop)

config.cfg
```diff
@@ -0,0 +1,52 @@
+# Copyright 2016-2016 the openage authors. See copying.md for legal info.
+
+set TOGGLE_CONSOLE `
```
TheJJ: cool stuff :)
do you think it's better placing this file to `dist/`? also, it's not really a `.cfg` with the unix-cfg-like-syntax, so that file suffix is maybe inappropriate (`.oac`, `.oacvar`, ...?)

config.cfg
```diff
@@ -0,0 +1,52 @@
...
+set SET_ABILITY_MOVE Ctrl x
+set SET_ABILITY_GATHER Ctrl c
+set SET_ABILITY_GARRISON g
+set SPAWN_VILLAGER v
+set KILL_UNIT Delete
+set BUILDING_1 q
+set BUILDING_2 w
+set BUILDING_3 e
+set BUILDING_4 r
+set BUILDING_5 Ctrl q
```
TheJJ: should we maybe clarify the syntax by `Ctrl+q` to "concat" the keys?

libopenage/cvar/cvar.cpp
```diff
@@ -0,0 +1,39 @@
+// Copyright 2013-2016 the openage authors. See copying.md for legal info.
+
+#include "cvar.h"
+#include "../engine.h"
+
+namespace openage {
+namespace cvar {
+
+bool CVarManager::create(const std::string &name,  const std::pair<get_func, set_func> accessors) {
```
TheJJ: the `accessors` param can also be `a const &`, then the move will be used instead of the copy i think.

libopenage/cvar/cvar.h
```diff
@@ -0,0 +1,43 @@
...
+ *     string get(string name) except +
+ *     void set(string name, string value) except +
+ */
+class CVarManager {
+	public :
+		bool create(const std::string &name, const std::pair<get_func, set_func> accessors);
+		std::string get(const std::string &name) const;
+		void set(const std::string &name, const std::string &value) const;
+	private :
+		std::unordered_map<std::string, std::pair<get_func, set_func>> store;
```
TheJJ: would be great if you could adhere to our current styleguide, the access labels at the same indent level of the class and the members at one tab indent.

config.cfg
elnabo: I'm not sure about where to put the file, I think the best would be either in the same place as the executable/launcher or in a subdirectory for config file. 

As for the extension, I used `.cfg` since it is a config file but anything could do. I prefer `.oac` over `.oacvar` since it is shorter.

config.cfg
elnabo: I choosed to use ` ` as separator since it was the only (or one of the few) character that you have to write. I was going to use `+` has you suggested but I didn't like the idea of having `Ctrl++`

config.cfg
TheJJ: how about using `cfg/keybinds.oac` then? that cfg folder is installed to `/usr/share` as a template to initialize the `~/.config/openage/cfg` folder then.

config.cfg
TheJJ: yea right. probably the best solution for now

TheJJ: How about something like `set BUILDING_6 "Ctrl w"`? Then the parsing is simpler and less error-prone. It's also better for the future when other than set commands are there I guess.

libopenage/cvar/cvar.h
```diff
@@ -0,0 +1,43 @@
...
+};
+
+// pxd: CVarManager &get_cvar_manager() except +
+CVarManager &get_cvar_manager();
+
+// pxd: PyIfFunc0[void] load_config
+extern pyinterface::PyIfFunc<void> load_config;
```
TheJJ: this is kinda bad, it should be a member of the cvarmanager and not a global function. the sane way would be having the engine return the cvar manager, on which we can then call `.load_config(file)`.

openage/cvar/cvar.pyx
```diff
@@ -0,0 +1,17 @@
+# Copyright 2016-2016 the openage authors. See copying.md for legal info.
+
+from libopenage.cvar.cvar cimport load_config as c_load_config, get_cvar_manager
+
+from .config_file import load_config_file as load_config
+
+cdef void load_config_file() except * with gil:
+    load_config("config.cfg")
```
TheJJ: here it would be `get_cvar_manager().load_config(...)`

openage/cvar/cvar.pyx
```diff
@@ -0,0 +1,17 @@
...
+
+cpdef set_cvar(name, value) with gil:
+    get_cvar_manager().set(name.encode('UTF-8'),value.encode('UTF-8'))
+
+cpdef get_cvar(name) with gil:
+    get_cvar_manager().get(name.encode('UTF-8'))
+
+def setup():
+    c_load_config.bind0(load_config_file)
```
TheJJ: the `c_load_config` is then a member of the cvarmanager and can still be bound that way.

libopenage/cvar/cvar.h
elnabo: When I access the engine from python, I have compilation problem with opus_multistream, this is why I don't call it from the engine (see the second commit of this PR)

elnabo:
TheJJ : It seems that having a python function local is impossible. If I try to bind the function by doing a call to python after the creation of the instance I get 

`openage.cppinterface.exctranslate.CPPException: Can't instantiate py interface component as non-global object.`

TheJJ:
Damnit. The problem is we _need_ to change the python interface to support such a thing. otherwise, dynamically allocated objects can never be used by any script provided by a mod.
When some villagers get some action hooks attached, the cpp engine must call to python, and the villager object must be accessible from there. And as this object was created on the fly in the c++ code, the pyiface function ptr has to be assigned after the object was created.

@mic-e do you think we can extend the python interface to support binding to non-global symbols?
Maybe with another indirection layer or something I guess it will work.

TheJJ:
You are lucky to create the first c++ component to have a python interface, so we now trigger all the problems we didn't think of when designing the python interface :)

elnabo:
So I don't think you can bind a local method without losing the reference to the instance. So I use c-style python binding (take a pointer to the instance as argument), called from a local wrapper.

A bit hacky, not so clean, but fully working.

openage/cvar/config_file.py
```diff
@@ -31,4 +30,5 @@ def load_config_file(path, loaded=None):
                 set_cvar(split[1], " ".join(split[2:]))
             elif split[0] == "load" and len(split) >= 2:
                 for sub_path in split[1:]:
-                    load_config_file(os.path.join(os.path.dirname(path), sub_path), loaded)
+                    new_path = os.path.join(os.path.dirname(path), sub_path)
+                    load_config_file(new_path, set_cvar, loaded)
```
TheJJ: i'd design the inclusion relative to the file that specifies the inclusion.
in `lol.cfg` i can then write `load bla/stuff.cfg`, where `lol.cfg` and the `bla` directory are in the same path.

TheJJ:
Cool, I'm ok with that. Anything left apart from my comments above (the file location, relative path)?

openage/cvar/config_file.py
```diff
@@ -0,0 +1,34 @@
...
+
+            strip = lstrip.rstrip()
+            split = strip.split()
+
+            if split[0] == "set" and len(split) >= 3:
+                set_cvar(split[1], " ".join(split[2:]))
+            elif split[0] == "load" and len(split) >= 2:
+                for sub_path in split[1:]:
+                    new_path = os.path.join(os.path.dirname(path), sub_path)
+                    load_config_file(new_path, set_cvar, loaded)
```
TheJJ: i'd design the inclusion relative to the file that specifies the inclusion.
in `lol.cfg` i can then write `load bla/stuff.cfg`, where `lol.cfg` and the `bla` directory are in the same path.

libopenage/cvar/cvar.h
```diff
@@ -0,0 +1,48 @@
...
+ *     PyIfFunc2[void, string, CVarManager*] load_config_file
+ */
+class CVarManager {
+
+public :
+	bool create(const std::string &name, const std::pair<get_func, set_func> &accessors);
+	std::string get(const std::string &name) const;
+	void set(const std::string &name, const std::string &value) const;
+	void load_config(const std::string &path);
+	static CVarManager &get();
```
TheJJ: with the new approach, we don't need the static `get` any more, right?

openage/cvar/config_file.py
elnabo: That is what is currently done

libopenage/cvar/cvar.h
elnabo: I'm not sure about the static `get`. I'm going to check after work.

openage/cvar/config_file.py
TheJJ: My bad, you are right. Btw you can check out the `pathlib` if you don't know it already, it's way cooler than `os.path`.

elnabo:
I put three places to look for a config file, the first existing file is loaded.

1- ~/.config or %APPDATA% (user local)
2- /etc or %ALLUSERPROFILES% (global)
3- binary folder (by default)

If no config file is found, the default hardcoded values will be used.
Can't test on Windows but it should work.

TheJJ:
Cool stuff!
The C-style object passing is the easiest solution for now I'd say.

We can merge this after you moved the default keybind file to `$projectroot/cfg/keybinds.oac` or some other location if you got a better idea.

elnabo:
Looks like the debian machine has problems

I moved the `.oac` to `openage/cfg/` as you suggested. At the moment, I think it has to be launched from the openage folder for `cwd` to give the correct path, but it seems to be a requirement from `run.py`

So if there is no real problem (i'm looking at you debian kevin), i'll squash the commits.

TheJJ:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=823337 is the problem....

TheJJ:
Seriously... I fixed it by replacing systemd-cron with cron, so python does not block the boot because it imports `random` which needs `getrandom()` which needs the kernels randomness pool, which is not initialized yet.

You can trigger a rebuild by amending or squashing :)

elnabo:
I squashed the commits but looks like the debian is still broken (missing the six python library and some c++ stuff).

TheJJ:
I hope I fixed it, please rebase on current master and try again :)

elnabo:
Ok this time it works and it looks like i didn't break anything

TheJJ:
Congratulations, this is the first Python - C++ API-like thing :dancers:

Issue #535, opened 2016-03-26
Segfault when queuing building


elnabo:
To reproduce : 

Select Villagers
Ask to build something (I used Barracks)
Use Shift click to ask to build more of the building on the same spot
If you spam hard enough it should Segfault

Trace : 

```
Traceback (most recent call last):
  File ?, in ? [0x40249f]
  File ?, in __libc_start_main+0xf5 [0x7f8b6e429b05]
  File ?, in main+0x360 [0x402410]
  File ?, in ? [0x403744]
  File ?, in PyInit_run+0x1040 [0x403620]
  File ?, in ? [0x7f8b71ddaec8]
  File ?, in PyEval_EvalCodeEx+0x619 [0x7f8b71dff849]
  File ?, in PyEval_EvalFrameEx+0x3fe4 [0x7f8b71df9a44]
  File ?, in PyEval_EvalFrameEx+0x413b [0x7f8b71df9b9b]
  File ?, in ? [0x7f8b66bd8ba1]
  File ?, in openage::run_game(openage::main_arguments const&)+0x668 [0x7f8b7199907d]
  File ?, in openage::Engine::run()+0x1f [0x7f8b719889db]
  File ?, in openage::Engine::loop()+0x25a [0x7f8b719884f4]
  File ?, in openage::GameMain::update()+0x10 [0x7f8b719c7c44]
  File ?, in openage::UnitContainer::update_all()+0x34 [0x7f8b71a46ecc]
  File ?, in openage::Unit::update()+0xa9 [0x7f8b71a45287]
  File ?, in openage::TargetAction::update(unsigned int)+0x4d [0x7f8b71a2f01f]
  File ?, in openage::BuildAction::update_in_range(unsigned int, openage::Unit*)+0xcc [0x7f8b71a2f50e]
  File ?, in openage::TerrainObject::place(openage::object_state)+0x305 [0x7f8b71a26e2b]
  File ?, in ? [0x7f8b6e43d120]
  File ?, in ? [0x7f8b719c5d18]
  File ?, in ? [0x7f8b6ed2ba41]
  File ?, in ? [0x7f8b6ed2b9f6]
  File ?, in openage::error::terminate_handler()+0xc7 [0x7f8b719c5bbb]
  File ?, in openage::error::StackAnalyzer::analyze()+0x7c [0x7f8b719c63c6]
```

Jon0:
Maybe becuase the GUI dispatches events concurrently. The game probably needs queue events between full state updates.

[...]

PR #539, opened 2016-03-31
Command queue, and fix building placement segfault

Jon0:
Replace `unit.invoke(Command &)` with `unit.queue_cmd(Command &)` which queues commands to be applied at the next update. queue_cmd is safe to call from multiple threads.
 Building placement is fixed, as described in https://github.com/SFTtech/openage/issues/535
Also fixed an issue where gatherers would only return wood.

### DIFF
--- a/libopenage/game_control.cpp
+++ b/libopenage/game_control.cpp
@@ -165,6 +165,7 @@ ActionMode::ActionMode()
 		auto player = engine.player_focus();
 		auto type = player->get_type(this->rng.probability(0.5)? 83 : 293);
 		Command cmd(*player, type);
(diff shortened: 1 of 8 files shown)


TheJJ:
Nicely done, I'd have needed a long time to track that down :+1:

Jon0:
The problem was each floating outline would be removed more than once. A 3x3 barracks outline attempted to be removed 9 times. Attempt to remove more than once caused a segfualt as the object was already deleted.

I implemented the command queue anyway, it should help later for networking.

The code style issue is now fixed, this can be merged.

TheJJ:
gud job :)

PR #543, opened 2016-04-09
Remove crossplatform folder

### LINKED ISSUE #485: Cross-platform separation
Currently, we try to use the `libopenage/crossplattform` folder for switches.
But this is kinda bad, ugly and hard to maintain.

New proposal:

Do everything within the subsystem, e.g. util/, job/, ... whatever.
The things currently in crossplattform/ should be migrated there.

Within a subsystem, the switches should be as clean and few as possible, e.g. one implementation for linux, win32, etc. Basically this equals the stubbing for windows we began.

That way, the code can stay in its subsystem and, if well coded, is better to understand as the subsystem locality is maintained and code jumps are minimized.

Thoughts?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29943463-cross-platform-separation?utm_campaign=plugin&utm_content=tracker%2F371436&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F371436&utm_medium=issues&utm_source=github).
</bountysource-plugin>

johanhelsing:
Fixes #485 

I moved everything to util. I'm not that familiar with the code yet, so please tell me if something is a better fit for another folder.


TheJJ:
nice :) util seems ok for now, once those files become bigger, we can move them to their own subsystem.

PR #544, opened 2016-04-14
Limit fps

jazzpi:
This adds a command line argument `--fps` (to the game subcommand).

When `--fps` is supplied and greater than 0, the main loop will delay after each frame until 1/`fps_limit` seconds have passed.

The FPS limit is applied on top of VSync so it should only come into effect when it is set below the refresh rate you'd need for VSync or if the window is in the background.


TheJJ:
cool stuff!

PR #546, opened 2016-04-30
Missing source in CMakeLists.txt in libopenage/util

sbrodehl:
Added missing os.cpp source.


TheJJ:
seriously? nicely done..

PR #550, opened 2016-05-04
Require harfbuzz version >= 1.x.x

sbrodehl:
Fixes compile error on OS X with clang 3.8 in debug mode.

Other things we need to take care of if we require a newer harfbuzz version?


TheJJ:
You need to store the correct email address in your commits (`git config --global user.name` and `user.email`, to change the commit you already pushed use `--amend --author "your name <your@mail.lol>` and add yourself to `copying.md`.

Also, please update the building documentation and state that we require harfbuzz `>= 1` now.

TheJJ:
And in the buildsystem itself, we should also require `>= 1.0.0`, our `FindHarfbuzz.cmake` already supports version checks.

TheJJ:
Thanks!

Issue #533, opened 2016-03-25
Building placement problem


elnabo:
It is possible to build on non buildable terrain (eg: water, ice) if it touch a walkable terrain. Destroying the building doesn't restore the previous terrain.

It is also possible to build stuff in a town center in some of its corner, try with house.

TheJJ:
building placement is all-new in the new simulation

PR #552, opened 2016-05-05
[WIP] Improving building logic

coffenbacher:
Do not place conflicting buildings and/or foundations. Conflict with unit is OK to place foundation, although building construction cannot begin until the site has been cleared. Addresses second half of #533.

Fixes the two behaviors seen here:

two_bad_building_behaviors


coffenbacher:
I think I also need to check to allow conflicts when it is another player's unbuilt foundation...otherwise Player 2 can just flood the map with foundations and Player 1 can't build.

coffenbacher:
This also causes foundations to be placeable on trees now. I'll try to fix that.

coffenbacher:
OK, this now properly fixes the two issues in the original screenshot. If it looks good I will squash and rebase.

Is there anything I can do to address the failing CI builds?

coffenbacher:
OK this now fully addresses #533. Squashed and rebased. Not sure how to deal with the CI issue? I'm not sure what's failing exactly.

elnabo:
The issue with the CI is that you didn't add yourself to the copying.md

 `WARNING author inconsistency: charles.offenbacher@gmail.com
	email appears in git log, but not in copying.md or .mailmap
`

coffenbacher:
Thanks @elnabo, I see the error now. Fixed.

libopenage/unit/producer.cpp
```diff
@@ -623,8 +623,29 @@ TerrainObject *BuildingProducer::place(Unit *u, std::shared_ptr<Terrain> terrain
...
 			}
+
+			// check if it's in "unbuildable terrain" blacklist
+			// TODO: match this with valid enum
+			if (tc->terrain_id == 1 || // water
+				tc->terrain_id == 4 || // shallows
+				tc->terrain_id == 14 || // deep water
+				tc->terrain_id == 15 || // medium water
+				tc->terrain_id == 18 // ice
+					) return false;
```
TheJJ: This terrain blacklist is called ["terrain restriction"](https://github.com/SFTtech/openage/blob/fe0a66dfa9f4850e1002998245d2d82b7f4d9728/openage/convert/gamedata/terrain.py#L39). We don't use it yet, but maybe you can figure out how it works.

Also, would be nice for consistency reasons if you indented the `tc->terrain_id == 4` by spaces starting at the indent level of `if` to line up with `tc->terrain_id == 1`, see [here](https://github.com/SFTtech/openage/blob/master/doc/code_style/tabs_n_spaces.md).

coffenbacher:
I refactored this to be based on the terrain restriction attribute of the object. As far as I can tell, the actual effect of the terrain restrictions (e.g. what does NO_ICE_0x08 actually allow for) was part of the engine and not part of the conversion, so for now I'm following that although this would be cool to be configurable by mods later. All of the logic is centralized in allowed_terrains now. 

Next, I hope to
- [ ] confirm the behavior of docks
- [ ] confirm the proper "shoreline" behavior as mentioned in the first part of #533 
- [ ] comment, squash, rebase

libopenage/unit/producer.cpp
```diff
@@ -25,27 +25,56 @@
...
+    }
+
+    // if we're using a blacklist, fill out a whitelist with everything not on it
+    if (blacklist.size() > 0) {
+        // Allow all terrains that are not on blacklist
+        for (int i = 0; i < 32; ++i) {
+            if (blacklist.count(i) == 0) {
+                whitelist.insert(i);
+            }
+        }
```
TheJJ: Your indentation here and in the lines above is wrong, please have a look at [our guide](https://github.com/SFTtech/openage/blob/master/doc/code_style/tabs_n_spaces.md).

libopenage/unit/producer.cpp
```diff
@@ -614,17 +643,33 @@ TerrainObject *BuildingProducer::place(Unit *u, std::shared_ptr<Terrain> terrain
 	 */
 	TerrainObject *obj_ptr = u->location.get();
 	std::weak_ptr<Terrain> terrain_ptr = terrain;
-	u->location->passable = [obj_ptr, terrain_ptr](const coord::phys3 &pos) -> bool {
+
+    // find set of allowed terrains
+    std::unordered_set<terrain_t> terrains = allowed_terrains(this->unit_data.terrain_restriction);
+
+    u->location->passable = [obj_ptr, terrain_ptr, terrains](const coord::phys3 &pos) -> bool {
```
TheJJ: indentation :)

coffenbacher: It's a good day when indentation is my worst problem :) I'm going to try finishing another feature (building docks) and then fix all of my changes per the style guide for merging

TheJJ:
Would be awesome if we can purge the hardcoding and implement the terrain restrictions from the gamespec directly.

coffenbacher:
For sure, I'm trying to replace the hardcoding with the equivalent gamedata. Slow going since I'm learning how the codegen and conversion works, but good practice :+1:

TheJJ:
Any progress here? :)

Issue #381, opened 2015-09-15
Trigger reconfigure when git commit hash changes


TheJJ:
When the value of `git describe` changes, we should trigger a reconfigure when typing `make`. This ensures the version generated to `config.h` matches with the actual repository version.

The "dirty" solution would be checking for the version each time, something better would be a update watch on some file in `./.git/`, but maybe you have a better idea how cmake may detect a changed HEAD efficiently.

bvbfan:
cmake-modules has it. https://github.com/rpavlik/cmake-modules/blob/master/GetGitRevisionDescription.cmake

castilma:
but this does not mean that _every_ file gets recompiled after a git commit/checkout, right?

TheJJ:
No. It just means that we must ensure that `config.cpp` is updated once the current git head changes, because that file is generated only when `cmake` thinks its time for it.

The only file that needs to be recompiled is `config.cpp` then.

Currently our `config.h` is recreated, but it is possible to skip it if the preprocessor defines didn't change. The configuration strings (e.g. compiler flags) could be stored in the `.cpp` only, and declared `extern` in the `.h`.

PR #560, opened 2016-05-29
[WIP] Another bold attempt at out of source builds

jprashanth:
- Moves python build process from cmake to setuptools
- Removes the clutter of cmake files from python packages
- Moves generation of pxd files, cython c++ files, python extension from source dir to binary dir
- Removes various buildsystem routines like pxdgen, check_py_file_list, compilepy, cythonize, inplacemodules
- Now, no configuration dependent files (config.h, config.cpp) are generated inside the libopenage directory
- Single point of version specification. Now, there is no need to invent a way to re-configure the project when the git HEAD changes
- Introduced a central source for storing and retrieving build metadata
- Removes the awkward cmake configuration for config.py.in
- Simpler approach to DEVMODE

Addresses #398 #381


libopenage/CMakeLists.txt
```diff
@@ -1,6 +1,20 @@
...
+	# add a C++ header file as pxdgen source file
+	foreach(HEADER_SOURCE ${ARGN})
+		if(NOT "${HEADER_SOURCE}" MATCHES ".*\\.h$")
+			message(FATAL_ERROR "pxdgen accepts only .h files. You provided: ${HEADER_SOURCE}")
+		endif()
+
+		get_filename_component(HEADER_SOURCE "${HEADER_SOURCE}" ABSOLUTE)
+		set_property(GLOBAL APPEND PROPERTY SFT_PXDGEN_SOURCES "${HEADER_SOURCE}")
+	endforeach()
+endfunction()
```
TheJJ: that function has nothing to do with libopenage, it should be somewhere in the `buildsystem/` folder.

libopenage/CMakeLists.txt
```diff
@@ -127,8 +144,9 @@ endif()
 
 get_config_option_string()
 
-configure_file(config.h.in ${CMAKE_CURRENT_SOURCE_DIR}/config.h)
-configure_file(config.cpp.in ${CMAKE_CURRENT_SOURCE_DIR}/config.cpp)
+configure_file(config.h.in config.h)
+configure_file(config.cpp.in config.cpp)
+include_directories(${CMAKE_BINARY_DIR})
```
TheJJ: this include does a lot of magic. would be cool if you commented a bit about it.

libopenage/CMakeLists.txt
```diff
@@ -163,3 +181,33 @@ if(NOT APPLE)
...
+# before the python extensions are built.
+add_custom_target(pxdgen
+	DEPENDS ${GENERATED_PXD_FILES}
+)
+
+add_custom_target(cleanpxdgen
+	COMMAND find "${CMAKE_BINARY_DIR}" -name "*.pxd" -type f -print -delete
+	COMMAND find "${CMAKE_BINARY_DIR}" -name "__init__.py" -type f -print -delete
+	WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}"
+)
```
TheJJ: the pxdgen stuff once again has to be in `buildsystem/`, not in the library folder

libopenage/build.cpp
```diff
@@ -0,0 +1,28 @@
+// Copyright 2016-2016 the openage authors. See copying.md for legal info.
+
+#include "build.h"
+
+#include <libopenage/config.h>
```
TheJJ: I understand why we need that include in the `<>` way, but still it's ugly.

libopenage/build.cpp
```diff
@@ -0,0 +1,28 @@
...
+	build_src_dir{config::buildsystem_sourcefile_dir},
+	compiler{config::compiler},
+	compiler_flags{config::compiler_flags},
+	cython_version{config::cython_version} {
+
+}
+
+
+build_configuration get_build_configuration() {
+	build_configuration config;
```
TheJJ: This must be static, as it's some compiletime stuff. We could just place all this as a struct in `config.cpp`,
so the things you copy from `config::` could be just a static struct in `config::`.

the `get_build_configuration` can then simply return that struct for py or whatever.

openage/assets.py
```diff
@@ -35,13 +35,23 @@ def get_assets(args):
     if args.asset_dir:
         return Directory(args.asset_dir).root
 
+    try:
+        from .build import get_build_configuration
+    except ImportError:
+        build_src_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+        global_asset_dir = '/usr/local/share/openage'
```
TheJJ: not sure if hardcoding it any way is a good idea. should always be provided by the build system, think of windows, nixos and whatever distro tries using us.

openage/build.pyx
```diff
@@ -0,0 +1,36 @@
...
+    cdef build_configuration *thisptr
+
+    def __cinit__(self):
+        self.thisptr = new build_configuration()
+
+    def __dealloc__(self):
+        del self.thisptr
+
+    property global_asset_dir:
+        def __get__(self): return self.thisptr.global_asset_dir.decode()
```
TheJJ: can't you use

```py
@property
def global_asset_dir(self):
    return self.thisptr.global_asset_dir.decode()
```
?

openage/config.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2014-2016 the openage authors. See copying.md for legal info.
+
+"""
+Project configuration
+"""
+
+import site
+
+
+DEVMODE = not any([__file__.startswith(path) for path in site.getsitepackages()])
```
TheJJ: Cool idea :) We should `normpath()` the `path` and the `__file__` to have a safe prefix check.

setup.py
```diff
@@ -0,0 +1,128 @@
+import os
+import sys
+import pkg_resources
+from setuptools import setup, find_packages, Distribution as _Distribution, Extension
```
TheJJ: That line is kinda ambigous. And I don't like `_` pre/suffixes.

TheJJ:
i'm just laughing really hard because there was a time (7ad1e94e8b409fb8f621b6610c82690cedd76892) when we had a `setup.py` file and basically implemented all the stuff you are undoing now :dancers:

Still, I want to have working out of source builds.

If I understand correctly, you purged all python stuff from our cmake code. And the flags etc are preserved because of the `setup.cfg`, which is awesome and apparently solves the problem we had and why we implemented all that stuff in cmake.

* Add setuptools to the dependency list
* Where are the html files for debugging the cython output?
* Is the compilation/pxdgen/bytecodegeneration still only triggered for files needing it?

openage/assets.py
jprashanth: The code in the `ImportError` exception block is run when you try to run the code from the source directory and not the `built image`.

jprashanth:
Hahaha. Did not know that we used `setuptools` before.

- Forgot about the cython annotated `html` files. Will add them.
- The extension checking/building is now handled by `cython` and `setuptools`. `cython` will take of only building/re-building extension if out-of-date. Similarly `setuptools`/`distutils` will take care building py modules. On the other hand, `pxdgen` is still in `cmake`. The pxd files are only generated when their corresponding header files change.

libopenage/build.cpp
jprashanth: Any ideas

openage/build.pyx
jprashanth: I don't think so. 

Source: http://docs.cython.org/src/userguide/extension_types.html#properties

libopenage/build.cpp
TheJJ: not really. we need it in the binarydir to avoid cluttering the binarydir.
we could add the `libopenage/` to `-I`, to then include "config.h" though.

jprashanth: > we could add the libopenage/ to -I, to then include "config.h", though.
This was my initial attempt, but libopenage/config.h looked more explicit and easy to understand.

TheJJ: indeed, so let's leave it this way.

buildsystem/pxdgen.py
```diff
@@ -366,42 +352,23 @@ def parse_args():
...
-    return args
+    return cli.parse_args()
 
 
 def main():
     """ CLI entry point """
     args = parse_args()
     cppdir = CWD + "/libopenage"
 
-    for filename in args.all_files:
```
TheJJ: We added support for multi file conversion because booting up the python VM for each file is very slow. Batch support is much faster for many files.

buildsystem/python.cmake
```diff
@@ -1,381 +1,52 @@
...
-			string(REGEX REPLACE "/" "_" TARGETNAME "${TARGETNAME}")
-
-			# generate the pretty module name
-			file(RELATIVE_PATH PRETTY_MODULE_NAME "${CMAKE_SOURCE_DIR}" "${source}")
-			string(REGEX REPLACE "\\.pyx?$" "" PRETTY_MODULE_NAME "${PRETTY_MODULE_NAME}")
-			string(REGEX REPLACE "/" "." PRETTY_MODULE_NAME "${PRETTY_MODULE_NAME}")
-			set(PRETTY_MODULE_PROPERTIES "")
-
-			if(EMBED_NEXT)
-				set(PRETTY_MODULE_PROPERTIES "${PRETTY_MODULE_PROPERTIES} [embedded interpreter]")
```
TheJJ: How do we handle interpreter embedding now? We used it for booting up the VM and do the memory sanitizing.

buildsystem/pxdgen.py
jprashanth: The single file approach ensures (with the help of cmake) that a pxd file is only regenerated if it's source header file has changed.

buildsystem/python.cmake
jprashanth: I am not handling embedding as of now (haven't figured it out), as it is not used anywhere except for run executable.

buildsystem/pxdgen.py
TheJJ: I understand, still the previous solution did the same thing, but with only one python invocation. Not sure if the overhead is worth the trouble though.

TheJJ:
Is is possible to just address the out-of-source stuff first? Internal buildsystem improvements can come later, addidionally, but are much harder to accomplish without loosing features.

I can imagine that it's not that complicated achieving an out of source build with the current approach with building all python components with `cmake` and not `distutils/setuptools`.

setup.py
```diff
@@ -94,16 +94,11 @@ def build_extension(self, extension):
     OpenageExtension('openage.util.fslike.cpp', sources=['openage/util/fslike/cpp.pyx'])
 ]
 
-def get_version():
-    with open('openage_version', 'r') as fp:
-        version = fp.read().strip()
-        if version.startswith('v'):
-            version = version[1:]
-        return version
+version = __import__('openage').__version__
```
TheJJ: Are you sure this doesn't fail when the project was not compiled yet?
The `openage` import will not succeed if the `libopenage` wasn't built.

jprashanth: version does not depend on build, longversion does. So, the import doesn't fail.

jprashanth:
@TheJJ When I have some free time, i will try to figure out embedding interpreter and creating executable.

castilma:
How much/what is missing?

TheJJ:
There's some changes that we have to think over again:
* Building the project with setuptools is something we undid some time ago: I don't remember why exactly but implementing the python build process in cmake instead was decided to be the better solution.
* DEVMODE definition and activation: right now it activates to use the asset dirs of the repo instead of the installed on. 
* `setuptools` (especially with `pxdgen`) might cause to rebuild the buildsystem more than necessary: currently it causes the minimal possible changes as cmake tracks the modified files. (This is just an assumption and maybe `setuptools` is cooler than I thought)

I'd have to spend some time to again understand the full impact of the proposed changes.

The overall goal is favorable though:
* No file will ever be created in the openage repo folder by the build system

And has these disadvantages:
* IDEs will have a hard time finding the generated files.
* All python files will have to be copied as they are to the dest-folder
* Debugging and launching has to happen in the dest-folder (our makefile can redirect that, though)

Whether it's a good idea to switch to `setuptools`, I can't tell. The proposed solution is better than our previous attempt in using `setup.py` which was then replaced by cmake in 7ad1e94e8b409fb8f621b6610c82690cedd76892.
You see that a lot of effort is needed to tell `setuptools` how to build our `pyx` modules, this was a lot easier in `cmake`.

My suggestion is to not use setuptools and instead try to slim down our cython handling where possible. This means:
* Always copy python files files to the builddir
* Create the generated files in the builddir (`.pyx-cpp`, `gamedata/*`, `config.*`)
* Create the `run` program in the builddir
* Still track and build all python files with cmake

TheJJ:
After digging a bit more, here's another problem with distutils/setuptools:

Currently, cython modules (i.e. compiled `pyx` files) are produced by cmake, which can do `RPATH` magic. That means cmake tracks where a library is located in the source tree, and adds this as default `LD_LIBRARY_PATH`.

Once installed, this `RUNPATH` is cleared from the library by cmake.
(see `readelf -a somecythonlib.so`: `(RUNPATH)            Library runpath: [/home/jj/devel/openage/.bin/clang++-debug-Oauto-sanitize-none/libopenage:]`) and https://cmake.org/Wiki/CMake_RPATH_handling

`distutils` can also add an `RPATH` to libraries it creates with `runtime_library_dirs` (see https://docs.python.org/3.6/distutils/apiref.html). But it can not remove that path upon installation. We would need to relink those libraries to not have the rpath upon installing, but installation is then also handled by `distutils`. Which leads to the conclusion that using cmake for building the python modules for our use case (running the project in the source or build dir via `RPATH`, and stripping `RPATH` from libraries once they're installed (after `make install`)) is the only way to go without much effort in tricking `distutils`.

TheJJ:
Another attempt is in progress at #1045.

PR #572, opened 2016-06-03
Clear the selection after exiting the ActionMode instead of after entering it

### LINKED ISSUE #531: Game crash after loading
To reproduce : 
- Generate a game
- Select all 3 blue villagers
- Ask them to build a house
- Make the 3 villagers build it on the same side (not sure if required)
- Save then quit the game (not openage) before the house is built
- Load the game (only 2 villagers should be present)
- Press 'M' to get back to the building mode, the game should crash

Might be related to #362   

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/32171184-game-crash-after-loading?utm_campaign=plugin&utm_content=tracker%2F371436&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F371436&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ghallak:
Fixes #531.
When ending the game the selection doesn't get cleared, so changing to Action Mode after starting a new game will call `selection.clear()` which tries to access selected units from previous game.
My fix is to call `selection.clear()` when leaving the Action Mode so that the current selection will be cleared before ending the game.

### DIFF
--- a/libopenage/game_control.cpp
+++ b/libopenage/game_control.cpp
@@ -564,13 +564,18 @@ Player* GameControl::get_current_player() const {
 
 void GameControl::set_mode(int mode_index) {
 	if (mode_index != -1) {
-		if (mode_index < std::distance(std::begin(this->modes), std::end(this->modes)) &&this->modes[mode_index]->available()) {
+		if (mode_index < std::distance(std::begin(this->modes), std::end(this->modes))
+		    && this->modes[mode_index]->available()
+		    && this->active_mode_index != mode_index) {
+
 			engine->get_input_manager().remove_context(this->active_mode);
 
-			// set the new active mode
+			// exit from the old mode
 			if (this->active_mode) {
 				this->active_mode->on_exit();
 			}
+
+			// set the new active mode
 			this->active_mode_index = mode_index;
 			this->active_mode = this->modes[mode_index];
 			this->active_mode->on_enter();
--- a/libopenage/game_control.cpp
+++ b/libopenage/game_control.cpp
@@ -36,6 +36,8 @@ std::string CreateMode::name() const {
 
 void CreateMode::on_enter() {}
 
+void CreateMode::on_exit() {}
+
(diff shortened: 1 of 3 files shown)


libopenage/game_control.cpp
```diff
@@ -562,6 +568,9 @@ void GameControl::set_mode(int mode_index) {
 			engine->get_input_manager().remove_context(this->active_mode);
 
 			// set the new active mode
+			if (this->active_mode) {
```
TheJJ: Just for future error-avoidance, can you change this to a check to test if the previous mode is not the new mode?
that way, a non-change in modes doesn't call the exit. This should probably be done up top in that function.

ChipmunkV:
In Dota2, if you switch to main menu and then back into game - you don't lose selection, I believe.

libopenage/game_control.cpp
ghallak: > can you change this to a check to test if the previous mode is not the new mode?
> that way, a non-change in modes doesn't call the exit.

It will not call `on_exit()`, but it will call `on_enter()`.
Maybe I should check if the previous mode is not the new mode in this if statement `if (mode_index < std::distance(std::begin(this->modes), std::end(this->modes)) &&this->modes[mode_index]->available())`. that way, non of these functions will be called if the mode is not being changed.
What do you think ?

TheJJ: Indeed. Go for it :)

TheJJ:
awesome :)

ChipmunkV:
Well, okay, nvm...

ghallak:
@ChipmunkV,  I haven't changed the behavior here, the selection was cleared anyway when entering the ActionMode.
I just cleared the selection when exiting instead of when entering since that will not change the behavior and will fix the bug.

Issue #277, opened 2015-04-14
Make all user interaction (keystrokes, mouse movements, scrolls) recordable/scriptable


franciscod:
This is related to networking and multiplayer, but also could be useful for reproducing bugs.

Thoughts?

TheJJ:
This is probably strongly correlated with creating an "action" subsystem as I suggested in #284.
All user actions are piped through that system and bound actions are executed. The recording can happen in between, the replay as well.

Btw if we're good we can even do gui and gameplay tests on travis that way, although spawning a virtual X screen will most likely be a pain in the ass.

franciscod:
If we manage to record/script user interactions via the action subsystem we
should be able to figure out Xvfb on travis!

mic-e:
This would indeed be great for reproducing bugs, and for automatic testing of in-game stuff.

boombatower:
Would it not be simpler to use existing record/replay systems for testing and such? I understand scripting is a different goal.

TheJJ:
I doubt we can be compatible to the existing game state storage structs, we can make use of the existing ideas and concepts of course.

boombatower:
I meant, input recording/replay tools, not aoe record files and the like.

TheJJ:
And what tools did you think of that could be of use?

janisozaur:
https://en.wikipedia.org/wiki/Xnee

janisozaur:
https://en.wikipedia.org/wiki/Xnee

ChipmunkV:
The replay should consist of: a replay of the 'network events' (what's actually happening in the game) and the replays of every player perspective. So, when viewing a match, you can fly around, use minimap, click on units to examine them... But also you can switch to the perspective of any player and see what he's doing and how the camera moves.

A replay will have a network data track and actions tracks for every player.

**The action system is ready for the implementation of the record/play of the player perspective.**

To do that - record everything in `InputContext::execute_if_bound`, for example.

Note: screen resolutions can be different for the viewer and the player, so the mouse movement should be recorded in the world coordinates. Probably don't expect the replay of the interactions with the UI overlays to be meaningful (clicks on the command buttons or the minimap), just record the action that it triggered.

simonsan:
For collecting thoughts on a replay container format:
https://pad.stusta.de/p/openage-replay-container

Issue #530, opened 2016-03-20
Network communication design


TheJJ:
tl;wr: the server does all calculations, clients only render the state they received. state is transmitted by keyframes at "predicted" end points.

https://pad.stusta.mhn.de/p/openage-communication

janisozaur:
In https://github.com/OpenRCT2/OpenRCT2 we have added multiplayer capability to the game, and included a simple lobby, *master server*, which lists online public games.

At first there was no authentication, but people abused publicly available servers to destroy maps. We have then added ability to password-protect servers and assign users into groups with specified permissions, but it turned out to not be enough, as we could not make those permissions persist. Our plan was to include centralised authorisation server, but this met some substantial resistance, as it was "against the spirit of open source", even when we wanted to publish the code for running your own auth service.

The solution we came up was to utilise OpenSSL's public-key cryptography to generate key on client and have the client sign server-generated token, client then sends signature to the server, gets verified and if that step is successful, the client is granted permissions he was assigned last time.

If you would like to employ similar scheme in your project, you can use https://github.com/OpenRCT2/OpenRCT2/pull/3699 as a reference.

TheJJ:
Thanks for the hints, doesn't sound bad. But this issue is rather about the simple server-client interaction, without all the lobby stuff. I just opened another issue for that: #562. I don't think we'll have a griefing problem, because games are rather short-lived and only spectators should be able to join afterwards. For client authentification, we thought about using GPG, so achievements can be issued to key owners for example. Match-rejoins after a disconnect can also be performed that way securely. But it's not that much a difference to OpenSSL, except maybe licensing fuckup.

janisozaur:
When I was designing the system, I found out about https://keybase.io/, this looks nice and something we could've probably used, but it still in invite-only mode. If you use crypto keys nevertheless, perhaps something worth taking a look at.

timohaas:
I tried to understand the proposed netcode:
[picture](https://img5.picload.org/image/rdgpgdgl/openage.png)
[dia file for editing](http://www.file-upload.net/download-11938929/openage.dia.html)

sources:
[aoe2 netcode](http://www.gamasutra.com/view/feature/131503/1500_archers_on_a_288_network_.php)  we dont want it (cheaters, performance problems)
[our netcode inspiration](https://blog.forrestthewoods.com/the-tech-of-planetary-annihilation-chronocam-292e3d6b169a#.lkgx11hvu)

janisozaur:
[...]

PR #584, opened 2016-07-09
[WIP] Input replay

ChipmunkV:
There is no established world-history protocol, so it's too early for the game replays.

The goal is to be able to replay the player's perspective by replaying the input. It's supposed to be a part of the real replay functionality.

Will do replay of actions and mouse/camera movements; handling switches between different players and 'free mode' during the replays.

Related to #277

Current state: dumps ActionMode events to stderr.


Vtec234:
The curve system from #740 and #530 should address this.

ChipmunkV:
Most of this PR is GUI and input adaptations to actually show the replay.

Also, just to make sure that we understand correctly: this is not a network replay. It's mouse movement, viewport movement (like jumping to units or locations with hotkeys) and action invocation. Should mention this somewhere in #740 or #530 (store and transfer as curves too?).

Vtec234:
It seems that there is no issue on general gameplay replays. It's indeed important to store local graphical input besides the world state in curves if we want to achieve completely accurate replays from the player's perspective. I'm unsure of the format though, curves seem a bit overkill.

PR #586, opened 2016-07-11
context_extraction osx patch

### LINKED ISSUE #585: Linking error building on OSX
make results in a linking error.

...
Undefined symbols for architecture x86_64:
  "qtsdl::extract_native_context(SDL_Window*)", referenced from:
      qtsdl::GuiRendererImpl::GuiRendererImpl(SDL_Window*) in gui_renderer_impl.cpp.o
ld: symbol(s) not found for architecture x86_64

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/35907308-linking-error-building-on-osx?utm_campaign=plugin&utm_content=tracker%2F371436&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F371436&utm_medium=issues&utm_source=github).
</bountysource-plugin>

vsimon:
Fixes #585

### DIFF
--- a/copying.md
+++ b/copying.md
@@ -82,6 +82,7 @@ _the openage authors_ are:
 | Sebastian Brodehl           | sbrodehl                    | sbrodehl@students.uni-mainz.de        |
 | Gaith Hallak                | ghallak                     | gaithhallak@gmail.com                 |
 | Pierre Hallot               | Hallot                      | hallotpierre@gmail.com                |
+| Vicken Simonian             | vsimon                      | vsimon@gmail.com                      |
 
 If you're a first-time commiter, add yourself to the above list. This is not
 just for legal reasons, but also to keep an overview of all those nicknames.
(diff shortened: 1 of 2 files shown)


ChipmunkV:
No way. And it works after that?

vsimon:
The issue was about getting osx compiling without errors again. This does
that.

As far as osx "working" at runtime, I think there are other problems
separate from the compiling problems to be able to say 'everything works'.

vsimon:
> And it renders the GUI correctly?

It doesn't get past the asset in extract_native_context: `assert(current_context);`. Again that should be a separate issue.

ChipmunkV:
Yes, that function was written in the blind mode. Good that it compiles now.

Maybe `[NSOpenGLContext currentContext]` instead of `[NSGraphicsContext currentContext]` will get it past the assert.

vsimon:
Thanks, I'll give that a shot. In the meantime, filed Issue #588.

PR #591, opened 2016-07-24
Selection box actions

ChipmunkV:
Don't test if the mouse button is down when dragging the selection box. Use new actions `BEGIN_SELECTION` and `END_SELECTION` instead.

To do that, add new `MOUSE_BUTTON_UP` and `MOUSE_BUTTON_DOWN` events to have them separately from the mouse clicks. Bind them to the selection actions.

Needed by #584, so during the input replay the `ActionMode` doesn't try to access the main `InputManager` by doing `engine.get_input_manager().is_down(input::event_class::MOUSE_BUTTON, 1)`.


TheJJ:
awesome!

PR #595, opened 2016-07-28
Render the GUI into an FBO

ChipmunkV:
Renders GUI to an FBO while using a separate context.

Using same context is slow at least on Linux.

- [x] render it
- [x] fix blending/filtering
- [x] API to get the texture ID to the main application
- [x] use VBO


TheJJ:
whee, now with one context and a separate gui fbo!

Longhanks:
Maybe this will fix crashing on OS X. Looking forward to test this when I get home!

ChipmunkV:
@Longhanks, no, it won't.

PR #597, opened 2016-07-30
GL debug option

ChipmunkV:
A `--gl-debug` option that creates a debug context instead of a normal one. Gets a more precise backtrace when error occurs.


TheJJ:
looks very good, thanks :)

PR #601, opened 2016-08-16
config: rebind STOP_GAME to Shift+ESCAPE

jazzpi:
ESC should just be the key to cancel things like building placement etc., pressing it to find the game closed is annoying 😄

### DIFF
--- a/cfg/keybinds.oac
+++ b/cfg/keybinds.oac
@@ -2,7 +2,7 @@
 
 set TOGGLE_CONSOLE `
 set START_GAME Return
-set STOP_GAME Escape
+set STOP_GAME Shift Escape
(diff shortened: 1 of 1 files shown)


franciscod:
silly question: why do we need a keystroke for this one? why not let the user do Alt+F4 or whatever their "close this window for me pls" combo is on their system?

zuntrax:
Having "quit game" on Shift+Esc won't hurt us for now imo. We can still unbind it later.

jazzpi:
I think having a way to close the game from within the game (without help of the WM or whatever) just makes the game feel more polished and is nice if you want to benchmark it by running it in pure X without any WM.

PR #602, opened 2016-08-19
noexcept annotation for all move constructors/assignment operators

### LINKED ISSUE #599: noexcept annotation for all move constructors/assignment ops
Annotate all move constructors/assignment operators should be annotated `noexcept`, because stl containers will not move the values when reallocating otherwise.

Instead, they will copy the values to ensure a valid state when the move might except.

-> Go through the whole code base (`grep` or `ag` for `&&`) and annotate all constructors/functions that actually don't except with the `noexcept` keyword.

http://en.cppreference.com/w/cpp/language/noexcept_spec

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/36659913-noexcept-annotation-for-all-move-constructors-assignment-ops?utm_campaign=plugin&utm_content=tracker%2F371436&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F371436&utm_medium=issues&utm_source=github).
</bountysource-plugin>

kev946:
Fixes #599

### DIFF
--- a/libopenage/util/stringformatter.h
+++ b/libopenage/util/stringformatter.h
@@ -257,18 +257,18 @@ public:
 		StringFormatter<FString>{this->buffer},
 		buffer{other} {}
 
-	FString(std::string &&other)
+	FString(std::string &&other) noexcept
 		:
 		StringFormatter<FString>{this->buffer},
-		buffer{other} {}
+		buffer{std::move(other)} {}
 
 	FString &operator =(const std::string &other) {
 		this->buffer = other;
 		return *this;
 	}
 
-	FString &operator =(std::string &&other) {
-		this->buffer = other;
+	FString &operator =(std::string &&other) noexcept {
+		this->buffer = std::move(other);
 		return *this;
 	}
 
--- a/libopenage/engine.h
+++ b/libopenage/engine.h
@@ -149,7 +149,7 @@ private:
 	/**
 	 * engine move operator.
 	 */
-	Engine &operator=(Engine &&other);
+	Engine &operator=(Engine &&other) noexcept;
 
 	// log sinks
 	std::unique_ptr<log::FileSink> logsink_file;
--- a/libopenage/gui/guisys/private/gui_subtree_impl.cpp
+++ b/libopenage/gui/guisys/private/gui_subtree_impl.cpp
@@ -195,7 +195,7 @@ void GuiEngineImplConnection::disconnect() {
 	}
 }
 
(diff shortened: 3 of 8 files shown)


libopenage/pyinterface/functional.h
```diff
@@ -63,7 +63,7 @@ class Func {
 
 	// for construction from lambdas and other callables (from C++).
 	template<typename F>
-	Func(F &&f) {
+	Func(F &&f) noexcept {
 		this->fptr = f;
```
iNaKoll: This will copy the function object and may actually throw. Note that strictly speaking this is not a move constructor.

libopenage/pyinterface/functional.h
```diff
@@ -77,13 +77,13 @@ class Func {
...
 	}
 
-	Func(std::function<ReturnType (ArgTypes ...)> &&f) {
+	Func(std::function<ReturnType (ArgTypes ...)> &&f) noexcept {
 		this->fptr = f;
```
iNaKoll: This may throw as well. This is the same thing bellow for Func operator=

libopenage/util/stringformatter.h
```diff
@@ -257,7 +257,7 @@ class FString : public StringFormatter<FString> {
 		StringFormatter<FString>{this->buffer},
 		buffer{other} {}
 
-	FString(std::string &&other)
+	FString(std::string &&other) noexcept
 		:
 		StringFormatter<FString>{this->buffer},
 		buffer{other} {}
```
iNaKoll: This will not move the string. Use` buffer{std::move(other)} ` instead.

libopenage/util/stringformatter.h
```diff
@@ -267,7 +267,7 @@ class FString : public StringFormatter<FString> {
 		return *this;
 	}
 
-	FString &operator =(std::string &&other) {
+	FString &operator =(std::string &&other) noexcept {
 		this->buffer = other;
```
iNaKoll: This will not move the string. Use `this->buffer = std::move(other) ` instead.

TheJJ:
All the `... noexcept = delete;` additions are not necessary, as these move constructors can't be used anyway.

kev946:
@TheJJ made those changes

TheJJ:
Thanks :) Why did you close it, I assume just accidentially?

PR #559, opened 2016-05-28
[WIP] Minimap, actions 

ChipmunkV:
Made some action buttons (hooked up 'kill' and 'garrison', haven't found the game logic for other ones).

Borrowed https://github.com/SFTtech/openage/pull/344 for the minimap. But can't understand why the player colours aren't bright. The player palette has only dark shades...

PR #605, opened 2016-08-24
Build Menus

jazzpi:
This adds the civilian and military build menus. Opening them is done via BUILD_MENU and BUILD_MENU_MIL, mapped to Q and W respectively. Opening them has no feedback except that one of two input contexts is registered that have the buildings mapped.

Issues I'd like to fix before merging:

- [x] Farms and docks don't work
- [x] Pressing M while in the build menu switches modes
- [x] There is no feedback when the menu is opened
- [x] The menus aren't accessible via the HUD


jazzpi:
The second issue is caused by events propagating upwards in the input contexts which means that all the binds that aren't overwritten by the build menu still work, so V for example also works. I'm not really sure how to fix that - Adding a propagate switch to the `InputManager` isn't enough since then for example the `ActionMode` doesn't get mouse movement events anymore so it can't render the building properly anymore.

Of course you could then catch mouse movement events in the build menu context and send them to the `ActionMode`, but that feels like a whole lot of redundant code... Any ideas?

jazzpi:
IIRC @ChipmunkV had some action buttons in #559 before. I guess using them to make the build menu accessible and then replace them with the building icons would fix points 3 and 4, but I can't find them anymore.

ChipmunkV:
I've made a branch with changes that were in that PR: https://github.com/ChipmunkV/openage/tree/actions

The implementation isn't too deep: these are just buttons, and no new mode added.

jazzpi:
Thanks, I'll have a look at it.

Farms do actually build but the texture isn't displayed while hovering and when they build they are dead, but I don't really know how to fix that. The dock now works by using unit ID 47 (DOCK3) instead of 45 (DOCK).

jazzpi:
Pressing M still changes modes, but the build menus get closed and if we're currently placing a building, that gets cancelled as well. That's good enough for me. I would however like to add some tests, since that seems like a good idea 😆

jazzpi:
I don't think I'll add tests since basically everything worth testing requires the engine to be set up and I don't really want to do that...

ChipmunkV:
Usually it's not recommended to call QML code from C++ (`emit_actions_ready()` and `set_icons_source()` cases).

There were some hacks for `emit_actions_ready()` situation, but there are better hacks that are based on the fact that models can take care of themselves without additional signals that tell that they are really changed:
https://github.com/ChipmunkV/openage/commit/1153cb43c7d1e029fb2e61e555b778dc8d9438e6

For the `set_icons_source()` situation I'd have just a variable of `ActionButtonsType` enum in C++ to control what kind of buttons to show. The data can be placed in `MilitaryUnitsListModel.qml`, `CivilianUnitsListModel.qml`, etc. So there would be:

    ListModel {
        property url iconsSource: "..."

        ListItem {
            ...
        }
        ...
    }

Then QML can use switch on that enum to choose the right model.

At least it won't be in C++. As I understand, there is no way to generate that from the gamedata.

jazzpi:
Good points. Qt is still a bit of a black box for me, so I tried changing as little about the existing code there as possible. I merged your changes now though and moved `iconsSource`.

TheJJ:
Whoa awesome!

libopenage/game_control.cpp
```diff
@@ -100,30 +118,102 @@ ActionMode::ActionMode(qtsdl::GuiItemLink *gui_link)
...
+				auto player = this->game_control->get_current_player();
+				this->type_focus = player->get_type(building);
+				if (&engine.get_input_manager().get_top_context() != &this->building_context) {
+					engine.get_input_manager().remove_context(&this->build_menu_mil_context);
+					engine.get_input_manager().register_context(&this->building_context);
+					this->announce_buttons_type();
+				}
+			}
+		});
+	};
```
TheJJ: this seems kinda redundant with the build_menu_context binding, maybe you have an idea how we can unite both?

libopenage/game_control.cpp
```diff
@@ -41,13 +41,31 @@ void CreateMode::on_exit() {}
...
+	action_mode(action_mode) {
+}
+
+void ActionModeSignals::on_action(const std::string &action_name) {
+	Engine &engine = Engine::get();
+	input::action_t action = engine.get_action_manager().get(action_name);
+	input::InputContext *top_ctxt = &engine.get_input_manager().get_top_context();
+	if (top_ctxt == this->action_mode || top_ctxt == &this->action_mode->building_context ||
+	    top_ctxt == &this->action_mode->build_menu_context || top_ctxt == &this->action_mode->build_menu_mil_context) {
+		using namespace input;
```
jazzpi: I swear I stole this from somewhere else but I can't find it right now - I believe the issue was that it would be `input::action_arg_t action_arg{input::Event{input::event_class::ANY, 0, input::modset_t{}}`... which is a tad verbose. Maybe there should be a function under openage::input that creates an `action_arg_t` like this?

jazzpi: Found it, I stole it from [this commit](https://github.com/jazzpi/openage/commit/3fb34d74fc74bec4efee0c23be2e978fc70bbbb8#diff-3cd26de8330566a41cadb63f7de53cccR350).

ChipmunkV: That commit wasn't reviewed.

jazzpi: I know, and it's been fixed in https://github.com/SFTtech/openage/pull/605/commits/8b489bac5ba4704fc967c4a6fca7df5b79bb5744

PR #606, opened 2016-09-05
Limit building completion to 1.0

jazzpi:
Currently, when completing a building it will usually display as a little bit more than 1.0 completed. This just sets the `completed` property to `1.0f` when the building is fully completed.


zuntrax:
Looks good.

PR #612, opened 2016-09-26
doc: Add section about range to keep building

jazzpi:
This might be related to the range for various other tasks (like keeping
chopping etc.), but I haven't tested that yet.


TheJJ:
Nice! If you have time, would be nice if you figured out if that was used in other task-searching situations too.

PR #620, opened 2016-10-02
[WIP] add quit button in menu mode

Shalzz:
fixes #616 

Just, can't seem to find the correct sender.

PR #621, opened 2016-10-03
[WIP]add quit button

marcobarbosa16:
Button is not yet closing the application.


TheJJ:
Thank you for your attempt, but we want to shut down the engine in a more clean way. See the comments in #620.

PR #627, opened 2016-10-05
Keep building after finishing a building

jazzpi:
When a unit's current action is to build a building and that building
completes, it will now search for the closest building and then queue
building that.

~~Currently, there are no range limitations as explained in
doc/reverse_engineering/building_placement.md and it's horribly
inefficient (it searches through ALL units on the map).~~

### DIFF
--- a/libopenage/terrain/terrain_search.cpp
+++ b/libopenage/terrain/terrain_search.cpp
@@ -9,8 +9,8 @@
 namespace openage {
 
 TerrainObject *find_near(const TerrainObject &start,
-                                         std::function<bool(const TerrainObject &)> found,
-                                         unsigned int search_limit) {
+                         std::function<bool(const TerrainObject &)> found,
+                         unsigned int search_limit) {
 
 	auto terrain = start.get_terrain();
 	auto tile = start.pos.draw.to_tile3().to_tile();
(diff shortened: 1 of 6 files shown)


jazzpi:
Should be done!

libopenage/terrain/terrain_search.cpp
```diff
@@ -52,6 +75,7 @@ void TerrainSearch::reset() {
 	std::swap(this->tiles, empty);
 	this->visited.clear();
 	this->tiles.push(this->start);
+	this->visited.insert(this->start);
```
TheJJ: why do we have to store the start in the visited list if the search is reset?

libopenage/unit/action.cpp
```diff
@@ -803,6 +803,31 @@ void BuildAction::update_in_range(unsigned int time, Unit *target_unit) {
...
+	auto valid = [this](const TerrainObject &obj) {
+		if (!obj.unit.has_attribute(attr_type::building) ||
+		    obj.unit.get_attribute<attr_type::building>().completed >= 1.0f) {
+			return false;
+		}
+		this->entity->log(MSG(dbg) << "Found unit " << obj.unit.logsource_name());
+		return true;
+	};
+
+	TerrainObject *new_target = find_in_radius(*this->entity->location, valid, 9.0f);
```
TheJJ: would be cool if you moved out the `9.0f` as a constexpr `float search_tile_distance` or something so we find it easier when replacing it someday.

libopenage/terrain/terrain_search.cpp
jazzpi: Because next_tile only adds tiles to visited when it adds them to this->tiles, so it will add the start tile again on its second run. So we'd check the start tile twice which is a) kind of useless and b) makes it more difficult to detect when the search was reset.

Now that I'm thinking about it, I'm not sure if it even makes any sense to have this->tiles and this>visited since they seem to always contain the same elements.

TheJJ:
To pass the build you need to rebase, please. there were some changes regarding the python codestyle checker, it was renamed from pep8 to pycodestyle upstream.

jazzpi:
Done.

TheJJ:
Did you think about the terrain search visited list again? It doesn't seem right to add that node again.

TheJJ:
After a short investigation i'd say your fix is correct, the start tile is "visited" on reset (which is called to initially fill the queue of tiles to visit).

TheJJ:
(The build continuing worked as expected btw, thanks man :)

PR #633, opened 2016-10-08
noexcept annotation for all move constructors/assignment operators

### LINKED ISSUE #599: noexcept annotation for all move constructors/assignment ops

kev946:
Fixes #599

### DIFF
--- a/libopenage/engine.h
+++ b/libopenage/engine.h
@@ -149,7 +149,7 @@ private:
 	/**
 	 * engine move operator.
 	 */
-	Engine &operator=(Engine &&other);
+	Engine &operator=(Engine &&other) = delete;
 
 	// log sinks
 	std::unique_ptr<log::FileSink> logsink_file;
--- a/libopenage/gui/guisys/private/gui_subtree_impl.cpp
+++ b/libopenage/gui/guisys/private/gui_subtree_impl.cpp
@@ -195,7 +195,7 @@ void GuiEngineImplConnection::disconnect() {
 	}
 }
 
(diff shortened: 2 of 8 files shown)


libopenage/engine.h
```diff
@@ -149,7 +149,7 @@ class Engine : public ResizeHandler, public options::OptionNode {
 	/**
 	 * engine move operator.
 	 */
-	Engine &operator=(Engine &&other);
+	Engine &operator=(Engine &&other) noexcept;
```
TheJJ: when I look at it, i can't find an implementation of this method, so you can remove the `noexcept` and `= delete` it as well.

kev946: @TheJJ Sure. Sorry this is taking so long, new to modern C++ features

Issue #632, opened 2016-10-07
Asset pack format


TheJJ:
We need a convention on how a asset pack looks like.

Q: What is an asset pack?
A: A collection of files that provide the engine with data to display.

Q: What is in an asset pack?
A: Some files, proposal:
- root mod metainformation file ("pack.nfo"?)
  - contains name, author, version of the mod (`name =`, `author =`, `version =`)
  - dependency information (`conflicts =`, `requires =`, `replaces =`)
  - nyan file name to load (`load = index.nyan`)
  - nyan object name of the mod information (which inherited from `Mod`) (`mod = MyMod`)
- mod data files in arbitrary directory hierarchy
  - images
  - nyan files to add content
  - qml files for adding gui components

example mod pack:

```
    only enforced file name: pack.nfo
    /
        pack.nfo <- "load = anothermod.nyan; mod = anothermod.MonsterMod"
        anothermod.nyan <- "import tentaclemonster.monster; MonsterMod(Mod)..."
        tentaclemonster/
            monster.nyan
            moving.png
            raping.png
            fleeing.png
            rape.py <- when a mod registers >=1 py file, it's a "code mod" that displays a big red ! if not signed by openage-authors.
```

---

nfo format idea:
```
name: (identifier) (must equal folder name?) (identifier in oa mod repository)
version: (major.minor.patch)
    
provides: (list of [identifier (= version)]))
conflicts: (list of [identifiers ((>=, <=) version)])
requires: (list of [identifiers ((>=, <=) version)])
    
description: <human-readable>
url: <mod homepage>
license: (license name or url or whatever)
authors: username <mail>, anotherguy <moremail>
    
load: folder/somedata.nyan
mod: folder.somedata.ModObject
```

examples and further ideas:

vanilla/pack.cfg:
```
name: vanilla
version: 1.0.0
provides: base
conflicts: sftflavouredvanilla
description: converted from AoC by openage alpha 1.2.3
url: http://openage.sft.mx
license: proprietary
authors: ensemble studios (via openage.convert)

load: data/main.nyan
mod: data.main.Main
```

villageroverhaul/pack.cfg:
```
modname: villageroverhaul
version: 4.2
requires: base   # you could also write 'requires: vanilla'.
description: makes your villager so much better you'll never want to play stock oa again!
url: http://forum.openage.sft.mx/mods/villager_overhaul

load: villager/mod.nyan
mod: villager.mod.BetterVillager
```

---

== not part of this task, just for your understanding: ==
[...]

PR #640, opened 2016-10-14
[WIP] doc: Nyan types

castilma:
I started a list of all base types that need to be implemented in the engine. Add your thoughts and make changes.

### DIFF
--- a/doc/nyan/base types.md
+++ b/doc/nyan/base types.md
@@ -51,13 +51,6 @@ Resource():
 	name : text
 	icon : file
 
-#do these 4 really belong in the engine?
-Gold(Resource)
-Food
-...
-...
-
-
 Building():
 	name : text
 
--- a/doc/nyan/base types.md
+++ b/doc/nyan/base types.md
@@ -1,0 +1,164 @@
+Base types
+==========
+
+These types are defined by the openage engine. Every nyan object in the game
+eventually inherits from one or more of these types.
+
+Normally you'd store one or more *NyanObject*s in a *.nyan file, but as these
+are defined in the engine and are here only for humans (like me, *beep blop*),
+they are stored here for an overview.
+
+Currently, this is a **proposal**.
+This file does not perfectly (if at all) represent the already defined types in
+libopenage, but hopefully will in the future. Many classes here, if not all,
+are currently work in progress. Prepare for changes.
+
+Shoutout:
+* Everyone: Comment! Give us good ideas.
+* libopenage-devs: change/erase this. You are the ones, who (are going to)
+implement this and know what's necessary and what should be where.
+
+```
+Mod():
+	# All patches in the 'patches' member are added on load time. So every mod
+	# needs to inherit from this class and new patches to 'patches'.
+
+	name : text
+	patches : set(NyanPatch) # Add your mod's patches to this member.
+
+Tech():
+	# Every patch that needs to be evaluated after a specific event in the game
+	# (e.g. research) needs to be added to 'updates'.
+
+	name : text              # ingame name
+	description : text       # ingame description
+	updates : set(NyanPatch) # Add your new researches (or other patches) here.
+
+#Proposal: call a `Tech' `Update' in nyan, and define Tech in openage like this:
+Update():
+	updates : set(NyanPatch)
+
+HUDElement():
+	#makes an element on the left of the HUD (building or unit menu)
+	name : text
+	description : text
+	icon: text
+Tech(Update, HUDElement):
+
+
+
+Resource():
+	name : text
+	icon : file
+
+#do these 4 really belong in the engine?
+Gold(Resource)
+Food
+...
+...
+
+
+Building():
+	name : text
+
+Ability(): # WIP
+	# Every unit (WIP: or even buildings, see 'CarryAbility',castles... ), that can do something, has an 'Ability'.
+	# WIP: what members belong here?
+
+HUDAbility(Ability): # WIP
+	# Creates a button on the left of the HUD. For Trebuchet packing and unpacking, unit patrol, formations.
+	name : text
+	description : text
+MoveAbility(Ability):
+	# You want to move? Use this class.
+
+	speed : float
+	instant : bool = False # WIP: how about checking (speed < 0) ⇒ instant before
+	                       # path searching?
+	range : float
+
+CooldownAbility(Ability):
+	# Any executed 'Ability', that inherits from this, can only be executed again
+	# after recharge_time in seconds.
+	recharge_time : float
+
+GatherAbility(Ability): # WIP
+	# Gathering resources requires this Ability
+	# WIP1: do we need a member for the current amount and type of resource here? or is that only inside the engine?
+
+	allowedResource : set(Resource) # Allowed ressources?
+	maxAmount : int # maximum amount  that can be carried
+	speed : float # gathered units per second
+
+	# WIP2: proposal: allow different maximum amounts for different resources:
+
+#	allowedResource : orderedset(Resource)
+#	maxAmount : orderedSet(int) # does <10,5,0,10> work? it is a set after all.
+
+	# WIP3: if WIP2 is not possible: implement an 'Ability' that allows to gather only one resource.
+	# A worker would then need 4 gather abilities added to it's 'abilities' set.
+	# And this would allow to reference
+
+	# WIP4: how do we reference Animations?
+#	animation : file # one picture that contains all frames or, if a directory,
+	                # contains all frames. # WIP5: naming scheme?
+#	renameMe : file # WIP6: how to read `animation'? describes, where in the big picture `animation' which frames are; whether to potentially mirror them.
+	                # possibilities: 1. $renameMe is a regular file: description of `animation' (syntax?)
+	                # 2. replace `renameMe' by more fields that describe `animation' (ugly imo, but maybe the best way)
+	                # 3. remove `renameMe':
+	                #    if $animation is a directory: look for '$animation/animate.nfo'
+	                #    else: look for "$(basename $animation png)nfo"
+
+CarryAbility(Ability): # WIP: find a better name?
+	# Used by priests for relicts.
+	# WIP1: how should this look like? keep it general? can maybe be used by buildings to let units find shelter in them? or by mods for adding new relict-like objects
+	# WIP2: can maybe merged with a modified GatherAbility?
+
+#	goodMemberName : Portable # what can be collected? make a 'Portable' class and let relict inherit that? could then be also used by resources?
+#	maxAmount
+
+	animation : file # what does a priest look like when he's carrying it?
+	                 # WIP: see GatherAbility.animation and nyan question 4 (further down).
+
+AttackAbility(Ability):
+	# Every attack shall inherit from this.
+	# WIP: how will damage calculation work in oa?
+
+	range : float
+	rate : float
+	damage : float # rename to baseDamage ?
+
+
+Unit(): # WIP
+	# Everything you can see in the game # (WIP: but not assets, which is basically everything you see. Maybe better:)
+	# Everything that has some 'Ability' inherits from this.
+
+	ability : set(Ability)
+
+
+
+#and probably more. if you know what's missing, add it.
+
+```
+
+Open nyan implementation questions:
+
+1. How do we say in Nyan, that a Towncenter attack depends on how many people are inside the building?
+2. What about the building 'Ability' to carry units? will this lead to recursive references?
+3. How do we reference hooks from Nyan/create new hooks in mods?
+
+  -> If the {Towncenter,Castle} has certain damage, contained units need to leave.
+  
+  -> A relict in a Church leads to increasing gold.
+4. How do we add new animations?
+
+  -> How would a mod add a new relict-like object that can be carried (the animation)? can we describe composition of frames? How is it done by the original game?
+  
+  -> We need a unified way/member-name to add a new animation to an 'Ability'
+  
+  -> animations are stored in one big picture (or allow one file per frame?). How do we tell where in the picture is which frame? How can you tell to mirror a frame?
+5. Of what type are the ingame objects, that don't move and give ressources? goldmine, fish, dead deer? How are deer and dead deer connected?
+
+Mod questions:
+
+1. Adding new units/buildungs: There are only blue units in the original assets. How does the engine now, what part of a frame needs new color?
--- a/doc/nyan/engine types.md
+++ b/doc/nyan/engine types.md
@@ -1,5 +1,5 @@
-Base types
-==========
+Engine types
+============
 
 These types are defined by the openage engine. Every nyan object in the game
 eventually inherits from one or more of these types.
--- a/doc/nyan/engine types.md
+++ b/doc/nyan/engine types.md
@@ -46,23 +46,50 @@ HUDElement():
 Tech(Update, HUDElement):
 
 
+Unit(): # WIP
+	# Everything that has a location ingame.
+
+	hp : float
 
 Resource():
 	name : text
 	icon : file
 
-Building():
+ResourceSpot():
+	# Anything with 'resource' in GatherAbility.allowedResource can gather this.
+	# WIP: since this has a location ingame, make this inherit from Unit? or make e.g. Goldmine(Unit, ResourceSpot)?
+
+	resource : Resource
+	amount : int # initial amount
+	droprate : float # in units per second.
+
+
+Building(Unit): # WIP: make another parent Object for Building?
 	name : text
+
+
+Animation(): # or Sprite() ?
+	# Everything that needs to be animated inherits from this class.
+
+	image : file # one picture that contains all frames or, if a directory,
+                 # contains all frames. # WIP5: naming scheme?
+
+#	metainfo : file # WIP6: how to read `animation'? describes, where in the big picture `animation' which frames are; whether to potentially mirror them.
+	                # possibilities: 1. $renameMe is a regular file: description of `animation' (syntax?)
+	                # 2. replace `renameMe' by more fields that describe `animation' (ugly imo, but maybe the best way)
+	                # 3. remove `renameMe':
+	                #    if $animation is a directory: look for '$animation/animate.nfo'
+	                #    else: look for "$(basename $animation png)nfo"
 
 Ability(): # WIP
 	# Every unit (WIP: or even buildings, see 'CarryAbility',castles... ), that can do something, has an 'Ability'.
 	# WIP: what members belong here?
 
-HUDAbility(Ability): # WIP
-	# Creates a button on the left of the HUD. For Trebuchet packing and unpacking, unit patrol, formations.
-	name : text
-	description : text
-MoveAbility(Ability):
+HUDAbility(Ability, HUDElement): # WIP
+	# Creates a button on the left of the HUD. For Trebuchet packing and
+	# unpacking, unloading units or relics. Maybe even unit patrol andformations.
+
+MoveAbility(Ability, Animation):
 	# You want to move? Use this class.
 
 	speed : float
@@ -75,9 +102,8 @@ CooldownAbility(Ability):
 	# after recharge_time in seconds.
 	recharge_time : float
 
-GatherAbility(Ability): # WIP
+GatherAbility(Ability, Animation): # WIP
 	# Gathering resources requires this Ability
-	# WIP1: do we need a member for the current amount and type of resource here? or is that only inside the engine?
 
 	allowedResource : set(Resource) # Allowed ressources?
 	maxAmount : int # maximum amount  that can be carried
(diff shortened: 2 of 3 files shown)


doc/nyan/base types.md
```diff
@@ -0,0 +1,164 @@
...
+
+Resource():
+	name : text
+	icon : file
+
+#do these 4 really belong in the engine?
+Gold(Resource)
+Food
+...
+...
```
TheJJ: no, they're defined by the base data pack (aka the original game conversion) :)

doc/nyan/base types.md
```diff
@@ -0,0 +1,164 @@
...
+
+
+#and probably more. if you know what's missing, add it.
+
+```
+
+Open nyan implementation questions:
+
+1. How do we say in Nyan, that a Towncenter attack depends on how many people are inside the building?
+2. What about the building 'Ability' to carry units? will this lead to recursive references?
```
TheJJ: References can be recursive of course, they can point to each other. But the carry ability would also be implemented in the engine i think, in a way that if the unit has the `CarryAbility`, nyan doesn't know about what the unit actually carries, but nyan only tells us it can carry `20 weight units, and that one unit type elephant will consume 5 weight units`. That means each unit must define its weight :).
The engine then checks for each unit to board if `current_weight + weight_of_new_unit_to_board >= max_weight` and decides by that if it can board.

doc/nyan/base types.md
```diff
@@ -0,0 +1,164 @@
...
+
+
+
+#and probably more. if you know what's missing, add it.
+
+```
+
+Open nyan implementation questions:
+
+1. How do we say in Nyan, that a Towncenter attack depends on how many people are inside the building?
```
TheJJ: The possibility of attacking buildings must also be implemented in the engine. Nyan provides the information `n units (of size x?) can garrison in the building`. Then those people can either regularly attack from there or the engine multiplies the attack damage for the building.

doc/nyan/base types.md
```diff
@@ -0,0 +1,164 @@
...
+2. What about the building 'Ability' to carry units? will this lead to recursive references?
+3. How do we reference hooks from Nyan/create new hooks in mods?
+
+  -> If the {Towncenter,Castle} has certain damage, contained units need to leave.
+  
+  -> A relict in a Church leads to increasing gold.
```
TheJJ: The damage info is a combination of nyan and the engine. The building itself must define a collapse threshold (in percent maybe) which the engine respects to send units out.

The relic gold is purely implemented in the engine which provides a "ResourceGenerator" object, which the relic inherits (`Relic(ResourceGenerator)`) and adds to the `set(resource_type) += {Gold}` and `rate = 10`. Nyan is not aware of the game mechanics, this is always the job of the engine (that was by the way the key enlightment i finally had to get a sufficient nyan design).

doc/nyan/base types.md
```diff
@@ -0,0 +1,164 @@
...
+* libopenage-devs: change/erase this. You are the ones, who (are going to)
+implement this and know what's necessary and what should be where.
+
+```
+Mod():
+	# All patches in the 'patches' member are added on load time. So every mod
+	# needs to inherit from this class and new patches to 'patches'.
+
+	name : text
+	patches : set(NyanPatch) # Add your mod's patches to this member.
```
TheJJ: Yes, that is the idea from the nyan doc currently, but we've come up with an alternative (which i still have to document):

If a mod is defined by instancing a child object of `Mod`, we have to support callback functions for objects if a child of them is created. This is doable, but kinda complicated.

New idea:
All the data (name, dependencies, etc) is defined as part of the mod metainformation, and the set of patches as well. Because then we don't have to implement that child hook and the initial patches can simply be applied by the engine directly. See #632

doc/nyan/base types.md
```diff
@@ -0,0 +1,164 @@
...
+	name : text
+	description : text
+	icon: text
+Tech(Update, HUDElement):
+
+
+
+Resource():
+	name : text
+	icon : file
```
TheJJ: jup, the resource is part of the engine and the data pack can then create arbitrary resources.

doc/nyan/base types.md
```diff
@@ -0,0 +1,164 @@
...
+#Proposal: call a `Tech' `Update' in nyan, and define Tech in openage like this:
+Update():
+	updates : set(NyanPatch)
+
+HUDElement():
+	#makes an element on the left of the HUD (building or unit menu)
+	name : text
+	description : text
+	icon: text
+Tech(Update, HUDElement):
```
TheJJ: This may be a very good idea. I've never really thought about the ui integration, but having another mixin for that may be the way.

doc/nyan/base types.md
```diff
@@ -0,0 +1,164 @@
...
+  -> If the {Towncenter,Castle} has certain damage, contained units need to leave.
+  
+  -> A relict in a Church leads to increasing gold.
+4. How do we add new animations?
+
+  -> How would a mod add a new relict-like object that can be carried (the animation)? can we describe composition of frames? How is it done by the original game?
+  
+  -> We need a unified way/member-name to add a new animation to an 'Ability'
+  
+  -> animations are stored in one big picture (or allow one file per frame?). How do we tell where in the picture is which frame? How can you tell to mirror a frame?
```
TheJJ: In the original game, the relic thing is actually a new unit: "monk with relic". This is kinda stupid because it assumes only monks can carry relics, which is find in age2 but very unsuitable for us.

So we can prepare the combination of a `Carryable` and some `Unit` in the sense that the carryable has to define its size and position, and we can then render the carryable in front of the unit. We can even define workaround replacement graphics that way in the sense that if it's a "known combination", the combination image is used (aka the monk with relic animation). How we model this in nyan I cannot tell right now (and haven't thought about it yet, so very good catch here).


The image metainformation should still be done in the good old `.docx` files as they're currently describing where a subimage is in one image. Things like animation speed, looping etc is defined in nyan though.
The engine will provide a nyan object `Sprite` which has a `file` member and others like the speed, etc. When this sprite is then required to be shown (e.g. because a unit uses it as its image) the engine will handle everything else internally, like parsing the subtexture information from the `.docx`, drawing it etc.

doc/nyan/base types.md
```diff
@@ -0,0 +1,164 @@
...
+5. Of what type are the ingame objects, that don't move and give ressources? goldmine, fish, dead deer? How are deer and dead deer connected?
```
TheJJ: I'd name them `ResourceSpot`, also defined by the engine (as it has to support giving out resources at that point).
Deer is a `Animal` which is a `Unit` i guess (so you can kill it), and if it's dead, like with all units it has to reference its dead form (to have the animations etc). So there is a `DeadDeer(Animal, ResourceSpot)` with `hp = 0` and a resource type and amount. The gathering itself, the amount decay etcetc is done in the engine, just the decay rate is specified by nyan again.

doc/nyan/base types.md
```diff
@@ -0,0 +1,164 @@
...
+
+Mod questions:
+
+1. Adding new units/buildungs: There are only blue units in the original assets. How does the engine now, what part of a frame needs new color?
```
TheJJ: Heh. Tricky thing this. Basically in the original data there is a "magical color" which is then recolored. Our approach is to paint that magical color as blue with an alpha value of `254`. The fragment shader checks if the alpha value is `254` at some point, then looks up the blue tone, then replaces it with the matching yellow or whatever tone, and draws it.

In the original game this was done in assembly, but we process it in opengl with the teamcolor shader.

doc/nyan/base types.md
castilma: https://github.com/SFTtech/nyan/blob/master/doc/nyan.md says: "Defined in the game engine"
Maybe I'll change that.

doc/nyan/base types.md
castilma: we still need a way to keep all the elements in order and allow 'pages'.

doc/nyan/base types.md
```diff
@@ -0,0 +1,164 @@
...
+#	allowedResource : orderedset(Resource)
+#	maxAmount : orderedSet(int) # does <10,5,0,10> work? it is a set after all.
+
+	# WIP3: if WIP2 is not possible: implement an 'Ability' that allows to gather only one resource.
+	# A worker would then need 4 gather abilities added to it's 'abilities' set.
+	# And this would allow to reference
+
+	# WIP4: how do we reference Animations?
+#	animation : file # one picture that contains all frames or, if a directory,
+	                # contains all frames. # WIP5: naming scheme?
```
castilma: Does every ability need an animation? Maybe Ability() should get a member `animation : Animation'.

jazzpi: I don't think so. For example, castles don't have an attack animation. And if training units/researching techs/attack stances etc. also become abilities, neither do they.

castilma: if for example buildings have an ability without animation, they could reference to their static picture. Maybe that's a dirty workaround. The only alternative I see is to include an AnimatedAbility class.

doc/nyan/base types.md
TheJJ: thx

doc/nyan/base types.md
TheJJ: I'd say the animation is a mixin, mainly used in abilities like `Walking(Ability, Animation)`. If that ability becomes active, we can check in the engine if the ability also inherited from `Animation`, then activate and display the data specified there. That way, we can add custom animations for nearly everything we want but do not enforce it and can fall back to the standard walk animation etc.

doc/nyan/base types.md
castilma: we have to tell somehow, what units can go inside: many units can garrison  the castle, but not the towncenter.

castilma: Maybe we can define types of units: infantry, cavalry, ships, siege. this could help with the allowed_targets member in attackability and with the upgrades from research.

doc/nyan/base types.md
castilma: Should I rename the Animation class to Sprite?

doc/nyan/base types.md
TheJJ: Exactly, the original game has those unit types available for us, we can represent it simply in the inheritance hierarchy. (`Huskarl(Cavalry)` e.g.)

doc/nyan/base types.md
TheJJ: I think animation is fine, we can extend it to a 3d model animation in a few years then hehehe.

TheJJ:
Can you rename the file to `engine_types.md` please?

doc/nyan/engine types.md
```diff
@@ -46,23 +46,50 @@ HUDElement():
...
 	icon : file
 
-Building():
+ResourceSpot():
+	# Anything with 'resource' in GatherAbility.allowedResource can gather this.
+	# WIP: since this has a location ingame, make this inherit from Unit? or make e.g. Goldmine(Unit, ResourceSpot)?
+
+	resource : Resource
+	amount : int # initial amount
+	droprate : float # in units per second.
```
TheJJ: this is called "decay" in the original, which i think is better.

TheJJ:
We don't wanna have a file name separated with spaces in the repo please :)

doc/nyan/base types.md
```diff
@@ -0,0 +1,164 @@
...
+	range : float
+	rate : float
+	damage : float # rename to baseDamage ?
+
+
+Unit(): # WIP
+	# Everything you can see in the game # (WIP: but not assets, which is basically everything you see. Maybe better:)
+	# Everything that has some 'Ability' inherits from this.
+
+	ability : set(Ability)
```
TheJJ: I think so, yes. Each `Unit` needs its idle look.

timohaas:
is it possible to model something like: if the building(or all of the same type) is destroyed where something was researched, than the tech is lost and must be researched again?

doc/nyan/base types.md
TheJJ: I'd say the order and paging is done by categories first and addition order. If mods remove elements the layout won't get fucked up and we can also easily add new elements. If we use numbers for positioning things will always conflict. Tough thing, but just storing it via orderedsets is the easiest way i think.

TheJJ:
Yes, but nyan would (if ever) just store the boolean and more information to enable that feature. The revert of the patch (aka undoing the tech) would require inversing patches, which we would have to add to nyan indeed. But the main work for this feature must be done in the engine (namely the "what happends when this building is destroyed).

timohaas:
How will this Stuff be implemented in the Engine? In my opinion, this should be implemented as Triggers:

Trigger:
type: always, once
conditions = expression
actions = [ list of action]
expression: True|False|condition and|or|xor True|False|condition

condition:

- unit x is killed (e.g. king)
- all military building of player x destroyed
- all reliquiens captured
- unit of player x discovers area x
- first blood
- first player researched tech x
- building x destroyed

action:

- change unit attribute to x
- remove tech x
- research tech x
- send message to player/team
- add food/wood to player x
- create unit x
- kill unit x
- victory for team x
- create other trigger
- only limited to your imagination

example: player klicks tech Y research

new trigger 1:
type: once
condition: building x destroyed
action: [remove tech research, delete trigger 2]

new trigger 2:
type: once
condition: research Y finished
action: [ delete trigger 1, update units]

TheJJ:
For non-general behavior modifications the trigger approach is awesome and we should definitively add those for building scenarios and map interactions.

But if we're talking about the general behavior of the game ("researches get lost when building is destroyed") then we should not have to add a trigger for each building x as you wrote, but instead for all buildings that do research. Nevertheless, it has to be implemented in the engine, and what amount of configuration data is stored in nyan for that is up to us.

doc/nyan/base types.md
castilma: I'm not sure why DeadDeer needs to inherit from from Animal. It needs:

0. A dying animation.
1. A rotting animation. After RessourceSpot.amount dropped to zero, play this and hold the last frame for like a minute, then remove this Unit completely.

Between 0 and 1 we need to decide if we want to hold the last frame from 0 or the first from 1. To me it looks like they're the same in the original assets.

My idea: We could add `dying : animation` to Unit. This is played when our unit dies. After that, create the referenced Deadunit(Unit). If it is a RessourceSpot, wait until RessourceSpot.amount dropped to zero.
Then play deadunit.dying (or deadunit.idle), which is the rotting animation.

TheJJ: yes, could also work. I think you're right that all units have a dying animation but many won't have one. also we somehow need to model that the living deer is not a resource spot.

Basically we have multiple possibilities to model this:
* store everything relevant for the dead deer in the living one already, aka `rotting_animation` and `resource_decay` and `resource_type` etcetc, this is the "bad" approach i'd say
* have a separate `DeadDeer` that inherits from `(original deer, resource spot)` and updates graphics etc. the original deer then references to that object as `on_death_become : Unit = DeadDeer`. this requires a forward declaration which will be allowed in the next revision of the nyan design.
* use abilities to test if the deer can even die. that ability then performs the death animation, another ability performs the resource collection, another one the decay.

I think the best approach is having an ability "being hunted" because it's different if a villager or a tribok killed it for the collectible food. That ability is implemented to change the unit type to `DeadDeer` once it is killed. The ability is included in the `Deer` `abilities` set, and also requires a forward declaration. It's the most overkill but also most extensible and awesome approach i'd say.

doc/nyan/engine_types.md
```diff
@@ -50,6 +49,10 @@ Unit(): # WIP
 	# Everything that has a location ingame.
 
 	hp : float
+	idleLook : animation
+	death : Unit # What is the dead form of this Unit?
+	# Proposal:
+#	dying : animation
```
TheJJ: as noted in the Deer-discussion above, not every unit has the ability to die. then we'd have to insert a placeholder or an empty value for `dying`, which we wanna avoid.

doc/nyan/engine_types.md
```diff
@@ -65,6 +68,7 @@ ResourceSpot():
 
 
 Building(Unit): # WIP: make another parent Object for Building?
+	# WIP2: Some buildings consist of a static frame and an animation on top. How do we reference that?
```
TheJJ: Probably an `AnimatedBuilding`, also the building annexes like the towncenter need to be modeled out of parts.
They have an ordered set of sprites (which may be animated) so the draw order is fixed.

doc/nyan/engine_types.md
```diff
@@ -75,9 +79,9 @@ Animation(): # or Sprite() ?
                  # contains all frames. # WIP5: naming scheme?
 
 #	metainfo : file # WIP6: how to read `animation'? describes, where in the big picture `animation' which frames are; whether to potentially mirror them.
-	                # possibilities: 1. $renameMe is a regular file: description of `animation' (syntax?)
-	                # 2. replace `renameMe' by more fields that describe `animation' (ugly imo, but maybe the best way)
-	                # 3. remove `renameMe':
+	                # possibilities: 1. `metainfo' is a regular file: description of `animation' (syntax?)
+	                # 2. replace `metainfo' by more fields that describe `animation' (ugly imo, but maybe the best way)
+	                # 3. remove `metainfo':
```
TheJJ: we can either keep the current `docx` format or add a new one to just contain the frame positions and ids. i wouln't store information about the texture itself in nyan, the implicit reference to `filename.anim` or something might be better. the current `docx` approach is very simple and works well, we should load it on demand though and not on launch.

doc/nyan/engine_types.md
castilma: so, remove `dying` and `death`, right?

doc/nyan/engine_types.md
```diff
@@ -0,0 +1,191 @@
...
+AttackAbility(Ability):
+	# Every attack shall inherit from this.
+	# WIP: how will damage calculation work in oa?
+
+	range : float
+	rate : float
+	damage : float # rename to baseDamage ?
+
+
+HealingAbility(Ability): #WIP
```
MaanooAk: Probably must be split. `Healing`, `Repairing` and `Healing_inside_a_building` must be separate because its have different logic behind it (eg. repair has cost).

`Healing_inside_a_building` could also be megred with `Garison`.

doc/nyan/engine_types.md
```diff
@@ -0,0 +1,191 @@
...
+	# Workers can repair buildings, ships and siege weapons. Buildings heal contained Units, Priests can heal moving units (including siege weapons?)
+	# WIP: priests and workers need an animation for this, buildings not.
+	# Should we write that as an attack with negative
```
MaanooAk: It will always have the same sign, so better positive which can be stored in a unsigned int

doc/nyan/engine_types.md
```diff
@@ -0,0 +1,191 @@
...
+# Alternative:
+Shelter(Ability):
+	# A 'Unit' which might hold other units. Used by Castle, Towncenter, TransporterShip, but not priest.
+
+	allowedUnits : set(UnitTypes) # what can be collected?
+	maxUnits : int
+	maxUnitsize : int # maximum size of each unit. If we want this, every movable unit should have a size member declared by a base class.
+	collapsingRate : float # percentage of health at which no units can be contained any longer.
+
+AttackAbility(Ability):
```
MaanooAk: - Friendly fire
- Min range
- Area of effect
- Consider projectiles

doc/nyan/engine_types.md
```diff
@@ -0,0 +1,191 @@
...
+
+#and probably more. if you know what's missing, add it.
+
+```
+
+Open nyan implementation questions:
+
+4. How do we add new animations?
+
+  -> How would a mod add a new relict-like object that can be carried (the animation)? can we describe composition of frames? How is it done by the original game?
```
MaanooAk: Original game: Two different units: `monk` and  `monk with relic`

doc/nyan/engine_types.md
```diff
@@ -0,0 +1,191 @@
...
+
+HUDElement():
+	#makes an element on the left of the HUD (building or unit menu)
+	name : text
+	description : text
+	icon: text
+Tech(Update, HUDElement):
+
+
+Unit(): # WIP
```
MaanooAk: - Cost
- Have limit (The max number of units of this type that a player can have at an instance)
- Had limit (The max number of units of this type that a player can create)

doc/nyan/engine_types.md
```diff
@@ -0,0 +1,191 @@
...
+
+```
+Mod():
+	# All patches in the 'patches' member are added on load time. So every mod
+	# needs to inherit from this class and new patches to 'patches'.
+
+	name : text
+	patches : set(NyanPatch) # Add your mod's patches to this member.
+
+Tech():
```
MaanooAk: - Cost

doc/nyan/engine_types.md
TheJJ: I'd do it with the ability to carry things, and a relic is a `Carryable`. To solve the asset combination the thing to be carried could be able to provide a overall carrying graphic, so the `monk-with-relic` graphic is activated as soon as it is picked up.

heinezen: > `ResourceProvider` when garrisoned in the monastry

@TheJJ Now that you mention it, I'd even go further than that and regard the relic as some kind of `BonusItem` or `BonusProvider`. While it's true that AoE2 see relics only as another source for gold income, AoM's relics also provide boni like cheaper unit creation cost, faster research time. RoN has something similar, where rare resources on the map provide a bonus to the player who places a merchant on it (and rare resources can be controlled by multiple people). I think modders would also get very creative with this construct.

doc/nyan/engine_types.md
castilma: what type should cost be?
`orderedset(int)` (what order?) or the unimplemented `dict(Ressource, int)` ?

doc/nyan/engine_types.md
castilma: what do you mean with "consider projectiles" and "area of effect"?

doc/nyan/engine_types.md
castilma: how would BonusProvider look like?

elnabo:
Maybe Unit should have a population field.

AoE1 and the last AoE2 extensions have infantries with 0.5 population usage.

doc/nyan/engine_types.md
heinezen: It would be similar to a `ResourceProvider`, but instead of giving a `Resource`, it provides a `Bonus` to the player who controls the `BonusProvider`. Boni could give stats upgrades (like 5% faster unit creation), area of effect buffs (like the generals in RoN), constant flow of resources (e.g. relics and feitoria in AoE2) or give a new ability to a hero who equips the `Mighty Staff of Midgets`, for example. Boni could also be temporary, permanent or tied to an item/unit/building. 

In `nyan` a `Bonus` would probably be utilized as a nyan patch. Similar to the `Tech()` patch. `BonusProvider` on the other hand would be like a `ResourceSpot`:

```
Item():
     name : text

Bonus():
     timeout : float
     updates : set(NyanPatch)

ResourceBonus(Bonus):
     resources : Resource
     rate : float

BonusProvider():
    condition : Condition
    is_temporary : bool
    boni : set(Bonus)

# Content
Relic(Item, BonusProvider):
     name = "relic"
     
     condition = garrisoned
     is_temporary = false
     boni = {RelicBonus}

RelicBonus(ResourceBonus):
     resource = {Gold}
     timeout = -1.0
     rate = 1.0

     GoldBoost<resource>():
         amount += 2.0

     patches = {GoldBoost}
```

Or something like that. I hope i got the nyan syntax right.

heinezen:
More in draft: https://pad.stusta.de/p/openage-mod-api

TheJJ:
Superseeded and continued by #1021 :)

PR #645, opened 2016-10-17
Heal action and attribute

MaanooAk:
I added the `HealAction` class and the `heal` attribute using as base the `AttackAction` and `attack` attribute.
I don't know if the the graphic must be heal or something else...

### DIFF
--- a/libopenage/unit/action.cpp
+++ b/libopenage/unit/action.cpp
@@ -99,7 +99,7 @@ void UnitAction::face_towards(const coord::phys3 pos) {
 void UnitAction::damage_object(Unit &target, unsigned dmg) {
 	if (target.has_attribute(attr_type::hitpoints)) {
 		auto &hp = target.get_attribute<attr_type::hitpoints>();
-		if ((hp.current - dmg) > 0) {
+		if (hp.current > dmg) {
 		    hp.current -= dmg;
 		}
 		else {
@@ -408,7 +408,7 @@ void IdleAction::on_completion() {}
 bool IdleAction::completed() const {
 	if (this->entity->has_attribute(attr_type::hitpoints)) {
 		auto &hp = this->entity->get_attribute<attr_type::hitpoints>();
-		return hp.current <= 0.0f;
+		return hp.current == 0;
 	}
 	else if (this->entity->has_attribute(attr_type::resource)) {
 		auto &res_attr = this->entity->get_attribute<attr_type::resource>();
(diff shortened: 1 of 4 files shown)


MaanooAk:
I suggest to merge this request and then I would to like to create a new one in order to move constant-hardcoded values from heal, attack and other actions to the corresponding attributes.

For example, the `rate_of_fire` is always `0.002f`, but it must be able to take different values (siege weapons have higher).
Other values: `damage, rate_of_heal, life healed, rate_of_repair` and so on...

TheJJ:
Keep in mind that all the values will be gathered from [nyan](https://github.com/SFTtech/nyan) someday. I'll review your code soon :)

libopenage/unit/action.cpp
```diff
@@ -1000,6 +1009,52 @@ void AttackAction::fire_projectile(const Attribute<attr_type::attack> &att, cons
...
+		this->heal(*target_ptr);
+	}
+
+	// inc frame
+	this->frame += time * this->current_graphics().at(graphic)->frame_count * this->rate_of_heal;
+}
+
+bool HealAction::completed_in_range(Unit *target_ptr) const {
+	auto &h_attr = target_ptr->get_attribute<attr_type::hitpoints>();
+	return h_attr.current >= h_attr.max; // is unit at full hitpoints?
```
TheJJ: comparison between signed and unsigned values :)

MaanooAk: `h_attr.current >= (int)h_attr.max` ?
or
`h_attr.current > 0 && (unsigned int)h_attr.current >= h_attr.max` ?

TheJJ: better solution would be to fix the types directly. current and max should both be unsigned i'd say.

MaanooAk: At source: `int current; // can become negative`, so I belive it may break things.
Also I am not fully understand what "extend the value fetching" means...

MaanooAk: After some investigation, `current` will never get a negative value, but at [gamedata-struct.md](https://github.com/SFTtech/openage/blob/master/doc/media/gamedata-struct.md) we have the hitpoints signed (from source: `int16_t hit_points;`)

MaanooAk:
The healing rate and amout are inside the attribute, only variable inside the action is the progress (`heal_percent`)

libopenage/unit/action.cpp
MaanooAk: Done [commit 5b4bfb7](https://github.com/SFTtech/openage/pull/645/commits/5b4bfb7e692796c8f9d75c73a136db38536aca7b)

Changed type to unsigned and tracked down all uses of it.

TheJJ:
merged with 6c48488

PR #646, opened 2016-10-20
Armor and damage calculation

MaanooAk:
- Added armor attribute (shared)
  - Changed some other attributes shared property
- Changed attack attribute damage variable type
- Created `void UnitAction::damage_object(Unit &target)` which takes the armor from the target, the damage from the entity (stored inside the action) and calculates the actual damage based on [reverse_engineering/damage.md](https://github.com/SFTtech/openage/blob/master/doc/reverse_engineering/damage.md)

(Kept the old `void UnitAction::damage_object(Unit &target, unsigned dmg)` for testing)


MaanooAk:
The checks fail from: `ImportError: No module named 'buildsystem.codecompliance.modes'` (bug?)
Everything else is ok.

TheJJ:
damnit, didn't commit that file. if you rebase your branch against the current master it will be fixed.

MaanooAk:
Fixed the files, have no idea how the executable bits changed. (Also fixed "syntax"(?) typos at reverse_engineering/damage.md)

libopenage/unit/attribute.h
```diff
@@ -114,6 +115,7 @@ class AttributeContainer {
 
 using attr_map_t = std::map<attr_type, std::shared_ptr<AttributeContainer>>;
 
+using typeamount_set = std::map<int, unsigned int>;
```
TheJJ: What is that? Please write a docstring. I guess it is the attack class to damage map.

Use `std::unordered_map` please, the `std::map` doesn't hash. And it is not a `set`, is it?

libopenage/unit/attribute.h
```diff
@@ -167,14 +169,44 @@ template<> class Attribute<attr_type::hitpoints>: public AttributeContainer {
...
+	// TODO remove (keep for testing)
 	Attribute(UnitType *type, coord::phys_t r, coord::phys_t h, unsigned int d, UnitType *reset_type)
 		:
 		AttributeContainer{attr_type::attack},
 		ptype{type},
 		range{r},
 		init_height{h},
+		stance{attack_stance::do_nothing},
+		attack_type{reset_type} {
+		damage[4] = d;
```
TheJJ: what is index 4? if it's a special value that could be named, please use an `enum class`

libopenage/unit/attribute.h
```diff
@@ -167,14 +169,44 @@ template<> class Attribute<attr_type::hitpoints>: public AttributeContainer {
...
+	}
+
+	Attribute(UnitType *type, coord::phys_t r, coord::phys_t h, typeamount_set d, UnitType *reset_type)
```
TheJJ: we should use a unified constructor for the attributes: the above (old) attribute constructor should directly call this one and create a `typeamount_set` with the one entry. Saves setting all the attributes separately.

libopenage/unit/attribute.h
MaanooAk: The 4 is the `gamedata::hit_class::UNITS_MELEE` which is not exported at the moment. However this constructor is `TODO remove` so no need to change the the whole gamedata exports.

libopenage/unit/attribute.h
```diff
@@ -115,7 +115,11 @@ class AttributeContainer {
 
 using attr_map_t = std::map<attr_type, std::shared_ptr<AttributeContainer>>;
 
-using typeamount_set = std::map<int, unsigned int>;
+/**
+ * An unordered_map with a int key used as a type id
+ * and a unsigned int value used as the amount
+ */
+using typeamount_map = std::unordered_map<int, unsigned int>;
```
TheJJ: but for what is this map? is it to store "attack/armor class with their corresponding damage/absorbtion value"?

libopenage/unit/attribute.h
```diff
@@ -198,10 +202,10 @@ template<> class Attribute<attr_type::attack>: public AttributeContainer {
 		init_height{h},
 		stance{attack_stance::do_nothing},
 		attack_type{reset_type} {
-		damage[4] = d;
+		damage[4] = d; // 4 = gamedata::hit_class::UNITS_MELEE (not exported at the moment)
 	}
```
TheJJ: Call `Attribute{type, r, h, {4, d}, reset_type}` as a shortcut here.

libopenage/unit/attribute.h
MaanooAk: Yes, but i didn't write that on the doc because it can be used for other things too.

TheJJ: ah i see, didn't know and realize that.

MaanooAk:
I have made all changes (no?). Also excellent grammar phrase: 

> [...] each attacker's attack classes is used only if the target has the matching armor class.

TheJJ:
merged with 1bd5703ca6a9caf298778579904644f79610e684

PR #647, opened 2016-10-21
Added my collection of memory mapping for the Genie-Engine

yvan-burrie:
These reversed engineered structures are for the age2_x1 EXE file (00.09.26.0809).
Created and researched using IDA over 3 years by myself with the enormous help of JustTesting1234.
I hope these are useful for Openage, and I will provide further improvements as there is much more information I yet must add.

### DIFF
--- a/doc/reverse_engineering/Memory_AOC_00_09_26_0809.txt
+++ b/doc/reverse_engineering/Memory_AOC_00_09_26_0809.txt
@@ -1,0 +1,1039 @@
+RGE_Command
+RGE_Task_list
+
+
+
+\Game (8240)
+	0	VOID	Name
+	4	CALL	005E4DA0
+	8	CALL	
+	12	CALL
+	24	CALL	
+	28	CALL	005E4720 (GetSetupStatus)
+	32	VOID	hCampaign
+	36	CALL	GetLanguageStringByID
+	40	VOID	Settings
+	44	VOID	HWND (WindowHandle)
+	48	DWORD	WindowStatus
+	56	DWORD	
+	72	DWORD	RoomTime
+	76	DWORD	
+	80	DWORD	???
+	88	CALL	???
+	92	CALL	???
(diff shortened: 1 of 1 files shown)


TheJJ:
Thanks! I'm gonna fix some whitespace errors before merging.
Did you export it or is it a handcrafted file?

Is this helpful to other IDA users, like, can they import the structs?

yvan-burrie:
Unfortunately this is a handcrafted file. If I had been more experienced and organised, I would have created C header structures that can be imported to IDA. Meanwhile I thought it would be best to upload this draft. I plan on redoing the entire memory mapping in a more convenient manner. There are also many important structures which I need to add, such as the scenario and player structures.

I have also reversed a whole bunch of file formats, including the entire SCN file correctly formatted, the database DAT file, and all other DAT files; TileEdge, BlkEdge, FilterMaps, LoQMaps, LightMaps, ViewIcm, STemplet and Blendomatic.

I am currently reversing the saved-game and recorded-game file formats.

TheJJ:
Whoa, especially the `TileEdge, BlkEdge, FilterMaps, LoQMaps, LightMaps, ViewIcm, STemplet and Blendomatic` are helpful for us, we only use blendomatic so far. I planned implementing hillshading the 3d-way though, with real elevations and regular "modern" shaders.

From the dat file we have most information already but surely you can further improve it (`openage/convert/gamedata`).

The savegame and gamerecord format aren't interesting for us at least, we'll have a custom format suiting our needs (#649 probably). Gamerecords will be the collection of all curves (for those, see #530).
The scn file is also nice to have, although we probably just reuse our savegame file format.

TheJJ:
Thank you for your data! If you have more findings about the `dat` file, we can very much use those.
We read it in `openage/convert/gamedata`, you'll notice the structs there.

I beautified your file a bit, looking forward to more findings :)

PR #649, opened 2016-10-21
savefile Format proposal

timohaas:
#613 It is written in Markdown and contains the first draft. I need the help of the other developers, I don't know what attributes need saving and what of what types they are.

### DIFF
--- a/doc/ideas/savefileformat.md
+++ b/doc/ideas/savefileformat.md
@@ -1,0 +1,446 @@
+# openage save format
+
+1.  [Questions](#1-questions)
+2.  [Overview](#2-overview)
+3.  [Header](#3-header)
+4.  [String Table](#4-string-table)
+5.  [Atrribute Table](#5-attribute-table)
+6.  [Game Information](#6-game-information)
+7.  [User Information](#7-user-information)
+8.  [Chunk Table](#8-chunk-table)
+9.  [Single Chunks](#9-single-chunks)
+10. [Triggers](#10-triggers)
+    1. [Event List](#101-event-list)
+    2. [Condition List](#102-condition-list)
+    3. [Action List](#103-action-list)
+11. [Crypto](#11-crypto)
+
+### 1 Questions
+
+Do we something like ingame time vs realtime if the game is played at a different speed?
+
+What do we want to track for stats?
+
+| category   | stats  |
+| ---------- | ------ |
+| ressources | collected food/wood/gold/stone, maximum res in the bank? |
+| units      | units/buildings killed/lost, units converted/lost to convertions, max units of each type |
+| map        | discovered, maximum simultanously viewable (discovered - fog of war) |
+| tech       | when was each tech/age researched? |
+| battle     | what is a battle? units killed per timeslot per close area 30 units died in 15 seconds in radius of 20 Tiles TODO: definitions, maybe add cost for unit for calculation of the size of battle |
+
+do we record timeseries? if yes fixed sizes(round robin database) or simple array snapshot per 10s/30s/60s/flexible?
+
+every minute we save unit count per player?
+
+crypto? multiplayer safe => server creates savefile, encrypts it, but one member of every team is needed to decrypt it and continue the game https://en.wikipedia.org/wiki/Secret_sharing
+
+- savefile vs recorded game what happens if a savefile is continued, new recorded game or should recorded game stored with the savefile? if we dont store it, maybe we can reference it, so it can be continued if available
+- how to store/reference stuff which is changed/modified/added to assets pack?
+
+
+what are we trying to achieve?
+
+- small size (compression per 'object')
+- fast access to individual chunks
+- everything should be checksummed(bitrot)
+- statusinformation must be readable really fast (ingame listing)
+- statsinformation should be easily be extensable, so no new format is needed
+
+### 2 Overview
+```
+|--------------------------|
+|    header                |
+|--------------------------|
+| string table             |
+|--------------------------|
+| key value table          |
+|--------------------------|
+| gameinformation          |
+|--------------------------|
+| userinformation          |
+| | userheader           | |
+| | user 1               | |
+| | user 2               | |
+|--------------------------|
+|        chunktable        |
+|--------------------------|
+| single chunks            |
+| | chunk 1 header        ||
+| | | chunk 1 object 1 | | |
+| | | chunk 1 object 2 | | |
+| | chunk 2 header        ||
+| | | chunk 2 object 1 | | |
+| | | chunk 2 object 2 | | |
+|--------------------------|
+|        Triggers          |
+|--------------------------|
+```
+
+types: default type is uint with width as seen in layout
+
+| type   | info                                                                                      |
+| ------ | ----------------------------------------------------------------------------------------- |
+| uint+  | (variable length)[https://en.wikipedia.org/wiki/Variable-length_quantity], maximum uint64 |
+| string | utf8 with uint+ length prefix                                                             |
+| f32    | float 32Bit                                                                               |
+
+endianess: Little-Endian
+
+### 3 Header
+
+| bytes__ | type      | name                              | info |
+| ------- | --------- | --------------------------------- | ----- |
+|   0- 10 | ascii[11] | openagesave                       | magic number( openage in ascii) |
+|  11- 11 | uint8     | version                           | <div>save file version number,</div><div>1: this document describes version 1,</div><div>0: should be used while developing(so they won't be used in the wild</div> |
+|  12- 15 | uint32    | string table offset               | |
+|  16- 79 | byte[64]  | sha512 over stringtable           | checksum over all compressed bytes specified in length(inclusive) |
+|  80- 83 | uint32    | key value table offset            | |
+|  84-147 | byte[64]  | sha512 over key value table       | checksum over all compressed bytes specified in length(inclusive) |
+| 148-151 | uint32    | gameinformation table offset      | |
+| 152-215 | byte[64]  | sha512 over gameinformation table | checksum over all compressed bytes specified in length(inclusive) |
+| 216-219 | uint32    | userinformation table offset      | |
+| 220-283 | byte[64]  | sha512 over userinformation table | checksum over all compressed bytes specified in length(inclusive) |
+| 284-287 | uint32    | chunktable table offset           | |
+| 288-351 | byte[64]  | sha512 over chunktable table      | checksum over all compressed bytes specified in length(inclusive) |
+| 352-355 | uint32    | single_chunks table offset        | |
+| 356-419 | byte[64]  | sha512 over single_chunks table   | checksum over all compressed bytes specified in length(inclusive) |
+| 420-423 | uint32    | triggers table offset             | |
+| 424-487 | byte[64]  | sha512 over triggers table        | checksum over all compressed bytes specified in length(inclusive) |
+| 488-488 | boolean   | crypto flag                       | |
+| 489-489 | uint8     | key count                         | how many keys are necessary to encrypt file? must be >= 2 or 0 |
+| 490-491 | byte[3]   | reserved                          | must be 0 |
+| 492-516 | byte[25]  | timestamp                         | ISO 8601: 2017-01-11T22:07:12+00:00 |
+| 517-772 | byte[216] | crypto key                        | must be 0, if crypto is not used |
+| 773-788 | byte[16]  | crypto iv                         | must be 0, if crypto is not used |
+| 789-852 | byte[64]  | sha512 over bytes 000-788         | header checksum, used to check if header is corrupted |
+
+### 4 String Table
+
+This table stores all the strings needed in the savegame, so they can compressed easily and simply referenced by a hashmap. The order of appearance in this list is there referrence id.
+
+all bytes are offsets to "string table offset" (Header Section)
+
+| bytes        | type                                 | name            | info                                                  |
+| ------------ | ------------------------------------ | --------------- | ----------------------------------------------------- |
+| 0-3          | uint32                               | length          | length of this compressed segment in bytes            |
+| 4-7          | uint32                               | string count    | string count                                          |
+| 8-(length-1) | string[string count] list of strings | these strings will be referenced by others:  id=count                   |
+
+### 5 Key Value Table
+
+all bytes are offsets to "key value table offset" (Header Section)
+
+| bytes        | type   | name                 | info                                       |
+| ------------ | ------ | -------------------- | ------------------------------------------ |
+| 0-3          | uint32 | length               | length of this compressed segment in bytes |
+| 4-7          | uint32 | key/value pair count | attribute count                            |
+| 8-(length-1) | ..     | list of single value | Single Value @5.1, order = id              |
+
+#### 5.1 Single Value
+
+
+| bytes | type  | name             | info                                  |
+| ----- | ----- | ---------------- | ------------------------------------- |
+| 0-0   | uint8 | basic type       | one of the basic type below           |
+| 1-..  | uint+ | string reference | reference to String Table (Section 4) |
+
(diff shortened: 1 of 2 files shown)


MaanooAk:
Must be saved for each unit:
- the non shared attributes 
- all the actions on the action queue

Notes:
The hitpoints, carrying and all of that are stored in the attributes.
Non shared attributes have the `shared()` return true
Keep in mind that both groups have are not fixed size

TheJJ:
The possible object attributes have to be dynamic as mods may add arbitraty new things to store per object, player, etc.

timohaas:
OK, like my professor always said: Every problem can be solved by adding a new layer of indirection.

There will be a table with String to uint assignment, so we dont have to store strings redundantly and all atttributes will be dynamicly stored. Further I will introduce a new type: [Variable Integer Encoding](https://en.wikipedia.org/wiki/Variable-length_quantity).

How do you like the documantation so far? Is the ASCII Art understandable?

TheJJ:
Yes, the documentation style is good, brief and informative.
Idea about the file magic: should be `openagesave${version}` because we might have other files with other uses, so we should differ between those with the magic as well.

For the time storage in the save files: We wanna have the curve design from #530, and I think it's a very good idea to record all those with timestamps to scroll back and forward in time for replays.

The format in this issue is for a snapshot of the above, it doesn't need to preserve history. So yes, we might want to store the passwd game time, but the history is stored in another file format (by maybe extending this at some later point).

timohaas:
I think I have found a good solution to the generic key-value problem with object attributes and stats.

I added a key-value table where a string to type assignment is possible.

@TheJJ magic number will be changed

in this format timestamp is only informational. Nothing should depend on it. Tech research time should be saved as percentage research etc...

Question: the snapshot doesnt need to save curves, it must save the action: "go from A to B" and not the actual path?

history stuff: we can add it later as a new segment. I can add it like the triggers as dummy segment for now, so we will be compatible in the future.

doc/savefileformat.md
```diff
@@ -0,0 +1,419 @@
...
+| units      | units/buildings killed/lost, units converted/lost to convertions, max units of each type |
+| map        | discovered, maximum simultanously viewable (discovered - fog of war) |
+| tech       | when was each tech/age researched? |
+| battle     | what is a battle? units killed per timeslot per close area 30 units died in 15 seconds in radius of 20 Tiles TODO: definitions, maybe add cost for unit for calculation of the size of battle |
+
+do we record timeseries? if yes fixed sizes(round robin database) or simple array snapshot per 10s/30s/60s/flexible?
+
+every minute we save unit count per player?
+
+crypto? multiplayer safe => server creates savefile, encrypts it, but one member of every team is needed to decrypt it and continue the game https://en.wikipedia.org/wiki/Secret_sharing
```
castilma: does every player get the encrypted savefile or is it only stored on the server? the former is better for the (public) server (no harddrive space for stroing games needed), but leads to (1) file-integrity and (2) user authentity questions. (1) the players could together decrypt the file and change the gamestate. if everyone [or one of every team] is involved, no one will call the other ones cheater. but they could make a weird gamestate that breaks the engine by creating a weird logic (leading to remote code execution[admittedly, in the worst case]) or results in a DoS (creating 10^9 units or something).

timohaas:
I'm sorry, I think I didn't describe my encryption plan precisly:

1) Gameserver encrypts savefile
2) encryption key will be split up into pieces
3) for every team, there will be the savefile with a piece of the decryption key
4) for continuing the game, both savefiles will be needed for recovering the key

=> no one can continue/look into/mainpulate the savegame until both agree to continue the game

=> if they never come to an aggreement, the savefile is lost

castilma:
In 4) you wrote 'both savefiles'. What do you mean by 'both'?

every player gets the encrypted gamestate and a part of the key, right? my point was, that the server loads gamestates provided by the players. how do we make sure, that those gamestates aren't some unrealistic states that need a lot of computational  server ressources?

ChipmunkV:
> how do we make sure, that those gamestates aren't some unrealistic states that need a lot of computational server ressources?

Let server owner sign it.

coffenbacher:
Do we need encryption / decryption? I feel like that may add undue complexity when there's an ugly failure state. If the clients each simply have a copy of the hash of the savefile, that performs the same anti-cheat function but with more flexibility. Each copy of the savefile is still stand-alone usable and intact, but clients know if someone is trying to load something they didn't expect.

ChipmunkV:
> Do we need encryption / decryption?

Someone will find your cheese if there is no encryption.

coffenbacher:
Ah, I suppose map hacking is a concern. I believe that's already an issue
with AoC but no reason not to design it out going forward, good point.

timohaas:
everything after the header is encrypted. Some modifications are necessary, otherwise nothing would be readable(not even playernames)

doc/savefileformat.md
```diff
@@ -0,0 +1,560 @@
...
+| 420-423 | uint32    | triggers table offset             | |
+| 424-487 | byte[64]  | sha512 over triggers table        | |
+| 488-488 | boolean   | crypto flag                       | |
+| 489-489 | uint8     | key count                         | how many keys are necessary to encrypt file? must be >= 2 or 0 |
+| 490-491 | byte[3]   | reserved                          | |
+| 492-747 | byte[216] | crypto key                        | must be 0, if crypto is not used |
+| 748-764 | byte[16]  | crypto iv                         | must be 0, if crypto is not used |
+| 765-825 | byte[64]  | sha512 over bytes 000-764         | |
+
+checksum over all compressed bytes specified in length(inclusive) in the segment
```
timohaas: thanks for spotting the mistake. Fixed it.

timohaas:
I updated the proposal.
Awaiting comments.

TheJJ:
Thinking about this, I'm not sure whether we should not do it the other way round: Improve our simulation and then see what we need to serialize for the save file.

Because what counts is the simulation. Designing some "arbitrary" format as a constraint where we need to connect two ends then may be too complex. The work you did is of course valuable for what we can do, but instead of a specific listing of exactly how we store, I would suggest we do that from the direction to dump and restore state the simulation stores. And then step by step extend the dump file with more data (and adapting the file format).

Do you have any objections to that idea?

TheJJ:
On a sidenote, the original savegame format (which also acts as a replay storage): https://github.com/stefan-kolb/aoc-mgx-format

timohaas:
Hi
I did a rebase and moved the document to the ideas folder. It is merge ready.

TheJJ:
Could you please do a quick rebase to eliminate the merge commit? Then we can add it to the ideas.

doc/ideas/savefileformat.md
```diff
@@ -0,0 +1,446 @@
...
+9.  [Single Chunks](#9-single-chunks)
+10. [Triggers](#10-triggers)
+    1. [Event List](#101-event-list)
+    2. [Condition List](#102-condition-list)
+    3. [Action List](#103-action-list)
+11. [Crypto](#11-crypto)
+
+### 1 Questions
+
+Do we something like ingame time vs realtime if the game is played at a different speed?
```
TheJJ: we just record keyframes. the timescale at which they are played afterwards doesn't matter.

doc/ideas/savefileformat.md
```diff
@@ -0,0 +1,446 @@
...
+
+What do we want to track for stats?
+
+| category   | stats  |
+| ---------- | ------ |
+| ressources | collected food/wood/gold/stone, maximum res in the bank? |
+| units      | units/buildings killed/lost, units converted/lost to convertions, max units of each type |
+| map        | discovered, maximum simultanously viewable (discovered - fog of war) |
```
TheJJ: the discovered area should probably be stored with an algorithm that splits up the area in maximum size rectangles. this should probably be per-chunk.

doc/ideas/savefileformat.md
```diff
@@ -0,0 +1,446 @@
...
+### 1 Questions
+
+Do we something like ingame time vs realtime if the game is played at a different speed?
+
+What do we want to track for stats?
+
+| category   | stats  |
+| ---------- | ------ |
+| ressources | collected food/wood/gold/stone, maximum res in the bank? |
+| units      | units/buildings killed/lost, units converted/lost to convertions, max units of each type |
```
TheJJ: all this information is keyframes: changing points are recorded, the middle values can be interpolated.

doc/ideas/savefileformat.md
```diff
@@ -0,0 +1,446 @@
...
+| map        | discovered, maximum simultanously viewable (discovered - fog of war) |
+| tech       | when was each tech/age researched? |
+| battle     | what is a battle? units killed per timeslot per close area 30 units died in 15 seconds in radius of 20 Tiles TODO: definitions, maybe add cost for unit for calculation of the size of battle |
```
TheJJ: the battle definition is done in the `reverse_engineering` part already. and we don't need to save it as this is information that is reconstructed.

doc/ideas/savefileformat.md
```diff
@@ -0,0 +1,446 @@
...
+
+do we record timeseries? if yes fixed sizes(round robin database) or simple array snapshot per 10s/30s/60s/flexible?
+
+every minute we save unit count per player?
```
TheJJ: no need for that, we have to recreate the whole game state.

doc/ideas/savefileformat.md
```diff
@@ -0,0 +1,446 @@
...
+
+crypto? multiplayer safe => server creates savefile, encrypts it, but one member of every team is needed to decrypt it and continue the game https://en.wikipedia.org/wiki/Secret_sharing
```
TheJJ: that's a pretty good idea!

doc/ideas/savefileformat.md
```diff
@@ -0,0 +1,446 @@
...
+
+- savefile vs recorded game what happens if a savefile is continued, new recorded game or should recorded game stored with the savefile? if we dont store it, maybe we can reference it, so it can be continued if available
```
TheJJ: the difference is that the record file stores all the keyframes (i.e. simulation output), the save file stores just the latest of all this information (except those components which time still matters (i can just think of timestamps of age advances and unit count snapshots for statistics))

doc/ideas/savefileformat.md
```diff
@@ -0,0 +1,446 @@
...
+- how to store/reference stuff which is changed/modified/added to assets pack?
```
TheJJ: that's impossible and also not part of the game state. if objects are missing, than we can try to continue without (like skyrim does) but most likely it will just fail.

doc/ideas/savefileformat.md
```diff
@@ -0,0 +1,446 @@
...
+
+
+what are we trying to achieve?
+
+- small size (compression per 'object')
+- fast access to individual chunks
+- everything should be checksummed(bitrot)
+- statusinformation must be readable really fast (ingame listing)
+- statsinformation should be easily be extensable, so no new format is needed
```
TheJJ: even more: the whole format should be extensible so when we discover that we wanna save additional things, old savegames must never break.

doc/ideas/savefileformat.md
```diff
@@ -0,0 +1,446 @@
...
+### 6 Game Information:
+
+gameinformation will be compressed: bytes 4 : length-1
+
+all bytes are offsets to gameinformation offset
+
+| bytes  | type        | name                     | info |
+| ------ | ----------- | ------------------------ | --- |
+| 0-3    | uint32      | length                   | length of this compressed segment in bytes |
+| 4-4    | int8        | speed                    | game speed relative to normal in 1/10 steps 90 => 9 times faster, -4 => 0.4 times normal speed |
```
TheJJ: as idea, it's ok, but we really can't guess how we can save our simulation efficiently. also, we can even represent all the game information as keyframes of our curves, which then just remains the same all the game.

So the much simpler approach is to create a generic curve storage, which we can serialize and unserialize and snapshot.

doc/ideas/savefileformat.md
```diff
@@ -0,0 +1,446 @@
...
+|   5 | intervall           | TIME uint+                      | -                            | fired every INTERVALL milliseconds, but max only once per tick                       |
+|   6 | timer               | TIME uint+                      | -                            | fired once after game is running >= TIME milliseconds                                |
+|   7 | object_in_area      | POLYGON uint+,uint+,uint+,uint+ | unit,playerid,teamid         | fired when an object is in area POLYGON                                              |
+|   8 | tech_researched     | -                               | tech,playerid,teamid         | fired when someone researched TECH                                                   |
+|   9 | unit_conversion     |                                 | unit,playerFrom,playerTo     | fired when a unit is converted and not killed (tech: Heresy)                         |
+|  10 | building_conversion |                                 | building,playerFrom,playerTo | fired when a building is converted and not destroyed (tech: Heresy)                  |
```
TheJJ: all this is just stored as keyframes, listing it explicitly will very likely never be done because we can't know our data fields yet, and want to extend them dynamically without pain and backward-compatibility-breaking.

doc/ideas/savefileformat.md
```diff
@@ -0,0 +1,446 @@
...
+everything from byte 826-end will be encrypted with aes.
+
+Key distribution:
+
+- encryption key will be randomly generated
+- server writes savegame unencrypted to disk
+- server encrypts everything after the last header byte with aes until file ending
+- server writes crypto IV in header field
+- generate team-count-1 times a random 256Bit(encryption key size) pseudo key
+- key team N(the one without pseudo key): team1 xor team2 xor .. teamN-1 xor original encryption key and destroys the original encryption key
```
TheJJ: this idea is nice, i'll show it to some of my crypto-experienced friends cause I can't tell if it has major loopholes.

doc/ideas/savefileformat.md
Tomatower: I think it would be possible to store the whole history in a file, just whenever a keyframe is surpassed by the renderer, it can be written into the savefile.
If we save the game, all "future" keyframes are written, the queues are written and the current time is stored.

castilma:
if you can, please read about [langsec](langsec.org). first skim through some slides and then read some articles. you can start with [For want of a nail](http://langsec.org/ForWantOfANail-h2hc2014.pdf).
intro:
langsec is a theoretical approach to secure input handling/ tries to give an explanation, why, after all the years, IT has still problems with SQL injections, XSS, buffer overflows that lead to arbitrary code executions and what the bigger problem after shellshock, heartbleed etc. is.

the problem is, that file formats and protocols are languages but are not treated as such. this becomes especially problematic, when they are turing complete: an attacker can use it to write arbitrary programs (crafted input, that works like bytecode on a virtuall machine), that are run on the 'weird machine' that a corresponding buggy parser implements.

to prevent these types of errors, langsec proposes:

1. use, where possible, 'simple' formats(/languages), according to the chomsky hierarchy: either regular or deterministic context-free.

2. 'full recognition before processing': don't intermix recognition of and computation on data. parsers that do this are called 'shotgun parsers'. see my [wave parser](https://github.com/castilma/openage/blob/pyopus/openage/convert/opus/opusenc.pyx#L217) for an example. this also saves cpu time if an invalid input is met.

3. use generated code for the recognition process. they provide a tool called hammer for the code generation from a format specification.

Note: Formats using length fields are not deterministic context-free. I did not understand yet how to use variable length parts without length fields. In one of the videos the female presenter mentions 'delimiters'. But I don't know how that looks like in reality. (does anyone know a format that uses delimiters?)

Another thought I had, which contradicts what I said above, is: Why don't we use XML/binary XML? If we encrypt, we could just encrypt the subtree and add it as one xml node in the savefile. when reading, we decrypt it and parse it as xml.

zuntrax:
This is not suited to our upcoming event system. Discussing savegame format before having worked out the core game logic is a bit premature. Once we get there, we can still necro this PR or salvage some ideas.

PR #648, opened 2016-10-21
Fix OSX Build

jazzpi:
I basically just took the fixes @Birch-san found in #643 and made them a bit more portable.

I'm not sure about the 4th problem... I didn't really adapt the fixes too much yet it didn't occur at all for me. It might be that Qt5 is now on 5.7.0, but again I'm not sure.

Also, this just fixes the build, the game still doesn't run sadly. It spams the terminal whith a whole bunch of

```
  File ?, in std::__terminate(void (*)())+0x8 [0x7fffd8869d69]
  File ?, in std::terminate()+0x33 [0x7fffd8869de3]
```

and you can't actually close it except for doing Ctrl-Z and `kill -9 %1`. This might be related to #608.

Closes #643.

Issue #643, opened 2016-10-16
Build break on (at least) Mac OS X 10.12 Sierra


Birch-san:
I had a try at building this on OS X Sierra. Since I was unable to find any screenshots, video, releases or status documentation since 2015, and wanted to find out what the project looks like today.

There were 4 problems so far. I have solutions for three of them.
## Problem 1: [configure] `Qt5Core` package configuration not found

```
Call Stack (most recent call first):
  CMakeLists.txt:88 (doxygen_configure)



cpp library
    libopenage                                          [212 sources] [15 generated]
CMake Error at libopenage/CMakeLists.txt:90 (find_package):
  By not providing "FindQt5Core.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Qt5Core", but
  CMake did not find one.

  Could not find a package configuration file provided by "Qt5Core"
  (requested version 5.4) with any of the following names:

    Qt5CoreConfig.cmake
    qt5core-config.cmake

  Add the installation prefix of "Qt5Core" to CMAKE_PREFIX_PATH or set
  "Qt5Core_DIR" to a directory containing one of the above files.  If
  "Qt5Core" provides a separate development package or SDK, be sure it has
  been installed.
```

Qt5 provides in its `lib/cmake` directory (i.e. `/usr/local/Cellar/qt5/5.6.2/lib/cmake`) some `cmake` scripts (e.g. `Qt5CoreConfig.cmake`). Our `cmake` needs to find these, since they tell it how to link against Qt5's modules (e.g. `Qt5::Core`).
#### Solution

Before running `./configure`, I set the following variable in my shell:

``` bash
CMAKE_PREFIX_PATH="$CMAKE_PREFIX_PATH:/usr/local/Cellar/qt5/5.6.2/lib/cmake"
```
#### Another solution

There is an alternative solution I found that for some reason works.

That is: add `/usr/local/opt/qt5/bin` to your `$PATH`.

I think I remember that this turned out to be only a partial solution. However it's possible my memory is failing me, and that the "remaining problem" with it was simply Problem 3 (which is a problem we encounter with either of the solutions).
## Problem 2: [make] `GL/gl.h` not found

```
➜  openage git:(master) 🍔 make
…
[  1%] Building CXX object libopenage/CMakeFiles/libopenage.dir/engine.cpp.o
In file included from /Users/birch/git/openage/libopenage/engine.cpp:17:
[...]

PR #658, opened 2016-10-30
[WIP] Add Travis CI build & testing for macOS

dusek:
This adds CI testing of the macOS build on Travis. It builds the project (basically `./configure && make`) and then runs the tests (`make tests`)

This is a WIP as it contains an adjusted form of fixes from #648 (all in second commit) in order to be able to demonstrate a [successful build](https://travis-ci.org/dusek/openage/builds/171836562). After #648 is merged, this PR will be updated (finished) to take into account the final form of fixes (basically drop the second commit, and adjust the first commit if necessary).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,21 @@
+os:
+  - osx
+before_install:
+  - set -o pipefail
+  - xcodebuild -version
+  - xcodebuild -showsdks
+install:
+  - brew update
+  - brew install caskroom/cask/brew-cask
+  - brew tap caskroom/fonts
+  - brew tap homebrew/versions
+  - brew tap homebrew/python
+  - brew update
+  - brew cask install font-dejavu-sans
+  - brew install python3 libepoxy freetype fontconfig harfbuzz sdl2 sdl2_image opus opus-tools opusfile
+  - brew install qt5
+  - pip3.5 install pygments cython numpy Pillow
+script:
+  - CMAKE_PREFIX_PATH=/usr/local/opt/qt5 ./configure --compiler=clang++ --prefix=/usr
+  - make -j4
+  - make tests
--- a/copying.md
+++ b/copying.md
@@ -89,6 +89,7 @@ _the openage authors_ are:
 | Johannes Walcher            | tomatower                   | johannes.walcher@stusta.de            |
 | Akritas Akritidis           | MaanooAk                    | akritasak@gmail.com                   |
 | Edgard Mota                 | edgardmota                  | edgardmota@gmail.com                  |
+| Boris Dušek                 | dusek                       | me@dusek.me                           |
 
 If you're a first-time commiter, add yourself to the above list. This is not
 just for legal reasons, but also to keep an overview of all those nicknames.
--- a/libopenage/CMakeLists.txt
+++ b/libopenage/CMakeLists.txt
@@ -6,6 +6,9 @@
 # are specified below the source file list.
 
 set(CMAKE_AUTOMOC ON)
+if(APPLE)
+        list(APPEND CMAKE_PREFIX_PATH /usr/local/opt/qt5)
+endif()
 
 declare_binary(libopenage)
 
@@ -139,6 +142,10 @@ get_config_option_string()
 configure_file(config.h.in ${CMAKE_CURRENT_SOURCE_DIR}/config.h)
 configure_file(config.cpp.in ${CMAKE_CURRENT_SOURCE_DIR}/config.cpp)
 
+if(APPLE)
+	find_path(QTPLATFORM_INCLUDE_DIRS QtPlatformHeaders "/usr/local/opt/qt5/include")
+endif()
+
 # directories for header inclusion
 include_directories(
 	${OPENGL_INCLUDE_DIR}
@@ -148,6 +155,7 @@ include_directories(
 	${SDL2_INCLUDE_DIR}
 	${SDL2IMAGE_INCLUDE_DIRS}
 	${HarfBuzz_INCLUDE_DIRS}
+        ${QTPLATFORM_INCLUDE_DIRS}
 )
 
 # link the executable to those libraries
--- a/libopenage/gui/guisys/public/gui_renderer.h
+++ b/libopenage/gui/guisys/public/gui_renderer.h
@@ -4,7 +4,11 @@
 
 #include <memory>
 
+#ifndef __APPLE__
 #include <GL/gl.h>
+#else
+#include <OpenGL/gl.h>
+#endif
 
 struct SDL_Window;
 


TheJJ:
Hm, i'm really not in favor of using travis ever again (#397). True, it does provide mac builds nowaday, but this run took 20 minutes, whereas our kevin build takes 2 on a kinda outdated server.

It's really reprehensible that the container has to be set up by us every time, the dependency setup takes way more time than the actual build. But I somehow agree that this may be the least painful way to have osx CI. I'd really like to just run a hackintosh vm for that, but Apple apparently doesn't want anyone except (free) commercial providers to provide Apple build environments for testing.

We had incredible pain packaging new stuff for travis before we migrated to kevin. And i really don't want to have that happen again. Maybe the Apple packaging stuff isn't as broken as Debian's apt and Ubuntus PPA, but we'll be very very cautious before ever touching travis again.

gabrielcsapo:
@TheJJ the issue with travis is the support? Shouldn't we open a support issue on the Travis repo for the problem you are having?

dusek:
> It's really reprehensible that the container has to be set up by us every time, the dependency setup takes way more time than the actual build

On the other hand it has the benefit of testing the dependency setup instructions too. I.e. testing that if someone follows the build instructions (what and how to install with `brew`), things will work.

> this run took 20 minutes, whereas our kevin build takes 2 on a kinda outdated server

I looked and found that the build itself (configure + make + make tests) takes already ~7.2minutes, dependency installation is the remaining ~11.8 minutes. Maybe this is all slow due to the fact that the Travis build for macOS is running with only 1 CPU (unlike the Linux ones some of which get 2 CPUs) - see https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments

The longest thing in installing dependencies is installing numpy (~5.5 minutes) and running `fc-cache -frv` (I haven't measured that alone, but I often see the build stuck at that command for some time).

Maybe using Travis cache could speed something like this up, but I have not used it (and it would probably mean that we are not testing the full build instructions).

> I'd really like to just run a hackintosh vm for that, but Apple apparently doesn't want anyone except (free) commercial providers to provide Apple build environments for testing

You can always (legally and for no additional cost) run macOS VMs on top of a native macOS instance (running on original Apple hardware); the "only" problem is of course the cost of such extra Apple hardware if you currently do not have one in the first place.

I think there are some providers that provide general macOS cloud hosting, like [MacStadium](https://www.macstadium.com).

> We had incredible pain packaging new stuff for travis before we migrated to kevin. And i really don't want to have that happen again.

In case of macOS (as proposed in this PR), the Travis build basically runs what we are asking the users to run in the docs. Currently installing dependencies for macOS is completely covered with homebrew (thankfully for both the users and the one who is writing .travis.yml 😄); not that it cannot change, but currently the .travis.yml for macOS is basically the result of straightforward copying the build instructions into .travis.yml (and correcting 1-2 hiccups along the way, like brew-linking pillow, for which I should actually submit another PR to correct the build instructions themselves).

TheJJ:
I agree that testing the dependency step is also important, but we wanna have a convenient developer workflow, and not wait multiple hours when travis servers are overloaded (we had that some time ago, may be better now).

My solution to this would be to finally integrate docker into kevin, so the VM spawns docker inside, and our Dockerfile contains the prerequisite steps. In the container, chantal is then run.
The Dockerfiles are not part of the openage repo but instead a new one (`openage-docker`?).

Doing this for mac is hard, again. I don't own a mac and I surely won't spend any money for providing/testing software on a platform that apparently doesn't want crossplatform software to be written for it. If we'd get a hosted mac provider that can run our kevin stuff (namely a on-mac falk instance), that would be very nice. But I really don't understand why the cloudy Apple doesn't allow/support running mac VMs on non-macs.

If we added travis (optionally) just for mac and keep kevin for the linux tests, we'd still have the annoying delay of roughly half an hour at hightimes until the feedback button is red or green, which is not good for the workflow, especially since kevin reports back after ~5 minutes.

Again, I'm not against having mac support or tests or whatever, it's just that they're making it really hard to do it in an easy and convenient way for existing infrastructure.

dusek:
> If we'd get a hosted mac provider that can run our kevin stuff (namely a on-mac falk instance)

Please what is "on-mac falk"? I did not succeed in googling any reasonable result (like some software, or cloud provider, etc.) for it.

dusek:
> Please what is "on-mac falk"? I did not succeed in googling any reasonable result (like some software, or cloud provider, etc.) for it.

OK, I found out it is probably some [component of kevin](https://github.com/SFTtech/kevin/tree/master/falk) 😄

TheJJ:
https://www.contrib.andrew.cmu.edu/~somlo/OSXKVM/ may allow running it in qemu with kvm?

janisozaur:
As someone who is having a project CI'd on Travis for macOS, I'd like to point out something you have not yet taken into account:

[selection_001](https://www.traviscistatus.com/)

Travis-CI has very limited farm and thus throughput for macOS builds. Just queuing for a slot to become available can take 0.5h.

That said as a proper FOSS citizen, you should have at least some testing on *that system* if possible.

mic-e:
@janisozaur: this sounds a lot like what we experienced with Travis's Linux build service. Our main experience with Travis was:

 - Builds took in excess of half an hour to even start to run
 - Builds were running slowly
 - (seriously, what would you expect of a free service that wants to sell its premium services?)
 - When you updated your pull request, Travis would stubbornly run the already-enqueued build, then start scheduling your update.
 - Travis was using (and probably still is, I won't even check to avoid throwing a tantrum) a by-now 4-year old Linux distribution which was becoming more and more of a pain to maintain with every passing day. Openage is a bleeding-edge project and requires the most recent programming language and compiler editions. This is the most evil feature of Travis because it made me waste literally hundreds of hours backporting packets in hopes that they would start offering recent Linux machines any time now. I'm not sure about whether the same issue will develop with their macOS machines, we will see with time. Damnit, threw the tantrum anyways.

Anyways, at some point I decided that the effort for completely re-writing Travis in a sane manner as a self-hosted solution would be less time-consuming and **a lot** more fun and lead to massively improved build times, and thus kevin-ci was born.

kevin-ci is capable of running macOS builds, all that it needs is a disposable Virtual Machine running said macOS. And thus the self-hosting issue: Nobody of us actually owns a Mac machine, and we're left with several sub-optimal options.

- Use a Hackintosh VM running on one of our Linux boxes. @TheJJ seems to think that this is very hard to achieve from a technical viewpoint, but I disagree. I believe it would take maybe 30 hours of work, less if assisted by someone who has experience with Hackintosh. Unfortunately, to my knowledge this is copyright infringement. EULAs have no power here in Germany, but I'd still have a really bad feeling about it. Maybe someone could provide some qualified legal advice?
- Pay some cloud provider to get a macOS machine that is launched on demand. This would probably be the easiest solution, but the build speeds would be sub-optimal and we'd have to pay some amount of money.
- Travel to some garbage disposal site and recover some 5-year-old macBook with a broken display and destroyed keyboard, connect it to the Internet and use it as our VM host. Downsides:
 - I don't like hanging around at garbage dumps
 - Will the current version of macOS run as a VM on the old machine?

I agree that we need some sort of solution though; the Mac build is breaking far too often.

janisozaur:
I completely understand your frustration, I have hit pretty much the same roadblocks along my way. Let me tell you how I have addressed it and what recent events have come to light.

* We have a total of 8 jobs per build
  * 2 with GCC from repo (native and multilib cross-compiler for 32 bit builds (we need them, don't judge us))
  * 1 with clang from repo 
  * 3 with GCC in docker
  * 1 with MinGW from repo
  * 1 with OS X (I don't maintain this one, but it isn't much different than the other ones)
* We use exclusively Ubuntu Trusty (14.04) in our Linux jobs, it is dated but much better than 12.04.
* Since I have no regard for Ubuntu/Debian and its ancient packages, I have set up a docker where I have installed minimal Arch.
  * The image is stored on https://hub.docker.com/ and downloaded on every commit
  * The image is rebuilt every time, but by hub.docker.com itself, they offer automated builds. It works fine, but requires some more work when a new dependency, namely changing the `Dockerfile` first, waiting till hub updates the image, re-triggering build in question.
  * Such image is less than 500MiB and downloading it takes less time than setting up all the packages we need on Ubuntu. It could probably be cached, but I haven't bothered setting it up.
  * I can thus ensure nobody breaks build on my rolling release.
* Ubuntu can actually easily install a much more up-to-date compiler, see https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test?field.series_filter=trusty
* Travis-CI builds for Linux jobs generally start within a reasonable time, up to 5 in parallel. In our case I have ordered longer running jobs earlier, resulting in larger overall throughput.
* Unlike AppVeyor (IRC), Travis-CI cannot discard old queued jobs from a PR that received new commits, but there are some 3rd party tools that address this problem: https://github.com/grosser/travis_dedup
  * In our case large number of developers use inferior systems and need help from GCC/clang (with `-Wall` and some others), I have not pursued this, so they get feedback on every push.
  * Usually a contention of builds happens in the "working hours" of our team, so we manually cancel stale builds if need be.
* Travis team have [recently (two days ago) provided an update](https://docs.travis-ci.com/user/build-environment-updates/2016-11-01/) where they start offering container-based images of Ubuntu 14.04, these container images are said to start much faster, but have some other restrictions: `sudo` is not available and only whitelisted set of PPAs and packages can be preinstalled on them.
  * If you need something not on such whitelist, you can request its inclusion or build and cache.
* Travis-CI's OS X images are [much more up-to-date](https://docs.travis-ci.com/user/osx-ci-environment/#OS-X-Version) than the Linux ones
  * Installing all dependencies with `brew` works, but may take a long time, especially if you use Qt. I recall recently seeing a script installing the official Qt binary, I can look around for it if you need it - let me know.

Other than the list above, I can offer some suggestion I have not yet tried, but may be viable:
* OS X can be cross-compiled to. See https://github.com/tpoechtrager/osxcross
* For speeding up builds, it may be worth checking out https://github.com/iboB/mtime_cache
* I haven't yet checked how well the new containerised Trusty image works, but it sounds interesting.
* Track https://github.com/travis-ci/travis-ci/issues/5821 to get updates on 16.04 availability. 12.04 has [less than half a year until EOL](https://www.ubuntu.com/info/release-end-of-life) now, I honestly believe Travis will have an answer by then.
* Judging purely by the number of commits (and transitively by amount of builds triggered), [our repo](https://github.com/OpenRCT2/OpenRCT2) is much larger than yours, you surely can cope with Travis ;)
* Depending on nature of the breakage on macOS (compilation error? linkage error? missing dependency?) you may partially alleviate the problem by turning all of `-Wall -Wextra` on Linux builds, though I imagine you already do that.

dusek:
A few comments/observations inspired by what has been said or what I achieved:
- I managed to fully remove the delay caused by compiling numpy (previously ~5 minutes) and Pillow (previously ~1 minute), by not installing them with `brew` but with `pip3`, which downloads pre-built binaries
- I managed to reduce the build time (i.e. `make`) from ~434s by using `make -j2`  to 279s; `make -j4` did not yield anything significant (~250s)
- with the 2 optimizations; the whole [build](https://travis-ci.org/dusek/openage/builds/172864412) now takes ~11 minutes (6 of which is dependency installation, remaining 5 the actual build); `fc-cache ...` takes ~2.3 minutes of the dependency installation; one sample, times do vary
- that said, Travis boot times can be an issue - yesterday evening (CET), I waited ~half an hour for a build to start; this morning, they started immediately (maybe they have most of their customer in the US?)
- Travis provides [Trusty VMs](https://blog.travis-ci.com/2015-10-14-opening-up-ubuntu-trusty-beta/), as of October 2015
- Travis AFAIK should have pre-installed latest versions of compilers, languages (Python) in the VM independent of the latest version in the distribution itself (e.g. latest clang, gcc, python, ruby, etc.); AFAIK with the Trusty image, they no longer have separate images based on what language one specified in `.travis.yml`
- macOS-wise, they seem to publish an image with a new version of Xcode a few days / max. weeks after it has been published; they even provide images with beta versions of Xcode during the beta testing period; the Xcode version is selectable in `.travis.yml` (the [`osx_image`](https://docs.travis-ci.com/user/osx-ci-environment/#OS-X-Version) key)
- to me as an outsider of the project, without meaning any disrespect, just presenting my impression, developing, running and maintaining a custom CI seems a little bit excessive
- Travis Linux can run docker, so that could solve the pre-installed dependencies if Travis for Linux is ever considered again
- [macOS EULA](http://images.apple.com/legal/sla/docs/macOS1012.pdf) handles virtualization in 2.B.III.; clearly running on non-Apple hardware or on Apple hardware currently running non-macOS OS as host OS is breaking of an EULA; there are other conditions that make it not clear for cases like using cloud hosting; I have not looked into that; maybe that is also one of the reasons why some people choose either their own office 24/7 Mac, or a service like Travis that has studied and sorted all the legal things on their side
- maybe some Mac cloud hosting company could sponsor openage; I looks like [OpenTTD](http://www.openttd.org) is somehow sponsored by OVH for some services, in exchange for displaying OVH logo on their site etc.; I bet such a Mac cloud hosting company must have studied the licensing and could answer questions on what is acceptable use (maybe they have some volume license of macOS with less restrictions), [MacStadium](http://macstadium.com) even advertise CI as a use case.
- I think the macOS situation w.r.t. binary package distribution is different when compared to Linux: apt-get is the official (and usually the only feasible) source of installing binary packages of anything on Linux; being official, they care about stability and keep the versions of packages frozen after the OS release; on the other hand, Homebrew on Mac is an unofficial package management system and provides always the latest versions of packages (provided there is some volunteer to package the new version, which is most often the case unless the package is very rarely used). So `apt-get install` on Linux would often install old version of dependent libraries, whereas `brew install` on Mac would usually install the latest version of dependent libraries; maybe this is a reason why the Travis build would be less of a PITA on macOS than on Linux

> But I really don't understand why the cloudy Apple doesn't allow/support running mac VMs on non-macs.

I guess they want more people to buy Macs 😄 

My overall impression is that Travis is an excellent simplification of CI worth some of its flaws.

dusek:
> Installing all dependencies with brew works, but may take a long time, especially if you use Qt. I recall recently seeing a script installing the official Qt binary, I can look around for it if you need it - let me know.

brew nowadays installs binaries ("bottles"); [Qt install in the Travis build of this PR](https://travis-ci.org/dusek/openage/builds/172864412#L588) downloads a binary too and takes ~1 minute, looks like unpacking the .tar.gz with the pre-built installation is what takes all the time.

> OS X can be cross-compiled to. See https://github.com/tpoechtrager/osxcross

I would a bit careful with solutions like these, I have not tested them but I would be worried whether they really provide an exact and up-to-date replica of the original Mac build environment.

janisozaur:
I recall installing Qt 5.5 from bottle with brew on Travis, it took ~10 minutes. I think this (Qt 5.7) was actually addressed by one of our team members.

Even though the compiler may be different, you can use proper OS X stuff: https://github.com/phracker/MacOSX-SDKs. We used to package it (I suggest `.tar.lzo` for quick decompression), host it on GitHub in a separate repo with prebuilts, then download, deploy and consume on Travis.

Side note: `brew` is a really silly name and makes it unnecessary hard to google.

Birch-san:
In case it's of interest: using the branch in this pull request (commit ccdbaeca13f9216673e83d0e57e4c6366c91e631 at the time of writing), I was able to successfully build on OS X using the standard steps.

Current theory is that the build problems I experienced before (#643) were caused by my using qt 5.6.

TheJJ:
So, what is our conclusion here? As long as the boot up time at travis can take half an hour or more in the evening, it's a no-go for me.
We should ask some mac-hoster if it sponsored a machine for us so we can deploy kevin/falk on it, so it can do our build like on our linux boxes. That seems the best way to go now (macstadium.com or ovh i presume).

TheJJ:
If somebody comes up with relations or motivation to care for a macOS build in some way that performs well, you're welcome to help. The "ideal" solution would be some machine where we have root access and can run our own kevin build VM then. A donated mac-mini maybe?

PR #659, opened 2016-10-30
GUI: Warn about possible missing includes

ChipmunkV:
Unwrapping of the GUI binding classes like `SomeClassLink` into `SomeClass` uses type trait `Unwrap` that is defined in the header with corresponding type. If the header is not included, then compiler considers it as a basic type and tries to pass it wrapped to a function that expects an unwrapped type. To warn about that possibility, PR adds a compile-time error message if the arguments don't match for any reason.


TheJJ:
oh c++, please have mercy with us. we might not understand your allmighty power yet, but need to create template fuckery at each and every corner to keep you under control. we are not worthy.

PR #660, opened 2016-10-31
Update build procedure #549

### LINKED ISSUE #549: Update build procedure
https://github.com/SFTtech/openage/blob/master/doc/building.md#build-procedure
- [ ] The `./configure` script no longer accepts `--cpp-compiler`
- [ ] The compiler versions should not be hardcoded, but it's probably difficult writing it in another way
- [ ] Create subfiles for all the distros and operating systems, so the file is less clobbered

edgardmota:
Performed as discussed on the related thread.

fixes #549

### DIFF
--- a/doc/building.md
+++ b/doc/building.md
@@ -54,15 +54,15 @@ Dependency list:
 
 ### Specific Prerequisite steps per Platform
 
-There are some prerequisite steps that need to be performed so *openage* can be succesfully build. Those steps vary from platform to platform, and are described bellow for some of the most common ones:
+There are some prerequisite steps that need to be performed so *openage* can be built successfully. Those steps vary from platform to platform, and are described bellow for some of the most common ones:
 
-- [Prerequisite steps for Ubuntu users (Ubuntu 15.04)](build_instructions/ubuntu_15.04.md);
-- [Prerequisite steps for Fedora users (Fedora 20, 21)](build_instructions/fedora_20_21.md);
-- [Prerequisite steps for Fedora users (Fedora 22)](build_instructions/fedora_22.md);
-- [Prerequisite steps for openSUSE users (openSUSE 13.2)](build_instructions/opensuse_13.2.md);
-- [Prerequisite steps for openSUSE users (openSUSE Tumbleweed)](build_instructions/opensuse_tumbleweed.md);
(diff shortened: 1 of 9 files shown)


doc/building.md
```diff
@@ -52,76 +52,17 @@ Dependency list:
...
-This command should provide required packages for Arch Linux installation:
-
-`sudo pacman -S --needed python python-pillow python-numpy python-pygments cython libepoxy ttf-dejavu freetype2 fontconfig harfbuzz cmake sdl2 sdl2_image opusfile opus-tools python-pylint qt5-declarative qt5-quickcontrols`
-
-If you don't have a compiler installed, you can select between these commands to install it:
- - `sudo pacman -S --needed gcc`
- - `sudo pacman -S --needed clang`
-
-You can install both compilers and select the one to be used by `./configure`.
+There are some prerequisite steps that need to be performed so *openage* can be succesfully build. Those steps vary from platform to platform, and are described bellow for some of the most common ones:
```
TheJJ: "so openage can be built successfully" :)

doc/building.md
```diff
@@ -52,76 +52,17 @@ Dependency list:
...
 
+- [Prerequisite steps for Ubuntu users (Ubuntu 15.04)](build_instructions/ubuntu_15.04.md);
+- [Prerequisite steps for Fedora users (Fedora 20, 21)](build_instructions/fedora_20_21.md);
+- [Prerequisite steps for Fedora users (Fedora 22)](build_instructions/fedora_22.md);
+- [Prerequisite steps for openSUSE users (openSUSE 13.2)](build_instructions/opensuse_13.2.md);
+- [Prerequisite steps for openSUSE users (openSUSE Tumbleweed)](build_instructions/opensuse_tumbleweed.md);
+- [Prerequisite steps for Mac OS X users (OS X 10.10 Yosemite)](build_instructions/os_x_10.10_yosemite.md);
+- [Prerequisite steps for Arch Linux users](build_instructions/arch_linux.md).
```
TheJJ: i think it's best to just write it like

```
- [Ubuntu (>=15.04)](...)
```

because the "users" and "prerequisite steps" is very redundant, the `;` or `.` is also not needed i'd say.

TheJJ:
aaaaaand it's live!

edgardmota:
Wow... proud of it, thanks! :laughing:

PR #664, opened 2016-11-03
Create FilterMaps_dat.md

yvan-burrie:


TheJJ:
Can you please place the file in either `doc/media` or `doc/reverse_engineering`, whereever it suits better?

TheJJ:
Do you have any idea what this is used for? What could it possibly "filter"?

yvan-burrie:
FilterMaps.dat contains 17 rhombus tiles, each mostly 49 rows high, with each row having a stride of columns as pixels, similar to Blendomatic tiles. Each pixel contains a reference to an ICM which uses corresponding LightMaps for shading. I still need to research more, however, it is very complicated since the developers obviously used ASM to code the entire graphics rendering which accounts for all those extra DAT files.

Besides, if Openage uses 3D terrain, then FilterMaps.dat is not useful.

PR #665, opened 2016-11-03
Developing with Qt Creator

ChipmunkV:
Following a suggestion by @TheJJ.


TheJJ:
rebase pls, the sanity check is fixed.

PR #668, opened 2016-11-11
Show current and required media version

### LINKED ISSUE #615: Show current and required media version
When a new media version is required, it just says "hey, reconvert" currently.
It should display the found media version and the new required version and what is going to be converted ("interface", "sound", ...).

kev946:
Fixes #615

### DIFF
--- a/openage/convert/main.py
+++ b/openage/convert/main.py
@@ -394,7 +394,7 @@ def conversion_required(asset_dir, args):
 
     else:
         if asset_version >= 0:
-            info("Converted assets outdated: Version %d" % asset_version)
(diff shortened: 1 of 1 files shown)


kev946:
I'm not sure if this is what we want

TheJJ:
I think this is a bit too complicated, but also not bad. I was more thinking of a message like "found media of version x, but now we need version y, which is why we need to convert {...}". The version of each component isn't relevant for the enduser and also not for debugging.

kev946:
Ok I'll make those changes when I get a chance

kev946:
example output:
INFO [py] Found converted assets with version 2, but need version 4
INFO [py] Converting interface

TheJJ:
merged with ca11d8f

Issue #551, opened 2016-05-05
Development cycle advice?


coffenbacher:
Is this the best way to make changes?

1. Make change in .cpp
2. make
3. ./run
4. Navigate to "Generate Game" and keep pushing it until assets have loaded
5. Test behavior

Just wondering, because I've done this about 50 times tonight, and if there are any recommended shortcuts (filesystem watch to make? launch with flags to automatically load game after assets load?) that would be good to know :+1: I'm on Ubuntu.

(Forgive me, I come from interpreted language world where this is a different ballgame.)

coffenbacher:
Well, installing a debugger helped enormously, since I don't have to rebuild so often. Any other helpful tips?

TheJJ:
My development cycle is using tests/demos.
e.g. for working on the renderer: `make && ./run test -v --demo renderer.tests.renderer_demo 1`

You can add arbitrary amounts of tests and demos to try something you're working on, which can later directly be used to test if it still works. The tests are non-interactive and just test some algorithm/procedure/whatever. The demos are interactive and used for e.g. displaying some window and working in it.

That demo/test can also be run in gdb then to debug things.
The demos/tests bypass the "normal" main function, so all the asset loading is not required. You have to design the function in a way you can test your modification, of course (e.g. show window, display unit, ...)

coffenbacher:
Thanks @TheJJ, good suggestions. I'll start trying to move in that direction as I get a handle on the codebase.

TheJJ:
Cool!

PR #669, opened 2016-11-12
GUI: Start game after assets have loaded

ChipmunkV:
Checkbox to generate the game and enter the ActionMode as soon as the assets are loaded.

Related to #551: click on the checkbox once instead of clicking the generate_game until it loads.

For more specific stuff it's probably useful to figure out how to write demos (like a demo to record replay and a demo to play it). But the autolaunch feature looks generic enough to be added to the generator's menu.

### DIFF
--- a/libopenage/game_control.cpp
+++ b/libopenage/game_control.cpp
@@ -19,6 +19,7 @@ void OutputMode::announce() {
 
 void OutputMode::set_game_control(GameControl *game_control) {
 	this->game_control = game_control;
+	this->announce();
 }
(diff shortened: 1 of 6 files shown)


TheJJ:
probably a good thing to have for now. later we'll have to trigger the creation manually in order to go to the editor and load existing games.

ChipmunkV:
It's still manual, but needs less clicking.

TheJJ:
ah, i misread it. it's about having action mode right away.

Issue #651, opened 2016-10-22
Split the hitpoints attribute


MaanooAk:
The one attribute must be shared and have the max hp (and possibly the size of the health bar)
The other attribute not shared and have the current hp

I am happy to implement it, but I need suggestions for the names of the two attributes.

elnabo:
How are the converted units taken into account for the max hp ?

MaanooAk:
Converted unit are not implemented yet and in the current system the cant (because no attribute could be shared). If the converted unit must not be affected by feature upgrades, a new unit type should be created. This applies even if we don't split the hitpoints attribute.
Check the unit_type.h and keep in mind that there are more attributes that need splitting.

TheJJ:
The units only store the concrete values for that one entity. This means their current hp, position, etc. The "class" information, i. e. information about the unit type, e.g. villager, militia, whatever are stored in nyan. Nyan is also used to handle the monk conversion by "baking" the unit type and assigning the converted unit to it. This reflects the original game behavior that converted units can't be upgraded any more.

TheJJ:
And to come back to your original question: The Unit must have a `hp` value and (just temporarily until nyan is merged) a `unit_type_max_hp` value which is shared and then replaced by the nyan object lookup.

MaanooAk:
`unit_type_max_hp` is quite long but ok, (I will get on it when i have some time)

TheJJ:
If you don't like it, chose another one suitable of course :D

MaanooAk:
This was done with #683

Issue #366, opened 2015-08-25
New pathfinding and grouping


zesterer:
Presumably you use some form of grid-based A\* algorithm? Maybe it is better to 'group' nearby searches to reduce the stress of re-calculating paths that are very similar.

TheJJ:
I plan improving the pathfinding after i finished the new renderer. It'll be a D*lite search with jump point search and for local optimization http://grail.cs.washington.edu/projects/crowd-flows/.
If anyone else wants to try their luck first, go for it.

zesterer:
I recommend you try out a game called 'Liquid War'. It's FOSS, and is in most mainstream distro repositories. It involves a game where you control this pathfinding fluid (composed of tens of thousands of entities) that move towards the mouse. Perhaps this would be a good bit of inspiration? Liquid War does it all at 60 fps, and often with very high resolution grids, so perhaps taking a peek within it's source code might be worthwhile?

TheJJ:
Never heard of it, but it [looks cool](https://www.gnu.org/software/liquidwar6/) and has probably nice inspiration for some areas.

Jon0:
Having a few modular path finding systems would be interesting, while having one path finding system as close to the original is good, it would be cool having many options like these newer techniques, which may considerably affect gameplay / balance

TheJJ:
The thing most complained about by aoe2 players i think is pathfinding bugs and especially blockages. So i guess we can try doing something awesome and wait for feedback. Imagine your huge armies just flowing through each other naturally without blocking :cake:

A simple approach to that would be that each unit has a discomfort hill in front of it. Other units get "pushed" out of the way as they wanna avoid that hill (the hill size is dependent on the unit speed i'd say).

The sum of all 2d discomfort functions is chosen for taking the path with the lowest discomfort, namely avoiding the hills. Might work like a charm, we have to limit the range drastically of course.

zesterer:
I think it's important with large groups with the same goal to preserve some kind of locality and flocking behaviour. If they are unable to share data among one-another, substantial optimisation will be impossible.

TheJJ:
http://www.gameaipro.com/GameAIPro/GameAIPro_Chapter23_Crowd_Pathfinding_and_Steering_Using_Flow_Field_Tiles.pdf

This is more or less the implementation design I was trying to come up with, but with awesome portals between the chunks. It's probably what we want!

some other example: https://vonwolfehaus.github.io/flow-field/

coffenbacher:
[...]

PR #672, opened 2016-11-27
Repair action & resource bundle

MaanooAk:
Repair of units based on the [doc/reverse_engineering/repair.md](doc/reverse_engineering/repair.md) (file in pull request).

In order to continue:
* ~~A data structure for multiple resource costs should be implemented (game_resource_bundle ?)~~
* The cost of the unit should be stored somewhere (inside unt_type ?)

~~Can I implement these? or just leave as it is and revisit after the nyan revolution?~~

EDIT: Also game_resource_bundle should replace the `game_resource type, int value` couples all over the place.

### DIFF
--- a/libopenage/unit/ability.cpp
+++ b/libopenage/unit/ability.cpp
@@ -322,7 +322,7 @@ bool HealAbility::can_invoke(Unit &to_modify, const Command &cmd) {
 }
 
 void HealAbility::invoke(Unit &to_modify, const Command &cmd, bool play_sound) {
-	to_modify.log(MSG(dbg) << "invoke repair action");
+	to_modify.log(MSG(dbg) << "invoke heal action");
 	if (play_sound && this->sound) {
 		this->sound->play();
 	}
--- a/libopenage/unit/ability.cpp
+++ b/libopenage/unit/ability.cpp
@@ -16,6 +16,11 @@ bool UnitAbility::has_hitpoints(Unit &target) {
 	       target.get_attribute<attr_type::hitpoints>().current > 0;
 }
 
+bool UnitAbility::is_damaged(Unit &target) {
+	return target.has_attribute(attr_type::hitpoints) &&
+	       target.get_attribute<attr_type::hitpoints>().current < target.get_attribute<attr_type::hitpoints>().max;
+}
+
 bool UnitAbility::has_resource(Unit &target) {
 	return target.has_attribute(attr_type::resource) &&
 	       target.get_attribute<attr_type::resource>().amount > 0;
@@ -268,6 +273,34 @@ void AttackAbility::invoke(Unit &to_modify, const Command &cmd, bool play_sound)
 
 	Unit *target = cmd.unit();
 	to_modify.push_action(std::make_unique<AttackAction>(&to_modify, target->get_ref()));
+}
+
+RepairAbility::RepairAbility(const Sound *s)
+	:
+	sound{s} {
+}
+
+bool RepairAbility::can_invoke(Unit &to_modify, const Command &cmd) {
+	if (cmd.has_unit()) {
+		Unit &target = *cmd.unit();
+		return &to_modify != &target &&
+		       to_modify.location &&
+		       target.location &&
+		       target.location->is_placed() &&
+		       is_damaged(target) &&
+		       is_ally(to_modify, target);
+	}
+	return false;
+}
+
+void RepairAbility::invoke(Unit &to_modify, const Command &cmd, bool play_sound) {
+	to_modify.log(MSG(dbg) << "invoke repair action");
+	if (play_sound && this->sound) {
+		this->sound->play();
+	}
+
+	Unit *target = cmd.unit();
+	to_modify.push_action(std::make_unique<RepairAction>(&to_modify, target->get_ref()));
 }
 
 ability_set UnitAbility::set_from_list(const std::vector<ability_type> &items) {
--- a/libopenage/unit/ability.cpp
+++ b/libopenage/unit/ability.cpp
@@ -385,7 +385,7 @@ bool ConvertAbility::can_invoke(Unit &to_modify, const Command &cmd) {
 }
 
 void ConvertAbility::invoke(Unit &to_modify, const Command &cmd, bool play_sound) {
-	to_modify.log(MSG(dbg) << "invoke repair action");
+	to_modify.log(MSG(dbg) << "invoke convert action");
 	if (play_sound && this->sound) {
 		this->sound->play();
 	}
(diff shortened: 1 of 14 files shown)


MaanooAk:
ResourceBoundle could be like [this (gist)](https://gist.github.com/MaanooAk/58cded6e8ac2fac9a48f66eda0630aba). Check the [main](https://gist.github.com/MaanooAk/58cded6e8ac2fac9a48f66eda0630aba#file-resource-h-L95) for how its used.

MaanooAk:
Implemented a ResourceBundle class and used it in player (kept old one with raw type and amount) and unit repair.

Next step is to add somewhere (unit type, unit attributes, ?) the cost of the unit. I have zero confidence implementing that so **this is as far I can go** (= removing WIP).

TheJJ:
Sorry for the delay, university has kept me very busy (http://www.move2space.de/MOVE-II/)...

However, I just tested your code and I couldn't get my villagers to repair their town center or mill.

The resource bundle is a good idea, as it's capable of adding resource types at runtime easily. We'll need that in the future.

The cost of the unit is drawn from the gamedata of course, which is accessible through the `std::vector<gamedata::empiresdat> gamedata;` in `libopenage/gamestate/game_spec.cpp`. But this is all going to be replaced by nyan. But nyan is still quite far from being ready so adding just another dirty hardcoding for it is ok. You'll find the structure of that struct in `openage/convert/gamedata` where we extract it.

MaanooAk:
I added the repair ability, that should do the work. I added and some other missing abilities while I was there.
And I implemented the repair gui button.

libopenage/unit/ability.cpp
```diff
@@ -270,6 +275,125 @@ void AttackAbility::invoke(Unit &to_modify, const Command &cmd, bool play_sound)
...
+		       to_modify.location &&
+		       target.location &&
+		       target.location->is_placed() &&
+		       is_enemy(to_modify, target);
+	}
+	return false;
+}
+
+void ConvertAbility::invoke(Unit &to_modify, const Command &cmd, bool play_sound) {
+	to_modify.log(MSG(dbg) << "invoke repair action");
```
TheJJ: copypasta :)

libopenage/unit/ability.h
```diff
@@ -264,6 +265,103 @@ class AttackAbility: public UnitAbility {
 	const Sound *sound;
 };
 
+/*
+ * initiates a repair action when given a valid target
+ */
```
TheJJ: pls write `/**` so it's parsed by doxygen

libopenage/unit/ability.h
```diff
@@ -264,6 +265,103 @@ class AttackAbility: public UnitAbility {
...
+
+	void invoke(Unit &to_modify, const Command &cmd, bool play_sound=false) override;
+
+private:
+	const Sound *sound;
+};
+
+/*
+ * initiates a heal action when given a valid target
+ */
```
TheJJ: same here

libopenage/unit/ability.h
```diff
@@ -264,6 +265,103 @@ class AttackAbility: public UnitAbility {
...
+
+private:
+	const Sound *sound;
+};
+
+/*
+ * initiates a research
+ * TODO implement
+ */
+class ResearchAbility: public UnitAbility {
```
TheJJ: and here

libopenage/unit/ability.h
```diff
@@ -264,6 +265,103 @@ class AttackAbility: public UnitAbility {
...
+
+private:
+	const Sound *sound;
+};
+
+/*
+ * initiates a patrol action when given a valid target
+ * TODO implement
+ */
+class PatrolAbility: public UnitAbility {
```
TheJJ: and here

libopenage/unit/ability.h
```diff
@@ -264,6 +265,103 @@ class AttackAbility: public UnitAbility {
...
+	void invoke(Unit &to_modify, const Command &cmd, bool play_sound=false) override;
+
+private:
+	const Sound *sound;
+};
+
+/*
+ * initiates a convert action when given a valid target
+ */
+class ConvertAbility: public UnitAbility {
```
TheJJ: and here

libopenage/unit/action.h
```diff
@@ -419,6 +420,11 @@ class RepairAction: public TargetAction {
 private:
 	bool complete;
 
+	float time;
+	float time_left;
+
+	ResourceBundle cost; // calculate and store the cost in the constructor
```
TheJJ: you can write `/** stores the cost of the repair */` or whatever is more suitable above it so it's parsed by doxygen, too.

MaanooAk:
The separate namespace should be done with a refactoring tool.
However I think the way attributes are defined is nice too.

TheJJ:
You mean templating the attributes? Yes, could also be done :)
Do whatever you like better and think suits us more.

Shall I merge your stuff before that or do you wanna integrate it also in this pull?

MaanooAk:
You should merge this (do you want me to rebase?)
I'm thinking of fixing the attributes first (more like #651) if that's ok.

TheJJ:
No I don't think you need to rebase, it doesn't conflict with the `master`.
@ attributes, sure go ahead! It's another basic foundation for the game logic.

MaanooAk:
I just realized I'm creating a ton of conflicts in my new pr (#683) with this one...
Can you merge this or you are not sure yet, can I do something?

libopenage/unit/ability.cpp
```diff
@@ -270,6 +275,125 @@ void AttackAbility::invoke(Unit &to_modify, const Command &cmd, bool play_sound)
...
+		       target.location &&
+		       target.location->is_placed() &&
+		       is_damaged(target) &&
+		       is_ally(to_modify, target);
+	}
+	return false;
+}
+
+void HealAbility::invoke(Unit &to_modify, const Command &cmd, bool play_sound) {
+	to_modify.log(MSG(dbg) << "invoke repair action");
```
TheJJ: still the wrong message here

MaanooAk:
I see... Just leave it, I will investigate and come back to you.
(May take some time, I have some serious Ubuntu craziness happening right now)

MaanooAk:
Added the ability, added 1000 of each res to players (so they can repair) and made gaia the Gaia player (prev: name1 was Gaia and gaia was some civ)

screenshot from 2016-12-30 22-45-34

Yanikore:
I'd like to add to this that there's a bug in the original game which still exists where you can repair (and garrison) a building from a diagonal location. It's called the wall corner glitch and I'm wondering if this bug also exists within openage and if it should either be fixed, or stay exactly the same.

TheJJ:
@Yanikore I think we'll have the same problem right now because we don't have any blocking physics currently, the pathfinder is really stupid and doesn't account for that. Go and help out at #366 :)

Yanikore:
@TheJJ I'd love to help out but I'm not capable for the C++/python side of this. I'm more a PHP developer than anything so I'll probably help out in the future when a website gets made. Right now I can only add some conceptual ideas to the situation.

TheJJ:
yay it works :)

PR #674, opened 2016-12-08
convert: read African Kingdoms expansion data when available

goto-bus-stop:
HD Edition 4.7 (and maybe earlier, I haven't yet checked) have two data files: `empires2_x1_p1.dat` and `empires2_x2_p1.dat`. Currently the convert script uses `x1`, but the new African Kingdoms units and techs are only present in `x2`. The data format for x2 is the same as x1. The only difference is that x2 has 100 terrains, instead of 42.

This patch uses `empires2_x2_p1.dat` when it is available. The convert script knows how many terrains to read by passing the `game_versions` list to the gamedata reader classes. I also added a bunch of unknown enum values that weren't used in `x1` but are used in `x2`.

`ArrayMember` was so far unused so I changed it to work with the existing code for reading raw C types. `ArrayMember("uint16_t", 10)` is the same as `uint16_t[10]`, but can have a dynamic length.

TODO:

 - [x] Check which exact versions have `empires2_x2_p1.dat`


Yanikore:
4.8 just came out and the x2 data file was updated. Not sure if that changes anything but you might have to check it out.

TheJJ:
I doubt they updated the structure (supporting 100 terrains) but instead only changed the contents. Would be cool if you could check if this branch still converts the 4.8 beta files.

goto-bus-stop:
Yeah. I forgot I was actually running this with 4.8 already, so it certainly works for the 4.8 beta.

https://github.com/tapsa/genieutils also supports both _x2 and _x1 files, and the only difference it implements is the terrain count.

I tried running it with HD Edition 4.3 resources just now, and it doesn't work yet, so I'll have a look at that :see_no_evil:

goto-bus-stop:
Okay, added an extra check for African Kingdoms data, and this works for me for HD Edition 4.3 and 4.7 and 4.8. I think this is ready to go now.

TheJJ:
i really wonder how the developers of aoe:hd can take that funny dat file for so many years without constant loss of manpower because of suicides.

PR #676, opened 2016-12-13
Purge engine singleton

### LINKED ISSUE #329: Purge engine singleton

TheJJ:
Fixes #329.

The decision to have a global engine object was really really bad. It lead to wrong design in some subsystems as the engine could just be reached directly.

TODOs:
* [x] Fix in-game crashes

### DIFF
--- a/buildsystem/codecompliance/pystyle.py
+++ b/buildsystem/codecompliance/pystyle.py
@@ -5,9 +5,9 @@ Checks PEP8 compliance, with some exceptions.
 """
 
 try:
-    from pep8 import StyleGuide
-except ImportError:
     from pycodestyle import StyleGuide
+except ImportError:
+    from pep8 import StyleGuide
 
 
 # these errors will be ignored by pep8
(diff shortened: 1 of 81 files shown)


TheJJ:
Follow-up will be #662 so the global coordinate state struct is purged.

PR #683, opened 2016-12-29
Attributes system improvements

MaanooAk:
### Attributes system improvements goals:

- [x] Abstract shared and ushared attributes
  - [x] share method
  - [x] ~~copy method (???)~~
- [x] Convert `attr_map_t` into a proper class (`Attributes`?)
  - [x] implement
    - [x] add `attributes.cpp` (?)
  - [x] refactor
- [x] Split attributes and refactor changes
  - [x] hitpoints -> hitpoints + current hitpoints (issue #651)
  - [x] building -> building + dropsite (semi-implemented)
  - [x] attack: move attack stance elsewher
- [x] Change the worker atrribute
  - [x] rename
  - [x] split into unshared and shared
    - [x] carrying resources
    - [x] move worker action graphics to somewhere else ($1)
  - [x] gather rate per resource
  - [x] move attack type (used only for villagers) from attack ($2)
- [x] Multitype attribute (for unit with subtypes: villages, trebuchets)
  - [x] implement
  - [x] refactor producers and fix $1 and $2
- [x] Change type bugs
  - [x] use `reinitialise` instead of `reinitialise` in order to keep the unshared attributes (eg. life)
- [x] ~~Actions attribute calls optimize~~
  - [x] ~~move has/get attribute in constructor (call once and store) (?)~~
  - [x] ~~move basic operations inside attributes~~
- [x] Add doc

### Notes

- All files including `attributes.h`: `unit_type.h`, `unit.h`, `action.h`
- `Attributes` methods:
  - bool add/has/removeAttribute (hide all map methods)
  - Attribute get<attr_type>

### Bugs found (and solved) on the way

- Health reset every time villager changes resources (eg. from wood to stone) (wait... no)
- One gather rate for all resources
- Search for new building to build includes enemy's building
- Towers auto attack
- ...

### DIFF
--- a/libopenage/unit/attribute.h
+++ b/libopenage/unit/attribute.h
@@ -128,22 +128,56 @@ template<attr_type T> Attribute<T> get_attr(attr_map_t &map) {
 	return *reinterpret_cast<Attribute<T> *>(map[T]);
 }
 
+/**
+ * Wraps a templated shared attribute
+ *
+ * Shared attributes are common across all units of
+ * one type
+ */
+class SharedAttributeContainer: public AttributeContainer {
+public:
+
+	SharedAttributeContainer(attr_type t)
+		:
+		AttributeContainer{t} {}
+
+	bool shared() const override {
+		return true;
+	}
+
+};
+
+/**
+ * Wraps a templated unshared attribute
+ *
+ * Shared attributes are copied for each unit of
+ * one type
+ */
+class UnsharedAttributeContainer: public AttributeContainer {
+public:
+
+	UnsharedAttributeContainer(attr_type t)
+		:
+		AttributeContainer{t} {}
+
+	bool shared() const override {
+		return false;
+	}
+
+};
+
 // -----------------------------
 // attribute definitions go here
 // -----------------------------
 
 class Player;
 
-template<> class Attribute<attr_type::owner>: public AttributeContainer {
+template<> class Attribute<attr_type::owner>: public SharedAttributeContainer {
 public:
 	Attribute(Player &p)
 		:
-		AttributeContainer{attr_type::owner},
+		SharedAttributeContainer{attr_type::owner},
 		player(p) {}
-
-	bool shared() const override {
-		return false;
-	}
 
 	std::shared_ptr<AttributeContainer> copy() const override {
 		return std::make_shared<Attribute<attr_type::owner>>(*this);
@@ -152,17 +186,13 @@ public:
 	Player &player;
 };
 
-template<> class Attribute<attr_type::hitpoints>: public AttributeContainer {
+template<> class Attribute<attr_type::hitpoints>: public UnsharedAttributeContainer {
 public:
 	Attribute(unsigned int i)
 		:
-		AttributeContainer{attr_type::hitpoints},
+		UnsharedAttributeContainer{attr_type::hitpoints},
 		current{i},
 		max{i} {}
-
-	bool shared() const override {
-		return false;
-	}
 
 	std::shared_ptr<AttributeContainer> copy() const override {
 		return std::make_shared<Attribute<attr_type::hitpoints>>(*this);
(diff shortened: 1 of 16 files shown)


MaanooAk:
After a lot of time, I cant find a way to improve the the copypaste `copy()` methods.

MaanooAk:
Issues:
- Only attribute left to split is `attack`. It has the attack stance inside which has to go somewhere else. Any suggestions on grouping attack stance with something else in order to avoid a whole attribute only for this?
- Many of the actions' update methods have many calls to `has_attribute` and `get_attribute`. It seems like they need to be reduced... Should we:
  - create pointers to the used attributes in the constructor?
  - move some basic logic from the actions to the attributes?
  - nothing for now?
- General feedback?

MaanooAk:
Should I move attack stance inside a formation attribute. This attribute will also store patrol and follow target command's information.

MaanooAk:
Any comments?

ChipmunkV:
You can also take a look at #642. Some of it is about attributes and position. I'm trying to record, serialize, transfer and replay them on the client. That PR has the quickest possible hacks to be able to do that.

Maybe you've got an idea how implementing record/replay has become simpler with your changes (or how to make it simpler, or how to approach the record/replay correctly).

TheJJ:
Sorry it takes so long, I'll have a look tomorrow hopefully.

TheJJ:
On a sidenote (as university and exams keep me busy..):  The shared attributes are the ones provided by nyan, the per-unit attributes are engine features and stored like we do currently.

MaanooAk:
Still the nyan should define the existence of a unshared attribute (eg. formation, resource, damaged) or somehow connect shared with unshared (eg. speed->formation, worker->resource, hitpoints->damaged).

EDIT: I thought I more weak connection between nyan and attributes was intended (eg. life: 10 -> hitpoints, damaged, armor) (am I correct?)

TheJJ:
My idea for that was to create nyan files that ship with the openage engine so that the c++ code can assume that members of those objects (from the shipped `.nyan` files) exist.
For example:

``` python
# this file is shipped with the engine
Unit():
    hp : int

# this file is then loaded at runtime
Villager(Unit):
    hp = 40
```

This means that each `Unit` definitely has a `hp` member.
So from C++ you can query:

``` cpp
nyan::Database db{};
db.load_files(engine_nyanfiles);
db.load_files(user_mods);
// get the "Villager", normally the query is more complicated,
// like looking up the object in some set of initial objects of the culture, whatever.
// now we just magically know that it's named "Villager"
nyan::Object *unit_type = db.get_obj("Villager")
openage::Unit *unit = simulation.get_unit(123);  // by id, location, whatever

// now, as VIllager is a child of Unit, values are guaranteed to be there
if (unit_type.is_child_of("Unit")) {
    int villager_hp = unit_type->get("hp");
    log::log(info << "villager has max hp: " << villager_hp);

    if (unit.damage > villager_hp) {
        unit->die();
    }
} else {
    // wtf internal error
}
```

TheJJ:
Sorry, the example not got a bit our of hands, but this is probably the first time that the dynamic interaction between nyan and c++ was ever documented. Hope it helps :)

MaanooAk:
So the next steps would be:
- [ ] Fix everything
  - [x] Attributes (this pr)
  - [ ] Actions
  - [ ] Abilities
- [ ] Create the "shipped" nyan files (after everything is working properly)
- [ ] Convert the converted data to nyan
- [ ] Redefine/Recreate the nyan in more modular way (= stop converting)

(I thing we should have this kind of task list all over the place)

TheJJ:
We have to do the transition slowly, so your shared attributes are mainly the nyan placeholder. But to simplify the transition, it would be nice if we could start the separation, if needed with placeholders, right now.

I'm doing my master's thesis about nyan this summer, so it will be done.
We have to check if my above example is indeed a feasible way to do.
Then start to organize the simulation code so that the shared attributes are clearly separated and then replaced, some references to "here be nyan" are already in the code (e.g. `libopenage/unit/unit_type.{h,cpp}`, which is similar to what i did above, but with one indirection layer more).

libopenage/unit/attribute.h
```diff
@@ -107,13 +126,61 @@ class AttributeContainer {
...
 	 */
 	virtual std::shared_ptr<AttributeContainer> copy() const = 0;
 };
 
-using attr_map_t = std::map<attr_type, std::shared_ptr<AttributeContainer>>;
+/**
+ * Contains a group of attributes.
+ * Can contain only one attribute of each type.
+ */
+class Attributes {
```
TheJJ: i would put the attributes class in a separate file `attributes.cpp,h`.

MaanooAk:
Moved to separate file and rebased.

TheJJ:
👍

PR #685, opened 2017-01-02
Load data 34 times faster!

MaanooAk:
From
screenshot from 2017-01-02 15-24-56
to
screenshot from 2017-01-02 16-06-28
(after the first time)

### How
I create a `cache.docx` file which contains all the other `*.docx`. Every time the read csv function is called the the file is added to the cache file. The first time (before the creation of the cache) the load time is like the old days and from the second time and after all the loading is happening from the cache structure much faster (= 34 times faster).

### Notes
- The cache file is only around 13 MB
- Fixed the timers wrong logic
- Comments are not cached

### DIFF
--- a/assets/.gitignore
+++ b/assets/.gitignore
@@ -1,5 +1,2 @@
 # converted game assets
 /converted
-
-# cache files
-cache.docx
--- a/assets/.gitignore
+++ b/assets/.gitignore
@@ -1,2 +1,5 @@
 # converted game assets
 /converted
+
+# cache files
+cache.docx
--- a/libopenage/gamestate/game_spec.cpp
+++ b/libopenage/gamestate/game_spec.cpp
@@ -1,4 +1,4 @@
-// Copyright 2015-2016 the openage authors. See copying.md for legal info.
+// Copyright 2015-2017 the openage authors. See copying.md for legal info.
 
 #include "../assetmanager.h"
 #include "../engine.h"
--- a/libopenage/gamestate/game_spec.cpp
+++ b/libopenage/gamestate/game_spec.cpp
@@ -38,7 +38,9 @@ bool GameSpec::initialize() {
 	util::Dir gamedata_dir = this->assetmanager->get_data_dir()->append(this->data_path);
 
 	log::log(MSG(info) << "Loading game specification files...");
-	this->gamedata = util::recurse_data_files<gamedata::empiresdat>(gamedata_dir, "gamedata-empiresdat.docx");
+	util::csv_file_map_t *tmp = load_multi_csv_file(gamedata_dir, "gamedata.docx");
+	this->gamedata = util::recurse_data_files<gamedata::empiresdat>(gamedata_dir, "gamedata-empiresdat.docx", tmp);
+	delete tmp;
 
 	this->on_gamedata_loaded(this->gamedata);
 	this->gamedata_loaded = true;
--- a/libopenage/gamestate/game_spec.cpp
+++ b/libopenage/gamestate/game_spec.cpp
@@ -36,7 +36,6 @@ bool GameSpec::initialize() {
 	util::Dir gamedata_dir = this->assetmanager->get_data_dir()->append(this->data_path);
 
 	util::csv_file_map_t *meta_file_map = load_multi_csv_file(gamedata_dir, "gamedata.docx");
-	util::csv_file_map = meta_file_map;
 
 	this->load_terrain(*this->assetmanager, meta_file_map);
 
--- a/libopenage/gamestate/game_spec.cpp
+++ b/libopenage/gamestate/game_spec.cpp
@@ -33,14 +33,13 @@ bool GameSpec::initialize() {
 	util::Timer load_timer;
 	load_timer.start();
 
+	util::load_csv_file_cache(*this->assetmanager->get_data_dir());
+
 	this->load_terrain(*this->assetmanager);
 
 	util::Dir gamedata_dir = this->assetmanager->get_data_dir()->append(this->data_path);
 
-	log::log(MSG(info)
-	         << "loading game specification files... "
-	         << "will be faster once we use nyan, so please help!");
-
+	log::log(MSG(info) << "Loading game specification files... (will take a while only the first time)");
 	this->gamedata = util::recurse_data_files<gamedata::empiresdat>(gamedata_dir, "gamedata-empiresdat.docx");
 
 	this->on_gamedata_loaded(this->gamedata);
--- a/libopenage/gamestate/game_spec.cpp
+++ b/libopenage/gamestate/game_spec.cpp
@@ -38,7 +38,7 @@ bool GameSpec::initialize() {
 	util::csv_file_map_t *meta_file_map = load_multi_csv_file(gamedata_dir, "gamedata.docx");
 	util::csv_file_map = meta_file_map;
 
-	this->load_terrain(*this->assetmanager);
+	this->load_terrain(*this->assetmanager, meta_file_map);
 
 	log::log(MSG(info) << "Loading game specification files...");
 	this->gamedata = util::recurse_data_files<gamedata::empiresdat>(gamedata_dir, "gamedata-empiresdat.docx", meta_file_map);
@@ -307,7 +307,7 @@ void GameSpec::load_projectile(const gamedata::unit_projectile &proj, unit_meta_
 	}
 }
 
-void GameSpec::load_terrain(AssetManager &am) {
+void GameSpec::load_terrain(AssetManager &am, util::csv_file_map_t *file_map) {
 
 	// Terrain data files
 	util::Dir *data_dir = am.get_data_dir();
@@ -315,7 +315,7 @@ void GameSpec::load_terrain(AssetManager &am) {
 	std::vector<gamedata::string_resource> string_resources;
 	util::read_csv_file(asset_dir.join("string_resources.docx"), string_resources);
 	std::vector<gamedata::terrain_type> terrain_meta;
-	util::read_csv_file(asset_dir.join("gamedata/gamedata-empiresdat/0000-terrains.docx"), terrain_meta);
+	util::read_csv_file(asset_dir.join("gamedata/gamedata-empiresdat/0000-terrains.docx"), terrain_meta, file_map);
 	std::vector<gamedata::blending_mode> blending_meta;
 	util::read_csv_file(asset_dir.join("blending_modes.docx"), blending_meta);
 
--- a/libopenage/gamestate/game_spec.cpp
+++ b/libopenage/gamestate/game_spec.cpp
@@ -33,14 +33,15 @@ bool GameSpec::initialize() {
 	util::Timer load_timer;
 	load_timer.start();
 
-	this->load_terrain(*this->assetmanager);
-
 	util::Dir gamedata_dir = this->assetmanager->get_data_dir()->append(this->data_path);
 
+	util::csv_file_map_t *meta_file_map = load_multi_csv_file(gamedata_dir, "gamedata.docx");
+	util::csv_file_map = meta_file_map;
+
+	this->load_terrain(*this->assetmanager);
+
 	log::log(MSG(info) << "Loading game specification files...");
-	util::csv_file_map_t *tmp = load_multi_csv_file(gamedata_dir, "gamedata.docx");
-	this->gamedata = util::recurse_data_files<gamedata::empiresdat>(gamedata_dir, "gamedata-empiresdat.docx", tmp);
-	delete tmp;
+	this->gamedata = util::recurse_data_files<gamedata::empiresdat>(gamedata_dir, "gamedata-empiresdat.docx", meta_file_map);
 
 	this->on_gamedata_loaded(this->gamedata);
 	this->gamedata_loaded = true;
--- a/libopenage/gamestate/game_spec.cpp
+++ b/libopenage/gamestate/game_spec.cpp
@@ -33,13 +33,11 @@ bool GameSpec::initialize() {
 	util::Timer load_timer;
 	load_timer.start();
 
-	util::load_csv_file_cache(*this->assetmanager->get_data_dir());
-
 	this->load_terrain(*this->assetmanager);
 
 	util::Dir gamedata_dir = this->assetmanager->get_data_dir()->append(this->data_path);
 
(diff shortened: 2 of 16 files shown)


TheJJ:
Very good. But I'd push this even further: As I can't see cache invalidation currently (which is very important), I'd like you to generate this one file only in the first place. Then it's no longer a cache and we just pack everything in this one file from the `Exportable`.

The entry call is in `convert_metadata` of `openage/convert/driver.py`. You may add a new parameter to the `add_data` that specifies that the output shall be just one file.

The file generation of data fed in by `add_data` happens in `openage/convert/dataformat/data_formatter.py` when the `export(target_folder)` is invoked.
This means if you add a flag so that `add_data` will store the data passed just in a single file, `export(target_folder)` will then actually write it, in just a single file.

In `export` there's the `files` dict, where you basically save to the same file (`files[single_file_name]`) (which is done by `add_snippet`), which has the name you passed as cache file name.

In other words:

``` python
# The snippet.file_name has to be "overridden" to the "cache" file
# and the target name is set when the data was added with add_data(...)

if snippet.file_name not in files:
    files[snippet.file_name] = GeneratedFile(snippet.file_name, format_)

files[snippet.file_name].add_snippet(snippet)
```

MaanooAk:
If I understand correctly you are saying to generate a single file instead of all of these docx files when the conversion is happening (Am I correct?)

I tried to do this but I could not understand the full thing... (also I'm not experienced with Python at all, my code compliance pr was my fist real thing written in python). So I made this thing instead, as a fast tmp solution.

I will try to take an other look at the convert code tomorrow (so I can understand what you are talking about).

**However**: Merging all the files after the conversion (one time thing) could be a much much simpler solution.

TheJJ:
Yes, exactly. Don't write thousands of files but instead just one, which has exactly the format you used for the cache file.

Writing the files as is and then merging them is easier currently, but this will take a very long time for the non-SSD users and also you don't have cache-invalidation currently if i didn't miss it. The exported game data may change any time (if we discover or enable new values), so we need the invalidation.

It shouldn't be too hard to change the `DataFormatter` to save snippets (pieces of text) from multiple files into one file where each part is annotated with your `## $filename` marker, I tried to give you hints in my comment above :) If you have further questions, just ask.

assets/.gitignore
```diff
@@ -1,2 +1,5 @@
 # converted game assets
 /converted
+
+# cache files
+cache.docx
```
TheJJ: if we now create the single file directly, this isn't needed any more.

libopenage/main.cpp
```diff
@@ -34,6 +34,8 @@ int run_game(const main_arguments &args) {
 	auto &cvar_manager = engine.get_cvar_manager();
 	cvar_manager.load_main_config();
 
+	util::load_csv_files(data_dir.join("converted/meta.docx"));
+
```
TheJJ: This is just a side-effect method without returning anything, so we should avoid it. Also, this should not happen in the `main.cpp`, instead in `game_spec.cpp` so it is accounted to the right timer and is also located in the right file.

libopenage/util/file.cpp
```diff
@@ -92,5 +92,32 @@ std::vector<std::string> file_get_lines(const std::string &file_name) {
 	return result;
 }
 
+std::unordered_map<std::string, std::vector<std::string>> csv_file_map;
```
TheJJ: ugh. global variables are a no-go.

libopenage/util/file.cpp
```diff
@@ -92,5 +92,32 @@ std::vector<std::string> file_get_lines(const std::string &file_name) {
...
+	}
+
+	std::vector<std::string> lines = file_get_lines(path);
+
+	std::string current_file = "";
+	for (auto& line : lines) {
+		if (line[0] == '#' && line[1] == '#' && line[2] == ' ') {
+			current_file = "./assets/converted" + line.erase(0, 3);
+			csv_file_map.emplace(current_file, std::vector<std::string>());
+		}
```
TheJJ: This method should rather return the map that is currently global. Then the returned map is passed to the csv reader as pointer, which is `nullptr` by default. If it's not `nullptr`, the map is used as cache.

That is, the entry call for loading the csv files (`recurse_data_files`) gets this map as new argument.

The `./assets/converted` should also not be hardcoded. The prefix is already known as game spec basedir (which is `gamedata_dir` in `game_spec.cpp`).

libopenage/gamestate/game_spec.cpp
```diff
@@ -37,10 +37,7 @@ bool GameSpec::initialize() {
 
 	util::Dir gamedata_dir = this->assetmanager->get_data_dir()->append(this->data_path);
 
-	log::log(MSG(info)
-	         << "loading game specification files... "
-	         << "will be faster once we use nyan, so please help!");
-
+	log::log(MSG(info) << "Loading game specification files...");
 	this->gamedata = util::recurse_data_files<gamedata::empiresdat>(gamedata_dir, "gamedata-empiresdat.docx");
```
TheJJ: here, the single-file is read and returned as map. this map is then passed as argument to `recurse_data_files`.

libopenage/util/file.h
```diff
@@ -7,6 +7,7 @@
 #include <fstream>
 #include <string>
 #include <vector>
+#include <unordered_map>
```
TheJJ: alphabetical ordering pls

libopenage/util/file.h
```diff
@@ -38,41 +42,68 @@ template<typename lineformat>
 void read_csv_file(const std::string &fname, std::vector<lineformat> &out) {
```
TheJJ: i'd add a new `read_csv_file` which does not open the file but instead uses the data from the map. we have to do this when deglobalizing the map.

alternatively, we can add another optional argument which provides the file data already (from the map), so the open and read is bypassed and the map data is used instead.

openage/convert/dataformat/data_formatter.py
```diff
@@ -184,5 +184,10 @@ def export(self, projectdir, requested_formats):
         # we now invoke the content generation for each generated file
         for gen_file in generate_files:
             file_name, content = gen_file.generate()
-            with projectdir[file_name].open('wb') as outfile:
-                outfile.write(content.encode('utf-8'))
+            if gen_file.format_ in {"csv"}:
+                with open('assets/converted/meta.docx', 'ab') as outfile:
+                    outfile.write(('## ' + file_name + '\n').encode('utf-8'))
+                    outfile.write(content.encode('utf-8'))
```
TheJJ: hehe no that's too simple. we generate way more files with this method which are not part of the gamespec and should still be separate. also, the `assets/converted/...` must not be hardcoded.

I would change the `def add_data(self, data_set_pile, prefix=None):` to have a new argument `single_output=None`.

This info is then passed around up to the `generate()` function, which returns another value, which is again the `single_output`. Then you can check if the single output was activated, and perform the write you did here.

Namely not just convert every `csv` to the `meta.docx`, but instead let each generated file specify where it wants to be packed to. And that information is provided by `add_data` already.

MaanooAk:
So this is what I did:
- All the game data files are now exported in `assets/converted/gamedata.docx` by implementing the `single_output` option:
```python
data_formatter.add_data(data_dump[0], prefix="gamedata/", single_output="gamedata")
```
- I stopped the sorting and dependacy things in the generation for csv files.
- All other `.docx` are exported as they did.

So this is what I did and what I din't do:
- I change all the file caching code, by the way this no longer a caching but a packing, but the global object (now a pointer) can't go away. 

The file map is passed as an argument but the code that call the reading function is **autogenerated**...

libopenage/gamestate/game_spec.cpp
```diff
@@ -33,15 +33,15 @@ bool GameSpec::initialize() {
...
-
 	util::Dir gamedata_dir = this->assetmanager->get_data_dir()->append(this->data_path);
 
-	log::log(MSG(info)
-	         << "loading game specification files... "
-	         << "will be faster once we use nyan, so please help!");
+	util::csv_file_map_t *meta_file_map = load_multi_csv_file(gamedata_dir, "gamedata.docx");
+	util::csv_file_map = meta_file_map;
+
+	this->load_terrain(*this->assetmanager);
```
MaanooAk: `load_terrain` needs the `meta_file_map` should I just pass it?

MaanooAk:
- Removed the global object
- Made file map optional argument
- Changed the gen files to use the file map (took me some time)

I think we are ready to go!

TheJJ:
Cool :)
Next idea: Now meld this together in the asset manager. This is another step towards nyan, and will simplify things even more. That way, any entity can request the load of some file from the asset manager, which will have this loaded-file-map and can also watch for file changes with inotify (it can do that already).
Then the data loading is finally at one central entity. Keep up the good work!

openage/convert/dataformat/struct_definition.py
```diff
@@ -35,6 +35,7 @@ def __init__(self, target):
         self.name_struct        = target.name_struct         # !< name of generated C struct
         self.struct_description = target.struct_description  # !< comment placed above generated C struct
         self.prefix             = None
+        self.single_output      = None
```
TheJJ: can you please comment those?

openage/convert/dataformat/members.py
```diff
@@ -671,8 +671,8 @@ def get_snippets(self, file_name, format_):
 
             # function to recursively read the referenced files
             txt.extend((
-                "int %s::recurse(openage::util::Dir basedir) {\n" % (self.type_name),
-                "\tthis->index_file.read(basedir); //read ref-file entries\n",
+                "int %s::recurse(openage::util::Dir basedir, openage::util::csv_file_map_t *file_map) {\n" % (self.type_name),
+                "\tthis->index_file.read(basedir, file_map); //read ref-file entries\n",
```
TheJJ: with the asset manager addition, one would recurse with a asset manager pointer here

openage/convert/dataformat/generated_file.py
```diff
@@ -144,6 +146,12 @@ def generate(self):
         prefs = self.default_preferences.copy()
         prefs.update(self.output_preferences[self.format_])
 
+        # if there is no need for sorting and resolving dependencies, just return the snippets
+        # as they have been added
+        if not prefs["process"]:
+            file_data = "\n".join(snippet.get_data() for snippet in self.snippets)
+            return prefs["folder"] + '/' + self.file_name + prefs["file_suffix"], file_data
```
TheJJ: very good idea... silly me didn't come up with that.

openage/convert/dataformat/data_formatter.py
```diff
@@ -58,13 +58,13 @@ class DataFormatter:
...
                 0: entry_parser.ParserTemplate(
-                    signature    = "int %srecurse(openage::util::Dir /*basedir*/)",
+                    signature    = "int %srecurse(openage::util::Dir /*basedir*/, openage::util::csv_file_map_t */*file_map*/)",
                     headers      = util.determine_header("engine_dir"),
                     impl_headers = set(),
                     template     = "$signature {\n\treturn -1;\n}\n"
                 ),
                 None: entry_parser.ParserTemplate(
-                    signature = "int %srecurse(openage::util::Dir basedir)",
+                    signature = "int %srecurse(openage::util::Dir basedir, openage::util::csv_file_map_t *file_map)",
```
TheJJ: those signatures would then just get the asset manager

openage/convert/changelog.py
```diff
@@ -35,6 +35,7 @@
     {"graphics"},
     {"interface"},
     {"interface"},
+    {"gamedata"},
```
TheJJ: we might wanna purge the old data, but we have no infrastructure for that yet. i.e. removing the thousands of files that will otherwise lay around forever.

MaanooAk:
This is a csv file map/packing code, why is this going to be used in nyan? This is hack, it should not be part of the asset manager. This started as small hack to improve the load time, a file containing pseudo file paths is even worse than a global variable.

A suggest to revisit the asset manager after we have a nyan file to work with, and not invest further on this.

openage/convert/dataformat/struct_definition.py
MaanooAk: done

openage/convert/dataformat/generated_file.py
MaanooAk: thx

openage/convert/changelog.py
MaanooAk: thousands of very small files, it ok ;)

TheJJ:
Hm yes, you have a point. Then the only think left is a proper cleanup of the old files, right?

MaanooAk:
You mean the .docx files right?

openage/convert/changelog.py
MaanooAk: If we add `metadata`, will be solved?

MaanooAk: Tested it didn't work but I think it should be there.

TheJJ: We need to remove the `gamedata` folder when gamedata is regenerated. That's the easiest and correct way i think. Btw the correct changelog name is "metadata", which doesn't quite fit.
So if we do
``` python
import shutil
shutil.rmtree('/path/to/your/dir/')
```
in `driver.py` right after
``` python
    if args.flag("no_metadata"):
        return
```

it will always purge the folder first.

MaanooAk: What is this `sane/path/t`

MaanooAk: I'm on it

openage/convert/changelog.py
```diff
@@ -35,7 +35,7 @@
     {"graphics"},
     {"interface"},
     {"interface"},
-    {"gamedata"},
+    {"gamedata", "metadata"},
```
TheJJ: There is no `gamedata` change, the available components are in the map above (`COMPONENTS`).

MaanooAk: Yes I saw that, I now am not sure to use the `targetdir` in deleting the folder:
```python
shutil.rmtree(args.targetdir ???? 'gamedata')
```

MaanooAk: done

MaanooAk:
I think we are ready to go! (again)

TheJJ:
```
find . | wc -l
45642
(...)
find . | wc -l
1
```

jup.

TheJJ:
Thank you very much. This improvement was _so_ overdue.

PR #686, opened 2017-01-04
Textfiles code compliance improvements

MaanooAk:
### Code compliance improvements:
- Print line number of an issue
- Added missing space issue (detects `){` and `if/for/while(`)
- Added extra space issue (detects spaces before `;`)
- Can be extended

> For the current master the `make checkfast` print:
>
> screenshot from 2017-01-04 17-06-26

### How it works:
First the whole file string is searched and if at least an issue is found somewhere, the file is searched line by line (in order to generate proper description with  the line and line number).
Regular expressions and generators are used when possible.

### DIFF
--- a/buildsystem/codecompliance/__main__.py
+++ b/buildsystem/codecompliance/__main__.py
@@ -171,7 +171,7 @@ def find_all_issues(args, check_files=None):
 
     if args.cppstyle:
         from .cppstyle import find_issues
-        yield from find_issues(check_files, ('openage', 'libopenage'))
+        yield from find_issues(check_files, ('libopenage',))
 
     if args.pylint:
         from .pylint import find_issues
--- a/buildsystem/codecompliance/__main__.py
+++ b/buildsystem/codecompliance/__main__.py
@@ -1,4 +1,4 @@
-# Copyright 2014-2016 the openage authors. See copying.md for legal info.
+# Copyright 2014-2017 the openage authors. See copying.md for legal info.
 
 """
 Entry point for the code compliance checker.
@@ -132,12 +132,15 @@ def main(args):
     else:
         check_files = None
 
-    issues_found = False
+    issues_count = 0
     for title, text in find_all_issues(args, check_files):
-        issues_found = True
+        issues_count += 1
         print("\x1b[33;1mWARNING\x1b[m {}: {}".format(title, text))
 
-    return not issues_found
+    if issues_count > 0:
+        print("A total of \x1b[33;1m{} issues\x1b[m were found.".format(issues_count))
+
+    return issues_count == 0
 
 
 def find_all_issues(args, check_files=None):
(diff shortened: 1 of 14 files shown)


MaanooAk:
The tests are failing because of the new issue detection. Before merge we should fix the warning (if this pr is in the right direction)

TheJJ:
You are: awesome.
This is very much in the right direction. The thing we didn't think about back then was adding exceptions for Cython generated files.

However: I would put the `.cpp` and `.h` checks in a separate file like `cppstyle.py`. And if the `textfiles.py` then needs to share code with the `cppstyle.py`, use the common functions by importing `textfiles` from `cppstyle.py`

Thanks :+1:

MaanooAk:
The `textfiles.py` is not compatible with the `only_changed_files`. Is there a reason or should I change it?

buildsystem/codecompliance/__main__.py
```diff
@@ -166,6 +169,10 @@ def find_all_issues(args, check_files=None):
         from .pystyle import find_issues
         yield from find_issues(check_files, ('openage', 'buildsystem'))
 
+    if args.cppstyle:
+        from .cppstyle import find_issues
+        yield from find_issues(check_files, ('openage', 'libopenage'))
```
MaanooAk: Ok (small change for next commit)

buildsystem/codecompliance/util.py
```diff
@@ -49,6 +49,26 @@ def readfile(filename):
...
+            if filename.startswith('.'):
+                continue
+
+            filename = os.path.join(path, filename)
+
+            if os.path.isdir(filename):
+                yield from find_all_files((filename,))
+                continue
+
+            yield filename
```
TheJJ: Why did you copy the function? Just make a new case if the extension is None, then the extension isn't checked. then you can remove the duplicated code.

buildsystem/codecompliance/util.py
```diff
@@ -68,3 +88,14 @@ def findfiles(paths, exts):
 
             if has_ext(filename, exts):
                 yield filename
+
+
+def issue_str(title, filename):
+    return (title, filename)
```
TheJJ: Can you factor out this function entirely? :)

TheJJ:
I don't think there's a reason, we just used it to check all remaining files. So we can already use the infrastructure to check only `.cpp` and `.h` files efficiently with the extension check.

buildsystem/codecompliance/util.py
MaanooAk: This function from which all (should/will) issues pass, so if we want to add other format or color we can just change this only function. You don't agree?

buildsystem/codecompliance/util.py
MaanooAk: I thought I could make it smaller with out the extension checking (turns out I cant) and then forgot about it... I will change it

MaanooAk: I thought I could make it smaller with out the extension checking (turns out I cant) and then forgot about it... I will change it

buildsystem/codecompliance/util.py
TheJJ: Ah, okay. Jup, then leave it and maybe add a comment what could be done there.

MaanooAk:
This should be all done!

PR #688, opened 2017-01-09
docs: Reverse engineering for Monks, Accuracy and Ram Speed

heinezen:
Made some docs for reverse engineering and added new info to other docs.

**Where does this information come from?**
Mainly from the Youtube Channels of [Spirit of the Law](https://www.youtube.com/channel/UChzLZJo-SxuPHz-oYKAIC_g) and [Zero Empires](https://www.youtube.com/channel/UCoMg6gEYvB9n5RaoIngFFJw). I provided the video links in the files, so that everybody can verify if I got something wrong.

**What do these docs cover?**

Accuracy:
- Relation of accuracy and distance/target size
- How inaccuracy/accuracy is defined by game mechanics
- Damage by stray arrows, scorpions and friendly fire

Damage:
- How and when does the game calculate damage from projectiles?

Market:
- Gold formula for trade carts
- Resource prices and tribute system
- Path finding for trade carts if market is destroyed

Monks:
- Converting buildings

Ram Speed:
- Which units affect the speed of a ram and by how much?

Also included are weird bugs and irritating behaviour that may or may not be kept in OpenAge for nostalgia.


TheJJ:
Whoa awesome!
Thank you very much!

PR #694, opened 2017-01-11
Market resource prices

MaanooAk:
This market resource buy and sell logic (not the building)

- The global base prices
- Calc prices for selling and buying
- Perform buy and sell

Based on [doc/reverse_engineering/market.md#prices](https://github.com/SFTtech/openage/blob/master/doc/reverse_engineering/market.md#prices).

### DIFF
--- a/libopenage/gamestate/market.cpp
+++ b/libopenage/gamestate/market.cpp
@@ -15,13 +15,13 @@ Market::Market() {
 // Price calculation is documented at doc/reverse_engineering/market.md#prices
 
 bool Market::sell(Player &player, const game_resource res) {
-	double mult = this->get_multiplier(player, false);
-
+	// deduct the standard MARKET_TRANSACTION_AMOUNT of the selling res
 	if (player.deduct(res, MARKET_TRANSACTION_AMOUNT)) {
 		// if deduct was successful
 		// calc the gold received from selling MARKET_TRANSACTION_AMOUNT of res
-		double amount = this->base_prices.get(res) * mult;
+		double amount = this->get_sell_prices(player).get(res);
 		player.receive(game_resource::gold, amount);
+
 		// decrease the price
 		this->base_prices[res] -= MARKET_PRICE_D;
 		if (this->base_prices.get(res) < MARKET_PRICE_MIN) {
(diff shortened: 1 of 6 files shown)


TheJJ:
Can you add documentation about the price calculation please?

MaanooAk:
Its on the [doc/reverse_engineering/market.md#prices](https://github.com/SFTtech/openage/blob/master/doc/reverse_engineering/market.md#prices) and its very simple:
```
buy -> increase value
sell -> decrease value
```
and there is a min and max value. Are you sure?

zuntrax:
The documentation there is fine. How about adding a comment to `libopenage/gamestate/market.cpp` referencing that file?
Also method and class docstrings in doxygen format would be nice.

MaanooAk:
I could add this to the class doc
```c++
/**
 * The global market prices.
 * 
 * Price calculation is documented at doc/reverse_engineering/market.md#prices
 */
class Market {
```
> method and class docstrings in doxygen format

If you mean this:
```c++
 * @param player the player
 * @param res the resource
```
I'm not a big fun of meaningless doc, but If you insist it's ok. Do you insist?

zuntrax:
I'd rather put it in the `cpp` file where it's actually implemented. Or in both the header and `cpp` file.
About the docstrings: I was stupid and looked for them in the `cpp` file and not in the header. The purpose of the parameters is quite obvious, so I think you can skip the `@param` lines. Sorry for the confusion.

MaanooAk:
Source files are not parsed by doxygen (I think), so both seems ok.

libopenage/gamestate/market.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+std::shared_ptr<ResourceBundle> Market::getBuyPrices(const Player &player) const {
+	return this->getPrices(player, true);
+}
+
+std::shared_ptr<ResourceBundle> Market::getSellPrices(const Player &player) const {
+	return this->getPrices(player, false);
+}
+
+std::shared_ptr<ResourceBundle> Market::getPrices(const Player &player, const bool is_buy) const {
```
TheJJ: I'm not sure how useful it is to have the resource bundle as `shared_ptr`. Isn't it sufficient as a stack object?

libopenage/gamestate/market.h
```diff
@@ -0,0 +1,71 @@
...
+	/**
+	 * The given player buys the given resource with gold.
+	 * Returns true when the transaction is successful.
+	 */
+	bool buy(Player &player, const game_resource res);
+
+	/**
+	 * Get the selling prices for a given player.
+	 */
+	std::shared_ptr<ResourceBundle> getBuyPrices(const Player &player) const;
```
MaanooAk: I always forget about that (not my fault, I was raised with Java)

TheJJ: 😀

TheJJ:
I fear you need to rebase because of some fixes we applied in the meantime.

PR #695, opened 2017-01-11
show timestamp in build output for easier VM troubleshooting

zuntrax:
#649 had a CI error because of a missing pep8 module. It would have been nice to see a timestamp in the build output.

### DIFF
--- a/.kevin
+++ b/.kevin
@@ -5,6 +5,7 @@
 
 
 sanity_check:
+	date -Iseconds
 	make checkall
 
 configure:


zuntrax:
We could as well put it into kevin source code before cloning the repo, but this is the easier fix.

zuntrax:
Just saw that there is a `_created` file containing a timestamp in the build results directory, but I'd still like to see it in the build output.

TheJJ:
Hm ok, just out of curiosity: why do you wanna see the timestamp?

zuntrax:
We'll add this in kevin.

PR #698, opened 2017-01-16
fix build with libbacktrace, clean up log includes

zuntrax:
* Fixed linker error when building with libbacktrace.
* `log/log.h` was implicitly imported by `error/error.h`.
* Reordered some includes to match "associated - global - local" structure.


zuntrax:
Blocks on #699.

PR #700, opened 2017-01-16
make pylint errors go away

### LINKED ISSUE #699: pylint complains about code from master
This leads to kevin failing all sanity checks in the Arch VM.
I'm fixing it right now.

zuntrax:
Closes #699

### DIFF
--- a/.kevin
+++ b/.kevin
@@ -5,7 +5,8 @@
 
 
 sanity_check:
-	make checkall
+	# skip in debian
+	apt --version > /dev/null || make checkall
 
 configure:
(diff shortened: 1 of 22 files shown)


zuntrax:
HRRRNGHHH! Old debian pylint complains about `redefined-variable-type`, but in version 2.0.0 this flag does no longer exist, so disabling it leads to another error. Happy fun times! 🍿 🔥

zuntrax:
Stuff is horribly broken on debian. I changed the sanity check to only run on arch.

PR #701, opened 2017-01-16
docs: Reverse engineering for Alarms, Pathfinding, Stats & more

heinezen:
More docs revealing some of the mysteries of Age of Empires II.

**Where does this information come from?**
From various sources on the internet and/or books and post mortems from developers. I provided links in the files, so everyone should be able to verify the info (and maybe add things that I missed).

**What do these docs cover?**

Alarm sounds:
- Differences in the 4 different "You are being attacked!" alerts and when they are played
- Battle sizes and alarm sounds
- Cooldown for alarms/alerts

Town Bell:
- Range
- Distribution of villagers algorithms

Waypoints:
- How they work
- Limits

Pathfinding:
- Short description of the three pathfinding algorithms and their underlying mechanics

Villagers:
- Build speed + Formula
- Switching tasks and losing resources
- Wolves and villagers (mostly focussed on balancing)

Unit stats:
Snapshot of available units and their stats from AoK, AoC, FE and African Kingdoms


doc/reverse_engineering/pathfinding.md
```diff
@@ -0,0 +1,39 @@
...
+
+```
+...............      ...............      ...............      ...............
+...............      ...............      ...............      ....XX.XXXXX...
+......T........ ---> ......T........ ---> ....XXTXXXXX... ---> ....XXTXXXXX...
+............... ---> ....XX.XXXXX... ---> ....XX.XXXXX... ---> ...............
+....XXXXXXX....      ....XX.XXXXX...      ...............      ...............
+....XXXXXXX....      ...............      ...............      ...............
+```
+Units will split up their formation if they encounter an obsruction such as a tree.
```
TheJJ: typo: `obsruction`

doc/reverse_engineering/pathfinding.md
```diff
@@ -0,0 +1,39 @@
...
+
+```
+..............      ..............      ..............      ..............
+..............      ..............      ..............      ....X....X....
+....YYYYYY.... ---> ....YYYYYY.... ---> ....YYYYYY.... ---> ...XYYYYYYX...
+.............. ---> ....XXXXXX.... ---> ...XXXXXXXX... ---> ...XXXXXXXX...
+....XXXXXX....      ....XXXXXX....      ....XX..XX....      ..............
+....XXXXXX....      ..............      ..............      ..............
+```
+When troops encounter an enemy formation they will try to go around to engabe in combat. Though if units get killed and create a gap, the pathfinder has some difficulties to distribute units efficiently.
```
TheJJ: typo: `engabe`

doc/reverse_engineering/pathfinding.md
```diff
@@ -0,0 +1,39 @@
...
+
+and *Ensemble Studios Official Strategies & Secrets to Microsoft's Age of Empires II: The Age of Kings*, Bruce Shelly, 1999, published by Sybex Inc.
+
+## Algorithms
+
+Age of Empires 2 uses three different algorithms for pathfinding plus two obstruction systems.
+
+The first (high-level-)pathfinder is meant for long distances and uses mip-mapping to create a general route across the map. Criteria for the mip-map subdivision is *passibility*. It ignores moving objects, e.g. units or wildlife, though it takes "static" object like buildings into account. Whenever a building is constructed the pathfinder gets notified and updates the corresponding mip-map chains.
+
+The other two pathfinders are for short and medium distances. One uses tile-based pathfinding while the other one is a polygonal pathfinder that dynamically generates convex hulls. These are used in combat situations or when units cross paths.
```
TheJJ: Do you have more information about the inner workings of all three algorithms?

doc/reverse_engineering/attacking_alarm.md
```diff
@@ -0,0 +1,33 @@
...
+    50313.wav  (Horn + monk) One of your units is being converted
+    50315.wav  (Horn) Your military units are attacked
+    50316.wav  (Bell) A villager or building of yours is attacked
+    50366.wav  (Horn + wolf) Wild animal attacks your units (not played for boar)
+
+All the alarms only trigger if you are attacked and not if you kill enemy units that have not hit you yet.
+
+## Battles
+
+There is only one alarm played per "battle". Battles start when the player is attacked, i.e. when the alarm sound plays. Their size is always 20x20 tiles with the center being at the place where the players' units where attacked first (i.e. a 10 tile radius). Units fighting in this area can not trigger more alarms until they either leave the battle field (they would have to move outside of the 20x20 area) or the battle is over (see **Cooldown**).
```
TheJJ: `where the players' units were attacked first` <- typo in where

doc/reverse_engineering/switching_villager_tasks.md
```diff
@@ -0,0 +1,19 @@
...
+
+AoE 2 : HD Edition further removed some instances where a player would lose resources. Now resources are only lost if the villager gets assigned to a resource that is different from what they are carrying. Food that was gathered in the role of a "shepard" will therefore carry over if the new role is "hunter", "gatherer" or "farmer".
+
+Looking at the first entry from **Weird AoE2 Quirks** it can be assumed that the HD Edition probably uses a dirty workaround to prevent the player from losing resources. The food is most likely put into the newly assigned sheep/boar/deer and then immediatly "collected" again.
+
+## Automatic Drop-off
+
+In addition to the manual method, workers will automatically drop off **any** resource they are carrying to the player's stockpile, if they constructed a gathering site. The type of gathering site does not matter. For example, villagers will drop off food at a mining camp, if they helped finishing the building. The same happens when they build a farm.
```
TheJJ: whoa. are you sure? i never noticed my villagers ever dropped wood at a mining camp.

doc/reverse_engineering/pathfinding.md
heinezen: Well most of the things you see here is stuff confirmed by developers. I did try to rev eng more: For example, I tried to find out when exactly they switch from the high-level to the low-level pathfinder. But there are so many variables that have an influence. Formation, distance to the goal, the obstruction type, etc... Ensemble definitely put a lot of work into this. I think with a lot more time I can figure out the polygonal and tile-based pathfinders but this requires extensive testing.

I added some further reading in pathfinding.md for anyone interested (in german though).

doc/reverse_engineering/attacking_alarm.md
heinezen: Fixed them all ;-)

doc/reverse_engineering/switching_villager_tasks.md
heinezen: Yes, I didn't know that either. Spirit tested it for AoE2:HD but it's also in AoC 1.0c. In AoK builders couldn't carry resources.

I modified the text because one could misunderstand this as "When they construct it they can now use it as a drop off point for any resource indefinitely". This only happens *after* they finished the site. Spirit explains this [here](https://youtu.be/FK__AnTL400?t=2m51s)

MaanooAk:
- The whole "automatic drop-off" thing makes no sense, shouldn't we consider it a bug?
- It would be nice to have all these stats in a more generic format (like csv, json or in the form you found them) in order to convert them into nyan at some point.

heinezen:
No it makes complete sense in AoE logic. Think of this situation:

1. Players pulls 3 villagers (role: lumberjacks) from forest. 
2. Makes the 3 (new role: builder) build a mining cam. Remember: Builders can hold resources since AoC
3. Villagers finish the mining camp and automatically start gold mining (new role: gold miner).

If this automatic drop off wouldn't be there, all three villagers would lose their wood because they switched roles. Gold miners cant hold anything except gold. 

I think this feature is very intentional and not a bug. The whole pupose of "builders can carry resources" is defeated, if players would lose the resources anyway by building a gathering site.

MaanooAk:
@heinezen Sure, but I was talking about the part where the villagers drop the **wood** into the **mining camp**.

> [...] all three villagers would lose their wood because they switched roles.

Why not? If you take a lumberjack and make him a miner he loses the wood, why make an exception?

heinezen:
Again, they do that only once and only in *one* specific edge case scenario. This had to be put into the game with thought. *It is not a bug*. It is there to reduce micro management

Let me guide you through the (possible) thought process of the developers:

"OK. So for AoC we want to reduce MicroManagement so that players have more fun."
"Ah yes. So how do we do that?"
"Let's let builders carry resources, so players dont have to send working villagers to their gathering site every time."
"Great. That has always been a tedious task."

Then there they did playtesting and probably observed from the recordings that players were still microing because:
- they lost resources when they built a gathering site (builders would switch roles and then lose resources)
- gathering sites are buildings you construct very often
- therefore players would still micro because they didn't want to lose resources

And then they put this behavior into the game because it
- doesn't change the balance of the game very much (it's usually 20-30 resources)
- removes the "I have to wait 1 sec in this intense game to micro them back to a gathering site"-moments
- works with the weird AoK logic where villagers lose resources when they switch roles
- helps new players that just want to play the game by not making their precious resources vanish

heinezen:
Maybe so. else can do some funky stuff with the .csv files. However, I'm done here. You can pull if you want.

PR #706, opened 2017-01-17
force colored cmake output in kevin builds

zuntrax:
SFTtech/kevin#14. For the time being, this should be an acceptable workaround.


TheJJ:
Indeed, and it is still equally readable in firefox if one doesn't use `curl` for viewing.

PR #718, opened 2017-01-23
Update unit.py

yvan-burrie:
Added clarifications and filled unknowns, only by comments.


TheJJ:
It's ok to also rename and update the variables directly :)
The build will fail if you rename an exported value that is used in the code.

You can use `curl -N build-details-link` on your terminal to see the build output.

Thanks for the updates though, I think our dat file can never be complete enough, so keep going!

TheJJ:
I'm integrating this in #723 because it will require a data reconversion anyway.

TheJJ:
Done.

Issue #691, opened 2017-01-10
Multiple searchpaths for assets and configuration


TheJJ:
Currently, we only focus on the in-devtree-version. But the `make install`'d version should catch up and also support the appropriate search paths:

Dev-mode detection currently works by the ability to `import devmode`, because `devmode.py` is not installed and thereby the import will fail.

Search paths:
* if in devmode, only(?): `$repo/`
* `~/.local/share/openage/` / `~/.config/openage/`
* `/usr/share/openage/` / `/etc/openage/`

And of course the matching parts on windows/macOS/whatever.
Basically, we will adhere to http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

The order depends on the data type to be located:

* config files are loaded global first, then user, ...: values get overidden and specialized then
* assets are searched local first, so that the most-custom-asset wins

This is also a step towards our modding API.
Assets can either be in `.tar.*` archives or extracted folders in the search paths. The mounting of archives is done transparently by #630.

As a first step, I'd suggest you install the game with `make install`, then **leave the repo folder** and try to run it with `python3 -m openage`. You should of course package the game to avoid filesystem clutter.

* [ ] searchpaths
* [ ] global launcher (something like `/usr/bin/openage`) #709

Fixes #610
Depends on #630
Will help #667
Will be used by #632 later
Fixes #709

TheJJ:
Done via #723

Issue #630, opened 2016-10-07
Filesystem abstraction C++ interface


TheJJ:
We have a cool filesystem abstraction in Python (`openage/util/fslike/`) which we also need to access from C++.

We wanna reuse the same code currently used in the convert script input file location.

After you [created a C++ interface](https://github.com/SFTtech/openage/blob/master/doc/code/pyinterface.md) for the `openage.util.fslike` with Cython, we will use it for searching for assets and configs in `/usr/share/openage`, your home folder, additional environment variables and so on.

This basically is the foundation for having multiple data sources spread accross your filesystem.

TheJJ:
Done in #723

TheJJ:
Done.

PR #723, opened 2017-01-26
Installation improvements

### LINKED ISSUE #431: Write slp parser in cython
To enhance the code optimizations, the SLP reader in `openage/convert/slp.py` should be written in cython. Currently, this is the bottleneck for the asset conversion.

And if that's too slow as well, we can think about using C++ for it...

### LINKED ISSUE #690: Normalize install prefixes in cmake
Please normalize the paths that are shown after the buildsystem configuration step:

observed:
```
   install prefix | /usr/local/../asdf/../test
py install prefix | /usr/lib/python-exec/python3.5/../../../lib64/python3.5/site-packages
```

expected:
```
   install prefix | /usr/test
py install prefix | /usr/lib64/python3.5/site-packages
```

### LINKED ISSUE #708: blacklist pythontex in the buildsystem
Pythontex is some kind of a python interpreter bundled with texlive. CMake finds it, but it's not suitable for what we are doing (obviously). We should blacklist it to get rid of clutter in the configure output. Also the "dropping invalid python interpreter" message (for the remaining possible cases) should not be a warning, but a normal message without modifiers.

<details>
<summary>Configure output</summary>
<pre>
[...]
Testing /usr/bin/python3
Testing /usr/bin/python3.6
Testing /usr/bin/python2.7
Testing /usr/bin/python3.6m
Testing /usr/share/texmf-dist/scripts/pythontex/pythontex.py
usage: pythontex.py [-h] [--version] [--encoding ENCODING]
                    [--error-exit-code {true,false}] [--runall [{true,false}]
                    | --rerun {never,modified,errors,warnings,always}]
                    [--hashdependencies [{true,false}]] [-j N] [-v]
                    [--interpreter INTERPRETER]
                    [--debug [<family>:<session>:<restart>] | --interactive
                    [<family>:<session>:<restart>]]
                    TEXNAME
pythontex.py: error: unrecognized arguments: -c
CMake Warning at buildsystem/modules/FindPython.cmake:149 (message):
  Dropping invalid python interpreter
  '/usr/share/texmf-dist/scripts/pythontex/pythontex.py'
  buildsystem/HandlePythonOptions.cmake:6 (find_package)
  CMakeLists.txt:69 (include)
[...]
</pre>
</details>

### LINKED ISSUE #709: Install global game binary
Currently, we don't install a game binary to e.g. `/usr/bin/openage`.
The only thing that complicates this is that we launch it from python, and installing some hacky bash script sucks.
Gentoo solves this with [`python-exec`](https://wiki.gentoo.org/wiki/Project:Python/python-exec), but we have to provide something similar for other distros.

The easiest way will probably be to just install the `run.py` script or the cythonized `run` executable.

The problem that needs to be solved is how we install that binary to the "correct" path without `distutils`, as we use `CMake` instead of `distutils` for compiling and installing the code.

In a nutshell:

* [x] Extend the buildsystem to figure out the distutils-equivalent install path (we can already do that for the openage python module itself)
* [x] modify `make install` to install the `run.py` as openage launcher script

TheJJ:
* [X] normalize python prefix (fixes #690)
* [X] install game launcher (fixes #709)
* [X] blacklist `pythontex.py` (fixes #708)
* [X] cache locating python and python modules
* [X] ninja build system support
* [X] enable repo assets directory depending on `devmode`
* [X] union-mount all possible asset search paths (`/usr/share`, `/home`, `--assets`, ...)
* [X] modernized file reading
  * use `std::istream`, `std::getline` etc
* [x] filesystem abstraction c++ interface (#630)
  * [X] abstraction layer through python
  * [x] abstraction layer through c++ stdlib
  * [x] filelike support through python
    * [X] reading
    * [x] writing
  * [x] filelike support through c++ stdlib
    * [x] reading
    * [x] writing
  * [x] wrapping the C++ api for python
    * the c++ objects can be passed again to python
    * filesystemlike and filelike api must be wrapped in the other direction again
* [x] integrate `util::Path` all over the codebase
  * [x] reading QML files and watch root dir
  * [x] improve docx reading
  * [x] audio manager
  * [x] asset manager
    * texture loading
    * sound loading
* [x] generic `.call()` method for python objects
* [x] random fixes and improvements
* [x] integration of #718 because this pullrequests requires a reconversion anyways
* [x] visgrep interface cutter optimizations (faster by factor ~50)
* [x] slp parser optimizations (faster by factor ~40) (fixes #431)
* [x] virtual file reading for audio system
* [x] cvar default loading

works towards #691


libopenage/util/csv.h
```diff
@@ -0,0 +1,160 @@
...
+		std::vector<std::string> lines = file_map->at(fname);
+
+		for (auto &line : lines) {
+			line_count += 1;
+
+			// create writable copy, for tokenisation
+			if (unlikely(strbuf.size() <= line.size())) {
+				strbuf.resize(line.size() + 1);
+			}
+			memcpy(strbuf.data(), line.c_str(), line.size() + 1);
```
Tomatower: Why memcpy here?

std::copy(line.c_str(), line.c_str()+line.size(), std::back_inserter(strbuf));

libopenage/util/csv.h
```diff
@@ -0,0 +1,160 @@
...
+			// ignore comments and empty lines
+			if (line.empty() || line[0] == '#') {
+				continue;
+			}
+
+			// create writable copy, for tokenisation
+			if (unlikely(strbuf.size() <= line.size())) {
+				strbuf.resize(line.size() + 1);
+			}
+			memcpy(strbuf.data(), line.c_str(), line.size() + 1);
```
Tomatower: memcpy?

libopenage/util/csv.h
```diff
@@ -0,0 +1,160 @@
...
+			line_count += 1;
+
+			// create writable copy, for tokenisation
+			if (unlikely(strbuf.size() <= line.size())) {
+				strbuf.resize(line.size() + 1);
+			}
+			memcpy(strbuf.data(), line.c_str(), line.size() + 1);
+
+			// use the line copy to fill the current line struct.
+			int error_column = current_line_data.fill(strbuf.data());
```
Tomatower: Actually - Why not give the line.data() directly into the fill() method?

libopenage/util/csv.h
```diff
@@ -0,0 +1,160 @@
...
+			}
+
+			// create writable copy, for tokenisation
+			if (unlikely(strbuf.size() <= line.size())) {
+				strbuf.resize(line.size() + 1);
+			}
+			memcpy(strbuf.data(), line.c_str(), line.size() + 1);
+
+			// use the line copy to fill the current line struct.
+			int error_column = current_line_data.fill(strbuf.data());
```
Tomatower: same here - use line.data() or

libopenage/util/csv.h
TheJJ: because the fill method will modify the data, and i think that line.c_str() is const.

TheJJ:
I'll clean the whole csv reading up, it's really a mess...

libopenage/util/csv.h
Tomatower: Why does the fill method modify the data? strbuf is rewritten for every line

TheJJ: It was based on `strtok` once, not sure if it still does.

Tomatower: Ah now I see what you mean... It is finally consumed [here](https://github.com/SFTtech/openage/blob/master/libopenage/util/strings.cpp#L90) - and there write access is required - which is still not completely broken - if the string is not modified in length one can use ```&line[0]````- and continue to avoid any copy

jonaspm:
Hello everyone, I am liking this project but... why use docx instead of a more documented and libre option as .odt?

thanks in advance.

elnabo:
If I remember correctly the .docx files are in fact csv files.

ChipmunkV:
@jamesmcm, in other words: it's not *that* docx.

TheJJ:
This file suffix is purely a troll, in fact the files are just plain text. We wanted to use them as as "temporary solution" until we have [nyan](https://github.com/SFTtech/nyan), but this turned out to be way more complicated, so we still have them as of today, but they _will_ go away at least after my master thesis about nyan. :stuck_out_tongue:

TheJJ:
I'm finally done here. Massive overhaul n stuff, let's see how it works out. Please test and break it, especially the path system.

openage/convert/game_versions.py
```diff
@@ -19,32 +19,45 @@ class GameVersion(enum.Enum):
...
-        "African Kingdoms",
+    age2_hd_ak = (
+        "AOE2 HD African Kingdoms",
         True,
         {'AoK HD.exe', 'resources/_common/dat/empires2_x2_p1.dat',
          'resources/_packages/african-kingdoms/config.json'},
     )
+    # HD Edition v5.1+ with Rise of Rayas
+    age2_hd_rayas = (
+        "AOE2 HD Rise of Rayas (Version 5.x)",
```
goto-bus-stop: Nit: s/Rayas/Rajas

goto-bus-stop:
Convert stuff looks correct to me so far, I'll try a few versions in a minute.

TheJJ:
Hm damn I'm encountering some issues when I do:
```
./configure --mode=debug --prefix=/usr/games --raw-cmake-args -DPYTHON=/usr/bin/python3.6
make install DESTDIR=/tmp/oatest
LD_LIBRARY_PATH=/tmp/oatest/usr/games/lib64/ PYTHONPATH=/tmp/oatest/usr/lib64/python3.6/site-packages python3.6 ./openage game --asset-dir /tmp/oatest/usr/games/share/openage/
```

Then the global asset path isn't resolved correctly. Should not be hard to fix, let's see...

openage/__main__.py
```diff
@@ -103,10 +102,17 @@ def main(argv=None):
...
+    try:
+        from . import config
+
+        if args.no_devmode:
+            config.DEVMODE = False
+        if args.devmode:
+            config.DEVMODE = True
+    except ImportError:
+        if args.no_devmode or args.devmode:
+            print("code was not yet generated, "
```
elnabo: This line looks weird, shouldn't the comma be out of the first string ?

TheJJ: dunno, how'd you write it? :)

TheJJ:
I think it works now. Packagers, please test if this branch is packageable and works when installed. It should generate assets then and everything just work as you would think it should.

openage/__main__.py
elnabo: Well why not just ` print("code was not yet generated, ignoring devmode activation request")`
It should still be under the 100 char limit.

TheJJ: Mainly because my editor window is not that wide, it ends at ~75 somewhere, so I try to break lines visually. It's 84 now, I changed it.

elnabo: I didn't knew this syntax existed for string concatenation existed in python, still stuck in 2.7.

TheJJ: It's like the compiletime string constant concat in C, very handy.

PR #726, opened 2017-01-29
Log Qt version and bump to Qt 5.5

ChipmunkV:
Qt 5.5 has introduced the INFO log level that fits for the version printing.

### DIFF
--- a/doc/building.md
+++ b/doc/building.md
@@ -45,7 +45,7 @@ Dependency list:
        S  pycodestyle (or pep8 (deprecated))
     C     pygments
        S  pylint
-    CR    qt5 >=5.4 (Core, Quick, QuickControls modules)
+    CR    qt5 >=5.5 (Core, Quick, QuickControls modules)
 
       A   An installed version of any of the following (wine is your friend).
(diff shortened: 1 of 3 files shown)


ChipmunkV:
There is a split. There are directories with files where it's:

* openage-only (no Qt header inclusions allowed)
* mixed
* Qt-only (no openage header inclusions allowed)

In the "mixed" I haven't found a place where this kind of global initialization happens. For the "openage-only" I would have to create a neutral interface to pass the version. "Qt-only" part has that place where the global `QApplication` is created plus the log forwarding (a needed interface) is already in place.

zuntrax:
Then we should create a place in the "mixed" layer for that. I'd assume that Qt log messages are directly printed to the console. It would be nice to have a `QtMessageHandler` that passes the messages to our own logging subsystem.

ChipmunkV:
@zuntrax the "log forwarding" is the thing that passes the messages to our own logging subsystem. It was there from the beginning of the Qt GUI.

zuntrax:
Well, nevermind then. I didn't take the time to understand the gui subsystem yet.

PR #731, opened 2017-02-02
Check for asdf in codecompliance

### LINKED ISSUE #710: Blacklist ASDF in the compliance checker as invalid
@mic-e and I started to write `ASDF` or `asdf` in the code to denote things that must be changed before the pull request is merged. This is mainly to mark something so we can easily find it again.

To prevent accidential merges of code containing `asdf`, the compliance checker should reject it. You have to implement that in `buildsystem/codecompliance`.

Yes, currently there's `asdf` in the upstream code, and we need to get rid of it. The `global coordinate asdf` is resolved by #662.

Btjones711:
Fixes #710

### DIFF
--- a/buildsystem/codecompliance/textfiles.py
+++ b/buildsystem/codecompliance/textfiles.py
@@ -33,6 +33,9 @@ def find_issues(dirnames, exts):
         if '\r\n' in data:
             yield "Windows EOL format", filename
 
+        if any(asdf in data for asdf in ("ASDF", "asdf")) and not filename.endswith(__file__):
+            yield "asdf found indicating an immediate TODO", filename
+
         if data.endswith('\n\n'):
             yield "Trailing newline at file end", filename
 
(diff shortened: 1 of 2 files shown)


mic-e:
We'll have to fix all the occurences that your checker has found before merging this; see [this search](
https://github.com/SFTtech/openage/search?utf8=%E2%9C%93&q=asdf).

The occurences in `libopenage/log/test.cpp` and `libopenage/console/test.cpp` will simply have to be replaced by other test strings, e.g. `foo`, `bar` or `foobar`...

Not sure about the occurence in `doc/project_structure.md`; @TheJJ?

Finally, the occurence in `libopenage/engine.h` is pending my pull request #662 which will be done any time now™

It would be nice if the string we're testing for, 'asdf', didn't actually occur in the file itself; that way we'd avoid the ugly filename.endswith(__file__) check. For example, how about something like this:

    for forbidden_string in 'as'+'df', 'AS'+'DF':
        ...

TheJJ:
I think it would be better if we did the check with a regular expression instead of reading through the data twice. Then it doesn't match the file itself and it will be faster.

The `asdf` in `project_structure.md` is a perfect example why we need this compliance check 😛.

TheJJ:
Obsoleted by #776

PR #737, opened 2017-02-10
Quaternion

TheJJ:
Something i did a long time ago...

### DIFF
--- a/libopenage/util/vector_test.cpp
+++ b/libopenage/util/vector_test.cpp
@@ -1,4 +1,4 @@
-// Copyright 2015-2016 the openage authors. See copying.md for legal info.
+// Copyright 2015-2017 the openage authors. See copying.md for legal info.
 
 #include "vector.h"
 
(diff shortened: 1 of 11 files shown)


libopenage/util/vector_test.cpp
```diff
@@ -10,33 +10,40 @@ namespace tests {
...
 		// we want to be able to reuse the variable names later.
 		const Vector<2> a(1.0, 2.0);
 		const Vector<2> b(3.0, 4.0);
 		Vector<2> c;
 
+		// test basic operators.
 		c = a + b;
-		TESTEQUALS(c[0], 4.0);
-		TESTEQUALS(c[1], 6.0);
+		c.equals({4.0, 6.0}) or TESTFAIL;
```
Tomatower: Test the c.equals function (that it can return true and false)

TheJJ: is done above, now more explicit with `not ... or TESTFAIL`.

PR #742, opened 2017-02-16
Population

MaanooAk:
- [X] Population attribute
- [x] Population logic
- [x] Population gui

Can anyone give a hint, how can I bind a variable (player.population) to the population slot on the gui/qml?

### DIFF
--- a/libopenage/gamestate/generator.cpp
+++ b/libopenage/gamestate/generator.cpp
@@ -289,6 +289,7 @@ void Generator::add_units(GameMain &m) const {
 			m.placed_units.new_unit(*mvtype, p, tile.to_tile3().to_phys3());
 			tile.se += 1;
 			m.placed_units.new_unit(*fvtype, p, tile.to_tile3().to_phys3());
+			// TODO uncomment when the scout looks better
 			//tile.se += 2;
 			//m.placed_units.new_unit(*sctype, p, tile.to_tile3().to_phys3());
 		}
--- a/libopenage/gamestate/generator.cpp
+++ b/libopenage/gamestate/generator.cpp
@@ -289,8 +289,8 @@ void Generator::add_units(GameMain &m) const {
 			m.placed_units.new_unit(*mvtype, p, tile.to_tile3().to_phys3());
 			tile.se += 1;
 			m.placed_units.new_unit(*fvtype, p, tile.to_tile3().to_phys3());
-			tile.se += 2;
-			m.placed_units.new_unit(*sctype, p, tile.to_tile3().to_phys3());
+			//tile.se += 2;
+			//m.placed_units.new_unit(*sctype, p, tile.to_tile3().to_phys3());
 		}
 	}
 }
--- a/libopenage/gamestate/generator.cpp
+++ b/libopenage/gamestate/generator.cpp
@@ -267,7 +267,8 @@ void Generator::add_units(GameMain &m) const {
 			auto tctype = p.get_type(109); // town center
 			auto mvtype = p.get_type(83);  // male villager
 			auto fvtype = p.get_type(293); // female villager
-			if (!tctype || !mvtype || !fvtype) {
+			auto sctype = p.get_type(448); // scout cavarly
+			if (!tctype || !mvtype || !fvtype || !sctype) {
 				break;
 			}
 
@@ -288,6 +289,8 @@ void Generator::add_units(GameMain &m) const {
 			m.placed_units.new_unit(*mvtype, p, tile.to_tile3().to_phys3());
 			tile.se += 1;
 			m.placed_units.new_unit(*fvtype, p, tile.to_tile3().to_phys3());
+			tile.se += 2;
+			m.placed_units.new_unit(*sctype, p, tile.to_tile3().to_phys3());
 		}
 	}
 }
--- a/libopenage/gamestate/player.cpp
+++ b/libopenage/gamestate/player.cpp
@@ -17,9 +17,10 @@ Player::Player(Civilisation *civ, unsigned int number, std::string name)
 	civ{civ},
 	name{name},
 	team{nullptr},
-	population{0},
-	population_cap{0} {
+	// TODO change, get population cap max from game options
+	population{0, 200} {
 	// starting resources
+	// TODO change, get starting resources from game options
 	this->resources[game_resource::food] = 1000;
 	this->resources[game_resource::wood] = 1000;
 	this->resources[game_resource::stone] = 1000;
@@ -128,10 +129,10 @@ void Player::active_unit_added(Unit *unit) {
 	if (unit->has_attribute(attr_type::population)) {
 		auto popul = unit->get_attribute<attr_type::population>().population;
 		if (popul < 0) {
-			population_cap += -popul;
+			population.add_capacity(-popul);
 		}
 		else if (popul > 0) {
-			population += popul;
+			population.add_population(popul);
 		}
 	}
 }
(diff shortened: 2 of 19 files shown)


zuntrax:
The glue code lives in`gui/*_link.{cpp,h}`. I still get confused about the details there, @ChipmunkV built almost all of it, he can tell you more about it.

ChipmunkV:
Kind of like `ActionMode::ability`. A signal in `ActionModeSignals` (`libopenage/game_control.h
`) and a `Q_PROPERTY` in `ActionModeLink`. And link them together as for `ability`. That will make `root.actionMode.population` available to use in `IngameHud.qml`. (don't forget to add it to `ActionMode::announce()`)

There isn't anything to trigger `ActionModeSignals::population_changed`, so it looks similar to how the resource amounts currently updated in `ActionMode::render()` by calling `announce_resources()` all the time.

MaanooAk:
The population for each player needs these variables (with example values):
- `population` (120)
- `population_cap_static` (0) (from civ bonuses) (gai should have infinity)
- `population_cap_total` (220) (`population_cap_static` + from buildings)
- `population_cap_max` (200) (game setup max cap + civ bonuses)
- `population_cap` (200) (`population_cap_total` limited by `population_cap_max`)

My point, should I just create a new class Population?

TheJJ:
`PopulationTracker` maybe, I dunno. But we need to watch the numbers in some class, so yea, go ahead.

libopenage/gui/game_control_link.cpp
```diff
@@ -124,13 +128,23 @@ void ActionModeLink::on_buttons_type_changed(const ActionButtonsType buttons_typ
 	emit this->buttons_type_changed(buttons_type);
 }
 
+void ActionModeLink::on_population_changed(int population, int population_cap) {
+	this->population = QString::number(population) + "/" + QString::number(population_cap);
+	log::log(MSG(warn) << "we are here " << this->population.toUtf8().constData());
```
MaanooAk: Code reaches here, but in the gui the population slot is empty

libopenage/gui/qml/IngameHud.qml
```diff
@@ -109,7 +109,7 @@ Item {
 					sourceComponent: resourceIndicator
 
 					onLoaded: {
-						item.amount = "13/42"
+						item.amount = root.actionMode.population
```
ChipmunkV: `onLoaded` is executed only once. Wrap it in `Qt.binding(function() { return root.actionMode.population })` like for the resources.

MaanooAk: Thanks a lot, you just saved me 20 hours of debugging :)

MaanooAk:
Should be revisited after unit triggers are implemented

libopenage/gamestate/generator.cpp
```diff
@@ -288,6 +289,8 @@ void Generator::add_units(GameMain &m) const {
 			m.placed_units.new_unit(*mvtype, p, tile.to_tile3().to_phys3());
 			tile.se += 1;
 			m.placed_units.new_unit(*fvtype, p, tile.to_tile3().to_phys3());
+			//tile.se += 2;
+			//m.placed_units.new_unit(*sctype, p, tile.to_tile3().to_phys3());
```
TheJJ: leftovers?

libopenage/gamestate/population_tracker.h
```diff
@@ -0,0 +1,72 @@
...
+	int population;
+
+	int capacity_static;
+	int capacity_real;
+	int capacity_max;
+
+	// generated values
+
+	int capacity_total;
+	int capacity;
```
TheJJ: documentation pls

libopenage/gamestate/player.cpp
```diff
@@ -111,5 +114,46 @@ void Player::initialise_unit_types() {
...
+	}
+
+	// TODO handle here building dependencies
+	// TODO handle here on death unit triggers
+
+	// population
+	if (unit->has_attribute(attr_type::population)) {
+		auto popul = unit->get_attribute<attr_type::population>().population;
+		if (popul < 0) {
+			population.add_capacity(popul);
```
TheJJ: to model buildings as negative population is kinda hacky :D 
then again, the only other way would be to have two attributes ("housing space", "housing demand"?).
Maybe we wanna have units that first require an empty population slot, but once created, it provides more living space.

libopenage/gamestate/player.cpp
```diff
@@ -111,5 +114,46 @@ void Player::initialise_unit_types() {
...
+		}
+		else if (popul > 0) {
+			population.add_population(-popul);
```
TheJJ: i'd also implement that as separate functions for freeing and aquiring a slot, e.g. `demand_population` or something like that.

libopenage/gui/qml/IngameHud.qml
```diff
@@ -87,6 +88,22 @@ Item {
...
+
+							visible: warning
+
+							color: "#80FFC100"
+
+							SequentialAnimation on opacity {
+								loops: Animation.Infinite
+								PropertyAnimation { from: 0; to: 1; duration: 250 }
+								PropertyAnimation { from: 1; to: 0; duration: 250 }
+							}
```
TheJJ: fancy.

libopenage/unit/producer.cpp
```diff
@@ -564,6 +569,14 @@ void BuildingProducer::initialise(Unit *unit, Player &player) {
 	unit->add_attribute(std::make_shared<Attribute<attr_type::hitpoints>>(this->unit_data.hit_points));
 	unit->add_attribute(std::make_shared<Attribute<attr_type::damaged>>(this->unit_data.hit_points));
 
+	// population
+	if (this->id() == 109 || this->id() == 70) { // Town center, House
+		unit->add_attribute(std::make_shared<Attribute<attr_type::population>>(-5));
+	}
+	else if (this->id() == 82) { // Castle
+		unit->add_attribute(std::make_shared<Attribute<attr_type::population>>(-20));
+	}
```
TheJJ: maybe you can get the exact amount from the gamedata?
https://github.com/SFTtech/openage/blob/5f6d4f2078b5a14d312041e14a8d8062f0b224c4/openage/convert/gamedata/unit.py#L233
But that's probably not the values that buildings grant, instead it seems that it's the slot price of a unit.

libopenage/gamestate/generator.cpp
MaanooAk: I thought it will be nice start with a scout (mostly) like the original, but then I saw the idle animation and it doesn't have the frame pause (#748 part 2) so it looks weird. If you don't mind leave there to uncomment when the scout looks better...

libopenage/gamestate/population_tracker.h
MaanooAk: done

libopenage/gamestate/generator.cpp
TheJJ: kay :) then add a comment that says why it was left there.

libopenage/gamestate/player.cpp
MaanooAk: Sounds good, I will split it to `population.demand` and `population.capacity`

libopenage/gamestate/player.cpp
MaanooAk: Are you thinking the same about capacity, capacity_static and capacity_max? (It will be too much I think)

TheJJ: :+1:

TheJJ:
Thank you for helping us help you help us all :+1:

Issue #732, opened 2017-02-04
[WIP] doc: Possible entries and options for a settings menu


heinezen:
Since OpenAge will support more options than the original game, we likely have to redesign the entire settings menu from scratch. Before we do that, it would be beneficial to enumerate all entries for our menu. Layouters or designers can then use this information to draft a GUI, e.g. for #728 .

The options we have to enumerate (as suggested by @coffenbacher) are:

- [ ] all useful settings available up through the latest UP / Voobly / HD 
- [ ] any settings we see demand for that aren't even in UP but will highly likely be in OpenAge
- [ ] any settings for any highly likely OpenAge-specific features

Additionally it would be nice, if we could sort all entries into categories.

Inspiration can also be drawn from other (RTS-)games with nice and highly configurable interfaces like DoTA2.

Feel free to suggest any options/entries/categories/nice-to-have-features in the comments. I'll add everything to this [gist](https://gist.github.com/heinezen/68b828708a5e2521b786b4e398d43f18).

heinezen:
I added the options that are supported in the current version of AoE2:HD.

heinezen:
Settings enumeration for the original game is now complete. Additionally I had a look at the settings of other strategy games (AoM, AoE3, Rise of Nations, CnC Generals, Company of Heroes, Starcraft) and added them to the gist, if they seemed appropriate. You can now add anything that ~~will probably result in feature creep~~ you think is still missing, if you wish ^^

@TheJJ @coffenbacher @ChipmunkV

TheJJ:
My updates: https://gist.github.com/TheJJ/34eb1fb188333d7613e9edb30f26d9d2

heinezen:
@TheJJ Nice! I wouldn't agree on excluding resolution and aspect ratio options, though. What if someone wants to play fullscreen but with a 5:4 ratio (and black bars on each side)? In addtion to this, a scalable UI can be tiny and unreadable on 4K screens. It can also be a little tedious to resize a window in some DEs (Unity, i'm talking about you), so a fue selectable resolution options could get rid of that problem.

ChipmunkV:
We've probably mentioned the 'font size' option. It actually helps against the 'unreadable UI'.

https://github.com/SFTtech/openage/issues/673

heinezen:
@ChipmunkV Ah, that's convenient. I'll add *Font Size* to the gist.

heinezen:
[...]

Issue #728, opened 2017-01-30
Graphical configuration interface


TheJJ:
Yes, we don't have many things to configure right now, but we need an interface.

* [ ] Use QtQuick to draw a config UI
  * Text entries
  * Sliders
  * Cancel and OK button
* [ ] Display settings from the cvar subsystem
* [ ] Write back changes done in the ui (cvar system probably doesn't support that yet)

heinezen:
Wouldn't it be better to start with working out everything that we want to be configured and then create a mockup first? That way we can seperate design prototypes and actual implementation. The problem is that the UI could easily become cluttered and confusing if more stuff is added later on. Creating a mockup should be easy if we use tools like [Pencil](https://github.com/evolus/pencil/). Design Process would probably look like this:

- Decide what options should be configurable
- Create a mockup that includes everything
- Review the usability of the mockup
- Implement the functionality of the options menu in Qt using the mockup as reference

Here is an example of a simple mockup made in pencil.

![mockup](https://cloud.githubusercontent.com/assets/6852422/22450654/76faa902-e768-11e6-83e5-b6b56c08fef4.png)

TheJJ:
Cool :)
I agree, but someday we have to create the ui in practice. Prototyping first is of course great, but we don't have that many config options yet, so we have to adjust the ui once more values end up there. Otherwise we'll prototype forever.

ChipmunkV:
I usually prototype straight in QML... :smile:

heinezen:
> we don't have that many config options yet, so we have to adjust the ui once more values end up there.

That's the point though. Creating an interface and then adding features to it as they come is very inefficient and leads to UIs that feel unintuitive and cluttered because you have to "tag on" every option. Or worse, the UI would have to be rewritten after 4-5 additional GUI elements.

> Otherwise we'll prototype forever.

Uuuh no ;-) . Designing the interface with a mockup would take 1-2 days max. I created the mockup from my above post in 30 mins. The beauty of mockups is that you can design the interface with every feature you possible want in the options and it's **fast** and easy to do. A well designed mockup could also last one year until every feature is implemented in the "real" GUI in Qt.

> I usually prototype straight in QML... :smile:

Too slow but I know what you mean ^^

[...]

Issue #673, opened 2016-12-03
HUD scaling for high-DPI displays


TheJJ:
Our Qt-GUI should easily be scalable by some slider to be big on high-DPI displays as well.

As the system supports arbitrary resolutions, it should also be zoomable to have the size the user desires.

ChipmunkV:
The UI scale is based around the font. Currently, the font is taken from the OS. So, if player can see the text on the desktop - then the scale of the game UI would be hopefully fine too.

A font size slider won't hurt ingame. That's also the beginning of the "settings" subsystem (or we have something like that already?).

TheJJ:
We have the cvar stuff, which is the base for the settings I'd say. But yes, that would be a good start to have some settings UI and also store the updated font factor then.

Issue #107, opened 2014-11-04
Translations


airon90:
Hi, is there a plan to translate openage? I could localize it in Italian and Esperanto

TheJJ:
https://doc.qt.io/qt-5/qtquick-internationalization.html

heinezen:
Best Practices for localization from the IGDA:

https://cdn.ymaws.com/www.igda.org/resource/collection/2DA60D94-0F74-46B1-A9E2-F2CE8B72EA4D/Best-Practices-for-Game-Localization-v22.pdf

https://cdn.ymaws.com/www.igda.org/resource/collection/2DA60D94-0F74-46B1-A9E2-F2CE8B72EA4D/Localization-ready-2.1.pdf

TakesTheBiscuit:
Worth having a quick think over the implications from the UI -- this is after all a game; and i've found i the past in certain Steam games (Team Fortress 2 used to suffer for example) that when someone uses a different language set in their name it explodes the visuals on screen and breaks the game.

I would imagine that certain Russian words and German words will be rather long and presents annoying truncation or worse strings exploding out of where they should be an overlaying or obscuring maps/units/actual gameplay.

Anyway </2centsOnly>

simonsan:
> We should probably use:
> 
> https://doc.qt.io/qt-5/qtlinguist-index.html
> https://doc.qt.io/qt-5/qtquick-internationalization.html

Just some thoughts: I know about tr(); in Qt and QtLinguist, just wanted to let Mozillas localisation system "Project Fluent" here - it also has full bidirectionality support, based on clear text files and with this is clearly addressing non-developers. Although it might be easier to implement basic translations with tr() and external Software like QtLinguist, it could make sense to use something where the community can easily help with the translations without the hurdle of another software and just by using a texteditor. But in the end I don't have any experiences with Fluent, but wanted to let it here as an "alternative". 

https://github.com/projectfluent/python-fluent

Syntax-Guide
https://projectfluent.org/fluent/guide/

On the other side, there are tools to use the Google API to help translate Qt Linguist TS files, so in the end, it might really be the easiest way for everyone to stay with Qt own functions.

duanqn:
I can sign up for translation from English to Chinese (simplified & traditional).

TheJJ:
[...]

PR #749, opened 2017-02-18
[WIP] Settings menu placeholder

ChipmunkV:
#673 
#728 
#732 
https://heinezen.github.io/openage/#g=1&p=home

Here it's like another `OutputMode`. Need to attach it to cvars.


lisacvuk:
Once compiled, how do I test it? Thanks!

lisacvuk:
Screenshots:    
http://i.imgur.com/LK912Ey.png  
http://i.imgur.com/oNTtv6E.png  
The rest of the tabs are empty.

heinezen:
@lisacvuk Testing of the whole UI happens in #751 ;)

libopenage/cvar/cvar.cpp
```diff
@@ -28,21 +56,146 @@ std::string CVarManager::get(const std::string &name) const {
...
+		if (!this->config_loading) {
+			std::lock_guard<std::mutex> lock(this->writeback->elements_mutex);
+
+			this->writeback->elements.emplace_back(name, value);
+
+			if (this->writeback->elements.size() == 1) {
+				auto writeback_shared = this->writeback;
+
+				this->job_manager->enqueue<int>([writeback_shared] {
+					std::lock_guard<std::mutex> lock(writeback_shared->elements_mutex);
```
ChipmunkV: Felt slow when was changing a parameter by scrolling the mousewheel.

heinezen:
OK, finally had a look at it myself :)

1. Instead of anchoring the tabs in the middle of the top, an anchor point further to the left would be nice. After all, this is were most players will look first.
2. Seeing how `Font Size` also affects UI scale, shouldn't we just call the option "UI Scale"? Furthermore the selection of a font size is very finicky and requires readjusting the mouse pointer every time. We either should have presets (like I did it in the mockup) or a `UI Scale` slider with percentage numbers.

TheJJ:
I changed the cvar manager a bit in #723, how should we proceed? Which one should go first? :)

ChipmunkV:
@TheJJ at a first glance, #723 doesn't deal with file contents. So it'll be easy.

I've got time to merge settings later.

TheJJ:
cvar stuff is merged, you may adapt this now please.

TheJJ:
Ping? :)

ChipmunkV:
Two things:
* can't figure out into what file to write the changes (so hardcoded the keybinds.oac)
* can't write the changes

TheJJ:
For those two things, I just created #813 :)

ChipmunkV:
For the second one it's about inability to edit files with the new path system. I still need to figure out why it doesn't accept `open("a+")`.

TheJJ:
Any luck finding out why `a+` doesn't work? Does `w` work?

TheJJ:
I implemented file append modes in #996.
I'm not sure though how we should edit our config files properly, because it's quite tricky to a) allow users to write the files as they like and b) update just the values at the right place automatically. Basically we'd have to parse the files and then rewrite them retaining all information including comments and whitespace, etc.

ChipmunkV:
> Basically we'd have to parse the files and then rewrite them retaining all information including comments and whitespace, etc.

That's roughly what this PR does. But it deletes the comment on the line where an option is overwritten (probably not the worst thing to do when I think about it).

TheJJ:
Ah okay, I didn't realize that since you asked for the append functionality.
Is something else missing or is this nearly reviewable/mergeable?

duanqn:
Should the text here go to openage-data ? What's our plan on translation?

TheJJ:
The translation issue is #107. Generally, engine-message-translations are within the `openage` repo, as openage-data is only for the default and free asset pack.

duanqn:
> The translation issue is #107. Generally, engine-message-translations are within the `openage` repo, as openage-data is only for the default and free asset pack.

Yeah I believe in #107 you mentioned some of these strings should go to openage-data but I'm not sure which ones should go. Maybe the settings is consistent across all mods so those strings should stay in this repo?

I think this pull request added some strings on the UI. We better store them in a resource folder somewhere so that it's easier to do translation afterwards.
Also I didn't read the code -- probably the author already did that -- but I think we should make sure to use wide char and UTF8 for all strings exposed to UI.

TheJJ:
Exactly: settings dialogs and any other engine text that remains the same and is independent of loaded games, modpacks, scripts, ... belong in this repo. Texts that are provided in a modpack has to be translated there. How we implement the translation system in general is not decided yet, and that discussion should happen in #107.

PR #754, opened 2017-02-25
GUI/QML:  qualified imports for the openage objects

ChipmunkV:
To distinguish between openage types and standard qt types in qml: "OA" prefix like `OA.ActionMode` or `OA.GameMain`.


ChipmunkV:
Probably fixed in 5.8:

https://bugreports.qt.io/browse/QTBUG-10822
https://bugreports.qt.io/browse/QTBUG-28639

Issue #670, opened 2016-11-14
Replace MurmurHash by SipHash


TheJJ:
https://emboss.github.io/blog/2012/12/14/breaking-murmur-hash-flooding-dos-reloaded/

instead, use https://en.wikipedia.org/wiki/SipHash

located in `libopenage/rng/rng.cpp`.

TheJJ:
done by #755 \o/

PR #755, opened 2017-02-28
Implement siphash

martinber:
This is my attempt on #670.

I want to know if I'm doing something incorrectly, I'm fairly new to programming. Later I'll implement some tests.

I implemented myself the siphash algorithm by looking into [the paper](https://131002.net/siphash/siphash.pdf), maybe you instead want to use an existing implementation?

I don't know how to handle the key for Siphash, that key should be always the same one. I think it should be on rng.h and change the value to something random (now I'm using the default key described on the paper). 

As far as I can see, Murmurhash was only used for seeding the RNG, I don't understand how the attack described in [the article](https://emboss.github.io/blog/2012/12/14/breaking-murmur-hash-flooding-dos-reloaded/) affects the RNG.

### DIFF
--- a/libopenage/rng/rng.cpp
+++ b/libopenage/rng/rng.cpp
@@ -4,9 +4,10 @@
 
 #include <random>
 #include <sstream>
-#include <string.h>
+#include <array>
 
 #include "../util/timing.h"
+#include "../util/hash.h"
 #include "../error/error.h"
 
 
@@ -20,6 +21,13 @@ namespace openage {
  * RNG, so don't use it as one. If you do, I'll tell my mom.
  */
 namespace rng {
+
+
+// Key for seed hashing, just some hardcoded key
+static const std::array<uint8_t, 16> seed_key{
+	0xba, 0xda, 0x55, 0x00, 0x5e, 0xed, 0x00, 0x00,
+	0xab, 0xcd, 0xef, 0x12, 0x34, 0x56, 0x78, 0x90
+};
 
 
 RNG::RNG(uint64_t v1) {
@@ -47,84 +55,20 @@ void RNG::seed(uint64_t val) {
 }
 
 
-/**
- * murmurhash3 bit avalancher - output is uniformly distributed
- * and changing one bit of input results in ~1/2 of output bits changed
- * This function is a good way of mixing up an input,
- * or mixing together two results with hash64(val1 ^ val2)
- *
- * https://en.wikipedia.org/wiki/MurmurHash
- */
-static uint64_t hash64(uint64_t val) {
-	// The two magic numbers have been determined heuristically.
-	// They avalanche all bits in val to within 0.25% bias.
-
-	val ^= val >> 33;
-	val *= 0xff51afd7ed558ccd;
-	val ^= val >> 33;
-	val *= 0xc4ceb9fe1a85ec53;
-	return val ^ (val >> 33);
-}
-
-
-/*
- * reinterpret_casts 8 chars from a buffer as a uint64_t.
- */
-static inline uint64_t from_cptr(const char *data) {
-	uint64_t result;
-	memcpy(reinterpret_cast<char *>(&result), data, sizeof(result));
-	return result;
-}
-
-
-/*
- * hashes an initial seed into the base state
- */
-static void init_hash(uint64_t *state, uint64_t seed) {
-	// Mix the initial seed into data for each state value
-	state[0] = hash64(seed);
-	state[1] = hash64(seed ^ state[0]);
-}
-
-
-/*
- * For each 8-byte block after the initial seed,
- * mix it with state[0] and set state[0] equal to the result.
- * Then, set state[1] to the mixture of state[1] and state[0]
- * This lets each bit from the input play a role in the output
- * w/o requiring that the size be a multiple of 8 or having extra
- * code for the non multiple-of-8 sections
- */
-static void hash_bytes(uint64_t *state, const char *data, size_t count) {
-	for (size_t i = 1; i <= count; i++) {
-		state[0] = hash64(state[0] ^ from_cptr(data+i));
-		state[1] = hash64(state[0] ^ state[1]);
-	}
-}
-
-
 void RNG::seed(const void *dat, size_t count) {
 	if (count == 0) {
 		throw Error(MSG(err) << "0 bytes supplied as seed data");
 	}
 
-	constexpr uint64_t default_seed = 0x0123456789abcdef;
-	constexpr size_t uint64_s = sizeof(uint64_t);
-	const char *data = static_cast<const char *>(dat);
+	const uint8_t *data = static_cast<const uint8_t *>(dat);
 
-	if (count > uint64_s) {
-		init_hash(state, from_cptr(data));
-		hash_bytes(state, data, count - uint64_s);
-	}
-	else {
-		uint64_t init_seed = 0;
-		memcpy(&init_seed, data, count);
-		init_hash(state, init_seed);
-	}
+	openage::util::Siphash siphash(seed_key);
 
-	if (!(state[0] || state[1])) {
-		state[0] = default_seed;
-	}
+	state[0] = siphash.digest(data, count);
+	// Previously, for state[1] we hashed (data ^ state[0])
+	// As `data` now has variable length I can't do that, I think that just
+	// hashing `state[0]` is enough
+	state[1] = siphash.digest(state[0]);
 
 	// This helps prevent low-zero states caused by a bad seed
 	this->discard(discard_on_seed);
--- a/libopenage/util/misc.h
+++ b/libopenage/util/misc.h
@@ -1,6 +1,13 @@
-// Copyright 2013-2016 the openage authors. See copying.md for legal info.
+// Copyright 2013-2017 the openage authors. See copying.md for legal info.
 
 #pragma once
+
+#include <string.h>
+#include <limits.h>
+#include <cmath>
+
+#include "../error/error.h"
+
 
 namespace openage {
 namespace util {
@@ -39,8 +46,8 @@ T mod(T x) {
  */
 template<typename T, int amount>
 T rol(T x) {
-	static_assert(sizeof(T) > amount && amount > 0, "invalid rotation amount");
-	return (x << amount) | (x >> (sizeof(T)*8 - amount));
+	static_assert(sizeof(T)*CHAR_BIT > amount && amount > 0, "invalid rotation amount");
+	return (x << amount) | (x >> (sizeof(T)*CHAR_BIT - amount));
 }
 
 
@@ -57,7 +64,7 @@ inline T div(T x, T m) {
  * generic callable, that compares any types for creating a total order.
  *
  * use for stdlib structures like std::set.
- * the template paramter has to be a pointer type.
+ * the template parameter has to be a pointer type.
  */
 template<typename T>
 struct less {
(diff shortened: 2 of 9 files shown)


libopenage/rng/rng.cpp
```diff
@@ -48,58 +49,93 @@ void RNG::seed(uint64_t val) {
...
-	val ^= val >> 33;
-	val *= 0xc4ceb9fe1a85ec53;
-	return val ^ (val >> 33);
+* Rotate left no carry
+*
+* http://blog.regehr.org/archives/1063
+*/
+inline uint64_t rotl64(uint64_t x, uint64_t n)
+{
+	assert (n<64);
```
TheJJ: we have a special `ENSURE` macro that you should use instead of `assert`.

libopenage/rng/rng.cpp
```diff
@@ -48,58 +49,93 @@ void RNG::seed(uint64_t val) {
...
-
-/*
- * reinterpret_casts 8 chars from a buffer as a uint64_t.
- */
-static inline uint64_t from_cptr(const char *data) {
-	uint64_t result;
-	memcpy(reinterpret_cast<char *>(&result), data, sizeof(result));
-	return result;
+// Maybe this should be a lambda inside of siphash()?
+inline void sipround(uint64_t *v) {
```
TheJJ: I think the compiler will like it more if leave it like this (maybe create a `static inline void`).
You can try though, create it as a `static auto round = [](uint64_t *v)` and use `objdump -d` on the `rng.o` file to see how it was compiled.

libopenage/rng/rng.cpp
```diff
@@ -48,58 +49,93 @@ void RNG::seed(uint64_t val) {
...
-/*
- * hashes an initial seed into the base state
- */
-static void init_hash(uint64_t *state, uint64_t seed) {
-	// Mix the initial seed into data for each state value
-	state[0] = hash64(seed);
-	state[1] = hash64(seed ^ state[0]);
-}
+// I don't know where to put this, or what value should we use
+static const uint64_t siphash_key[2]{0x0706050403020100ull, 0x0f0e0d0c0b0a0908ull};
```
TheJJ: As I understood it, siphash can (and should) be used with a secret key. This means we should not hardcode it, instead I'd say we wrap siphash in another class, which is initialized by this key.
The key is then determined either fixed (e.g. for consistent map generation) or random (e.g. for each game).

libopenage/rng/rng.cpp
```diff
@@ -48,58 +49,93 @@ void RNG::seed(uint64_t val) {
...
+		size_t rem_bytes{len_data};
+		const uint8_t *idx{data}; // Points to the start of the current block being copied
+		uint64_t block;
+		bool finished{false};
+
+		do {
+			block = 0;
+			if (rem_bytes < 8) {
+				finished = true;
+				memcpy(reinterpret_cast<uint8_t *>(&block), idx, rem_bytes);
```
TheJJ: I'm thinking about how it could be possible to work read-only on the input data and just get back the result then (to avoid `memcpy`)

libopenage/rng/rng.cpp
```diff
@@ -48,58 +49,93 @@ void RNG::seed(uint64_t val) {
...
+				block |= ending;
+			}
+			else {
+				memcpy(reinterpret_cast<uint8_t *>(&block), idx, sizeof(block));
```
TheJJ: I think you can avoid this memcpy by just assigning the value of reinterpreted values, be careful about the memory alignments though :)

libopenage/rng/rng.cpp
```diff
@@ -110,18 +146,18 @@ void RNG::seed(const void *dat, size_t count) {
...
-		memcpy(&init_seed, data, count);
-		init_hash(state, init_seed);
-	}
-
+	const uint8_t *data = static_cast<const uint8_t *>(dat);
+
+	// Previously, for state[1] we hashed (data ^ state[0])
+	// I think that now the easiest way to do that is xoring the key
+	uint64_t key[2];
+	memcpy(key, siphash_key, uint64_s * 2);
```
TheJJ: why do we need the siphash_key anyway here? can't we just use the seed data?

libopenage/rng/rng.cpp
martinber: That's nice, I did not realize I can do that. I can replace it with:

```
block = *reinterpret_cast<const uint64_t *>(idx);
```

Yesterday I started reading about alignments, that's new for me, and I'm not sure if removing the memcpy can give me problems. I tried and it works, but I'm afraid of doing something machine-dependent. Also now I'm afraid of having problems about endianness, I'll try to learn more about that.

libopenage/rng/rng.cpp
martinber: As far I understand that the options we have are (correct me if I'm wrong :P):

- Use a hardcoded key, but it's going to be visible to everyone who reads the code
- Generate a key from something like current time, but players will not be able to share seeds
- Generate a key from the seed, but I think that it defeats the purpose of having a key

I think that any of those option are okay

libopenage/rng/rng.cpp
martinber: I tried both options and the generated rng.cpp.o was the same apart of some register names. So I made it a lambda because of the scope, so now sipround() it's inside of siphash().

libopenage/rng/rng.cpp
```diff
@@ -48,58 +48,93 @@ void RNG::seed(uint64_t val) {
...
-	val ^= val >> 33;
-	val *= 0xff51afd7ed558ccd;
-	val ^= val >> 33;
-	val *= 0xc4ceb9fe1a85ec53;
-	return val ^ (val >> 33);
+* Rotate left no carry
+*
+* http://blog.regehr.org/archives/1063
+*/
+static inline uint64_t rotl64(uint64_t x, uint64_t n)
```
martinber: I just realized that there is already a rotate left function in `libopenage/util/misc.h`, I'll use that one instead

martinber:
Now I think that it's almost ready.

For now the key for siphash it's hardcoded on `rng.cpp`, I don't  know what to do with that. I think that the options are:

- Use a hardcoded key, but it's going to be visible to everyone who reads the code
- Generate a key from something like current time, but players will not be able to share seeds
- Generate a key from the seed, but I think that it defeats the purpose of having a key

Then I don't know if memory alignments can give me problems when I reinterpret uint8_t arrays to uint64_t or vice versa. I think that the safe way would be to use memcpy. What do you think?

TheJJ:
I think it's ok to hardcode it for the rng for now, we seed it anyway.
We won't need sharing seeds as the simulation is authoritatively manged by the server, which has the seed. Players may have local rngs for some visualizations.

libopenage/util/misc.h
```diff
@@ -66,5 +73,95 @@ struct less {
 	}
 };
 
+
+// Size of uint64_s measured in bytes
+static constexpr size_t uint64_s = 8;
```
TheJJ: size of `uint64_t` i assume? And you can do it with `= sizeof(uint64_t)`, as this is the amount of chars, and we use `CHAR_BIT`, so it is consistent and portable.

libopenage/util/misc.h
```diff
@@ -66,5 +73,95 @@ struct less {
...
+ * Convert a C-style array of uint8_t to a uint64_t. Does not change endianness.
+ * Fills the remaining bytes with zeroes.
+ *
+ * @param start Pointer to start of the input data.
+ * @param count Number of bytes to read.
+ * @return Input data as a 64 bit number.
+ */
+inline uint64_t
+array8_to_uint64(const uint8_t *start, size_t count) {
+	if (count > 8) {
```
TheJJ: here you can use `uint64_s` then.

libopenage/util/misc.h
```diff
@@ -66,5 +73,95 @@ struct less {
...
+		throw Error(MSG(err) << "Tried to copy more than 8 bytes");
```
TheJJ: `more than " << uint64_s << " bytes"`

libopenage/util/misc.h
```diff
@@ -66,5 +73,95 @@ struct less {
...
+uint64_to_array8(const uint64_t value) {
+	std::vector<uint8_t> result(uint64_s, 0);
+	memcpy(reinterpret_cast<uint64_t *>(result.data()), &value, uint64_s);
+	return result;
+}
+
+
+// Get return array size for array8_to_array64()
+// Otherwise
+inline constexpr size_t array64_size(size_t count) {
```
TheJJ: doxygen comment may also be helpful here.

libopenage/util/misc_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+
+void array_conversion() {
+	{
+		static const std::vector<uint8_t> test_array8{
+			0x01, 0x23, 0x45, 0x67, 0x89, 0xab, 0xcd, 0xef,
+			0xf0, 0xf1, 0xf2, 0xf3, 0xf4, 0xf5, 0xf6, 0xf7
+		};
+		// The same array but as uint64_t
+		// ASDF: Can we just assume that everything is little-endian?
```
TheJJ: hehe no :) you'd need to add one case for each endianness (for runtime switch, you maybe wanna use http://stackoverflow.com/a/1001330/1348535)

libopenage/util/misc_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+		array8_to_array64(test_array8.data(), 16) == test_array64 or TESTFAIL;
+		array64_to_array8(test_array64.data(), 2) == test_array8 or TESTFAIL;
+	}
+	{
+		static const std::vector<uint8_t> test_array8{
+			0x01, 0x23, 0x45, 0x67, 0x89, 0xab, 0xcd, 0xef,
+			0xf0, 0xf1, 0xf2, 0xf3, 0xf4, 0xf5, 0x00, 0x00
+		};
+		// The same array but as uint64_t
+		// ASDF: Can we just assume that everything is little-endian?
```
TheJJ: same as above

libopenage/util/misc_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+		static const std::vector<uint64_t> expected{
+			0xf1f0efcdab896745, 0x0000f7f6f5f4f3f2
+		};
+
+		// Now I'm trying to convert misaligned bytes
+		// ASDF: Or It's just nonsense?
+		array8_to_array64(test_array8.data() + 2, 14) == expected or TESTFAIL;
```
TheJJ: i think this will work out fine as you use `memcpy` and not a direct cast. The alignment problems occur as far as I understood it when you have instructions that move multibyte values but the beginning of the multibyte is not aligned. The memcpy moves single bytes though.

libopenage/util/misc.h
martinber: In this case, using 8 is okay. Because i'm using it for conversions between uint8_t and uint64_t. I think that the variable name I used is misleading, maybe just replacing every ocurrence of uint64_s is better.

libopenage/util/misc_test.cpp
martinber: I think you are right. I've been reading a little about that and now I understand better. Anyway now I'm not using memcpy (and reinterpreted types) anymore because of endianness.

TheJJ: hm not sure if that is an improvement now ;)
I think the memcpy is fine there, as it is independent of the byte order. It just copies the bytes and doesn't care how they are interpreted then. If we copy less bytes than what the target expects we run into problems of course.
But we should work on this top-down: We want to hash data chunks that are less than `uint64_t` in size. Padding this with 0 is just natural and the right way I assume, but it was certainly solved elegantly by somebody already, endian-independently.

martinber: The problem I found with memcpy is that I need to reinterpret types anyway, for example, in function `array8_to_uint64`:

```
memcpy(reinterpret_cast<uint8_t *>(&result), start, count);
```

So now I'm just copying bytes into the uint64_t number memory, so far everything it's the same in every machine, but if I understand correctly, when I read that number as a whole the interpreted result depends on the endianness of the machine.

I've been looking into stackoverflow ([1](http://stackoverflow.com/questions/9695720/how-do-i-convert-a-64bit-integer-to-a-char-array-and-back), [2](http://stackoverflow.com/questions/2350099/how-to-convert-an-int-to-a-little-endian-byte-array)) and I think the best way to do this independently of endianness is by shifting as I did today.

I'm doing all this stuff because Siphash works on 64bit numbers, so at least I need `array8_to_uint64`, I thought that the remaining three functions are handy :stuck_out_tongue: 

Thank you for spending your time helping me, I really appreciate it

martinber:
I think I finished it. Now everything should work well independently of the endianness of the machine.

I don't know if you like those functions on `misc.h`, they convert between `uint64_t` and `uint8_t` arrays. Maybe they are a little redundant? Now they can interpret the byte array as big or little-endian.

I'm currently using `uint64_to_array8()` and `array8_to_uint64()`, the other two are unused. I need little-endian conversions on the siphash algorithm because the specification requires it, for everything else I use big-endian because I think that's more intuitive.

TheJJ:
So the hash is now the same if you feed in data from a little endian machine and compare it to the hash generated on a machine that fed in the same data on a big endian machine? Also, did you run valgrind on your tests to make sure no access errors occur?

martinber:
I don't know of any big-endian machine, x86 is little-endian, ARM can work as big or little-endian but looks like just everyone uses little-endian. I should try with something [like a VM.](http://omega.cs.iit.edu/~bharatkris/blog/?p=211)

I've never used Valgrind, looks cool. I've run my tests with:

```
valgrind --leak-check=yes ./run test util::tests::siphash
valgrind --leak-check=yes ./run test util::tests::array_conversion
```

The output shows _0 definitely and indirectly lost bytes_ and something like _~13000 probably lost bytes_ in both tests. I tried the `util::tests::vector` (from upstream/master) too and looks the same. Anyways I'm not dynamically allocating memory anywhere.

martinber:
Today I tested my code on a PPC emulated in QEMU, because it is big-endian. Steps I did:

1. `sudo apt install qemu`
1. Download Debian Wheezy Standard from [here](https://people.debian.org/~aurel32/qemu/powerpc/)
1. `qemu-system-ppc -hda debian_wheezy_powerpc_standard.qcow2 -redir tcp:2222::22`

On QEMU:

4. Account: `root`, Password: `root`
1. `apt-get update`
1. Install ssh server: `apt-get install dropbear`. Starts the server automatically.

On your (host) terminal:

7. Connect to QEMU on SSH: `ssh -p 2222 root@localhost`
1. If works OK, `exit` so I can copy the test file with scp
1. `scp -P 2222 test.cpp root@localhost:~`
1. `ssh -p 2222 root@localhost`

On QEMU via SSH:

11. `apt-get install g++`
1. `g++ -std=c++0x test.cpp -o test`
1. `./test`

According to a response from [this stackoverflow question](http://stackoverflow.com/questions/1001307/detecting-endianness-programmatically-in-a-c-program#1001373), that test ran on a big-endian machine
```
if ( htonl(47) == 47 ) {
  // Big endian
} else {
  // Little endian.
}
```

You can see the contents of `test.cpp` from [this gist](https://gist.github.com/martinber/4399c631e10f241e604b85f7a5c693d6)

Now I can't wait until openage support for MIPS... :stuck_out_tongue:

TheJJ:
hehe, awesome.
so we're ready to merge?

martinber:
There I rebased into a single commit. I hope I did not mess up something with git.

PR #756, opened 2017-03-01
GUI: Also try a separate-shared-context mode

ChipmunkV:
Try to init GUI as usual, fallback to the shared context mode if it fails.

### DIFF
--- a/libopenage/gui/guisys/private/gui_ctx_setup.h
+++ b/libopenage/gui/guisys/private/gui_ctx_setup.h
@@ -1,0 +1,94 @@
+// Copyright 2017-2017 the openage authors. See copying.md for legal info.
+
+#pragma once
+
+#include <stdexcept>
+#include <memory>
+#include <functional>
+
+#include <QOpenGLContext>
+#include <QOffscreenSurface>
+
+struct SDL_Window;
+
+QT_FORWARD_DECLARE_CLASS(QOpenGLDebugLogger)
+
+namespace qtsdl {
+
+class CtxExtractionException : public std::runtime_error {
+public:
+	explicit CtxExtractionException(const std::string &what_arg);
+};
+
+/**
+ * Abstract base for the method of getting a Qt-usable context.
+ */
+class CtxExtractionMode {
+public:
+	virtual ~CtxExtractionMode() {
+	}
+
+	/**
+	 * @return context that can be used by Qt
+	 */
+	QOpenGLContext* get_ctx();
+
+	/**
+	 * Function that must be called before rendering the GUI.
+	 */
+	virtual void pre_render() = 0;
+
+	/**
+	 * Function that must be called after rendering the GUI.
+	 */
+	virtual void post_render() = 0;
+
+protected:
+	QOpenGLContext ctx;
+};
+
+/**
+ * Use the same context to render the GUI.
+ */
+class GuiUniqueRenderingContext : public CtxExtractionMode {
+public:
+	explicit GuiUniqueRenderingContext(SDL_Window *window);
+
+	virtual void pre_render() override;
+	virtual void post_render() override;
+};
+
+/**
+ * Create a separate context to render the GUI, make it shared with the main context.
+ */
+class GuiSeparateRenderingContext : public CtxExtractionMode {
+public:
+	explicit GuiSeparateRenderingContext(SDL_Window *window);
+	virtual ~GuiSeparateRenderingContext();
+
+	virtual void pre_render() override;
+	virtual void post_render() override;
+
+private:
+	/**
+	 * GL context of the game
+	 */
+	QOpenGLContext main_ctx;
+
+	/**
+	 * GL debug logger of the GL context of the GUI
+	 */
+	std::unique_ptr<QOpenGLDebugLogger> ctx_logger;
+
+	/**
+	 * Function to make the game context current
+	 */
+	std::function<void()> make_current_back;
+
+	/**
+	 * Surface that is needed to make the GUI context current
+	 */
+	QOffscreenSurface offscreen_surface;
+};
+
+} // namespace qtsdl
--- a/libopenage/gui/guisys/private/gui_rendering_setup_routines.cpp
+++ b/libopenage/gui/guisys/private/gui_rendering_setup_routines.cpp
@@ -1,0 +1,68 @@
+// Copyright 2017-2017 the openage authors. See copying.md for legal info.
+
+#include "gui_rendering_setup_routines.h"
+
+#include <cassert>
+
+#include <QDebug>
+
+#include "gui_ctx_setup.h"
+
+namespace qtsdl {
+
+GuiRenderingSetupRoutines::GuiRenderingSetupRoutines(SDL_Window *window) {
+	try {
+		this->ctx_extraction_mode = std::make_unique<GuiUniqueRenderingContext>(window);
+	} catch (const CtxExtractionException&) {
+
+		qInfo() << "Falling back to separate render context for GUI";
+
+		try {
+			this->ctx_extraction_mode = std::make_unique<GuiSeparateRenderingContext>(window);
+		} catch (const CtxExtractionException&) {
+			assert(false && "setting up context for GUI failed");
+		}
+	}
+}
+
+GuiRenderingSetupRoutines::~GuiRenderingSetupRoutines() {
+}
+
+QOpenGLContext* GuiRenderingSetupRoutines::get_ctx() {
+	return this->ctx_extraction_mode->get_ctx();
+}
+
+void GuiRenderingSetupRoutines::pre_render() {
+	this->ctx_extraction_mode->pre_render();
+}
+
+void GuiRenderingSetupRoutines::post_render() {
+	this->ctx_extraction_mode->post_render();
+}
+
+GuiRenderingCtxActivator::GuiRenderingCtxActivator(GuiRenderingSetupRoutines &rendering_setup_routines)
+	:
+	rendering_setup_routines{&rendering_setup_routines} {
+
+	this->rendering_setup_routines->pre_render();
+}
+
+GuiRenderingCtxActivator::~GuiRenderingCtxActivator() {
+	if (this->rendering_setup_routines)
+		this->rendering_setup_routines->post_render();
+}
+
+GuiRenderingCtxActivator::GuiRenderingCtxActivator(GuiRenderingCtxActivator&& o)
+	:
+	rendering_setup_routines{o.rendering_setup_routines} {
+
+	o.rendering_setup_routines = nullptr;
+}
+
+GuiRenderingCtxActivator& GuiRenderingCtxActivator::operator=(GuiRenderingCtxActivator&& o) {
+	this->rendering_setup_routines = o.rendering_setup_routines;
+	o.rendering_setup_routines = nullptr;
+	return *this;
+}
+
+} // namespace qtsdl
--- a/libopenage/gui/guisys/private/gui_rendering_setup_routines.h
+++ b/libopenage/gui/guisys/private/gui_rendering_setup_routines.h
@@ -1,0 +1,57 @@
+// Copyright 2017-2017 the openage authors. See copying.md for legal info.
+
+#pragma once
+
+#include <memory>
+
+#include <QtGlobal>
+
+struct SDL_Window;
+
+QT_FORWARD_DECLARE_CLASS(QOpenGLContext)
+
+namespace qtsdl {
+
+class CtxExtractionMode;
+
+class GuiRenderingCtxActivator;
+
+/**
+ * Returns a GL context usable by Qt classes.
+ * Provides pre- and post-rendering functions to make the context usable for GUI rendering.
+ */
+class GuiRenderingSetupRoutines {
+public:
+	explicit GuiRenderingSetupRoutines(SDL_Window *window);
+	~GuiRenderingSetupRoutines();
+
+	QOpenGLContext* get_ctx();
+
+private:
+	friend class GuiRenderingCtxActivator;
+	void pre_render();
+	void post_render();
+
+	std::unique_ptr<CtxExtractionMode> ctx_extraction_mode;
+};
+
+/**
+ * Prepares the context for rendering the GUI for one frame.
+ * Activator must be destroyed as soon as the GUI has executed its frame render call.
+ */
+class GuiRenderingCtxActivator {
+public:
+	explicit GuiRenderingCtxActivator(GuiRenderingSetupRoutines &rendering_setup_routines);
+	~GuiRenderingCtxActivator();
+
+	GuiRenderingCtxActivator(GuiRenderingCtxActivator&& o);
+	GuiRenderingCtxActivator& operator=(GuiRenderingCtxActivator&& o);
+
+private:
+	GuiRenderingCtxActivator(const GuiRenderingCtxActivator&) = delete;
+	GuiRenderingCtxActivator& operator=(const GuiRenderingCtxActivator&) = delete;
+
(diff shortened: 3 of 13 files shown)


libopenage/gui/guisys/private/platforms/context_extraction_x11.cpp
```diff
@@ -35,4 +35,23 @@ std::tuple<QVariant, WId> extract_native_context(SDL_Window *window) {
...
+	GLXContext current_context;
+	SDL_SysWMinfo wm_info;
+	SDL_VERSION(&wm_info.version);
+
+	if (SDL_GetWindowWMInfo(window, &wm_info)) {
+		assert(wm_info.info.x11.display);
+
+		current_context = glXGetCurrentContext();
+		assert(current_context);
+	}
```
TheJJ: if this if is skipped, we should abort, right? (same above in the old function i think)

libopenage/gui/guisys/private/gui_event_queue_impl.cpp
```diff
@@ -24,7 +28,12 @@ GuiEventQueueImpl* GuiEventQueueImpl::impl(GuiEventQueue *event_queue) {
 
 void GuiEventQueueImpl::process_callbacks() {
 	assert(QThread::currentThread() == this->thread);
+#ifdef __APPLE__
+	if (QThread::currentThread() != QCoreApplication::instance()->thread())
+		this->callback_processor.processEvents();
+#else
 	this->callback_processor.processEvents();
```
TheJJ: why does apple need a special treatment here? (maybe even add a comment about the reason)

libopenage/gui/guisys/private/gui_ctx_setup.h
```diff
@@ -0,0 +1,75 @@
...
+
+class GuiUniqueRenderingContext : public CtxExtractionMode {
+public:
+	explicit GuiUniqueRenderingContext(SDL_Window *window);
+
+	virtual void pre_render() override;
+	virtual void post_render() override;
+};
+
+class GuiSeparateRenderingContext : public CtxExtractionMode {
```
TheJJ: documentation for those classes pls

libopenage/gui/guisys/private/gui_ctx_setup.h
```diff
@@ -0,0 +1,75 @@
...
+#include <QOpenGLContext>
+#include <QOffscreenSurface>
+
+struct SDL_Window;
+
+QT_FORWARD_DECLARE_CLASS(QOpenGLDebugLogger)
+
+namespace qtsdl {
+
+class CtxExtractionException : public std::runtime_error {
```
TheJJ: would be a good job for the `error::Error` so we had nice backtraces etc.

libopenage/gui/guisys/private/gui_application_impl.cpp
```diff
@@ -26,7 +26,9 @@ GuiApplicationImpl::~GuiApplicationImpl() {
 
 void GuiApplicationImpl::processEvents() {
 	assert(std::this_thread::get_id() == this->owner);
+#ifndef __APPLE__
 	this->app.processEvents();
+#endif
```
TheJJ: comment about the reason pls, and where does apple process its events then?

libopenage/gui/guisys/private/gui_rendering_setup_routines.h
```diff
@@ -0,0 +1,49 @@
...
+
+QT_FORWARD_DECLARE_CLASS(QOpenGLContext)
+
+namespace qtsdl {
+
+class CtxExtractionMode;
+
+class GuiRenderingCtxActivator;
+
+class GuiRenderingSetupRoutines {
```
TheJJ: documentation pls

libopenage/gui/guisys/private/opengl_debug_logger.cpp
```diff
@@ -0,0 +1,131 @@
...
+			auto cb = reinterpret_cast<void (*)(GLenum, GLenum, GLuint, GLenum, GLsizei, const GLchar*, void*)>(params.callback);
+			QObject::connect(&*logger, &QOpenGLDebugLogger::messageLogged, [cb] (const QOpenGLDebugMessage &debugMessage) {
+				GLenum source;
+				GLenum type;
+				GLenum severity;
+
+				switch (debugMessage.source()) {
+				case QOpenGLDebugMessage::APISource:
+					source = GL_DEBUG_SOURCE_API;
+					break;
```
TheJJ: we already have that in `libopenage/error/gl_debug.cpp`, we should not duplicate it. any idea how we could unify them?

libopenage/gui/guisys/private/gui_event_queue_impl.cpp
ChipmunkV: Let's leave it out and see what mac users say. Can't really test this part.

libopenage/gui/guisys/private/platforms/context_extraction_x11.cpp
ChipmunkV: Yes, it's a good unification.

libopenage/gui/guisys/private/gui_ctx_setup.h
ChipmunkV: It's too deep to include the `error`. Replace the exception with an assert?

libopenage/gui/guisys/private/opengl_debug_logger.cpp
ChipmunkV: Can remove the switch.

libopenage/gui/guisys/private/gui_ctx_setup.h
TheJJ: as you catch it yourself it should be ok like that.

libopenage/gui/guisys/private/opengl_debug_logger.cpp
```diff
@@ -0,0 +1,49 @@
...
+#else
+	Q_UNUSED(current_source_context);
+#endif
+
+	return params;
+}
+
+void apply_opengl_debug_parameters(gl_debug_parameters params, QOpenGLContext &current_dest_context)  {
+#ifndef __APPLE__
+	if (params.is_debug && params.callback) {
```
TheJJ: can you elaborate a bit why apple needs this special treatment of debug function config?

ChipmunkV: Should have cleaned this up like in that other place.

AFAIK, apple doesn't support this GL version, so there may be some problems. Maybe Qt can deal with them by itself without the ifdefs.

libopenage/gui/guisys/private/gui_ctx_setup.h
ChipmunkV: Ok, that's done. They won't escape.

TheJJ:
Simple question: what is the separate-shared-context mode? :grin: Is it to have the gui opengl context separately?

ChipmunkV:
Yes, that creates another context just for gui.

libopenage/gui/guisys/private/gui_rendering_setup_routines.cpp
```diff
@@ -0,0 +1,68 @@
...
+	try {
+		this->ctx_extraction_mode = std::make_unique<GuiUniqueRenderingContext>(window);
+	} catch (const CtxExtractionException&) {
+
+		qInfo() << "Falling back to separate render context for GUI";
+
+		try {
+			this->ctx_extraction_mode = std::make_unique<GuiSeparateRenderingContext>(window);
+		} catch (const CtxExtractionException&) {
+			assert(false && "setting up context for GUI failed");
```
TheJJ: probably a debugging leftover

libopenage/gui/guisys/private/platforms/context_extraction_cocoa.mm
```diff
@@ -1,4 +1,4 @@
-// Copyright 2015-2016 the openage authors. See copying.md for legal info.
+// Copyright 2015-2017 the openage authors. See copying.md for legal info.
```
TheJJ: btw why does the cocoa extraction have to be a `.mm` file?

libopenage/gui/guisys/private/platforms/context_extraction_x11.cpp
```diff
@@ -24,15 +24,38 @@ std::tuple<QVariant, WId> extract_native_context(SDL_Window *window) {
 
 		current_context = glXGetCurrentContext();
 		assert(current_context);
+
+		return std::make_tuple(
+			QVariant::fromValue<QGLXNativeContext>(
+				QGLXNativeContext(current_context,
+								  wm_info.info.x11.display,
+								  wm_info.info.x11.window)),
```
TheJJ: indent

libopenage/gui/guisys/private/gui_rendering_setup_routines.cpp
ChipmunkV: The whole assert or the comment in it?

That's what I called "replace the exception with an assert".

It's kind of a spiritual successor to this one:

https://github.com/SFTtech/openage/blob/816fa62150da6e6c17461b538c4f8888f8b2519d/libopenage/gui/guisys/private/platforms/context_extraction_x11.cpp#L26

libopenage/gui/guisys/private/platforms/context_extraction_cocoa.mm
ChipmunkV: That's a valid question. With current approach there is probably no need to do ObjC there.

libopenage/gui/guisys/private/gui_rendering_setup_routines.cpp
TheJJ: hm, i was just confused by the `false &&`, but it [seems alright](https://stackoverflow.com/questions/5867834/assert-with-message) :)

libopenage/gui/guisys/private/platforms/context_extraction_cocoa.mm
TheJJ: Do you wanna fix that? Would be good :)

Birch-san:
I tried to compile this on latest Sierra, but there were some missing `#define`s. To fix this: I provide a pull request for your pull request:

https://github.com/ChipmunkV/openage/pull/2

Birch-san:
**Good news**

launches (with my pull request):

```
➜  openage git:(gui-gl-context) make run
[  1%] Automatic moc for target libopenage
[  1%] compiling .py files to .pyc files
[1/1] Compiling /Users/anon/git/openage/openage/convert/hdlanguagefile.py to /Users/anon/git/openage/openage/convert/__pycache__/hdlanguagefile.cpython-36.pyc
./run game
INFO [py] launching openage v0.3.0-861-g85e62658
INFO [py] compiled by AppleClang 8.0.0.8000042
INFO [py] running in DEVMODE
INFO launching engine with Path(Union(Directory(/Users/anon/git/openage/cfg).root @ (b'cfg',), Directory(/Users/anon/git/openage/assets).root @ (b'assets',)):) and fps limit 0
INFO Using audio device: default [freq=48000, format=32784, channels=2, samples=4096]
INFO Compiled with Qt 5.8.0 and run with Qt 5.8.0
INFO Falling back to separate render context for GUI
INFO loading configuration files...
INFO [py] loading config file [Union(Directory(/Users/anon/git/openage/cfg).root @ (b'cfg',), Directory(/Users/anon/git/openage/assets).root @ (b'assets',))]:cfg/keybinds.oac...
INFO Loading time [engine]: 0.531 s
INFO Loading time   [game]: 0.613 s
INFO [T1] Loading game specification files...
INFO [T1] Loaded multi-csv file: 99984 sub-files
INFO [T1] Loading textures...
INFO [T1] Loading sounds...
INFO [T1] Loading time  [data]: 19.449 s
```

And it doesn't do the infinite stack trace thing. CPU usage is 0%, so that's good.

**Bad news**

Can't kill it with Ctrl-C.

Nothing renders, ever. Nor are any sounds made.

image

Birch-san:
is there a way to run openage with placeholder assets? since maybe the black screen just indicates that the assets weren't extracted correctly. admittedly my `assets/converted/graphics` looks legit, but still it would be nice to rule it out as a problem.

TheJJ:
Unfortunately not, if the log doesn't say anything about load-failures they should be loaded correctly. I guess that the separate gui context is failed to be displayed.

ChipmunkV:
@Birch-san, we also discussed in this PR some changes to the input handling (in outdated comments): libopenage/gui/guisys/private/gui_application_impl.cpp and libopenage/gui/guisys/private/gui_event_queue_impl.cpp.

Dropped that commit because it was unexplainable. I've put it here:

https://github.com/ChipmunkV/openage/commit/d577f5681bf61e09cd7d8175661be2df389b9b4d

Birch-san:
that commit d577f5681bf61e09cd7d8175661be2df389b9b4d certainly helps!

image

image

Birch-san:
Encountered these errors whilst using the game:

https://gist.github.com/Birch-san/cc9220818094b61cc9b817925da1e036

Birch-san:
Here's a gameplay video.

https://youtu.be/9GmwKreWg6I

Colours are off in a strange way. I wonder if it's an endianness problem? I saw color shifts like this once when compiling a Game Boy Advance emulator for Mac, and it was totally endianness.

I wasn't able to interact with anything in the game, but the HUD buttons seemed to be clickable (though they spent a lot of their time disappearing).

Birch-san:
image

Birch-san:
The exact code I was running is here:
https://github.com/Birch-san/openage/commit/e83a4e37ea8f177edbb971fcafb3b3e203cbcd17

Which (at the time of writing) is the HEAD of this branch:
https://github.com/Birch-san/openage/tree/gui-gl-context-fix-3

ChipmunkV:
A theory about the disappearing buttons: it's probably related to that input handling fix - some events can't pass through if there is not enough real input activity (so the GUI isn't updated).

The debug overlay (the one with the graph) has correct colors. The GUI too.
 * do the asset png files look correctly?
 * is the ingame HUD ok? (the one with the population and resource coun)

Birch-san:
> do the asset png files look correctly?

alpha works correctly:

image

z-sorting works correctly:

image

animation works correctly:

openage-animation

> is the ingame HUD ok? (the one with the population and resource coun)

I've never seen it. it's not a z-sorting problem (I tried scrolling the entire map off-screen, and even then I cannot see any HUD).

again, here's a full screen recording:

https://youtu.be/9GmwKreWg6I

Birch-san:
oh, I was able to see the HUD by clicking "change mode" a few times. it's a bit damaged.

image

Birch-san:
when I changed into a different mode, I was able to select units:

image

Birch-san:
action mode:

image

Birch-san:
> some events can't pass through if there is not enough real input activity (so the GUI isn't updated).

certainly I found that sometimes I had to click buttons several times before they would work.

Birch-san:
I fixed the color problem. I was able to confirm in Photoshop that it was just a case of "red & blue channels' being swapped". I was able to fix the pixel output format.

See my pull request:

https://github.com/ChipmunkV/openage/pull/3

TheJJ:
@ChipmunkV Do you still want to get rid of the objective C file?

ChipmunkV:
@TheJJ, yes, but I only see how to remove a part of it.

ChipmunkV:
Added [missing GL_DEBUG definitions to fix compile on Mac](https://github.com/Birch-san/openage/commit/dd5f40f8e5cc3b8e793c3551bce0abe6380178d1) and minimized the ObjC file to just getting the current context and putting it into Qt.

PR #758, opened 2017-03-02
Score & Unit lifecycle

MaanooAk:
- [x] Score system
  - [x] get
  - [x] Sub-scores
  - [x] Team score update
- [x] Unit costs
- [x] Damage refactors and kill credit
- [x] Keep track of units counts (for the buildings dependecies and others)

(This will take a lot)

### DIFF
--- a/libopenage/gamestate/game_main.cpp
+++ b/libopenage/gamestate/game_main.cpp
@@ -22,13 +22,13 @@ GameMain::GameMain(const Generator &generator)
 	// players
 	unsigned int i = 0;
 	for (auto &name : generator.player_names()) {
-		this->players.emplace_back(this->add_civ(i), i, name);
+		this->players.push_back(new Player(this->add_civ(i), i, name));
 		i++;
 	}
 
 	// initialise types only after all players are added
 	for (auto &p : this->players) {
-		p.initialise_unit_types();
+		p->initialise_unit_types();
 	}
 
 	// initialise units
@@ -36,14 +36,18 @@ GameMain::GameMain(const Generator &generator)
 	generator.add_units(*this);
 }
 
-GameMain::~GameMain() {}
+GameMain::~GameMain() {
+	for (auto &p : this->players) {
+		delete p;
+	}
+}
 
 unsigned int GameMain::player_count() const {
 	return this->players.size();
 }
 
 Player *GameMain::get_player(unsigned int player_id) {
-	return &this->players.at(player_id);
+	return this->players.at(player_id);
 }
 
 unsigned int GameMain::team_count() const {
--- a/libopenage/gamestate/game_main.cpp
+++ b/libopenage/gamestate/game_main.cpp
@@ -20,9 +20,10 @@ GameMain::GameMain(const Generator &generator)
 	spec{generator.get_spec()} {
 
 	// players
+	this->players.reserve(generator.player_names().size());
 	unsigned int i = 0;
 	for (auto &name : generator.player_names()) {
-		this->players.push_back(new Player(this->add_civ(i), i, name));
+		this->players.push_back(std::make_shared<Player>(this->add_civ(i), i, name));
 		i++;
 	}
 
@@ -37,9 +38,6 @@ GameMain::GameMain(const Generator &generator)
 }
 
 GameMain::~GameMain() {
-	for (auto &p : this->players) {
-		delete p;
-	}
 }
 
 unsigned int GameMain::player_count() const {
@@ -47,7 +45,7 @@ unsigned int GameMain::player_count() const {
 }
 
 Player *GameMain::get_player(unsigned int player_id) {
-	return this->players.at(player_id);
+	return this->players.at(player_id).get();
 }
 
 unsigned int GameMain::team_count() const {
--- a/libopenage/gamestate/player.cpp
+++ b/libopenage/gamestate/player.cpp
@@ -128,8 +128,9 @@ void Player::active_unit_added(Unit *unit) {
 		return;
 	}
 
+	this->units_have[unit->unit_type->id()] += 1;
+	this->units_had[unit->unit_type->id()] += 1;
 	// TODO handle here building dependencies
-	// TODO handle here on create unit triggers
 
 	// population
 	if (unit->has_attribute(attr_type::population)) {
@@ -149,6 +150,9 @@ void Player::active_unit_added(Unit *unit) {
 	} else if (unit->has_attribute(attr_type::building) || unit->has_attribute(attr_type::population)) { // building, living
 		this->score.add_score(score_category::economy, unit->unit_type->cost.sum() * 0.2);
 	}
+
+	// TODO handle here on create unit triggers
+	// TODO check for unit based win conditions
 }
 
 void Player::active_unit_removed(Unit *unit) {
@@ -157,8 +161,8 @@ void Player::active_unit_removed(Unit *unit) {
 		return;
 	}
 
+	this->units_have[unit->unit_type->id()] -= 1;
 	// TODO handle here building dependencies
-	// TODO handle here on death unit triggers
 
 	// population
 	if (unit->has_attribute(attr_type::population)) {
@@ -178,6 +182,9 @@ void Player::active_unit_removed(Unit *unit) {
 	} else if (unit->has_attribute(attr_type::building) || unit->has_attribute(attr_type::population)) { // building, living
 		this->score.remove_score(score_category::economy, unit->unit_type->cost.sum() * 0.2);
 	}
+
+	// TODO handle here on death unit triggers
+	// TODO check for unit based win conditions
 }
 
 void Player::killed_unit(const Unit & unit) {
@@ -190,6 +197,20 @@ void Player::on_resources_change() {
 	this->score.update_resources(this->resources);
 
 	// TODO check for resource based win conditions
+}
+
+int Player::get_units_have(int type_id) const {
+	if (this->units_have.count(type_id)) {
+		return this->units_have.at(type_id);
+	}
+	return 0;
+}
+
+int Player::get_units_had(int type_id) const {
+	if (this->units_had.count(type_id)) {
+		return this->units_had.at(type_id);
+	}
+	return 0;
 }
 
 } // openage
--- a/libopenage/gamestate/player.cpp
+++ b/libopenage/gamestate/player.cpp
@@ -25,6 +25,7 @@ Player::Player(Civilisation *civ, unsigned int number, std::string name)
 	this->resources[game_resource::wood] = 1000;
 	this->resources[game_resource::stone] = 1000;
 	this->resources[game_resource::gold] = 1000;
+	this->on_resources_change();
 }
 
 bool Player::operator ==(const Player &other) const {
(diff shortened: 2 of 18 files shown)


libopenage/gamestate/team.cpp
```diff
@@ -61,4 +64,12 @@ member_type Team::get_member_type(Player &player) {
 	return member_type::none;
 }
 
+std::vector<const Player*> Team::get_players() const {
```
MaanooAk: Any ideas on this

libopenage/gamestate/score.cpp
```diff
@@ -0,0 +1,111 @@
...
+	Score(),
+	player{player} {
+}
+
+void ScorePlayer::update_score() {
+	Score::update_score();
+	// update team score
+	// TODO fix player->team should be nullptr
+	if (player->team) {
+		// TODO rework team update
```
MaanooAk: The problem here is that the players  are passed to the Team as consts. One solution could be to have the players in Team not const

libopenage/gamestate/team.cpp
TheJJ: you could just give back a reference to the members list?

MaanooAk:
Changed how the players are stored in order to have a static memory location, previously the vector would change the location of the objects making impossible to use a pointer to a Player.

libopenage/gamestate/team.cpp
MaanooAk: `warning: reference to local variable`

It will be optimized into a pointer pass anyway, I was talking more about the vector construction adding one by one the players. But its ok, it's a small list...

libopenage/gamestate/score.h
```diff
@@ -0,0 +1,139 @@
...
+
+	/** Used by update_resources. */
+	int score_resources;
+};
+
+
+/**
+ * The score of a player
+ */
+class ScorePlayer : public Score {
```
TheJJ: `PlayerScore`?

libopenage/gamestate/score.h
```diff
@@ -0,0 +1,139 @@
...
+private:
+
+	Player *player;
+};
+
+
+/**
+ * The score of a team
+ */
+class ScoreTeam : public Score {
```
TheJJ: `TeamScore`?

libopenage/gamestate/team.cpp
TheJJ: Ah, sorry. I think there's no better way to return dict keys separately, if the dict would support returning its keys, the internal implementation would be exactly this.

http://www.lonecpluspluscoder.com/2015/08/13/an-elegant-way-to-extract-keys-from-a-c-map/

If you like, you can create the mentioned template function to our `libopenage/util/` library.

libopenage/unit/action.cpp
```diff
@@ -96,24 +96,18 @@ void UnitAction::face_towards(const coord::phys3 pos) {
...
 
-void UnitAction::damage_object(Unit &target) {
 	if (target.has_attribute(attr_type::damaged)) {
 		auto &dm = target.get_attribute<attr_type::damaged>();
 
-		if (target.has_attribute(attr_type::armor) && this->entity->has_attribute(attr_type::attack)) {
+		// this is the damage calculation system
+
+		if (dm.hp == 0) {
+			// already killed, do nothing
```
TheJJ: we should test this with a `Unit::is_dead()` function someday, we'll need it very often.

libopenage/gamestate/game_main.cpp
```diff
@@ -22,28 +22,32 @@ GameMain::GameMain(const Generator &generator)
 	// players
 	unsigned int i = 0;
 	for (auto &name : generator.player_names()) {
-		this->players.emplace_back(this->add_civ(i), i, name);
+		this->players.push_back(new Player(this->add_civ(i), i, name));
```
MaanooAk: - Player is created at the start and destroyed at the end of the game, no need to increase complexity with unique or shared_ptr.
- I think the vector is already using the heap

> no polymorphism involved here

Can you explain this?

libopenage/gamestate/player.cpp
```diff
@@ -120,19 +128,31 @@ void Player::active_unit_added(Unit *unit) {
 		return;
 	}
 
+	this->units_have[unit->unit_type->id()] += 1;
+	this->units_had[unit->unit_type->id()] += 1;
```
MaanooAk: The default initializer of the int gets invoked, so it's guaranteed to be 0. Its not like the:
```
int i;
```
Its like
```
int i = int();
```

libopenage/gamestate/score.cpp
```diff
@@ -0,0 +1,109 @@
...
+	score_exploration{0},
+	score_resources{0} {
+}
+
+void Score::add_score(const score_category cat, double value) {
+	this->add_score(cat, (int) std::lround(value));
+}
+
+void Score::add_score(const score_category cat, int value) {
+	this->score[(int) cat] += value;
```
MaanooAk: I will take look on this...

libopenage/gamestate/score.h
MaanooAk: I like to do that for two reasons:
- its logic based and not language based
- its sorted better (docs, autocomplete):

```
... (other classes)
Score
ScorePlayer
ScoreTeam
...
```
vs
```
...
PlayerScore
...
Score
...
TeamScore
...
```

libopenage/unit/action.cpp
MaanooAk: How the attributes is used in actions (always getting the same attribute from the attributes set) needs to be improved all together 

(my next pr "coming soon")

libopenage/gamestate/score.cpp
MaanooAk: Changed it and also the same thing or resources

libopenage/gamestate/game_main.cpp
LordAro: References > pointers. I don't see any particular reason to need pointers here - surely the number of players doesn't change after initialisation? Shouldn't be any issue with invalidated iterators

TheJJ: The `shared_ptr`/`unique_ptr` actually removes comlexity as you don't need to delete the players in the deconstructor then. And it is exception-safe only then.
The `vector` uses the heap internally but with the `new Player` the vector stores the pointer list on the heap which point to Players allocated somewhere else.

We have two choices:

store the player in a `shared_ptr`
```
    this->players.reserve(generator.player_names().size());
    for (auto &name : generator.player_names()) {
        this->players.push_back(std::make_shared<Player>(this->add_civ(i), i, name));
    }
```
or store the player data directly in the vector
```
    this->players.reserve(generator.player_names().size());
    for (auto &name : generator.player_names()) {
        this->players.push_back(Player{this->add_civ(i), i, name});
    }
```

The `shared_ptr` storage has the advantage that when the vector invalidates its iterators/pointers because of resizing, the raw pointers to players would remain the same.
But nobody needs to hold raw pointers to `Player`s, then the direct storage is better (and more cache-friendly).

libopenage/gamestate/score.h
TheJJ: That makes sense. I'd still say we should focus on human-compatibility though, although the autocomplete argument is a good one, then again, how often do we need to complete those classes? :)

libopenage/gamestate/player.cpp
TheJJ: okay, good.

libopenage/gamestate/team.cpp
Tomatower: Can you give the list to the output parameter as argument like ```get_players(std::vector<const Player*>)``` then you can avoid a potential copy.

MaanooAk:
I'm storing the shared_pointer in the vector but `get_player` still returns a raw pointer. I'm not comfortable enough to change that.

libopenage/gamestate/player.cpp
```diff
@@ -141,19 +161,56 @@ void Player::active_unit_removed(Unit *unit) {
...
 		}
 		if (popul.capacity > 0) {
-			population.remove_capacity(popul.capacity);
+			this->population.remove_capacity(popul.capacity);
 		}
 	}
+
+	// score
+	// TODO improve selectors
+	if (unit->unit_type->id() == 82 || unit->unit_type->id() == 276) { // Castle, Wonder
```
MaanooAk: All this information should be inside the UnitType, which I'm not touching because of the upcoming nyan revolution...

libopenage/gamestate/player.cpp
```diff
@@ -141,19 +161,56 @@ void Player::active_unit_removed(Unit *unit) {
...
+		// nothing
+	} else if (unit->has_attribute(attr_type::building) || unit->has_attribute(attr_type::population)) { // building, living
+		this->score.remove_score(score_category::economy, unit->unit_type->cost.sum() * 0.2);
```
MaanooAk: Same here

libopenage/gamestate/game_main.cpp
MaanooAk: About the polymorphism check #842

libopenage/gamestate/game_main.cpp
```diff
@@ -37,17 +38,14 @@ GameMain::GameMain(const Generator &generator)
...
-	}
 }
 
 unsigned int GameMain::player_count() const {
 	return this->players.size();
 }
 
 Player *GameMain::get_player(unsigned int player_id) {
-	return this->players.at(player_id);
+	return this->players.at(player_id).get();
```
Tomatower: I tried to change this to ```shared_ptr<Player>``` locally - it does not seem to break anything.

Whay do you mean with not comfortable enough?

MaanooAk: - How it does not break anything...? All `Player*` must become `std::shared_ptr<Player>`, returning a shared_ptr and calling get() right after has no point.
- "not comfortable enough" = 15% I don't know if I should + 15% I don't know if I can + 70% I don't want to
But if you incιst I can try

Tomatower: What i have tried was to change the return type to ```std::shared_ptr<Player>``` - it did not fail. 
That means, it seems not to be called, or using auto all over the place. 

So my question is, if it were to break any existing code to change it

MaanooAk: I think you are not in my branch, you must be on the master. 
The get_player was not used and the vector was public (wrong design, you shouldn't be able to change the player list), so I made private and the used the get_player.
Now its used here:
````
libopenage/gamestate/generator.cpp
libopenage/gamestate/game_main.cpp
libopenage/gamestate/game_save.cpp
libopenage/gamestate/game_main.h
libopenage/game_control.cpp
````
And then all of these pass the Player* here:
```
libopenage/gamestate/score.cpp
libopenage/gamestate/score.h
libopenage/gamestate/team.cpp
libopenage/gamestate/team.h
libopenage/game_control.cpp
libopenage/game_control.h
```

Comment (Tomatower) [inline libopenage/gamestate/game_main.cpp]: ah yes - I was on master! 
on second thought - if the player is never stored anywhere (like in a class variable) there should be no trouble using ```shared_ptr::get```. The only place where this might be useful is to store the owning player of each unit - and avoiding circular references is not something you want to waste your brainpower on. 

I agree with your solution!

File: libopenage/gamestate/score.h [hunk 12]
```diff
@@ -0,0 +1,139 @@
...
+	 * TODO update gui here
+	 */
+	virtual void update_score();
+
+	// Getters
+
+	int getScore(const score_category cat) const { return score[static_cast<int>(cat)]; }
+	int getScore(const int index) const { return score[index]; }
+
+	int getScoreTotal() const { return score_total; }
```
TheJJ: sorry for nitpicking, but those three methods shouldn't be camelcase

MaanooAk:
Amended the case error...

TheJJ:
Again a bit closer to victory!

Issue #760, opened 2017-03-08
'Out of range' exception thrown when attacking with ranged units


heinezen:
Attacking an enemy unit with some of the ranged units will throw an exception. The exception seems to occur at the same time as the projectile hits the target. 

Affected units: Onagers, Scorpions, Hand Cannoneer, Bombard Cannon
Unaffected units: (All?) Archers, Cavalry Archers, Skrmishers

```
FATAL: terminate has been called

uncaught exception

std::exception of type std::out_of_range: map::at

current stack

Traceback (most recent call last):
  File ?, in __libc_start_main+0xf1 [0x7f29efdaa3f1]
  File ?, in main+0x3b9 [0x558bf7df21f9]
  File ?, in ? [0x558bf7df3b22]
  File ?, in PyInit_run+0x13f3 [0x558bf7df3a83]
  File ?, in ? [0x558bf7df265a]
  File ?, in ? [0x558bf7df2490]
  File ?, in PyEval_EvalFrameEx+0x81cb [0x7f29f067b64b]
  File ?, in PyEval_EvalFrameEx+0x8257 [0x7f29f067b6d7]
  File ?, in PyCFunction_Call+0xe9 [0x7f29f0573339]
  File ?, in ? [0x7f29dea6029f]
  File ?, in openage::pyinterface::translate_exc_cpp_to_py()+0x1e [0x7f29eca13c27]
  File ?, in __cxa_rethrow+0x49 [0x7f29f01e05e9]
  File ?, in ? [0x7f29f01e0381]
  File ?, in ? [0x7f29f01e0336]
  File ?, in openage::error::terminate_handler()+0x16e [0x7f29ec9bda7f]
  File ?, in openage::error::StackAnalyzer::analyze()+0xc6 [0x7f29ec9be136]

standard terminate handler

terminate called after throwing an instance of 'std::out_of_range'
  what():  map::at
```

TheJJ:
I really hate it that c++ standard exceptions don't collect any backtrace...

ChipmunkV:
I just never use `at()`.

MaanooAk:
I use `at()` all the time... I may caused that bug :)

MaanooAk:
Found it

TheJJ:
wheee

PR #761, opened 2017-03-11
Bug fix: Ownerless projectiles

MaanooAk:
Fix to #760. Some projectile do not have owners, added checks for that.

In the future (when the friendly fire and accuracy is (re)implemented) this should be revisited.

EDIT: the `active_unit_removed` and `active_unit_added` are reworked in #758.
EDIT2: also now the kill cannot be credited to a player


TheJJ:
Thanks. Can you amend the commit just to trigger a new build please? Kevin's back.

TheJJ:
thx, nice find :)

PR #762, opened 2017-03-11
Bug fix: Missing tau

MaanooAk:
Added tau.

I think @mic-e approves ([math.py#L12](https://github.com/SFTtech/openage/blob/master/openage/util/math.py#L12))


mic-e:
actually, on second thought, there are a few more expressions in the codebase where `math::PI` should be replaced by `math::TAU` or a new `math::degreesperrad` constant, especially those that involve the number `180`.

MaanooAk:
Anything else?

By the way, is kevin dead?

TheJJ:
Yea kevin was dead, lives again after the connection was restored :P

PR #767, opened 2017-03-16
Surrounded Engine::run with try catch

### LINKED ISSUE #766: Engine::run should be guarded by a try_catch block
This way we can try to stop all workers before we continue to throw the exceptions.

This way workers are prevented on running with deinitialized data and segfaulting, which hides the original excpetion.

Example:

In `Engine::loop` an exception is thrown because of an OpenGL `OUT_OF_MEMORY` error. Because the excpetion is not catched the Engines destructor and therefor in the end the destructor for an asset manager is called. However a job performing loading operations like `GameSpec::load_terrain` still has a referenced to the now invalid AssetManager and works with him in a normal fashion. This results in a segfault at some point e.g. when merging directory strings.

One way to prevent this would be using `shared_ptr` so that the necessary resources are present for the jobs as long as they need them. However because the excpetion will kill the game we dont want to still loads things afterwards. So just stoping all workers if an exception is thrown and then continuing with the "exception-flow" should be sufficient.

citron0xa9:
- This allows to stop all workers
 - after that is done the exception is rethrown

fixes #766

### DIFF
--- a/copying.md
+++ b/copying.md
@@ -95,6 +95,7 @@ _the openage authors_ are:
 | Marco Savelli               | Piruzzolo                   | svlmrc@gmail.com                      |
 | Yvan Burrie                 | yvan-burrie                 | yvan.burrie@hotmail.com               |
 | Martin Bernardi             | martinber                   | martinbernardi@openmailbox.org        |
+| Erik Griese                 | citron0xa9                  | erik.griese@yahoo.de                  |
 
 If you're a first-time committer, add yourself to the above list. This is not
 just for legal reasons, but also to keep an overview of all those nicknames.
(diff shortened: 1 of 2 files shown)


TheJJ:
The ci check didn't succeed because your commits were done with a different email address than you entered in the `copying.md`. I suggest you rebase and change your git config so the commits get the correct mail.

citron0xa9:
thank you

TheJJ:
It's still conflicting. Did you fetch `upstream/master` and rebased to it?

citron0xa9:
no I forgot that^^

TheJJ:
Now this email of the commit is not affiliated with your github account, you can simply add the mail addres to your github mail list and the commit will be credited to you.

Issue #772, opened 2017-03-26
Make not working


bradleyJohnson55:
I followed the steps for the build procedure, but when I get to "make" in the terminal, it stops at 88% and I get these errors.

![openage_error](https://cloud.githubusercontent.com/assets/26680994/24327637/fbe54752-118a-11e7-9703-981c4e32738d.png)

Using Ubuntu 16.04 LTS

How do I proceed?
Thanks

castilma:
I'm on Ubuntu 16.04. Building from commit 93f9022c works. 
My libopenage/util/path.pxd looks like yours. Maybe try updating cython:
`pip3 install --upgrade cython`

TheJJ:
See #773, you need `cython >=0.25`, sorry :)

TheJJ:
Was the cython update successful? Did it build afterwards?

bradleyJohnson55:
Yes, it was successful. Thank you!

PR #773, opened 2017-03-26
doc: Update required Cython version in Build docs.

goto-bus-stop:
Master now uses Cython `ctypedef`s nested inside `cppclass`es which is only supported from version 0.25 and up.

Cython 0.25 is not currently in the standard package repositories on Ubuntu 16.04, but is available through pip, so I changed the Ubuntu build doc to account for that.

Ref #772


goto-bus-stop:
No, 16.10 is also still on 0.24. I think 17.04 will have 0.25

ChipmunkV:
It's in 17.04.

PR #775, opened 2017-04-02
Fixing our DoublyLinkedList

castilma:
Our (currently unused) DoublyLinkedList had some real issues.
While writing tests, I discovered many things, but the real interesting things were those, that should have raised a compilation error, but didn't. I guess they didn't, because we defined everything in a header and the functions containing uncompilable code were never called(referenced).

I'm not exactly sure, what we can learn from this.
I would suggest using header files for declaration and a separate file for definition. When trying to create `doubly_linked_list.o` from `doubly_linked_list.cpp`, that should have raised an error.
Besides that: If we keep everything in one header, doesn't that lead to every compilation unit recreating the machine code for that class?


libopenage/datastructure/doubly_linked_list.h
```diff
@@ -85,6 +85,10 @@ class DoublyLinkedList {
 
 		node->next->previous = new_node;
 		node->next           = new_node;
+
+		if (this->last == node) {
+				this->last = new_node;
+		}
```
castilma: like insert_after(), I'm not sure, if we want to make insert_*() check for this, or let the caller take responsibility for that case.

libopenage/datastructure/doubly_linked_list.h
```diff
@@ -121,14 +129,14 @@ class DoublyLinkedList {
...
+		if (this->size() > 0) {
 			do {
 				if (current->data == item) {
-					this->erase_node(current);
+					this->erase(current);
 					return;
 				}
 				current = current->next;
-			} while (current != this->end);
+			} while (current != this->last);
```
castilma: `last` and `first` maybe should be callled `back` and `front`.

libopenage/datastructure/doubly_linked_list.h
```diff
@@ -240,14 +248,17 @@ class DoublyLinkedList {
 	/**
 	 * Delete all list nodes.
 	 * O(n)
+	 * asdf: don't call erase: this safes relinking all nodes.
 	 */
 	void clear() {
 		node_t *delete_now = this->first;
 		while (this->node_count > 0) {
-			node_t *deleted_next = this->first->next;
```
castilma: this is just wrong.

libopenage/datastructure/doubly_linked_list.h
```diff
@@ -114,28 +118,42 @@ class DoublyLinkedList {
...
 	/**
 	 * Fully delete a node and fix the neighbor pointers.
+	 * Update #first or #last, if necessary.
 	 * O(1)
+	 * @pre \p node belongs to this list.
 	 */
 	void erase(node_t *node) {
-		node->previous->next = node->next;
-		node->next->previous = node->previous;
+		assert(this->node_count != 0);
```
castilma: remove this assert?

libopenage/datastructure/doubly_linked_list.h
```diff
@@ -114,28 +118,42 @@ class DoublyLinkedList {
 
 	/**
 	 * Fully delete a node and fix the neighbor pointers.
+	 * Update #first or #last, if necessary.
 	 * O(1)
+	 * @pre \p node belongs to this list.
 	 */
 	void erase(node_t *node) {
```
castilma: Do we want the erase() functions to correct the first/last pointer if necessary? 
there can be a protected unlink() function, which does the same without setting first and last.

TheJJ:
I added that list back when I was writing the heap, mainly as i was toying with fibonacci heaps. If it is unused now, we should just remove it?

castilma:
I'm a fan of removing code. I'd say yes.

TheJJ:
K, go for it :)

TheJJ:
Fixed by removal :stuck_out_tongue:

zuntrax:
Fix by removal is best fix indeed.

PR #776, opened 2017-04-03
check for asdf in codecompliance

castilma:
basically PR #731, but Btjones711 did not continue.


castilma:
The codecompliance test says the python module re has no member I. it sure does have it on my system and in the [documentation](https://docs.python.org/3/library/re.html#module-contents)...

TheJJ:
Pylint is not always right. Append a `# pylint: disable=no-member` to the "wrong" line.

TheJJ:
I think you can remove the remaining asdfs and replace them by a simple todo. @mic-e knows he has to replace it in #662 :)

PR #787, opened 2017-04-09
ignore compiled QML

Birch-san:
There's a bunch of compiled qml files in my assets folder. I assert that we need to exclude these from version control.


TheJJ:
Maybe we should add that to the root `.gitignore`, what do you think?

Birch-san:
hm, whenever I'm "not sure if it's better", I decide: "make just the change you _know_ you need".

admittedly `.qmlc` is _always_ compiler output, and we _never_ want compiler output under version control… except when we do. For example: we might want to put compiled output under source control if we have a sample compiled file for unit tests to run against.

I vote to make the most specific change. especially since this will reveal situations where we are producing `.qmlc` in locations where we do not expect it to appear.

PR #428, opened 2015-12-25
Monk conversions instead of killings

bombadie:
I hope this fixes #375 
priests don’t get attack ability any more but instead convert ability. I also added a attribute remove function so that the owner attribute can be replaced. 
This is my first contribution so go easy on me :)

PR #789, opened 2017-04-13
Monk conversions instead of killings

### LINKED ISSUE #375: Monk kills enemy unit instead of converting to own team
and he doesn't keep distance to enemy unit.

LordAro:
Supercedes #428. No functional changes (I haven't even tested it beyond compilation & tests!), and @bombadie retains primary commit authorship

I got very bored.

Fixes #375

### DIFF
--- a/libopenage/unit/unit.h
+++ b/libopenage/unit/unit.h
@@ -175,6 +175,12 @@ public:
 	void add_attribute(std::shared_ptr<AttributeContainer> attr);
 
 	/**
+	 * delete a given attribute this this unit
+	 * this is used to delete/change things like color, hitpoints and speed
+	 */
+	void remove_attribute(const attr_type type);
+
+	/**
 	 * Give new attributes to this unit.
 	 * This is used to add the default attributes
 	 */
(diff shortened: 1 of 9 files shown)


MaanooAk:
There are more things that need to be done.
It you change only the owner upgrades will keep applying and the population tracker and the player base triggers will not work.
I believe the right method would be to:
- remove the unit
- create a copy of the unit type (find a way to not create a lot of copies)
- add a new unit (through the new owener player)

LordAro:
Yeah, figured it wouldn't be so easy :)

I'll take a look and see what I can do (probably not very soon though)

libopenage/unit/unit.h
```diff
@@ -175,6 +175,12 @@ class Unit : public log::LogSource {
 	void add_attribute(std::shared_ptr<AttributeContainer> attr);
 
 	/**
+	 * delete a given attribute this this unit
+	 * this is used to delete/change things like color, hitpoints and speed
```
castilma:
```
Delete a given attribute from this unit.
^                        ^~~~
```
```
This is used to delete/change things like color, hitpoints and speed.
^
```
Thank you.

LordAro: I noticed this at the time, but (most of) the rest of the doccomments in this file have a similar lack of punctuation. something to fix elsewhere perhaps?

TheJJ: If you like, everywhere :P

LordAro: I meant in a different prq ;)

TheJJ: me too :D

LordAro: deferred to @castilma in #832 :p

castilma: hey!

simonsan:
Will most likely be superseded by new gamestate implementation. Will mark it as `zombie` for later usages though. Thank you.

PR #790, opened 2017-04-14
Fix RGB endianness used in pixel transfer

Birch-san:
Fix pixel transfer format for loading an OpenGL texture from an SDL surface. On Mac, this swaps the red & blue channels (so RGB becomes BGR).

Before:

image

After:

image

Basically I just copied [this stackoverflow](http://stackoverflow.com/a/4340330/5257399).


TheJJ:
Btw, if you select the older release in the steamdownload or use the non-hd version, all the X will go away and the UI is complete :)

Issue #735, opened 2017-02-06
Scripting API


TheJJ:
As our ultimate goal is cool moddability, and not everything can be done with data mods (with nyan #734)  we also need scripting support.

This consists of multiple parts that need to be done:

- [ ] Load and run scripts
  - [ ] Load scripts via nyan
  - [ ] Install script hooks via nyan
  - [ ] Let the script manipulate the hooks it installed
- [ ] Game manipulation
  - [ ] Apply patches in nyan
  - [ ] Perform actions with game entities
  - ...
- [ ] higher level AI scripting
  - [ ] Python interface
  - [ ] even higher level lisp-like interface like the original's ai scripting language
  - ...
- [ ] Caution users against unsgned mod packs containing scripts
  - We can't and want to sandbox python, you should be able to use http://scikit-learn.org/ and the like
  - The openage-dev-team should be able to sign mods that have approved code: We can still hack your computer with the regular openage code, so we can also sign mod code.

timohaas:
if the api has access to the whole gamestate

it can determine if a player has lost and will be removed from game
=> so every gametype can is possible by simple providing a python script

TheJJ:
Yes, it should be possible then. What other gametypes do you have in mind?

timohaas:
The Standard ones:
Defend the Wonder, Conquest, King of the Hill, Conquest

New types:
- Tower Defense: all Players against AI (where does the Emeny appear? how many? who do the attack?)
- HelmsKlamm: 2 Players defend a position and get 2xressources per villager as the other 6
- Capture the Flag/Relic: go to the enemy base and capture the relic
- toxic map: units/defense structures/buildings loose their health points over time (until 20% or 30% or only 10 points?)
- civ hopping: every 5min the players civ changes randomly (got to work out details like civ1 research paladin and civ2 has not, but the next civ3 you can have paladin again)
- Wonder Race with fighting
- Kill the Founding Fathers, kill all the villagers from the game start (nothing else matters) (to the players, they are indistinguishable from normal villagers), you have to remember them or place them in towers and give up ressource gathering, good times for raiding with horses

other stuff:
- environment changes, new lakes, forests, shallow crossings, rivers disappear
- technology upgrades only for newly created units.
- technology stealing, monk converts unit => you get all technologies this unit has
                                monk converts blacksmith/stable.. => you get all technologies which has been researched in this type of building (option: other player looses it, must research it again)
[...]

PR #796, opened 2017-04-19
lua script api

timohaas:
labels: experimental,proofofconcept
needs: liblua5.1-0-dev

I wanted to look into lua and did some experiments how to integrate with c++. I choose the [sol2](https://github.com/ThePhD/sol2) framework, because its a simple header file. For compiling you need the lua 5.1 dev files of your distro. The whole thing are event based hooks, so if theres nothing to during the gameloop, there will be nothing done.

A maintainer should add labels experimental, proofofconcept. This pull request is **not** for merge. The script.cpp is ugly and would need some serious rewriting.

How does it work?
There is a singleton to register lua scriptfiles(you can have more than 1). The hooks we decide to implement are written in c++ and call the scripts via member method. The method does all the work with lua code and change the gamestate if necessary.

example hook:
``` cpp
for(auto script : ScriptSingleton::instance().scripts) {

 script->onResourceChange(this->color,resource,amount,this->resources[resource]);

}
```

I think lua is the better choise for ingame scripting. It has better security than python and I dont think python can be integrated so nicely like a language which was designed as embedded language.

It is a counter proposal to @TheJJ #735

### DIFF
--- a/libopenage/CMakeLists.txt
+++ b/libopenage/CMakeLists.txt
@@ -44,6 +44,7 @@ add_subdirectory("pathfinding")
 add_subdirectory("pyinterface")
 add_subdirectory("renderer")
 add_subdirectory("rng")
+add_subdirectory("scripting")
 add_subdirectory("shader")
 add_subdirectory("terrain")
 add_subdirectory("testing")
@@ -146,6 +147,8 @@ if(APPLE)
 	find_path(QTPLATFORM_INCLUDE_DIRS QtPlatformHeaders "/usr/local/opt/qt5/include")
 endif()
 
+find_package(Lua51 REQUIRED) 
+
 # directories for header inclusion
 include_directories(
 	${OPENGL_INCLUDE_DIR}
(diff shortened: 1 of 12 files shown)


TheJJ:
I agree that the one advantage over Python is the better sandboxing of the code. But we should really really be careful of adding yet another language to the project. Also, many people were happy about some game _not_ using lua but Python instead.

The plan was to add big red warnings for mods that add Python code, those go away if we (the project) sign this mod release (i.e. we confirm the code won't delete your homefolder etc).

We also put huge effort into our Python interface (https://github.com/SFTtech/openage/blob/master/doc/code/pyinterface.md), which would not be obsolete with Lua, but it was designed to provide a nice bi-directional interface to C++.

Calls like
```cpp
    for(auto script : ScriptSingleton::instance().scripts) {
        script->update(lastframe_duration, this->gametime);
    }
```
are totally possible with our Python interface. We should do them without some crappy singleton though, instead scripts should register their hooks at some scriptmanager, which is then a member of the `Engine`.

Independent of the language, hooks for scripts and functions they can call is the way we should design the api.

Another sidenote: I'm very opposed to bundling any 3rd-party software into the repo, it should be a regular dependency instead.

ChipmunkV:
AFAIK, there is no openage developer who is fluent in all languages that openage uses.

TheJJ:
I think we can conclude that pulling in another language for a purpose we already have a language for is no use. The API ideas can be implemented directly with our Python interface.

PR #797, opened 2017-04-20
Prefix-matching for tests

Tomatower:
When running tests, like ```util::tests::constinit_vector``` only one test is run. 
This enables to say run test ```util::tests``` to run all tests prefixed with ```util::tests```: 
```
util::tests::constinit_vector
util::tests::enum_
util::tests::init
util::tests::matrix
util::tests::quaternion
util::tests::vector
util::tests::siphash
util::tests::array_conversion
```

This makes testing one module much more convenient.

### DIFF
--- a/openage/testing/main.py
+++ b/openage/testing/main.py
@@ -54,9 +54,8 @@ def process_args(args, error):
         error("you have to run all tests, "
               "otherwise I don't care if you have assets")
 
-    if args.run_all_tests:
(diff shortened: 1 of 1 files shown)


Vtec234:
While you're changing this file, you could also fix line 141, which should state `_, _, _, demofun = test_list[args.demo[0], 'demo']` instead :)

PR #801, opened 2017-04-23
Several spelling fixes and code style corrections.

castilma:


doc/code_style/mom.cpp
```diff
@@ -120,7 +120,7 @@ class HalfLife3 : public ValveGame {
 	}
 
 private:
-	/** links to the story of the game */
+	/** Links to the story of the game. */
 	Story *story;
 
 	/** time in seconds estimated left until release */
```
TheJJ: Can you update that comment as well?

castilma:
That should do it.

PR #802, opened 2017-04-23
Buildsystem: Added Benchmarking

### LINKED ISSUE #167: Create benchmark function registering in buildsystem
The buildsystem already supports `demo` and `test` for interactive and static execution of features.
`buildsystem/testing.cmake` needs to be extended to support a third category: `benchmark`.

This can (later?) be enhanced by an arbitrary amount of awesome features like plotting, uploading to a comparison server, regression testing etcetc.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/6429404-create-benchmark-function-registering-in-buildsystem?utm_campaign=plugin&utm_content=tracker%2F371436&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F371436&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Tomatower:
fixes #167

added ```./run test -b BENCHMARK ``` for benchmarking the code. 

It runs the benchmark until the whole block takes longer than 5 seconds, totalling in 10 seconds for the whole process. Minimum of 7 Iterations in total.

This enables us, to benchmark also more complex systems that take a longer while, and still keep the same logic for high-performance tests.

It will produce Output like: 

```
$ ./run test -b testing.benchmark.benchmark
Iterations Total time    Average time per execution
         1  0.10013296s  0.10013296s
         2  0.20025852s  0.10012926s
         4  0.40048001s  0.10012000s
         8  0.80094380s  0.10011798s
        16  1.60193363s  0.10012085s
        32  3.20378083s  0.10011815s
        64  6.40750377s  0.10011725s

Benchmark Results:
------------------
Iterations Total time    Average time per execution
       127 12.71503352s  0.10011837s
```

### DIFF
--- a/libopenage/testing/CMakeLists.txt
+++ b/libopenage/testing/CMakeLists.txt
@@ -1,5 +1,6 @@
 add_sources(libopenage
 	testing.cpp
+	benchmark_test.cpp
 )
 
 pxdgen(
--- a/openage/testing/benchmark.py
+++ b/openage/testing/benchmark.py
@@ -1,0 +1,48 @@
+# Copyright 2017-2017 the openage authors. See copying.md for legal info.
+
+""" Benchmarking tools for the tests. """
+
+from timeit import timeit
+from sys import stdout
+from time import sleep
+
+
+def benchmark_test_function():
+    """ Simple function to call in for benchmarking. """
+    sleep(0.1)
+
+
+def benchmark(func):
+    """
+    Benchmark the given function. Repeated execution helps to give a maximum to
+    the consumed time, until one iteration takes more than 5s, summed up 10s.
+    """
+
+    result = [(0, 1)]
+    number = 1
+    total = [0, 0]
+    str_row_format = "{:10} {:12}  {:11}"
+    row_format1 = "{:10} "
+    row_format2 = "{:11.8f}s  {:10.8f}s"
+    row_format = row_format1 + row_format2
+
+    print(str_row_format.format("Iterations", "Total time", "Average time per execution"))
+    while number < 4 or result[-1][0] < 5 and number < 65537:
+        print(row_format1.format(number), end="")
+        stdout.flush()
+
+        time = timeit(stmt=func, number=number)
+        result.append((time, number))
+        print(row_format2.format(time, time / number))
+        total[0] += number
+        total[1] += time
+        stdout.flush()
+        number *= 2
+
+    del result[0]
+
+    print()
+    print("Benchmark Results: ")
+    print("------------------")
+    print(str_row_format.format("Iterations", "Total time", "Average time per execution"))
+    print(row_format.format(total[0], total[1], total[1] / total[0]))
--- a/openage/testing/main.py
+++ b/openage/testing/main.py
@@ -6,8 +6,9 @@ import argparse
 
 from ..util.strings import format_progress
 
+from .benchmark import benchmark
 from .testing import TestError
-from .list_processor import get_all_tests_and_demos
+from .list_processor import get_all_targets
 
 
 def print_test_list(test_list):
@@ -16,7 +17,7 @@ def print_test_list(test_list):
     """
     namelen = max(len(name) for name, _ in test_list.keys())
 
-    for current_type in ['test', 'demo']:
+    for current_type in ['test', 'demo', 'benchmark']:
         for (name, type_), (_, lang, desc, _) in test_list.items():
             if type_ == current_type:
                 print("[%s %3s] %-*s  %s" % (type_, lang, namelen, name, desc))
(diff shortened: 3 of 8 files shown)


openage/testing/main.py
```diff
@@ -42,29 +42,32 @@ def init_subparser(cli):
...
 
-    if args.run_all_tests and (args.test or args.demo):
-        error("can't run individual test or demo when running all tests")
+    if args.run_all_tests and (args.test or args.demo or args.benchmark):
+        error("can't run individual test or demo or benchmark when running "
+              "all tests")
 
-    if args.test and args.demo:
-        error("can't run a demo _and_ tests")
+    if args.test and (args.demo or args.benchmark) or args.demo and args.benchmark:
```
TheJJ: why the `or args.demo and args.benchmark` at the end?

openage/testing/main.py
```diff
@@ -138,3 +144,8 @@ def main(args, error):
     if args.demo:
         _, _, _, demofun = test_list[args.demo[0], 'demo']
         exit(demofun(args.demo[1:]))
+
+    if args.benchmark:
+        from openage.testing.benchmark import benchmark_func
```
TheJJ: why not import ealier at top-level? and it should be `from .benchmark import benchmark_func`
The function to perform the benchmark should be called `benchmark` i'd say, the testing function `benchmark_test`, then the naming of the "relevant" function would be better.

openage/testing/benchmark.py
```diff
@@ -0,0 +1,47 @@
...
+    row_format1 = "{:10} "
+    row_format2 = "{:11.8f}s  {:10.8f}s"
+    row_format = row_format1 + row_format2
+
+    print(str_row_format.format("Iterations", "Total time", "Average time per execution"))
+    while number < 4 or result[-1][0] < 5 and number < 65537:
+        print(row_format1.format(number), end="")
+        stdout.flush()
+
+        time = timeit(stmt=func, number=number)
```
Tomatower: I could not find anything related in timeit (https://docs.python.org/3/library/timeit.html) , and since "execute, double, repeat" is not that hard of a logic - I did not search deeper.

openage/testing/benchmark.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    Benchmark the given function. Repeated execution helps to give a maximum to
+    the consumed time, until one iteration takes more than 5s, summed up 10s.
+    """
+    result = [(0, 1)]
+    number = 1
+    total = [0, 0]
+    str_row_format = "{:10} {:12}  {:11}"
+    row_format1 = "{:10} "
+    row_format2 = "{:11.8f}s  {:10.8f}s"
```
Tomatower: This is there, that I can just do ```row_format1+row_format2``` for the output in the end, and I don't have to change that many locations.

libopenage/testing/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 add_sources(libopenage
-	testing.cpp
+	"testing.cpp"
+	"benchmark_test.cpp"
```
Tomatower: cmake and I have a long history of hate and shouting - I always do quotes because of it.

TheJJ: :D they make sense if there are variables involved, but for consistency reasons we should skip them in add_sources like the rest of the project does.

openage/testing/benchmark.py
TheJJ: Ah, I mixed that up with the [google benchmark toolkit, which does that](https://github.com/google/benchmark)

openage/testing/main.py
Tomatower: How would you run a demo and a benchmark at the same time?

I want to stop if somebody starts with ```-d demo -b benchmark```

TheJJ: Right. I'd add more `()` for clarification though. e.g. `(args.test and (args.demo or args.benchmark)) or (args.demo and args.benchmark)`

Tomatower: if bool(args.test) ^ bool(args.demo) ^ bool(args.benchmark): - xor-ing is the way to go.

TheJJ: even better :)

castilma: The following logic throws an error if only one of the 3 is passed...

openage/testing/testlist.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2015-2017 the openage authors. See copying.md for legal info.
+# Copyright 2017-2017 the openage authors. See copying.md for legal info.
```
castilma: What happened here?

openage/testing/main.py
```diff
@@ -42,29 +43,32 @@ def init_subparser(cli):
...
-    if args.run_all_tests and (args.test or args.demo):
-        error("can't run individual test or demo when running all tests")
+    if args.run_all_tests and (args.test or args.demo or args.benchmark):
+        error("can't run individual test or demo or benchmark when running "
+              "all tests")
 
-    if args.test and args.demo:
-        error("can't run a demo _and_ tests")
+    if bool(args.test) ^ bool(args.demo) ^ bool(args.benchmark):
+        error("can only run one of demo, benchmarks tests")
```
castilma: if only one of them is true, this throws an error.
you want 
`bool(args.test) + bool(args.demo) + bool(args.benchmark) != 1`

openage/testing/testlist.py
TheJJ: Good question. It should remain 2015. We disabled the start-date-check a while ago because of performance reasons parsing the whole git history for every single file to figure out its oldest date.

openage/testing/main.py
TheJJ: uuuuuuh yes you are right... wtf did i think

PR #804, opened 2017-04-25
Fixes for GCC7

janisozaur:
This is not exhaustive, but remaining warnings seem to be triggered in generated code.

See https://ptpb.pw/7wcN

### DIFF
--- a/libopenage/console/buf.cpp
+++ b/libopenage/console/buf.cpp
@@ -617,6 +617,7 @@ void Buf::process_escaped_cp(int cp) {
 	switch (cp) {
 	case 'c': //RIS (full reset)
 		this->reset();
+		break;
 	default:
 		//unimplemented or invalid codepoint
 		this->escape_sequence_aborted();
--- a/libopenage/engine.cpp
+++ b/libopenage/engine.cpp
@@ -352,6 +352,7 @@ void Engine::loop() {
 						}
 					}
 				}
+				break;
 			}
 
 			default:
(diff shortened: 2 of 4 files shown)


libopenage/util/compress/lzxd.cpp
```diff
@@ -500,7 +500,8 @@ void LZXDStream::init_next_block() {
 		}
 		this->htaligned.make_decode_table();
 
-		// fall through: rest of aligned header is same as verbatim
+		// Falls through.
```
janisozaur: https://dzone.com/articles/implicit-fallthrough-in-gcc-7

libopenage/engine.cpp
```diff
@@ -352,6 +352,7 @@ void Engine::loop() {
 						}
 					}
 				}
+				break;
```
Tomatower: this changes the logic - Do we want that?

libopenage/console/buf.cpp
```diff
@@ -617,6 +617,7 @@ void Buf::process_escaped_cp(int cp) {
 	switch (cp) {
 	case 'c': //RIS (full reset)
 		this->reset();
+		break;
```
Tomatower: This changes the logic, @mic-e pls help how the cp needs to be processed. I guess the new `break` is correct, though.

TheJJ: I think this is okay, as a full reset should not imply an aborted escape sequence.

libopenage/engine.cpp
TheJJ: Nowhere else a windowevent is matched, so thins doesn't seem to break things.

PR #806, opened 2017-04-25
convert: enhance filter options

castilma:
This adds an option to ignore AI and Random Map scripts.

Interface files where not filtered correctly. I found this by running
`$ python3 -m openage convert --no-graphics --no-interface --no-metadata  -v  --force`
There are some `bin` files belonging to animations in `interface.drs`. This documented in `doc/media/drs-file.md`.

There was no option to ignore the random map scripts. There is now.

### DIFF
--- a/openage/convert/driver.py
+++ b/openage/convert/driver.py
@@ -241,14 +241,8 @@ def convert_media(args):
     """ Converts the media part """
 
     # set of (drsname, suffix) to ignore
-    # if drsname is None, it matches to any naoe
-    ignored = set()
-    if args.flag("no_sounds"):
-        ignored.add((None, '.wav'))
-    if args.flag("no_graphics"):
-        ignored.add((frozenset({'graphics', 'terrain', 'gamedata'}), '.slp'))
-    if args.flag("no_interface"):
-        ignored.add((frozenset({'interface'}), '.slp'))
+    # if dirname is None, it matches to any name
+    ignored = get_filter(args)
 
     files_to_convert = []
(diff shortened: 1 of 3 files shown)


openage/convert/driver.py
```diff
@@ -279,7 +282,7 @@ def convert_media(args):
             # do the dir "renaming"
             if dirname == "gamedata" and filepath.suffix == ".slp":
                 output_dir = "graphics"
-            elif dirname == "gamedata" and filepath.suffix == ".wav":
+            elif dirname == "interface" and filepath.suffix == ".wav":
```
castilma: My aoe installation (aoe II from the collectors edition with conquerors expansion) has no `wav` files in `gamedata.drs`, but in `interfaces.drs`. I assume, this was a simple error from TheJJ. Please check you're installation.

castilma:
well, since pylint thinks, 13 branches are to many, we need a way to reorganize some logic or remove the script filter...

castilma:
I can extract the filter creating part into a new function.

openage/convert/driver.py
TheJJ: I have `.wav` files in my gamedata and interface mounts (see for yourself with `python3 -m openage convert -i`).
That condition should therefore be `elif dirname in {"interface", "gamedata"} and filepath.suffix = ".wav"` I'd say/

TheJJ:
Indeed, the `ignored = set()` generation can be done in a new function and return the set.

PR #810, opened 2017-04-30
speed up initial loading of the textures

stiopaa1:
use several threads for loading SDL_Surface

### DIFF
--- a/libopenage/assetmanager.cpp
+++ b/libopenage/assetmanager.cpp
@@ -2,6 +2,9 @@
 
 #include "assetmanager.h"
 
+#include <map>
+#include <SDL2/SDL_image.h>
+
 #if WITH_INOTIFY
 #include <sys/inotify.h>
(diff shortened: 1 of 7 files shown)


stiopaa1:
>Hm, so you're dispatching the loading to worker threads? Good idea, might speed up things (but I can't tell if it really is a bottleneck :) If you like, try our new benchmark functions :D

I haven't looked at benchmark functions yet, but this speeds up initial loading on my machine from around ~12s to ~6s.

stiopaa1:
@TheJJ 
Is that how JobManager should be used? There aren't too many examples in the code...

ChipmunkV:
I'm just using `enqueue()` function in #749:

https://github.com/SFTtech/openage/pull/749/files#diff-37e9ae7e38604588a1695b75b95c39c8R103

Btw, the GL stuff is still in the render thread? AFAIK, it should remain there.

Issue #434, opened 2015-12-29
Use opus/libogg API via cython instead of opusenc binary


TheJJ:
With our Cython interface, we can easily use the `opus` api directly to convert `wav` files to `opus`. For packing it in the `ogg` container, we must use `libogg`.
Should be a bit faster than the current approach of calling the `opusenc` binary for each conversion.

edit: changed from "use opusfile" to "use opus" because opusfile can't encode.

castilma:
opusfile only supports decoding.

TheJJ:
Right you are, thanks. https://github.com/xiph/opusfile I'll update the description.

castilma:
I started working on this and I asked you, how to create a buffer(char *). You gave me these links.
[bytes - python doc](https://docs.python.org/3/c-api/bytes.html)
[example usage](https://github.com/SFTtech/openage/blob/master/openage/cabextract/lzxd.pyx#L104) [note line(s 6 and) 7]

But because byterrays are easier to handle, I tried to use them. 
[bytearray - python doc](https://docs.python.org/3/c-api/bytearray.html)

note, that i searched for header files that contain the functions mentioned in the bytearray doc but didn't find any. 
maybe, you don't really need those functions:
[encoding 'text' to bytes](http://docs.cython.org/en/latest/src/tutorial/strings.html#encoding-text-to-bytes)

[please look at 'write_opus_header'](https://github.com/castilma/openage/tree/pyopus)
this doesn't compile yet. partly, because dereferencing `op` throws an error.
`*op.packet =...`
but op is a pointer, '->' doesn't work either. how is that supposed to be?

and did I make an error by using bytearray because of it's easier usage?

2 problems so far, and I still need to get my head around cython.

TheJJ:
Basically we have to create a simplified version of [the opusenc tool](https://git.xiph.org/?p=opus-tools.git;a=blob;f=src/opusenc.c).
For the header creation, they use `opus_header_to_packet`.

You should handle all the data in the conversion as bytes directly. In `opusenc`, they also know the block size beforehand, which is why I suggested that you use `bytes` instead of a `bytearray`. But it should not matter. Nevertheless, I think it will be easier to stick to `bytes`.

`bytes` can directly be casted to `unsigned char *` (see the `lzxd.pyx`) (not sure about `bytearray`).

If you create regular python objects, then they are garbagecollected because of their reference counting.

For dereferencing memory, you can use the `dereference` function:
[...]

PR #814, opened 2017-05-01
Use libopus instead of opusenc binary for conversion.

castilma:
#This is in response to #434.

Things todo:
- [x] Make the resampler faster.
- [x] Comment my code. (I'm supposed to do that.)
- [x] Comment on my code. (You're supposed to do that.)
- [x] Checkout which `todo`s from the `todo` file should be done and do them.
- [x] Remove opus-tools requirement from buildsystem and documentation. [Who is responsible for the [arch packet](https://aur.archlinux.org/packages/openage-git/)? @linkmauve or @FabioLolix?]
- [x] Register cython files into buildsystem.
- [x] Rebase and squash leaving only a few commits. (One for the code, one for the buildsystem changes, one for ...) Please don't merge before I ticked this.

Optionally:
- [ ] Come up with an idea how to unit test this code.


doc/building.md
```diff
@@ -56,7 +56,7 @@ Dependency list:
 ### Prerequisite steps for Ubuntu users (Ubuntu 15.04)
 
  - `sudo apt-get update`
- - `sudo apt-get install cmake libfreetype6-dev python3-dev libepoxy-dev libsdl2-dev libsdl2-image-dev libopusfile-dev libfontconfig1-dev libharfbuzz-dev opus-tools python3-pil python3-numpy python3-pygments python3-pip qtdeclarative5-dev qml-module-qtquick-controls`
```
castilma: libopusfile-dev already depends on libopus-dev. I'm not sure if it is supposed to be installed in an explicit way.
And I don't know about other distros.

openage/opus/opusenc.pyx
```diff
@@ -0,0 +1,415 @@
...
+#cython: language_level=3
+#^- asdf: remove if unnecessary?
+
+import time
+from libc.string cimport memcpy, memset
+from cpython.mem cimport PyMem_Malloc, PyMem_Free
+from cpython.bytes cimport PyBytes_FromStringAndSize
+
+from ogg cimport *
+from opus cimport *
```
TheJJ: we should avoid `*`-imports as they are very indirect and confusing where which functions comes from. I'd suggest to import them as `ogg` and `opus`, the 4 or 5 letters more won't hurt.

openage/opus/opusenc.pyx
```diff
@@ -0,0 +1,415 @@
...
+    if rate > 24000:
+        coding_rate = 48000
+    elif rate > 16000:
+        coding_rate = 24000
+    elif rate > 12000:
+        coding_rate = 16000
+    elif rate > 8000:
+        coding_rate = 12000
+    else:
+        coding_rate = 8000
```
TheJJ: why don't those rates match to their lower bounds?

openage/opus/opusenc.pyx
```diff
@@ -0,0 +1,415 @@
...
+    divisor = gcd(target_rate, opts['input_rate'])
+    num = target_rate / divisor
+    den = opts['input_rate'] / divisor
+
+    osmpls = ((ismpls - 1) * num /den).__trunc__() + 1 # asdf: explain
+    outsize = osmpls * wframe_len
+
+    out = <char *> PyMem_Malloc(outsize)
+    if out == NULL:
+        return "Out of memory."
```
TheJJ: i'd raise an exception here.

openage/opus/opusenc.pyx
```diff
@@ -0,0 +1,415 @@
...
+    cdef int r = -1
+    while r != 0:
+        r = m % n
+        m, n = n, r
+    return m
+
+
+def debug(*args):
+    if verbose:
+        print(args[0].format(*args[1:]))
```
TheJJ: we have a nice logging system, you know? :)  (`from openage.log import debug`)

openage/opus/opustest.py
```diff
@@ -0,0 +1,39 @@
...
+        print("Error: {}".format(out))
+        return 1
+
+    outname = wavname.replace("wav", "opus") if wavname.endswith("wav") else wavname + ".opus"
+    opus = open(outname, mode='wb')
+    opus.write(out)
+    print("Wave size: {}\tOpus size:{}\tcompression: {:.2n}\ttime {}".format(len(wav),len(out),len(out)/len(wav),t))
+
+if __name__ == "__main__":
+    convert()
```
TheJJ: might be good if you converted this file into a demo, then you can launch it with `./run test -d demoname`.

openage/opus/setup.py
```diff
@@ -0,0 +1,14 @@
...
+from Cython.Build import cythonize
+
+setup(
+    name = 'opusencoder',
+    ext_modules = cythonize([
+        Extension("opusenc",["opusenc.pyx"],
+                   libraries=["ogg","opus"])
+        ])
+)
+
```
TheJJ: the buildsystem integration is very simple, just look like e.g. `./openage/cabextract/CMakeLists.txt` does it.

doc/building.md
TheJJ: If we make use if libopus directly, I'd add it, who knows how this may be packaged in the future. We can't avoid the explicit avoiding, developers will need to install those packages, unfortunately.

openage/opus/opusenc.pyx
castilma: I don't understand what your question is. Are you asking, why I didn't choose e.g. coding_rate = 12000 for 12000 <= rate  < 16000 ?
That could lead to aliasing, which sounds horrible, if I didn't lowpass the input below half the coding_rate before resampling. By upsampling we avoid that filtering.

openage/opus/opusenc.pyx
castilma: only here, or at other allocations too? or replace every error message with an exception?

openage/opus/opusenc.pyx
TheJJ: Yes, that was the question. I have no idea how to resample properly :)

openage/opus/opusenc.pyx
TheJJ: Otherwise that function would just return a error string, is that handled peoperly then? Python exceptions sound like the right way to abort after the error.

openage/opus/setup.py
castilma: I tried, but `cimport ogg` fails. ogg.pxd is not found. even though it is in the same directory.

openage/opus/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+	opusenc.pyx
+)
+
+add_pxds(
+	ogg.pxd
+	opus.pxd
+)
+
+add_py_modules(
+	__init__.py
```
castilma: I don't know, that file is necessary.

openage/opus/todo
```diff
@@ -0,0 +1,31 @@
...
+
+remove debug messages from opusenc.py?
+
+remove unused error constants in opus.pxd ?
+
+limit bandwidth for inputsignals with low samplerate ? check, if this is done automatical.
+
+copy metadata from Inputfile?
+
+see, why the upsampler is so slow.
```
TheJJ: you can profile cython code: see here: https://cython.readthedocs.io/en/latest/src/tutorial/profiling_tutorial.html

openage/opus/setup.py
TheJJ: `from . cimport ogg` should be the right statement for this relative import

castilma: where do I declar the necessary libraries to link to?

castilma: okay, `set(PYEXT_LINK_LIBRARY ogg opus)` is the right way to go. but I don't know how it linked to opus without that in CMakeLists.txt

openage/opus/todo
castilma: I read the opusenc.cpp and declared the variable 'ch' as with cdef int. for a 6 minute song the resampling now takes only 3 seconds instead of 30. total encoding time is now 15 seconds instead of 42.

the cython generated code looks pretty horrible to me. I hope the compiler optimizes most of. Where are the compilersettings for cython generated files?

zuntrax: `/buildsystem/HandlePythonOptions.cmake`
That stuff is built with mostly the same options as hand-written C++ code, but with reduced warnings.
Just checked the optimization flags, `-O0` and `-O3` are in the compiler invocation, but I guess `-O3` wins because it comes later.

openage/opus/CMakeLists.txt
TheJJ: It is, otherwise the folder is not accepted as a python module.

openage/opus/setup.py
TheJJ: Hmm, the `pyext_link_library` seems to be non-ideal, as the variable value would descend the recursion of the directory traversal. But it is the only way currently linking against a custom library. Those libraries need to be found first (see the buildsystem folder and the `libopenage/CMakeLists.txt` on how that is done).

Hardcoding `ogg` and `opus` won't work on all platforms, that's why the locations are "found" first and then inserted there.

castilma:
the second last commit will be changed. I'm asking you to run
`$ time ./run convert --no-graphics --no-interface --no-scripts --no-metadata   --force`
twice. before running it the second time, change the comments in `convert/driver.py:400` so it uses the module.
my times:
`real	0m41.558s
user	0m34.936s
sys	0m3.608s` with the module
`real	0m23.106s
user	0m21.476s
sys	0m9.496s` with opusenc binary.

TheJJ:
```
opusenc.pyx with ./configure --mode=debug
./run convert --no-graphics --no-interface --no-scripts --no-metadata --force  41.81s user 6.76s system 112% cpu 43.310 total

opusenc binary:
./run convert --no-graphics --no-interface --no-scripts --no-metadata --force  16.50s user 9.56s system 281% cpu 9.245 total
```

TheJJ:
For profiling, I added the following. For "better" results, I loop the conversion 100 times.

```patch
diff --git a/openage/convert/driver.py b/openage/convert/driver.py
index f76c935e..09a198ae 100644
--- a/openage/convert/driver.py
+++ b/openage/convert/driver.py
@@ -407,7 +407,22 @@ def convert_wav(filepath, dirname, args):
 #       raise Exception("opusenc failed")
 
     tic = time.time()
-    outdata = opusenc.encode(indata)
+
+
+    import cProfile, pstats, io
+    pr = cProfile.Profile()
+    pr.enable()
+
+    for i in range(100):
+        outdata = opusenc.encode(indata)
+
+    pr.disable()
+    s = io.StringIO()
+    sortby = 'cumulative'
+    ps = pstats.Stats(pr, stream=s).sort_stats(sortby)
+    ps.print_stats()
+    print(s.getvalue())
+
     tic = time.time() - tic
     info("opusenc module took {:.3f} seconds".format(tic))
     if isinstance(outdata, (str, int)):
@@ -420,6 +435,8 @@ def convert_wav(filepath, dirname, args):
     with args.targetdir[out_parts].with_suffix('.opus').open_w() as ou
         outfile.write(outdata)
 
+    exit(0)
+
 
 def convert_mediafile(filepath, dirname, names_map, converter_pool, ar
     """
diff --git a/openage/convert/opus/opusenc.pyx b/openage/convert/opus/op
index 05f8fbce..6a1d8f6e 100644
--- a/openage/convert/opus/opusenc.pyx
+++ b/openage/convert/opus/opusenc.pyx
@@ -1,5 +1,7 @@
 # Copyright 2016-2017 the openage authors. See copying.md for legal in
 
+# cython: profile=True
+
 import time
 from libc.string cimport memcpy, memset
 from cpython.mem cimport PyMem_Malloc, PyMem_Free
```

castilma:
If I convert the assets using multiple threads, it gets slower a few seconds (~25 seconds instead of 20). No matter how many threads I choose, cpu usage is only at 60% according to my tray bar (i.e. only one core, the rest is my music player and stuff). Is that normal?
With the binary opusenc and one thread, it takes about 30 seconds. Using more than one, my cpu usage is at 100% (both cores busy) and conversion takes 19 seconds.

Can you check how it is on your machine?

I will check if that's case with the rest that's being converted, too.

openage/convert/opus/ogg.pxd
```diff
@@ -0,0 +1,32 @@
+# Copyright 2016-2017 the openage authors. See copying.md for legal info.
+
+cdef extern from "ogg/config_types.h":
+    # TODO hope the best
+    ctypedef short ogg_int16_t
+    ctypedef long ogg_int64_t
```
castilma: What do we do about that?
libogg has a lot of `#ifdef`s. I didn't find a way to tell cython to just use the definition from the header.
cython is really limiting portability.
can someone maybe come up with a workaround? see the comment on opus.pxd for inspiration.

openage/convert/opus/opus.pxd
```diff
@@ -0,0 +1,29 @@
...
+    # TODO replace int by int32_t or hope that this works
+    ctypedef int opus_int32
+    ctypedef short opus_int16
+    OpusEncoder *opus_encoder_create(opus_int32 Fs, int channels,
+                                     int application, int *error)
+    void opus_encoder_destroy(OpusEncoder *st)
+    opus_int32 opus_encode(OpusEncoder *st, const opus_int16 *pcm,
+                           int frame_size, unsigned char *data,
+                           opus_int32 max_data_bytes)
+    int opus_encoder_ctl(OpusEncoder *st, int request)
```
castilma: opus_encoder takes actually more than two arguments, but one has to use a macro for the second argument, which evaluates to commata and arguments. but there are no macros in cython. what do? see below! just say the macro is a function. cython will just leave the 'function call' there, and that way, the c preprocessor can see it.

openage/convert/opus/opusenc.pyx
```diff
@@ -0,0 +1,406 @@
+# Copyright 2016-2017 the openage authors. See copying.md for legal info.
+
+import time
+from libc.string cimport memcpy, memset
+from cpython.mem cimport PyMem_Malloc, PyMem_Free
+from cpython.bytes cimport PyBytes_FromStringAndSize
+
+from ...log import dbg, spam
+
+from . cimport ogg, opus
```
castilma: `make cythonize` gives me this warning:
`/home/castilma/.repo/ap/openage/openage/convert/opus/opusenc.pyx: cannot find cimported module '.'`

openage/convert/opus/opusenc.pyx
```diff
@@ -0,0 +1,406 @@
...
+
+    err = opus.opus_encoder_ctl(oe, opus.OPUS_GET_LOOKAHEAD(&lookahead))
+    if err != opus.OPUS_OK:
+        free_buffers(&os, oe, NULL, NULL)
+        return 1
+
+    inopt['pre_skip'] = lookahead / (48000 / coding_rate)
+
+    if write_opus_header(&os, &op, inopt):
+        free_buffers(&os, oe, NULL, NULL)
```
castilma: there is a potential memory leak here and in other places. write_opus_header creates a bytearray. if there is no memory left, that fails and a MemoryError is thrown. this is not catched, and free_buffers will not be called.
since this will quit the conversion (and release all memory), I think, this does not need to be fixed.
But I will fix it, if requested.

zuntrax: If you are sure that it gets disposed of in another place, that would be OK.
On the other hand, a simple try-finally construct would better communicate what we expect to happen to the buffers.

TheJJ: Would be better to always fix those problems beforehand, or, at least write some comment so if we change the code someday, that won't stay unnoticed.

openage/convert/opus/opusenc.pyx
```diff
@@ -0,0 +1,430 @@
...
+        if outsize == 0:
+            free_buffers(&os, oe, NULL, NULL)
+            raise MemoryError("Could not write opus header.")
+
+    err = write_opus_comment(&os, &op)
+    if err:
+        free_buffers(&os, oe, NULL, NULL)
+        if err == 1:
+            raise MemoryError("Could not create opus comment.")
+        return "Could not write opus comment."
```
castilma: So, now I catch all (?; see other comment) MemoryErrors in write_opus_comment et al. But I could catch them here. I'm not sure, what's more eloquent?

openage/convert/opus/opusenc.pyx
```diff
@@ -0,0 +1,428 @@
...
+        buf = bytearray(19)
+    except MemoryError:
+        return 1
+
+    buf[0:8] = b'OpusHead'
+    buf[8] = 1                # Opus encapsulating version.
+    buf[9] = inopt['channels']
+    buf[10:12] = inopt['pre_skip'].to_bytes(2, 'little')
+    buf[12:16] = inopt['input_rate'].to_bytes(4, 'little')
+    buf[16:18] = b'\x00\x00'  # Output gain
```
castilma: Does anyone know, if this could raise a MemoryError, or if Python is smart enough to directly write into buf and avoid creating b'\x00\x00'?

zuntrax: You could check the generated C++ code in context of the python source. `opusenc.html` will be **extremely** useful there. The darker the yellow background, the more stuff is happening in the python interpreter. Also you can click on the line to expand the generated code. (I hope I'm not telling you stuff you already know...)

castilma: I knew about `opusenc.html`, but (for some reason) didn't think that it would be useful.
Checked it now; it looks good.

openage/convert/opus/opusenc.pyx
TheJJ: All errors should raise `Exceptions`, those are easier to handle than to decide if a return value was a string or some number, although the effect is the same.

As all our files are `c++`, you maybe could use `make_unique` instead of the `pymalloc` so no free call is needed at all. That buffer you can still pass the raw pointer (`pointer.get()`) to opus and python.

openage/convert/opus/opusenc.pyx
```diff
@@ -0,0 +1,428 @@
...
+from libc.string cimport memcpy, memset
+from cpython.mem cimport PyMem_Malloc, PyMem_Free
+
+from ...log import dbg, spam
+
+from .bytearray cimport (PyByteArray_AS_STRING, PyByteArray_GET_SIZE,
+                         PyByteArray_Resize)
+cimport ogg, opus
+
+def encode(inputdata=None):
```
TheJJ: I wouldn't default it. the function is useless if inputdata is `None`, right?

openage/opus/setup.py
castilma: So I need to make a cmake module for finding libogg? The FindOpusfile.cmake seems to also find libopus.
side question: will I need to call `find_package(Opusfile)` again, since this is under `openage` and not under `libopenage/CMakeLists.txt`, where it is called first?

openage/convert/opus/opusenc.pyx
```diff
@@ -0,0 +1,428 @@
...
+    if oe != NULL:
+        opus.opus_encoder_destroy(oe)
+    if pymem1 != NULL:
+        PyMem_Free(pymem1)
+    if pymem2 != NULL:
+        PyMem_Free(pymem2)
+    return
+
+
+cdef upsample(indata, dict opts, int target_rate):
```
TheJJ: i think you can speed up things if you declare the `indata` type or use a memoryview.

openage/convert/opus/opusenc.pyx
```diff
@@ -0,0 +1,428 @@
...
+    # with num / den we get the index of the last output sample. In general,
+    # this is not a whole number. Therefore we take only the integer part.
+    # Because indices are zero-based, add one for the number of output samples.
+    osmpls = ((ismpls - 1) * num / den).__trunc__() + 1
+    outsize = osmpls * channel * 2  # we convert to int16
+
+    try:
+        ret = bytearray(outsize)
+    except MemoryError:
+        return "Could not allocate buffer."
```
TheJJ: I think you can safely just let that exception raise, other memory errors are raised anyway.

openage/convert/opus/opusenc.pyx
```diff
@@ -0,0 +1,428 @@
...
+    if bit_depth > 8:
+        for iout in range(0, osmpls):
+            if (iin + 1) * num < iout * den:
+                iin += 1
+
+            alpha = iout * den / num - iin
+            for ch in range(0, channel):
+                a = (<short *> inp)[iin * channel + ch]
+                b = (<short *> inp)[(iin + 1) * channel + ch]
+                (<short *> out)[(iout + ch) * channel] = (<short> (a + alpha * (b - a))) & mask
```
TheJJ: those casts are always 16 bit, right? a `uint16_t` cast might be more platform compatible.

openage/convert/opus/opusenc.pyx
```diff
@@ -0,0 +1,428 @@
...
+    op.bytes = <long> len(buf)
+    op.b_o_s = <long> 0
+    op.e_o_s = <long> 0
+    op.granulepos = <ogg.ogg_int64_t> 0
+    op.packetno = <ogg.ogg_int64_t> 1
+
+    return ogg.ogg_stream_packetin(os, op)
+
+
+def read_wav(wav):
```
TheJJ: cythonizing this function will surely give some more speed boost.

openage/convert/opus/opusenc.pyx
castilma: indata may be bytes or bytearray; declaring would force one type, right?

openage/convert/opus/opusenc.pyx
castilma: I think it has to be signed, not sure though. the best would be to use opus.opus_int16.

openage/convert/opus/opusenc.pyx
castilma: what do you mean by cythonizing? Have you profiled it?

openage/convert/opus/opusenc.pyx
castilma: I don't have that much experience with c++. what does make_unique and why is no freeing necessary? I found [this](http://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique), but it doesn't help me much. does make_unique allocate on the stack?

Vtec234: It's a 'smart pointer' that deallocates the memory for you when it goes out of scope. So no freeing is necessary, since when the `std::unique_ptr` is destroyed, it calls `free` internally. It uses the heap.

openage/convert/opus/opusenc.pyx
zuntrax: Making it a `cdef` and using a few more of the pyx language extensions. Basically every cdef'd variable is handled by native C++ code without invoking the python interpreter, which is way faster.

TheJJ: Such a `cdef` method can't be called from python code, only within cython code itself. There's the hybrid `cpdef` which is both.
You can learn a lot about cython here if you haven't looked into that already: http://docs.cython.org/en/latest/src/userguide/language_basics.html

buildsystem/modules/FindOgg.cmake
```diff
@@ -0,0 +1,28 @@
+# Copyright 2017-2017 the openage authors. See copying.md for legal info.
+# asdf: this file was copied from FindOpus.cmake. check legal stuff.
```
TheJJ: this file contains very basic cmake invocations that are common to just finding some library, so i don't think we need to attribute the findopus author.

openage/convert/opus/opusenc.pyx
castilma: I'm really trying not to hate cython. I have a little problem with make_unique. At first I [practiced](https://codepaste.net/raa4nb) in c++. that worked. but in cython, the templating doesn't work as I expected (It uses [type] instead of <type>). figured that out, but then
[error](https://codepaste.net/aodn91)
[diff](https://codepaste.net/4jz9ff)
any ideas why `char[]` worked in the variable declaration, but not in `make_unique`?

TheJJ: I think if you use `ctypedef char[] chararray` and then use `make_unique[chararray]` it could work. But I can't tell why it won't work, sorry.
If the smart pointer is too much effort, I think it's okay to stick with the `bytearray` object from python, it is refcounted and free'd automatically.

castilma: the typedef worked, thanks. but there is another problem. enabling the conversion from c++ exceptions to python exceptions breaks compilation. from `~/.local/lib/python3.5/site-packages/Cython/Includes/libcpp/memory.pxd`:

>     # Temporaries used for exception handling break generated code
>     unique_ptr[T] make_unique[T](...) # except + 

if I remove the hash, compilation fails:
```
/home/mcd/.repo/ap/openage/openage/convert/opus/opusenc.cpp: In function ‘PyObject* __pyx_pf_7openage_7convert_4opus_7opusenc_encode(PyObject*, PyObject*)’:
/home/mcd/.repo/ap/openage/openage/convert/opus/opusenc.cpp:2756:20: error: use of deleted function ‘std::unique_ptr<_Tp [], _Dp>& std::unique_ptr<_Tp [], _Dp>::operator=(const std::unique_ptr<_Tp [], _Dp>&) [with _Tp = char; _Dp = std::default_delete<char []>]’
       __pyx_v_spad = __pyx_t_20;
                    ^
In file included from /usr/include/c++/5/memory:81:0,
                 from /home/mcd/.repo/ap/openage/libopenage/pyinterface/../error/error.h:6,
                 from /home/mcd/.repo/ap/openage/libopenage/pyinterface/exctranslate.h:21,
                 from /home/mcd/.repo/ap/openage/libopenage/pyinterface/hacks.h:21,
                 from <command-line>:0:
/usr/include/c++/5/bits/unique_ptr.h:599:19: note: declared here
       unique_ptr& operator=(const unique_ptr&) = delete;
                   ^
```
I don't know, if someone from cython is working on that.
do we want to allow uncatched exceptions here?

TheJJ: Looks like the copy-assignment-operator is invoked here. I would just use a `std::shared_ptr` then. Or the python `bytearray` as storage :)
`unique_ptr`s and cython don't seem to play well together...

castilma: there is no way to allocate a shared pointer and a dynamic buffer with it, like there is for unique_ptr. using shared_ptr results in a compilation error:
```
/usr/include/c++/5/ext/aligned_buffer.h:85:34: error: invalid application of ‘sizeof’ to incomplete type ‘char []’
     : std::aligned_storage<sizeof(_Tp), std::alignment_of<_Tp>::value>
                                  ^
```

bytearray it is.

iNaKoll: http://en.cppreference.com/w/cpp/memory/default_delete

iNaKoll: Sorry for my "link response" above. I must have deleted my message when I paste the link... Anyway, to use a dynamic array with a shared_ptr one solution is to use std::default_delete (the default deleter of std::unique_ptr) : 
```
#include <algorithm>
#include <iostream>
#include <memory>

int main()
{
	std::shared_ptr<int> shared_good(new int[10], std::default_delete<int[]>());
	std::generate_n(shared_good.get(), 10, [i=0]() mutable { return ++i; });
	std::cout << shared_good.get()[9] << std::endl;
	return 0;
}
```

castilma:
someone started writing [libopusenc](https://opus-codec.org/release/dev/2017/05/24/libopusenc-0_1.html).

TheJJ:
I also can't see about the old versions, but as they're no longer listed they may be out of support already anyway.

https://apps.fedoraproject.org/packages/libogg-devel
https://apps.fedoraproject.org/packages/opus-devel

https://en.wikipedia.org/wiki/Fedora_(operating_system)#Releases

==> somebody should remove the old fedoras from our instructions anyway.

openage/convert/opus/opusenc.pyx
castilma: thanks. I tried it, but failed. cython seems to support the new operator, but translating your posted solution (and variations) doesn't build in cython.
seems like someone smarter than me has to do it, or we keep the current way of allocating.

openage/convert/opus/opusenc.pyx
castilma: I really don't feel like this needs to be made faster. The header parsing takes only a fraction of the conversion time. I think putting effort into reducing it takes time, but will give us barely any speed boost.

TheJJ: okay :) then we can come back to this later.

doc/building.md
```diff
@@ -42,7 +42,8 @@ Dependency list:
     CR    sdl2
     CR    sdl2_image
     CR    opusfile
-      A   opus-tools
+      A   libopus
+      A   libogg
```
TheJJ: i think they are both `CRA`, because opus is needed at compile and runtime (via opusfile).

openage/convert/opus/opus.pxd
```diff
@@ -0,0 +1,24 @@
...
+
+cdef extern from "opus/opus_defines.h":
+    #Errors
+    const int OPUS_OK = 0
+    const int OPUS_ALLOC_FAIL = -7
+    #Application
+    const int OPUS_APPLICATION_AUDIO = 2049
+    #Macros
```
TheJJ: nitpick: python comments should have a space being the `#` before text starts

openage/convert/opus/opusenc.pyx
```diff
@@ -0,0 +1,415 @@
...
+        for iout in range(0, osmpls):
+            if (iin + 1) * num < iout * den:
+                iin += 1
+
+            alpha = (<float> iout * den) / num - iin
+            for ch in range(0, channel):
+                a = (<char *> inp)[iin * channel + ch]
+                b = (<char *> inp)[(iin + 1) * channel + ch]
+                (<opus.opus_int16 *> out)[iout * channel + ch] = \
+                        (<opus.opus_int16> (a + alpha * (b - a))) << 8 & mask
```
TheJJ: the only difference is the `<< 8`, right? we could init that number to 0 or 8 depending what bit_depth is.

```python
if bit_deptn > 8:
    shift_correct = 0
else:
    shift_correct = 8
```

and always do `<< shift_correct`.

openage/convert/opus/demo.py
```diff
@@ -0,0 +1,46 @@
...
+    if isinstance(out, (str, int)):
+        crit("Encoding failed: {}".format(out))
+        return 1
+
+    outname = args.out
+    if outname is None:
+        outname = wavname[:-3] + "opus" if wavname[-3:] == "wav" else wavname + ".opus"
+
+    info("Writing {}.".format(outname))
+    open(outname, mode='wb').write(out)
```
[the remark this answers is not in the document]
castilma: I thought files get closed, when the file object gets garbagecollected

openage/convert/opus/opusenc.pyx
castilma: and the casts are to `char *` instead of `short *`

TheJJ: Oh, didn't see that, sorry. Can you add a comment in the code what differences are needed and why?

openage/convert/opus/demo.py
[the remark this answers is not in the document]
TheJJ: Indeed they are. But pypy warns about this:
http://doc.pypy.org/en/latest/cpython_differences.html#differences-related-to-garbage-collection-strategies

openage/convert/opus/opusenc.pyx
castilma: done

castilma:
I still need to check the refcounting thing in the todo file, but then this should be it.

castilma:
the macos dependencies still need to be changed. but otherwise this is finished.

zuntrax:
#967 is merged.

zuntrax:
@castilma what is the isshue with the macos dependencies? You highlighted me in that context.

castilma:
sorry. I only mentioned you to approve the rebase.

TheJJ:
Improvements like paralellism can be added now that it is merged :)

FabioLolix:
Arch package have been updated

castilma:
looks like #434 can be closed.

PR #821, opened 2017-05-02
Fix build on macOS computers that don't have Sierra && XCode 8

Birch-san:
Since we are setting time to "NOW", we can just use a nullptr. We can also ditch `time.h`, since we no longer need `clock_gettime()`.

### DIFF
--- a/libopenage/util/fslike/directory.cpp
+++ b/libopenage/util/fslike/directory.cpp
@@ -11,7 +11,6 @@
 #include <unistd.h>
 
 #ifdef __APPLE__
-#include <time.h>
 #include <sys/time.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


Birch-san:
Oh, was I missing an `== 0` this whole time?

The original code I wrote for the Mac side was based on this:
https://github.com/mpartel/bindfs/issues/6

Birch-san:
`utimes`:
> RETURN VALUES
>     Upon successful completion, the value 0 is returned; otherwise the value -1 is returned and the global variable
>     errno is set to indicate the error.

Birch-san:
I've added the `== 0`. I suspect that this function never gets used. otherwise it would've been failing catastrophically this whole time.

PR #823, opened 2017-05-02
testing: fix copyright year and wrong logic

castilma:
Fixes problems in #802.

### DIFF
--- a/openage/testing/main.py
+++ b/openage/testing/main.py
@@ -61,8 +61,8 @@ def process_args(args, error):
         error("can't run individual test or demo or benchmark when running "
               "all tests")
 
-    if bool(args.test) ^ bool(args.demo) ^ bool(args.benchmark):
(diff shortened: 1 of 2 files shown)


openage/testing/main.py
```diff
@@ -61,8 +61,8 @@ def process_args(args, error):
         error("can't run individual test or demo or benchmark when running "
               "all tests")
 
-    if bool(args.test) ^ bool(args.demo) ^ bool(args.benchmark):
-        error("can only run one of demo, benchmarks tests")
+    if bool(args.test) + bool(args.demo) + bool(args.benchmark) != 1:
```
TheJJ: `+ bool(args.run_all_tests)` should do it.

castilma: since that is already tested a few lines before, I took > 1

PR #832, opened 2017-05-07
Spelling

castilma:
some things I found over time.
we could keep this pr open for a while and I will add stuff over time.

I don't know, if you want the change in print in logging.py. If you do, I will squash it, if not, remove it.

### DIFF
--- a/buildsystem/modules/FindOpusfile.cmake
+++ b/buildsystem/modules/FindOpusfile.cmake
@@ -1,7 +1,7 @@
 # This file was taken from Unvanquished,
 # Copyright 2000-2009 Kitware, Inc., Insight Software Consortium
 # It's licensed under the terms of the 3-clause OpenBSD license.
-# Modifications Copyright 2014-2015 the openage authors.
+# Modifications Copyright 2014-2017 the openage authors.
 # See copying.md for further legal info.
 
 # - Find opus library
@@ -12,8 +12,15 @@
 #  OPUS_FOUND          - True if opus is found.
 
 # find the opusfile header, defines our api.
-find_path(OPUS_INCLUDE_DIR
+find_path(OPUSFILE_INCLUDE_DIR
 	NAMES opus/opusfile.h
+	DOC "Opusfile include directory"
+)
+mark_as_advanced(OPUSFILE_INCLUDE_DIR)
+
+# find the opus header
+find_path(OPUS_INCLUDE_DIR
+	NAMES opus/opus.h
 	DOC "Opus include directory"
 )
 mark_as_advanced(OPUS_INCLUDE_DIR)
@@ -33,11 +40,11 @@ find_library(OPUS_LIBRARY
 mark_as_advanced(OPUS_LIBRARY)
 
 
-# handle the QUIETLY and REQUIRED arguments and set OPUSFILE_FOUND to TRUE if
+# handle the QUIETLY and REQUIRED arguments and set OPUS_FOUND to TRUE if
 # all listed variables are TRUE
 include(FindPackageHandleStandardArgs)
 find_package_handle_standard_args(Opus DEFAULT_MSG OPUSFILE_LIBRARY OPUS_LIBRARY OPUS_INCLUDE_DIR)
 
 # export the variables
 set(OPUS_LIBRARIES "${OPUSFILE_LIBRARY}" "${OPUS_LIBRARY}")
-set(OPUS_INCLUDE_DIRS "${OPUS_INCLUDE_DIR}" "${OPUS_INCLUDE_DIR}/opus")
+set(OPUS_INCLUDE_DIRS "${OPUSFILE_INCLUDE_DIR}" "${OPUS_INCLUDE_DIR}" "${OPUS_INCLUDE_DIR}/opus")
--- a/libopenage/audio/dynamic_resource.h
+++ b/libopenage/audio/dynamic_resource.h
@@ -39,7 +39,7 @@ struct chunk_info_t {
 	/** The chunk's current state. */
 	std::atomic<state_t> state;
 
-	/** The chunk's actual size. */
+	/** The chunk's count of valid entries (samples). */
 	size_t size;
 
 	/** The chunk's buffer. */
--- a/libopenage/audio/dynamic_resource.h
+++ b/libopenage/audio/dynamic_resource.h
@@ -94,7 +94,7 @@ public:
 	 */
 	static constexpr int DEFAULT_PRELOAD_AMOUNT = 10;
 
-	/** The default used chunk size in bytes (100ms). */
+	/** The default used chunk size in bytes (100ms for 48kHz stereo audio). */
 	static constexpr size_t DEFAULT_CHUNK_SIZE = 9600*2;
 
 	/** The default number of chunks, that can be loaded at the same time. */
--- a/libopenage/engine.h
+++ b/libopenage/engine.h
@@ -253,7 +253,7 @@ public:
 	time_nsec_t lastframe_duration_nsec() const;
 
 	/**
-	 * render text with the at a position with specified font size
+	 * render text at a position with the specified font size
 	 */
 	void render_text(coord::viewport position, size_t size, const renderer::Color &color, const char *format, ...) ATTRIBUTE_FORMAT(5, 6);
 
--- a/libopenage/engine.h
+++ b/libopenage/engine.h
@@ -158,13 +158,6 @@ public:
 	void end_game();
 
 	/**
-	 * draw the current frames per second number on screen.
-	 * save the current framebuffer to a given png file.
-	 * @param filename the file where the picture will be saved to.
-	 */
-	void save_screenshot(const char* filename);
-
-	/**
 	 * Draw the game version and the current FPS on screen.
 	 */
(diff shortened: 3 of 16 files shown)


openage/log/logging.py
```diff
@@ -62,9 +62,8 @@ def log(lvl, msg, stackframes=1):
         # the C++ interface is uninitialized.
         with LOCK:  # pylint: disable=not-context-manager
             if lvl.value >= Level.current.value:
-                print("\x1b[" + level_colorcode(lvl) + "m" +
-                      lvl.name.upper().rjust(4) + "\x1b[m " +
-                      msg)
+                print(''.join("\x1b[", level_colorcode(lvl), "m",
+                              lvl.name.upper().rjust(4), "\x1b[m ", msg))
```
castilma: pylint says too many positional arguments. I'd say we ignore it.

> E:65,22: Too many positional arguments for method call (too-many-function-args)

zuntrax: Maybe wrap it in `[ ]`, then pylint will stop complaining.

TheJJ:
The `logging.py` `join`-improvement is good, we'll keep it :)

castilma:
I'd keep this open for a couple weeks/months.

openage/log/logging.py
```diff
@@ -63,9 +63,8 @@ def log(lvl, msg, stackframes=1):
         # the C++ interface is uninitialized.
         with LOCK:  # pylint: disable=not-context-manager
             if lvl.value >= Level.current.value:
-                print("\x1b[" + level_colorcode(lvl) + "m" +
-                      lvl.name.upper().rjust(4) + "\x1b[m " +
-                      msg)
+                print(''.join(["\x1b[", level_colorcode(lvl), "m",
+                               lvl.name.upper().rjust(4), "\x1b[m ", msg]))
```
TheJJ: you don't have to join the string, print can have multiple arguments that will be printed sequentially

castilma:
what is our if else style?
```if (istrue()) {
    do();
} else {
    do_else();
}
```
or
```if (istrue()) {
    do();
}
else {
    do_else();
}
```
?

TheJJ:
I think it's best to allow both (https://github.com/SFTtech/openage/blob/master/doc/code_style/mom.cpp#L192), because I can't see any big advantage for using either of them.

libopenage/audio/dynamic_resource.h
```diff
@@ -39,7 +39,7 @@ struct chunk_info_t {
 	/** The chunk's current state. */
 	std::atomic<state_t> state;
 
-	/** The chunk's actual size. */
+	/** The chunk's actual amount of values in buffer. */
```
castilma: is that clear? what would you understand from that?

TheJJ: It is the number of bytes that are stored in this chunk of audio. Namely the size of `buffer`.

simonsan:
Can we please resolve the conflicts and merge this before it get's completely outdated? Last activity was 2 years ago.

libopenage/audio/dynamic_resource.h
castilma: What I wanted to say is "it's the number of valid entries the array". An entry is a int16_t. 
in dynamic_resource.cpp you see this in the constructor:
```
size{0},
buffer{std::make_unique<int16_t[]>(buffer_size)} {}
```
size is initialized to zero, because there are no values in the beginning.
the buffer_size argument determines the maximum number of values (not bytes) the buffer can contain (if i understood make_unique correctly).

castilma:
So I had to remove one deleted file and adjust some legal headers as well as commit dates. If this passes CI I'm done.

buildsystem/modules/FindOpusfile.cmake
```diff
@@ -33,11 +40,11 @@ find_library(OPUS_LIBRARY
...
-# handle the QUIETLY and REQUIRED arguments and set OPUSFILE_FOUND to TRUE if
+# handle the QUIETLY and REQUIRED arguments and set OPUS_FOUND to TRUE if
 # all listed variables are TRUE
 include(FindPackageHandleStandardArgs)
 find_package_handle_standard_args(Opus DEFAULT_MSG OPUSFILE_LIBRARY OPUS_LIBRARY OPUS_INCLUDE_DIR)
 
 # export the variables
 set(OPUS_LIBRARIES "${OPUSFILE_LIBRARY}" "${OPUS_LIBRARY}")
-set(OPUS_INCLUDE_DIRS "${OPUS_INCLUDE_DIR}" "${OPUS_INCLUDE_DIR}/opus")
+set(OPUS_INCLUDE_DIRS "${OPUSFILE_INCLUDE_DIR}/opus" "${OPUS_INCLUDE_DIR}/opus")
```
simonsan: Why does this change from `${OPUS_INCLUDE_DIR}` as (1) to **(1)/opus** here?

libopenage/engine.h
```diff
@@ -157,13 +157,6 @@ class Engine : public ResizeHandler, public options::OptionNode {
 	 */
 	void end_game();
 
-	/**
-	 * draw the current frames per second number on screen.
-	 * save the current framebuffer to a given png file.
-	 * @param filename the file where the picture will be saved to.
-	 */
-	void save_screenshot(const char* filename);
-
```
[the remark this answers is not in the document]
castilma: I moved this function to be part of the ScreenshotManager Class. See libopenage/screenshot.{h,cpp}

buildsystem/modules/FindOpusfile.cmake
castilma: I had problems building openage with nix. It was because opusfile.h and opus.h were in different packages (namely opusfile and libopus), and nix put those in different directories. So I changed this file to look explicitly for both files. That's why I added `"${OPUSFILE_INCLUDE_DIR}/opus"`

As far as I can remember, I thought "${OPUS_INCLUDE_DIR}" was unnecessary/wrong, because the related headerfiles are in "${OPUS_INCLUDE_DIR}"/opus and are included as <opusfile.h> or <opus.h>.
Fact check: 
```
$ ag 'opus[a-z_-]*\.h'
buildsystem/modules/FindOpusfile.cmake
10:#  OPUS_INCLUDE_DIRS   - where to find opus/opus.h, opus/opusfile.h, etc
16:	NAMES opus/opusfile.h
23:	NAMES opus/opus.h
[...]

libopenage/audio/opus_loading.h
7:#include <opus/opusfile.h>

libopenage/audio/opus_in_memory_loader.cpp
[...]
5:#include <opusfile.h>
[...]

libopenage/audio/opus_dynamic_loader.cpp
[...]
5:#include <opusfile.h>

openage/convert/opus/opus.pxd
3:cdef extern from "opus/opus.h":
17:cdef extern from "opus/opus_defines.h":
```
So we have a mixed inclusion style. opus_in_memory_loader.cpp and opus_dynamic_loader.cpp are different from the others. As we are looking in FindOpusfile.cmake for opus/opusfile.h, we should include it everywhere is that.
I propose changing this line to
```
set(OPUS_INCLUDE_DIRS "${OPUSFILE_INCLUDE_DIR}" "${OPUS_INCLUDE_DIR}")
```

TheJJ: ah, good catch.

I'd say we should unify the layout to only use `#include <opus/opusfile.h>` and `<opus/opus.h>`, because that's how we did it with all other software packages (hopefully).

For example we used:
```
#include <gperftools/profiler.h>
#include <eigen3/Eigen/Dense>
#include <epoxy/gl.h>
#include <SDL2/SDL.h>
#include <SDL2/SDL_image.h>
...
```

`sdl2` and `sdl2_image` are also separate packages, just like `opus` and `opusfile`.

`opusfile` and `opus` should provide us with their includedirs, and we should add both to the search path in such a way we can write `#include <opus/opusfile.h>` in cpp/h files. Agreed?

castilma: Yes.
But it doesn't work:
opusfile.h includes <opus_multistream.h>. [Explanation](https://gitlab.xiph.org/xiph/opusfile/commit/9bc622bbe1822584eb8d774700a11b592f462b49)
So looks like we need 
```
set(OPUS_INCLUDE_DIRS "${OPUSFILE_INCLUDE_DIR}" "${OPUS_INCLUDE_DIR}" "${OPUS_INCLUDE_DIR}/opus")
```
Did that

TheJJ: hm weird, why do they assume this? :)
anyway, thanks for the fix, that should do it.

libopenage/audio/dynamic_resource.h
TheJJ: yea that's true, and your (updated) code comment says the right thing :)

libopenage/engine.h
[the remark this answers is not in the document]
TheJJ: kay, thanks for the investivations :D

simonsan:
thank you! <3

buildsystem/modules/FindOpusfile.cmake
castilma: I guess pkg-config does that, so that you can have several versions installed in different subdirs under include/.

PR #833, opened 2017-05-08
doc: Integrate Tutorials 

heinezen:
We've been given permission by Jan dc to integrate his [3D modelling tutorials](http://aok.heavengames.com/cgi-bin/forums/display.cgi?action=st&fn=9&tn=44487&st=recent&f=9,44487,0,10) into our repository. They need to be converted into Markdown first and we should include a reference to the kind author. Overall this will bring us a step further to create our own asset pack.

- [x] Introduction
- [x] [Part 1 - Blender Setup and Rendering](http://aok.heavengames.com/cgi-bin/forums/display.cgi?action=ct&f=26,42291,,10)

Next step would be rewriting the text so that it fits to our other docs.


TheJJ:
I think we should de-personalize it though, not just have a 1:1 copy :smile:

heinezen:
I think it's fine if we leave it personalized and include a disclaimer that this is originally from Jan and not from us. After all it is his work that we shouldn't change without permission. At least we should ask him beforehand about changes before we butcher his texts :D

TheJJ:
Kay, I interprete his "Btw of course you can integrate my tutorials." as we can also change it, but we can of course ask him explicitly. If we bring merge the tutorials, I would very much reword them quite a bit and add e.g. advanced humor(tm). The "I" perspective of the guide won't match the other project docs otherwise.

heinezen:
Better be safe than sorry. Can you ask him for the `.blend` file too? I don't think he uploaded it anywhere.

TheJJ:
Hmmmm, when I see all those binaries I'm thinking of a having a new repo for such things so that the good and sane main code repo doesn't get binary-blobbed. What do you think?

Vtec234:
@TheJJ: what about redoing openage-pr to contain modding tutorials and related docs (like this one)? That kind of thing is usually available on a project's website.

citron0xa9:
Another option would be using [git large file storage](https://git-lfs.github.com/)

heinezen:
I agree with @Vtec234 that we could have a new repository that contains all the stuff that is not directly related to the codebase. Something like openage-pr or openage-modding. Or we create a `/doc`-folder in the openage-data repo and put tutorials there.

TheJJ:
I'd go either for `openage-pr` or a new `openage-modding`. The `data` repo is for the resulting work only, and `git lfs` depends on github I think.
If I had to choose, I'd go for `openage-modding`, with markdown files containing all sorts of guides.

heinezen:
Let's do it ;)

TheJJ:
License? It should be some copyleft variant.
* https://en.wikipedia.org/wiki/GNU_Free_Documentation_License
* Creative commons share-alike?

heinezen:
CC-BY-SA would be a good idea since it's the most common one and well adapted. I don't know much about the GNU FDL other than that there were some controversies.

TheJJ:
Repo is initialized!

heinezen:
This PR has moved [here](https://github.com/SFTtech/openage-modding/pull/2).

PR #836, opened 2017-05-11
Added mouse controlled window movement to supplement arrow keys

cdw33:
Added functionality to move game window by moving the cursor to its corresponding edge.


libopenage/input/input_manager.h
```diff
@@ -138,6 +138,25 @@ class InputManager : public InputHandler {
...
 
 	/**
+	 * Edges enum for use with is_mouse_at_edge
+	 */
+	enum Edges {
+		EDGE_LEFT,
+		EDGE_RIGHT,
+		EDGE_UP,
+		EDGE_DOWN
+	};
```
TheJJ: please use an `enum class` :) That would be
``` cpp
enum class Edge {
    LEFT,
    RIGHT,
    UP,
    DOWN
};
```

libopenage/engine.cpp
```diff
@@ -378,19 +378,19 @@ void Engine::loop() {
...
-				input.is_mouse_at_edge(input.EDGE_RIGHT, coord.window_size.x)) {
+				input.is_mouse_at_edge(input.Edge::RIGHT, coord.window_size.x)) {
 				mov_x = cam_movement_speed_keyboard;
 			}
 			if (input.is_down(SDLK_DOWN) ||
-				input.is_mouse_at_edge(input.EDGE_DOWN, coord.window_size.y)) {
+				input.is_mouse_at_edge(input.Edge::DOWN, coord.window_size.y)) {
 				mov_y = cam_movement_speed_keyboard;
 			}
 			if (input.is_down(SDLK_UP) ||
```
Tomatower: Is it possible to store the keys (```SDLK_LEFT```) in some kind of map, that the input can be stored in a input configuration?

TheJJ: That would mean those events must be tied into the cvar-manager. If you want, @cdw33 you can try, but that's also possible as a later improvement.

copying.md
```diff
@@ -98,6 +98,7 @@ _the openage authors_ are:
 | Erik Griese                 | citron0xa9                  | erik dawt griese à yahoo dawt de                  |
 | Alex Birch                  | Birch-san                   | vengeance dawt m dawt x+openage à gmail dawt com  |
 | Michal Jarzabek             | stiopa                      | stiopa à gmail dawt com                           |
+| Christopher Wilson          | cdw33                       | cdw33 à zips dawt uakron dawt edu                           |
```
TheJJ: the last `|` is not aligned :P

libopenage/input/input_manager.cpp
```diff
@@ -257,6 +257,28 @@ void InputManager::set_relative(bool mode) {
 	}
 }
 
+bool InputManager::is_mouse_at_edge(Edge edge, int window_size ) {
+	int x, y;
+	SDL_GetMouseState(&x, &y);
+
+	//how mant pixels in from edge to consider "edge"
+	int edge_offset = 0;
```
TheJJ: Isn't `0` too less? How does the original game handle it?

cdw33:
Will resubmit

TheJJ:
You don't need to resubmit a pull request. If you follow our rebasing guideline then you can update the commits in a pullrequest afterwards :)
I guess you didn't do a force-push, which is the only way how to "forget" the current commits in a pull and replace them with the new ones.

cdw33:
Yeah obviously had some issues with the rebase. Got frustrated and just resubmitted. You might be right, I don't think I did a force-push..

PR #838, opened 2017-05-13
Added mouse controlled window movement to supplement arrow keys 

cdw33:
Resubmitting #836 with fixes. Updated edge offset to 10px to match AoE mechanics.


TheJJ:
Continued in #843. Thank you!

PR #839, opened 2017-05-13
buildsystem: single point of version specification

### LINKED ISSUE #381: Trigger reconfigure when git commit hash changes

TheJJ:
This embeds the version into Python, so version is determined at runtime, not compiletime.
I'm unsure about the git-version determination, before it was `git-describe`, now it's the last commit epoch, which doesn't seem suitable.

Taken from #560.
Fixes #381.


Tomatower:
Is it really the best to have the version of the cpp-binaries in the script-interface in python? 
This opens up a whole now fuckitup-vector for people manually changing the version and loading savefiles/networking.

Defining it only at one single point is a really good idea - but what is wrong at loading it from a version-file into a config.h?

TheJJ:
That's true, and the current mechanism does all this at configure-time, not runtime.
I wasn't really sure if this commit is an actual improvement, the current approach seems better though.

PR #841, opened 2017-05-15
Constmap fix

Vtec234:
This fixes type errors in the `constexpr ConstMap` structure.

### DIFF
--- a/libopenage/datastructure/constexpr_map.h
+++ b/libopenage/datastructure/constexpr_map.h
@@ -1,9 +1,12 @@
-// Copyright 2015-2015 the openage authors. See copying.md for legal info.
+// Copyright 2015-2017 the openage authors. See copying.md for legal info.
 
 #pragma once
 
 #include <functional>
 #include <utility>
+
+#include "../error/error.h"
+
 
 namespace openage {
 namespace datastructure {
@@ -61,21 +64,21 @@ public:
 	 * Return the number of entries in this map.
 	 */
 	constexpr int inline size() const {
-		return this->has_duplicates(), this->position();
+		return this->verify_no_duplicates(), pos;
 	}
 
 	/**
 	 * Tests if the given key is in this map.
 	 */
 	constexpr bool inline contains(const K &key) const {
-		return this->has_duplicates(), this->has_entry(key);
+		return this->verify_no_duplicates(), this->has_entry(key);
 	}
 
 	/**
 	 * Return the stored value for the given key.
 	 */
 	constexpr const inline V &get(const K &key) const {
-		return this->has_duplicates(), this->must_contain(key), this->fetch(key);
+		return this->verify_no_duplicates(), this->fetch(key);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


ChipmunkV:
Why it's even there if it's not used anywhere?

Vtec234:
I use it in the renderer rewrite.

libopenage/datastructure/constexpr_map.h
```diff
@@ -268,15 +238,13 @@ class ConstMap<K, V, 0> {
 	 * Fetching a value from the end element aborts as it has no entry.
 	 */
 	constexpr const inline V &fetch(const K &) const {
-		return this->missing_key();
+		// A key is not in the map.
+		// This function should never be called. If it is called,
+		// you requested a nonexisting key.
+		return *static_cast<const V*>(nullptr);
```
citron0xa9: is this so that a compile error is generated?

Vtec234: Yeah, if this function gets called the key is not there and should error at compile time.

TheJJ: We can't use `static_assert` because of some reason I no longer remember. If that reason is nonexistent, it would be even better than remembering it.

Vtec234: It has to have a return type and static_assert returns void.

TheJJ: Hm and to have both?

Tomatower: what about ```std::static_assert(false, "You did something terribly wrong");```?

http://en.cppreference.com/w/cpp/language/static_assert

Vtec234: It also always fails :P static_assert is always evaluated, while *nullptr only if the branch is called

TheJJ: Aaaah, that was the thing :D Question is: can we do better? Not having a proper compiletime error message seems kinda annoying...

Tomatower: This (second answer) might be usable here

https://stackoverflow.com/questions/37440370/is-it-possible-to-check-if-const-value-is-known-at-compile-time

citron0xa9: one possibility is also doing this:
```cpp
template <typename T>
class FalseTemplate : public std::false_type {};

template <typename T>
T* function_to_fail() {
  static_assert<FalseTemplateType<T>::value, "MSG");
  return nullptr;
}
```

Vtec234: That won't work either, because the template is instantiated on use, unless you work out some type-level `get` that only instantiates `fetch` with the key that we're trying to find.. But it's pointless unless you're interested in template hackery, and I realized a constexpr map is not really what we need, since I use it at runtime anyway, so I think I'll revert to using switch statements.

TheJJ: If we don't need it then I'd say just remove it. That nullptr-crashing-hackery really shouldn't be remain.

Vtec234: Um, I just realized that we can just throw an exception. Nice error at runtime and "throw ABC non-constant expression" erro at compile time, understandable enough.

TheJJ: .... yeah. Not as good as compiletime though, but that seems rather hard to accomplish :)

Vtec234: Well, it does give a compile time error in constexpr context if a wrong key is requested.

TheJJ:
I'm sorry but you need to rebase for the conflict in `copying.md`.

Vtec234:
Rebased.

PR #845, opened 2017-05-17
Research (aka Technologies)

MaanooAk:
Classes
- **ResearchType** immutable (contains the name, cost, patch ...)
  - will be extented by **NyanResearchType**
- **Research** muttable (is active, is completed ...)
- **ResearchAbility** and **ResearchAction**

Note: Support for repeating research

### DIFF
--- a/libopenage/unit/action.cpp
+++ b/libopenage/unit/action.cpp
@@ -25,7 +25,7 @@ IntervalTimer::IntervalTimer(unsigned int interval, int max_triggers)
 	interval{interval},
 	max_triggers{max_triggers},
 	time_left{interval},
-	triggers{triggers} {
+	triggers{0} {
 }
 
 void IntervalTimer::skip_to_trigger() {
@@ -1323,7 +1323,7 @@ void ProjectileAction::update(unsigned int time) {
 	coord::phys3 new_position = this->entity->location->pos.draw + d_attr.unit_dir * time;
 	if (!this->entity->location->move(new_position)) {
 
-		// TODO implemnt friendly_fire (now friendly_fire is always on), attack_attribute.friendly_fire
+		// TODO implement friendly_fire (now friendly_fire is always on), attack_attribute.friendly_fire
 
 		// find object which was hit
 		auto terrain = this->entity->location->get_terrain();
@@ -1338,7 +1338,7 @@ void ProjectileAction::update(unsigned int time) {
 			}
 		}
 
-		// TODO implemnt area of effect, attack_attribute.area_of_effect
+		// TODO implement area of effect, attack_attribute.area_of_effect
 
 		has_hit = true;
 	}
--- a/libopenage/unit/action.cpp
+++ b/libopenage/unit/action.cpp
@@ -17,16 +17,30 @@ namespace openage {
 
 IntervalTimer::IntervalTimer(unsigned int interval)
 	:
+	IntervalTimer{interval, -1} {
+}
+
+IntervalTimer::IntervalTimer(unsigned int interval, int max_triggers)
+	:
 	interval{interval},
-	time_left{interval} {
+	max_triggers{max_triggers},
+	time_left{interval},
+	triggers{triggers} {
+}
+
+void IntervalTimer::skip_to_trigger() {
+	this->time_left = 0;
 }
 
 bool IntervalTimer::update(unsigned int time) {
-	this->time_left -= time;
-	if (this->time_left > 0) {
+	if (this->triggers == this->max_triggers) {
+		return false;
+	} else if (this->time_left > time) {
+		this->time_left -= time;
 		return false;
 	} else {
-		this->time_left += this->interval;
+		this->time_left += this->interval - time;
+		this->triggers += 1;
 		return true;
 	}
 }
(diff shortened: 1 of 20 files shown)


libopenage/unit/research.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+void Research::apply() {
+	// apply the patch
+	this->type->apply();
+
+	// perform category based actions
+	if (type->category() == research_category::age_advance) {
+		type->owner.advance_age();
+
+	} else if (type->category() == research_category::age_advance) {
```
TheJJ: we can do the research type-specific directly with nyan by testing the patch hierarchy:
```
nyan::Object research = ...;  // the research nyan object that now shall apply
// apply the research patch
nyan::Transaction tx = database.transaction();
for (auto &patch : research.get<nyan::Patch>("patches")) {
    tx.add(research);
}
tx.apply();

if (research.extends("openage.research.AgeAdvance")) {
    // do things for age advancements that nyan didn't patch (i don't know what that would be, maybe gui updates?)
}
```

MaanooAk: (I'm back)

EDIT: This was actually a typo: `advance_age` is handled at the first if, this was supposed to be `generic`, thinks that couldn't be handled just with nyan (eg. research that allows to see the enemy line of sight). 
However my following explanation for the `advance_age` still stands

The code you posted will be 
inside NyanResearchType::apply -> 
which ovverides ResearchType::apply -> 
which is called inside Reaseach::apply -> 
which is called from the ReaseachAction::update

Notes:
  ResearchType immutable (contains the name, cost, patch ...)
  Reaseach muttable (is active, is completed ...)

Based on that when the gui will have to draw the progress bar at the top of the screen it need only `action.research.get_categroy() == advance_age` and `action.timer.get_progress()`

No need for the gui to dive into the nyan hierarchy...

libopenage/unit/action.h
```diff
@@ -8,13 +8,77 @@
...
+
+	unsigned int interval;
+
+	int max_triggers;
+
+	unsigned int time_left;
+
+	int triggers;
+
+};
```
TheJJ: I think this should go into a separate file. All those monster files deserve to be split up, actually :)

This timer triggering should be synched with @Tomatower's event triggers, where such a timer would be registered and you'd get a function call for the right time. Might be a very good start for integrating into the event-architecture.

libopenage/unit/attribute.h
```diff
@@ -284,13 +284,18 @@ template<> class Attribute<attr_type::attack>: public SharedAttributeContainer {
 		:
 		Attribute{type, r, h, {{4, d}}} {}
 
-	Attribute(UnitType *type, coord::phys_t r, coord::phys_t h, typeamount_map d)
+	Attribute(UnitType *type, coord::phys_t r, coord::phys_t h, typeamount_map d,
+		      coord::phys_t min_range = 0, bool friendly_fire = false,
+			  coord::phys_t area_of_effect = 0)
```
TheJJ: indent, and default value assignments don't have spaces around the `=` please

libopenage/unit/research.cpp
TheJJ: Welcome back :)
Ah now I get it. Yes, we need to implement some research types in the engine, e.g. the line of sight. This would be done by specifying an object layout in nyan, which is then used by the c++-part to evaluate the data changes, e.g. the line of sight or whatever is needed.

The call hierarchy is correct I'd say. It would be a nice abstraction over the nyan backend.
I also like the difference between `Research` and `ResearchType` (i.e. ongoing vs its type).

To make the gui nyan-independent should be a high goal :D 

This wrapping of nyan-types is a great idea to write the interface to nyan: Then we'd have C++-classes that do the intended access to nyan objects (=> they're tightly coupled) and the rest-code can just use the C++-interface. The way it should be done.

libopenage/unit/action.h
MaanooAk: @Tomatower code is far away and a huge changes will have to be made in order to be able to move into curves (I will write more at his pr)

MaanooAk: @TheJJ I'm not making a new file because the timer is a stepping stone to transition better to Events or Curves (I prefer Curves, more at #744)
(no problem if you insist)

Tomatower: The events are actually designed for that, and the only thing that is missing is a nice Api in the game engine - hopefully without major changes to the existing codebase - so inspire me for the best possbile api 😃

libopenage/unit/action.cpp
```diff
@@ -10,10 +10,57 @@
...
+	:
+	IntervalTimer{interval, -1} {
+}
+
+IntervalTimer::IntervalTimer(unsigned int interval, int max_triggers)
+	:
+	interval{interval},
+	max_triggers{max_triggers},
+	time_left{interval},
+	triggers{triggers} {
```
TheJJ: where does `triggers` come from? currently its the uninitialized member.

MaanooAk: Why can this even compile? It must obviously be zero

Issue #826, opened 2017-05-05
Add python utility module for quick and easy profiling


TheJJ:
The "usual" code for profiling some Python call is:

``` python
import cProfile, pstats, io
pr = cProfile.Profile()
pr.enable()

# call here what should be profiled

pr.disable()
s = io.StringIO()
sortby = 'cumulative'
ps = pstats.Stats(pr, stream=s).sort_stats(sortby)
ps.print_stats()
print(s.getvalue())
```

Then, you'll see a nice table of what function took how long.
We should add a utility module in `openage/util/profiler.py` so one can directly do

``` python
from openage.util.profile import Profiler

profiler = Profiler()
with profiler:
    # do something we wanna analyze

print(profiler.report(sortby="cumulative"))
```

PR #854, opened 2017-06-06
[WIP]Python profiling utility

ohn0:
This is for #826. It adds a profiling class that profiles all the calls made with a 'with' statement and generates a report to a file or as a string.

 I'm not sure whether I should be adding more functionality to it.

### DIFF
--- a/copying.md
+++ b/copying.md
@@ -101,6 +101,7 @@ _the openage authors_ are:
 | Christopher Wilson          | cdw33                       | cdw33 à zips dawt uakron dawt edu                 |
 | Wojciech Nawrocki           | Vtec234                     | wjnawrocki à protonmail dawt com                  |
 | Folkert van Verseveld       | methos, medicijnman         | folkert dawt van dawt verseveld à gmail dawt com  |
+| Neel Patel                  | ohn0                        | silverskinx à gmail dawt com                      |
 
 If you're a first-time committer, add yourself to the above list. This is not
 just for legal reasons, but also to keep an overview of all those nicknames.
--- a/openage/util/CMakeLists.txt
+++ b/openage/util/CMakeLists.txt
@@ -7,6 +7,7 @@ add_py_modules(
 	fsprinting.py
 	iterators.py
 	math.py
+	profiler.py
 	strings.py
 	struct.py
 	system.py
--- a/openage/util/profiler.py
+++ b/openage/util/profiler.py
@@ -1,45 +1,44 @@
 import cProfile, pstats, io
 class Profiler:
-	'''
-		A class for quick and easy profiling.
-		Usage:
-			p = Profiler()
-			with p:
-				#call methods that need to be profiled
-			p.report()
-			p.close_stream()
-	'''
-	
+    '''
+        A class for quick and easy profiling.
+        Usage:
+            p = Profiler()
+            with p:
+                #Call methods that need to be profiled.
+            p.report()
+    '''
+    
     profile = None
     profile_stats = None
     profile_stream = None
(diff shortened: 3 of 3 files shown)


copying.md
```diff
@@ -101,7 +101,7 @@ _the openage authors_ are:
 | Christopher Wilson          | cdw33                       | cdw33 à zips dawt uakron dawt edu                 |
 | Wojciech Nawrocki           | Vtec234                     | wjnawrocki à protonmail dawt com                  |
 | Folkert van Verseveld       | methos, medicijnman         | folkert dawt van dawt verseveld à gmail dawt com  |
-
+| Neel Patel                  | ohn0                        | neel1093 à hotmail dawt com
```
TheJJ: please leave the empty line in there, and the last `|` is missing.

openage/util/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ add_py_modules(
 	fsprinting.py
 	iterators.py
 	math.py
+    profiler.py
```
TheJJ: indent with one tab pls

openage/util/profiler.py
```diff
@@ -0,0 +1,50 @@
+# Copyright 2015-2017 the openage authors. See copying.md for legal info.
```
TheJJ: copyright is 2017-2017, unless you started in 2015 already :)

openage/util/profiler.py
```diff
@@ -0,0 +1,50 @@
...
+
+"""
+A Profiling Utility.
```
TheJJ: `Profiling utilities`?

openage/util/profiler.py
```diff
@@ -0,0 +1,50 @@
...
+"""
+
+import cProfile
+import pstats
+import io
+
+
+class Profiler:
+    """A class for quick and easy profiling.
+        Usage:
```
TheJJ: our python docstring style is:
``` python
"""
A class for quick and easy profiling.
Usage:
    p = Profiler()
    with p:
        # Call methods that need to be profiled.
        ...
    p.report()
"""
```

openage/util/profiler.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def __init__(self, oStream=None):
+        # oStream can be a file if the profile results want to be saved.
+        self.profile = cProfile.Profile()
+        self.profile_stream = oStream
+
+    def __enter__(self):
+        # begins profiling
+        self.profile.clear()
+        self.profile.enable()
```
TheJJ: I think it would be cool if the enable and disable functions are not only exposed by the context manager (`with`) but also with `.enable()` and `.disable()` methods provided by the `Profiler` class. That way one could, sometimes, avoid indenting larger blocks. What do you think?

openage/util/profiler.py
```diff
@@ -0,0 +1,50 @@
...
+    profile_stream = None
+
+    def __init__(self, oStream=None):
+        # oStream can be a file if the profile results want to be saved.
+        self.profile = cProfile.Profile()
+        self.profile_stream = oStream
+
+    def __enter__(self):
+        # begins profiling
+        self.profile.clear()
```
TheJJ: Does the `profile_stream` also need clearing then?

openage/util/profiler.py
ohn0: I totally didn't copy paste the tag from one of the other .py files.

openage/util/profiler.py
ohn0: Sounds good, I changed it.

TheJJ:
I did the squash in #859. Thank you :)

PR #856, opened 2017-06-15
Added line to media convert page

Haroldoa:
It was not clear to me that upon restarting steam it would download. I updated the instructions so that nobody else would need to click the link at the top just to figure that out.


TheJJ:
As I assume you created the changes with the github-editor, I squashed your commits and made another pull-request in #858.

PR #860, opened 2017-06-22
fix units getting stuck when dropsite isnt found

Jon0:
Gatherers stop responding to commands after collecting 10 resources and not having a dropsite.


Jon0:
Ok, its rebased now.

TheJJ:
awesome!
welcome back :smile:

PR #862, opened 2017-06-23
units can get removed before on_complete runs

Jon0:
Minor correction to last pr. Using the pointer to target unit in on_completion() is not a good idea, they may already be removed. Instead pass id of the type.

### DIFF
--- a/libopenage/unit/action.cpp
+++ b/libopenage/unit/action.cpp
@@ -236,7 +236,7 @@ void TargetAction::on_completion() {
 	// retask units on nearby objects
 	// such as gathers targeting a new resource
 	// when the current target expires
-	this->on_completion_in_range(this->target.get());
+	this->on_completion_in_range(this->target_type_id);
 }
 
 bool TargetAction::completed() const {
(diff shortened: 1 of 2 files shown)


libopenage/unit/action.cpp
```diff
@@ -1036,13 +1036,13 @@ void GatherAction::update_in_range(unsigned int time, Unit *targeted_resource) {
...
 
-void GatherAction::on_completion_in_range(Unit *) {
+void GatherAction::on_completion_in_range(int target_type) {
 
 	// find a different target with same type
 	TerrainObject *new_target = nullptr;
 	new_target = find_near(*this->entity->location,
-		[this](const TerrainObject &obj) {
-			return obj.unit.unit_type->id() == this->get_target_type_id() &&
+		[this, target_type](const TerrainObject &obj) {
```
TheJJ: then you don't need to capture `this`.

Jon0:
I have heard attack commands prefer units of the same type as the next target, not sure if this is true. I do know gatherers will not auto re-task between some different types of food sources (sheep, hunt), but will between others (fish, forage). The type id can be used to lookup object class and resource types which should what is needed.

TheJJ:
btw how'd you like our "new" kevin-ci? :D

PR #865, opened 2017-06-27
save alpha-channel in screenshots

castilma:
the memcpy takes only 6 ms, the manual copy took 14 ms. `IMG_SavePNG` unfortunately still takes about a second. 

I adjusted the imports to match our style.

### DIFF
--- a/libopenage/screenshot.cpp
+++ b/libopenage/screenshot.cpp
@@ -1,18 +1,18 @@
 // Copyright 2014-2017 the openage authors. See copying.md for legal info.
 
 #include "screenshot.h"
-#include "util/strings.h"
 
+#include <ctime>
+#include <epoxy/gl.h>
 #include <math.h>
+#include <memory>
 #include <stdlib.h>
 #include <SDL2/SDL.h>
 #include <SDL2/SDL_image.h>
 
-#include <epoxy/gl.h>
-
+#include "util/strings.h"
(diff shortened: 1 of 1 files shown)


TheJJ:
Nice. The sdl-savepng could also be replaced by libpng directly, maybe it's faster then. And I guess adjusting the compression ratio of the png encoding could speed things up, but make the screenshots larger.

lisacvuk:
hey would saving bitmaps be faster than saving PNGs? Without transparency, of course.

TheJJ:
As they're not compressed, it would be faster I think. But compressing them would save space, and everyone likes to not have a 47.5 MiB file (`(1920*1080*24) / (1024 ** 2)`) for a fullhd-screenshot ;)

castilma:
those are Mib, not mebi_bytes_.  it's 'just' 5.93 MiB.
I should note, that the screenshots I took did only contain pixel with the alpha channel set to 0xFF. not saving the alpha channel would be a good idea, but after a quick web search I did not find a way to pass options to IMG_SavePNG().

TheJJ:
Right, mybad :)

castilma:
what about starting a thread right after copying the current frame just to encode the picture?

TheJJ:
yup, that would work, good idea. i'd just dispatch it to our job manager.

PR #582, opened 2016-07-03
context_extraction win32 patch

YellowOnion:
Untested as full program doesn't compile, but would hate for this to go to waste.

Issue #480, opened 2016-01-15
Missing util/compiler implementation for WIN32


franciscod:
The `dlfcn` headers are provided by this package https://github.com/dlfcn-win32/dlfcn-win32 that wraps winapi calls into the posix api.

However, `dlfcn-win32` doesn't provide dladdr (because it is not posix, more info here https://github.com/dlfcn-win32/dlfcn-win32/issues/21) so we should work around it on WIN32.

franciscod:
Is this crucial for the pyinterface? @mic-e @TheJJ

franciscod:
also related to #479, there's LLVM's libunwind with a [mingw-w64 port](https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-libunwind-svn/PKGBUILD)

TheJJ:
So the `dladdr` is missing, right? But it must be possible on windows to somehow do the symbol lookup, so we should simply add that, or am i wrong?

franciscod:
Yeah, `dladdr` and some data structures related to the internals of `dlfcn`.

It should be possible!

tusharpm:
Closed in #982.

Issue #479, opened 2016-01-15
Missing StackAnalyzer implementation for WIN32


franciscod:
Currently, if the `WITH_BACKTRACE` flag is set by CMake, the modern `backtrace.h` is used. Else, the GNU's `execinfo.h` is tried.

When building on MSYS2, it tries to include `execinfo.h` and fails, because `execinfo` isn't available for windows.

[Here](http://spin.atomicobject.com/2013/01/13/exceptions-stack-traces-c/) is some example code for catching exceptions and getting stack traces on windows platforms.

This should be implemented in `libopenage/error/stackanalyzer.cpp`.
##

TheJJ:
I think it should be placed in `stackanalyzer.cpp`, that crossplattform subfolder is getting more and more suspicious to me..

tusharpm:
Closed in #982.

PR #867, opened 2017-07-02
Windows native support using MSVC - Visual Studio 2017

### LINKED ISSUE #508: Visual Studio 2015 build
Hi,

Continuing the discussion from IRC, making a note here. I have been trying to generate the build solution using cmake here: https://github.com/SFTtech/openage/compare/master...synth2014:master

I was able to generate the solution file. It would be great if some else could give it a try too. 

There were a number of projects generated in the VS solution: http://s12.postimg.org/4tz4oiivh/projects.png

I am now trying to get rid of the 800+ errors that I get with VS 2015 community edition in libopenage: http://pastebin.com/sn5bXv4V

The changes work in Linux and the makefile, and the build completes successfully.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/31058051-visual-studio-2015-build?utm_campaign=plugin&utm_content=tracker%2F371436&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F371436&utm_medium=issues&utm_source=github).
</bountysource-plugin>

tusharpm:
A step towards #9.
Fixes #508 with a modification - VS2017 instead of VS2015.
Supersedes #582 (credits to @YellowOnion).

<del>
## The procedure (this is here for my recollection later)
- Get the dependencies using `vcpkg`.
- Build/install the missing dependency `opusfile` manually.
- Copy `dirent.h` from `vcpkg`'s downloads directory - it's named `fontconfig-dirent.h`.
- Build using CMake - the usual workflow.
- Install the `DejaVu Serif Book` font and copy the configuration to the fontconfig directory.
- Modify the environment variables required to make the program run (hacks required before proper packaging).
</del>

## A summary of the changes
- A majority of the changes are for selectively compiling some code using preprocessor macros `_MSC_VER` or `_WIN32`. The former for MSVC specific code, and the latter if it is applicable to other compilers on Windows also.
- Introduction of ~~`LIBOPENAGE_API`~~`OAAPI` for decorating the DLL's entry points for Windows. `__declspec(dllexport)` when compiling the DLL's sources, `__declspec(dllimport)` for the code including the header files only - the API user.
  - Added an introduction in `doc/code/pyinterface.md`.
- ~~**[Disrupting change]** In `libopenage/error/error.h`, the variadic macro `ENSURE` has been changed to a two-argument macro. I left a `TODO` - kindly respond with the approach to choose here.~~
- Python changes are mostly for path handling - the native path separator for Windows is a `'\\'` (weird, agreed).
- CMake changes include path handling, finding dependent libraries and packages.
- **[Coordinate system hacks]** The macros CONCAT and LOOP retain their variadic implementation, with the condition that they have a smaller arguments supported limit (3 for both). This should be superseded by the coordinate system overhaul in #662.
- `libopenage/error/stackanalyzer.cpp` contains the skeleton code for #479. Without #480 though, I don't expect it to provide a lot of useful information. The change includes a bug-fix in the GNU execinfo section.
- Included `ciso646` which provides the expansion for `and`, `or`, `not`, `xor`, etc.
- Added a document describing the build instructions: `doc/build_instructions/windows_msvc.md`.

I've left a bunch of `TODO`s and `HACK`s around in the code. I can try to address the critical ones pointed out by the reviewers, the rest can be addressed subsequently.
I've built the project on Windows (using MSVC) only. I'll rely on the CI reports for other platforms (apologies for the cache thrashing). Please let me know if I need to run some test manually.


openage/convert/dataformat/data_definition.py
```diff
@@ -93,7 +93,7 @@ def generate_csv(self, genfile):
                     entry = os.path.relpath(
                         entry,
                         os.path.dirname(self.name_data_file)
-                    )
+                    ).replace(os.path.sep, '/') # HACK: Change to better path handling
```
TheJJ: no fear, this will totally be dropped with `nyan`.

.gitignore
```diff
@@ -22,6 +28,8 @@ __pycache__
 # build system
 /bin
 /.bin
+build
+deps
```
tusharpm: I prefer disallowing such generic names throughout the repository (I found it useful for embedding thirdparty libraries). I can just make the change here.

buildsystem/check_py_file_list.py
```diff
@@ -37,12 +37,12 @@ def main():
...
     for filename in sorted(actual - listed):
         success = False
-        print("file was not listed via add_py_module: " +
+        print("file was not listed via add_py_module: " + filename + " as " +
               os.path.relpath(filename, cwd))
 
     for filename in sorted(listed - actual):
         success = False
-        print("file was listed via add_py_module but does not exist: " +
+        print("file was listed via add_py_module but does not exist: " + filename + " as " +
```
tusharpm: I put them in for my debugging. I left them there because I thought they'd be useful to others too.

buildsystem/cpp.cmake
```diff
@@ -88,7 +89,7 @@ function(finalize_binary target_name output_name type)
 				message(FATAL_ERROR "finalize_binary flag 'allow_no_undefined' is only valid for libraries!")
 			endif()
 
-			if(NOT "${CMAKE_CXX_FLAGS}" MATCHES "-fsanitize")
+			if(NOT "${CMAKE_CXX_FLAGS}" MATCHES "-fsanitize" AND NOT MSVC)
```
tusharpm: I'm confused. Undefined symbols would result in link-time errors.
Can you please elaborate what errors we're looking for?

buildsystem/modules/FindPython.cmake
```diff
@@ -170,19 +170,24 @@ endforeach()
 # test all the found interpreters; break on success.
 foreach(PYTHON ${PYTHON_INTERPRETERS})
 
-	# ask the interpreter for the essential extension-building flags
-	py_get_config_var(INCLUDEPY PYTHON_INCLUDE_DIR)
-	py_get_config_var(LIBDIR PYTHON_LIBRARY_DIR)
-	py_get_lib_name(PYTHON_LIBRARY_NAME)
+	if (PYTHON STREQUAL PYTHON_EXECUTABLE)
+		find_package(PythonLibs REQUIRED)
```
tusharpm: From `find_package(PythonInterp <other arguments>)` around line 73. Afterwards from the cache.
I'll have to tweak some pieces of this file to make the buildsystem happy on all platforms. You can skip reviewing this file for now.

buildsystem/modules/FindPython.cmake
```diff
@@ -170,19 +170,24 @@ endforeach()
...
 	# there's a static_assert that tests the Python version.
 	# that way, we verify the interpreter and the library version.
 	# (the interpreter provided us the library location)
 	try_compile(PYTHON_TEST_RESULT
 		"${CMAKE_BINARY_DIR}"
 		SOURCES "${CMAKE_CURRENT_LIST_DIR}/FindPython_test.cpp"
-		LINK_LIBRARIES "${PYTHON_LIBRARY_NAME}"
-		CMAKE_FLAGS "-DINCLUDE_DIRECTORIES=${PYTHON_INCLUDE_DIR}" "-DLINK_DIRECTORIES=${PYTHON_LIBRARY_DIR}" "-DCMAKE_CXX_STANDARD=14"
+		LINK_LIBRARIES ${PYTHON_LIBRARIES}
+		CMAKE_FLAGS "-DINCLUDE_DIRECTORIES=${PYTHON_INCLUDE_DIRS}" "-DCMAKE_CXX_STANDARD=14"
```
tusharpm: The variable `PYTHON_LIBRARIES` has the full path of the libraries (actually just one library per configuration) that need to be linked. I made (rather, tried to make) the variable consistent for other ways to locate the Python interpreter and libraries. I must've missed something that made kevin fail.

libopenage/util/compiler.h
```diff
@@ -8,18 +8,44 @@
...
+ */
+#if defined(_WIN32)
+	#if defined(libopenage_EXPORTS)
+		#define LIBOPENAGE_API __declspec(dllexport)
+	#else
+		#define LIBOPENAGE_API __declspec(dllimport)
+	#endif /* libopenage_EXPORTS */
+#else
+	#define LIBOPENAGE_API
+#endif
```
tusharpm: AFAIK, this is the only way to sanely* link to a DLL('s import library) with C++ interfaces in VS.
The alternatives while still using a shared library include:
- Restricting to C-style interfaces (like extern "C" foo). Then using `LoadLibrary` and `GetProcAddress` (equivalents of `dlopen` and `dlsym` respectively) for using the DLL at runtime.
- Go completely off-book and implement mangling for C++ identifiers (which is bound to break in the next release of VS).

Another alternative is to change libopenage to a static library - the problem would be it'll be linked to every python extension (.pyd) we build - bloating up the package when we build one.

Usually this macro doesn't scatter as much if the interface is small and consolidated. In our case, the interface with python is spread across the library.

*Sanity is relative. Essentially an alien concept to VS users.

libopenage/util/fds.h
```diff
@@ -20,7 +19,7 @@ class FD {
 	/**
 	 * wraps an existing FD
 	 */
-	FD(int fd=STDOUT_FILENO);
+	FD(int fd);
```
tusharpm: Not really. Just with `STDOUT_FILENO`. Usually, it's defined in `unistd.h` which is just absent for VS. Turned out `#define STDOUT_FILENO 1` was too much to ask for.

buildsystem/cpp.cmake
TheJJ: Undefined symbols are only discovered at link-time for executables, not for libraries like `libopenage`. Those cflags ensure that errors are shown nevertheless. I guess msvc has something that does the same.

tusharpm: Can you confirm or deny if `libopenage` is actually a shared library (possibly named `libopenage.so` or `libopenage.dylib` on other platforms)? If it is a shared library then the linker should complain about any missing symbols. AFAIK VS doesn't have an equivalent of `-undefined dynamic_lookup`, so no worries there. Otherwise, if it is a static library (like `libopenage.a`), I'll have to search in the MSVC docs for a flag which can do that, if one exists.

libopenage/gui/guisys/private/gui_renderer_impl.h
```diff
@@ -7,6 +7,7 @@
 #include <mutex>
 #include <condition_variable>
 
+#include <epoxy/gl.h>
```
ChipmunkV: The GUI code must not depend on epoxy.

buildsystem/cpp.cmake
TheJJ: openage is and has to be a dynamic library so python can load it (`import libopenage`). This can never be done for a static library.

On Linux it did not complain, which is why we added those options. Hmm.

libopenage/gui/guisys/private/gui_subtree_impl.cpp
```diff
@@ -1,8 +1,10 @@
 // Copyright 2015-2017 the openage authors. See copying.md for legal info.
 
 #include "gui_subtree_impl.h"
+#include "gui_renderer_impl.h"
 
 #include <cassert>
+#include <ciso646>
```
ChipmunkV: Never seen this header in the wild.

If it's about the `and` and `or` operators then it's better to replace them with the C++ operators.

libopenage/gui/guisys/public/gui_renderer.h
```diff
@@ -4,11 +4,7 @@
 
 #include <memory>
 
-#ifndef __APPLE__
-#include <GL/gl.h>
-#else
-#include <OpenGL/gl.h>
-#endif
+#include <epoxy/gl.h>
```
ChipmunkV: The GUI code must not depend on epoxy.

libopenage/gui/guisys/private/gui_subtree_impl.cpp
TheJJ: http://en.cppreference.com/w/cpp/language/operator_alternative

> Because in C++ these are built into the language, the C++ version of <iso646.h>, as well as <ciso646>, does not define anything.

-> no need to replace them with symbol operators and no need to include the header

tusharpm: http://en.cppreference.com/w/cpp/header/ciso646
> ... in a conforming implementation this header is empty.
It may still be required in order to have the alternative operator representations in old or non-conformant compilers.

And you would've guessed it already, MSVC is non-conforming at least in this case.

TheJJ: AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA

libopenage/error/stackanalyzer.cpp
```diff
@@ -245,7 +291,8 @@ void StackAnalyzer::get_symbols(std::function<void (const backtrace_symbol *)> c
 	symbol.lineno = 0;
 
 	if (reversed) {
-		for (size_t idx = this->stack_addrs.size(); idx > 0; idx--) {
+		// `for (auto pc : this->stack_addrs | <ranges-ns>::view::reverse)` after Ranges-TS
+		for (size_t idx = this->stack_addrs.size(); idx-- > 0;) {
```
Tomatower: Why do you do the for like that?

libopenage/error/gl_debug.cpp
```diff
@@ -10,7 +10,7 @@ namespace openage {
 namespace error {
 
 namespace {
-void callback(GLenum source, GLenum, GLuint, GLenum, GLsizei, const GLchar *message, const void*) {
+void APIENTRY callback(GLenum source, GLenum, GLuint, GLenum, GLsizei, const GLchar *message, const void*) {
```
Tomatower: `APIENTRY` is not defined in a linux environment

libopenage/shader/program.cpp
```diff
@@ -1,8 +1,7 @@
-// Copyright 2013-2016 the openage authors. See copying.md for legal info.
+// Copyright 2013-2017 the openage authors. See copying.md for legal info.
 
 #include "program.h"
 
-#include <unistd.h>
```
Tomatower: do we prefer to surround it with `#ifndef _MSC_VER`

libopenage/error/stackanalyzer.cpp
tusharpm: The next line does a dereference using `idx`. I was initially planning to use a reverse iterator, but then got a hint from line 181 above to make a smaller change.

libopenage/error/gl_debug.cpp
tusharpm: It must be (I'm assuming) in `epoxy/gl.h`. I was relying on the CI reports to ensure that it is.

libopenage/shader/program.cpp
tusharpm: I used that option only as a last-resort thing. The first (and IMO better) option is to identify and eliminate platform-dependent dependencies like this (technically `unistd.h` is part of the POSIX standard, but VS doesn't provide one). Same as above, if the CI is happy, we can remove this, right?

libopenage/error/gl_debug.cpp
Tomatower: did not knew that. You are right that CI will see such things.

libopenage/shader/program.cpp
Tomatower: I agree with you, to eliminate such dependencies is best.

libopenage/error/error.h
```diff
@@ -152,7 +152,8 @@ inline std::string no_ensuring_message()
 }
 
 // ENSURE(condition, errormessage << variable << etcetc)
-#define ENSURE(...) do if (unlikely(not OPENAGE_ENS_FIRST(__VA_ARGS__))) throw ::openage::error::Error(MSG(err) OPENAGE_ENS_REST(__VA_ARGS__)); while (0)
+// TODO: Switch to variadic template if needed (or drop the variadic behaviour, if acceptable)
```
ChipmunkV: It may or may not be a macro.

Calling it with just one parameter (the 'condition') should produce a compiler warning.

libopenage/gui/guisys/public/gui_renderer.h
```diff
@@ -5,6 +5,10 @@
 #include <memory>
 
 #ifndef __APPLE__
+#ifdef _MSC_VER
```
ChipmunkV: The `libopenage/gui/guisys` deals only with Qt and SDL. It doesn't include anything else from openage or its dependencies like epoxy. The `openage/libopenage/gui/integration` (and `openage/libopenage/gui/integration/private/gui_make_standalone_subtexture.h` in it) is for connecting together Qt and openage, hence it uses epoxy that's used by openage textures.

This #ifdef boilerplate can be probably fixed by including some GL Qt header in `gui_renderer.cpp` file before the `#include "gui_renderer.h"`. Qt basically does the same thing as epoxy.

libopenage/gui/guisys/public/gui_renderer.h
```diff
@@ -5,6 +5,10 @@
...
+#define NOMINMAX
+#include <windows.h> // for WINGDIAPI, APIENTRY
```
ChipmunkV: The `libopenage/gui/guisys` deals only with Qt and SDL. It doesn't include anything else from openage or its dependencies like epoxy. The `openage/libopenage/gui/integration` (and `openage/libopenage/gui/integration/private/gui_make_standalone_subtexture.h` in it) is for connecting together Qt and openage, hence it uses epoxy that's used by openage textures.

This #ifdef boilerplate can be probably fixed by including some GL Qt header in `gui_renderer.cpp` file before the `#include "gui_renderer.h"`. Qt basically does the same thing as epoxy.

libopenage/gui/guisys/public/gui_renderer.h
```diff
@@ -4,11 +4,7 @@
 
 #include <memory>
 
-#ifndef __APPLE__
-#include <GL/gl.h>
-#else
-#include <OpenGL/gl.h>
-#endif
+#include <QtGui/qopengl.h>
```
ChipmunkV: Can't really see the `QtGui/qopengl.h` in the Qt docs. It's probably more like a private header. I remember that I used `#include <QOpenGLFunctions>` somewhere instead.

The other thing is that the `gui_renderer.h` is a pimpl wrapper to isolate the openage code from Qt. So it can't include any Qt stuff. I can't try it right now, but that include should work in the `gui_renderer.cpp`.

tusharpm: [A Qt examples page](http://doc.qt.io/qt-5/qtopengl-hellogl2-logo-h.html) shows that it can be used directly [Sorry, they don't have line numbers; and I don't know the protocol for sharing their code without involving the copyrights/licences]. Given, I used a more explicit full path - which I can change to the shorter one if needed. OTOH, `QOpenGLFunctions` provides some wrappers which aren't required. I'm not well versed with Qt to take a decision either way.

I observed the pimpl idiom used in the gui code (here and in a few other places). For the inclusion of decided-upon file in `gui_renderer.cpp`:
Option 0: Before `gui_renderer.h` directly contradicts [mom's recommendation](https://github.com/SFTtech/openage/blob/d94876028d28f80a1ee7e72846133e6d7fdc9965/doc/code_style/mom.cpp#L17-L18). I'd prefer not to do that.
Option 1: After `gui_renderer.h` results in compilation failure for all units including it (`error C3646: 'render': unknown override specifier` and some others). :x:

Option (try larger change): Change/remove the `GuiRenderer::render` member to make it work with a forward-declared type `GLuint`. I'll need more information to make this work.
Option (none of these): Revert back to the expanded form before my last change.

ChipmunkV: Second best for me is the 'expanded form'.

libopenage/util/compiler.h
```diff
@@ -8,25 +8,55 @@
...
+
+#if defined(_MSC_VER)
+#ifndef HAVE_SSIZE_T
+typedef int ssize_t;
```
TheJJ: Hm I think that is not correct, as `int` is not always the size of `size_t`.

http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.htmlhttp://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html

> The type `ssize_t` shall be capable of storing values at least in the range `[-1, {SSIZE_MAX}]`.

Strictly speaking, we've used `ssize_t` wrong in some places in codebase, so we should fix that someday :)

On my machine `SSIZE_MAX` is defined as `LONG_MAX`, which is `2 ** 63 - 1`, and bigger than `int` (`2 ** 31  - 1`).

tusharpm: I agree. And I'd like nothing more than a (POSIX?) standard file providing this declaration.
For instance, MSVC claims to provide `SSIZE_T` in `BaseTsd.h`. My problem with that, however, is with a disagreeing [re]definition from `pyconfig.h`. So I went with a non-conflicting definition, and wrapped it in `_MSC_VER` preprocessor guards. To be completely honest, I haven't even tried building this code for x64 yet. That would be another error to deal with.

I can check if we can reduce the usages of `ssize_t`. But, we would still need a definition that agrees with Python's headers, otherwise it could lead to other problems in the build.

libopenage/util/compiler.h
```diff
@@ -8,25 +8,55 @@
...
  * the expression is expected to be true (=likely) or false (=unlikely).
  *
  * btw, this implementation was taken from the Linux kernel.
  */
+#if defined(__GNUC__)
 #define likely(x)    __builtin_expect(!!(x), 1)
 #define unlikely(x)  __builtin_expect(!!(x), 0)
+#else
+#define likely(x)   (x)
+#define unlikely(x) (x)
```
tusharpm: I actually took it from a cython generated cpp file:
```
/* Test for GCC > 2.95 */
#if defined(__GNUC__)     && (__GNUC__ > 2 || (__GNUC__ == 2 && (__GNUC_MINOR__ > 95)))
  #define likely(x)   __builtin_expect(!!(x), 1)
  #define unlikely(x) __builtin_expect(!!(x), 0)
#else /* !__GNUC__ or GCC < 2.95 */
  #define likely(x)   (x)
  #define unlikely(x) (x)
#endif /* __GNUC__ */
```
I'm not aware of a branch prediction tuning feature for MSVC, if one exists.

libopenage/util/compiler.h
TheJJ: Ah, the `ssize_t` is 32 bit wide from python because of your 32 bit platform i guess. So if we define it the way python does, it should be fine?

libopenage/util/compiler.h
tusharpm: Apparently it isn't. I don't think there's a large enough convention to follow out there. I used what I knew from experience. I'm certainly open to using `OAAPI` as that strikes a nice balance between a larger and a descriptive name. `OAFUNC`, not so much, because this macro is used for classes and functions alike (actually even for `extern`ed objects, but I don't know what common name all of them can use). If you're okay with `OAAPI`, I can just do a find-replace all.

libopenage/error/stackanalyzer.cpp
```diff
@@ -190,6 +190,52 @@ void StackAnalyzer::get_symbols(std::function<void (const backtrace_symbol *)> c
...
+	std::vector<void *> buffer{64};
+	auto count = RtlCaptureStackBackTrace(base_skip_frames, buffer.size(), buffer.data(), NULL);
+	if (count < buffer.size()) {
+		buffer.resize(count);
+	}
+	this->stack_addrs = std::move(buffer);
+}
+
+// TODO: exact copy of the GNU execinfo counterpart below. Move to common place.
+void StackAnalyzer::get_symbols(std::function<void (const backtrace_symbol *)> cb,
```
tusharpm: I made the changes.

libopenage/util/compiler.cpp
```diff
@@ -14,6 +16,10 @@ namespace util {
 
 
 std::string demangle(const char *symbol) {
+#ifdef _MSC_VER
+	// TODO: demangle names for MSVC;
+	return symbol;
```
tusharpm: There is: [`UnDecorateSymbolName`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms681400(v=vs.85).aspx). However, it has a problem mentioned in the Remarks section: it is single threaded. I can make this work with some synchronization primitives, but I'd prefer that to be in a subsequent PR, so the focus here remains on the build and we can get more opinions before going that way.
Suggestions?

libopenage/util/compiler.cpp
```diff
@@ -33,6 +40,9 @@ std::string addr_to_string(const void *addr) {
 
 
 std::string symbol_name(const void *addr, bool require_exact_addr, bool no_pure_addrs) {
+#ifdef _MSC_VER
+	return no_pure_addrs ? "" : addr_to_string(addr);
```
tusharpm: Similar to above (although, clearly I forgot to add a `TODO` - I'll do that).
There's [`SymFromAddr`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms681323(v=vs.85).aspx), which has the same pitfalls as the one mentioned above.
This problem is actually discussed in #480, and I would like to know their opinions on using this approach.

libopenage/util/compiler.h
tusharpm: Python's headers make it bad to use as an exact copy:
```cpp
/* _W64 is not defined for VC6 or eVC4 */
#ifndef _W64
#define _W64
#endif

/* Define like size_t, omitting the "unsigned" */
#ifdef MS_WIN64
typedef __int64 ssize_t;
#else
typedef _W64 int ssize_t;
#endif
```
Where `MS_WIN64` is defined if another macro (`_WIN64`) is defined. I just made my definition work with the x86 compilation; For x64, I'll probably use the other definition and `_WIN64` directly. Is that okay?

TheJJ: Hm, if there's no direct way, then so be it. But it seems strange that such definitions are necessary. Should be part of the standard library.

PR #877, opened 2017-07-23
doc: Network protocol of AoC 1.0c

heinezen:
**EDIT:** Screw it! Let this PR be a (full) documentation of the network communication in AoC.

Motivation: Revealing otherwise hidden mechanics in the game engine, understanding the implementation of other game mechanics, determining what can lead to an out-of-sync error. Could result in finding a lot of ridiculous bugs.

- [x] General structure
- [x] Protocol headers
- [x] Synchronization
    - [x] In-game Sync 
    - [x] Lobby
- [x] Chat Messaging (already in repo)
- [x] Spoofing and forging messages
- [x] Player-issued commands
    - [x] Unit interaction
        - [x] Primary actions
        - [x] Movement
        - [x] Formations
    - [x] Unit creation
    - [x] Buildings
    - [x] Market
    - [x] Others
        - [x] Tribute
        - [x] Diplomacy
        - [x] Flares
    - [x] AI actions

Things for later PRs:
* AI actions refined
The AI is a pain to test which leads to action `0x1f` being mostly unknown bytes
* Wireshark dissector
who wouldn't want this?
* ~~C structs~~
~~If we or someone else would want to use this protocol, for whatever reason.~~ Done.


happyleavesaoc:
Might be useful -- my `mgz` parsing implementation: https://github.com/happyleavesaoc/aoc-mgz

Here's the actions specifically: https://github.com/happyleavesaoc/aoc-mgz/blob/master/mgz/body/actions.py

It draws from a number of sources, some already attributed in this PR.

heinezen:
@happyleavesaoc hey thanks, i indeed did not discover your work before. I'm pretty much done with the actions, but another source for double-/triple-checking can't hurt :)

happyleavesaoc:
@heinezen you're missing an action, `0x18` multiplayer save. There's also `0x70`, unknown thus far.

heinezen:
@happyleavesaoc Multiplayer save is `0x1b` (for saving through menu and "Save & Exit" after a de-sync). According to the [recanalyst research](https://github.com/goto-bus-stop/recanalyst/blob/b259b954bed5bb936b24f6e7bb647e2a871c6de0/src/Analyzers/BodyAnalyzer.php#L78), `0x18` breaks a formation up.

Interestingly I've never seen `0x70` in the wild and didn't even know it existed. Do you have a link to an example?

happyleavesaoc:
@heinezen 
I was using https://github.com/stefan-kolb/aoc-mgx-format/blob/master/spec/body/actions/18-save.md as  reference for multiplayer save. It appears that file does not have the right action ID. The referenced `mgx` files contain `0x1b` `save and exit` as you suggest. So that is solved.

I looked for a rec with `0x70` and found two. In one of those recs is also unknown `0x0F`. How should I send these to you?

heinezen:
@happyleavesaoc I think the best solution would be, if you could post the actions' hex values for `0x70` and `0x0F` right here in the comments. Looking at the whole `.mgx` file isn't necessary.

(An action is a multiple of 4 byte, if that helps)

goto-bus-stop:
@happyleavesaoc  I'd be interested in seeing those recs too for RecAnalyst, if you zip them up you can drag&drop them into a Github comment and it'll upload them.

happyleavesaoc:
Action `0x0F`:

```
    type = action (total 6)
    op = 1
    length = 12
    action = Container:
        type = 15
        bytes = \x00\x00\x00\x01\x00\x01\x00\x00\x00\x00\x00 (total 11)
```

Action `0x70`:

```
    type = action (total 6)
    op = 1
    length = 16
    action = Container:
        type = 112
        bytes = \x00\x00\x00\x1b\x02\x02\x01\xfd\x13\x00\x00\xab\x15\x00\x00 (total 15)
```

There's also Action `0x20`, "chapter", which isn't listed in your document. It has just one byte, which I think is `player_id`.


Full files: [0x0F and 0x70.zip](https://github.com/SFTtech/openage/files/1377382/0x0F.and.0x70.zip)
.

heinezen:
> There's also Action 0x20, "chapter", which isn't listed in your document. It has just one byte, which I think is player_id.

This document only covers network actions, which is why it's not covered ;)

> 0F 00 00 00 --> just the command? usually there's at least player_id somewhere  
> 01 00 01 00 --> possibly two int16  
> 00 00 00 00 --> there's one or more values here  

> 70 00 00 00 --> again just the command?  
> 1B 02 02 01 --> could be an int32 or two int16  
> FD 13 00 00 --> looks like the id of a unit/building  
> AB 15 00 00 --> same as above  

hmm.. no idea yet what they exactly represent.

goto-bus-stop:
I tried looking for 0x70 in the 1.0c disassembly but I can't figure out where it comes from haha. Maybe they're userpatch only? could be flipping patrol style config or so?

goto-bus-stop:
Asked UP folks, 0x0F and 0x70 are UP-only:

```
16:47 <+scripter16> message 0x0F is used for certain spec messages, I think
16:47 <+scripter16> to notify for spec connections, I think
16:48 <+scripter16> 0x70 might be an MQ train message
16:49 <+scripter16> there may also be 0x35? I think for various AI commands
```

heinezen:
@goto-bus-stop nice, thanks! that explains it.

happyleavesaoc:
@goto-bus-stop Yep, thanks. MQ train makes sense, since very few games are played with it on.

doc/reverse_engineering/networking/13-other.md
```diff
@@ -0,0 +1,351 @@
...
+Always has the value `0x67`.
+
+*:action_type*  
+Represents the specific action that was taken (diplomatic, changing speed or cheating).
+
+Hex Value | Action
+----------|-------
+0x00      | Diplomacy
+0x01      | Change Game Speed
+0x04      | Cheat Response (if cheat is activated for all players, e.g. "aegis")
```
goto-bus-stop: I think this is _only_ used for "aegis". It's called "quick build" internally. The packets should look like `0x67 0x04 0x01` when quick build is enabled, and `0x67 0x04 0x00` when quick build is disabled. (toggles every time you use "aegis").

happyleavesaoc: This agrees with my research.

heinezen: Yes, that's true. Have to update the description of *:source_player_number* to reflect that. This action drove me crazy because of its multi purpose values :D

goto-bus-stop:
If I'm reading the disassembly correctly, there are a few more ways 0x67 can be sent:

 - Changing the Allied Victory setting in the diplomacy tab (0x05)
   ```c
   struct {
     uint8_t action_identifier; // 0x67
     uint8_t action_type; // 0x05
     uint16_t player_id;
     uint16_t status; // 1 or 0 …
     // … zeroes
   }
   ```
 - Researching Treason in a Regicide game (0x0A)
   ```c
   struct {
     uint8_t action_identifier; // 0x67
     uint8_t action_type; // 0x0A
     uint8_t player_id;
     // … zeroes
   }
   ```
 - Queuing (0x0D) and unqueuing (0x0E) farm reseeds in the mill
   ```c
   struct {
     uint8_t action_identifier; // 0x67
     uint8_t action_type; // 0x0D || 0x0E
     uint16_t player_id;
     uint16_t amount; // amount to queue or unqueue
     // … zeroes
   }
   ```

 - AI setting some strategic numbers (0x1B)? need to look into this further. It's not sent for every strategic number change.
   ```c
   struct {
     uint8_t action_identifier; // 0x67
     uint8_t action_type; // 0x1B
     uint16_t player_id;
     uint16_t strategic_number; // ID of the strategic number
     int32_t value; // new value for the sn.
   }
   ```
 - Something with `action_type` 0x09 … can't tell what it's doing yet. Has a `player_id` and a boolean.
   ```c
   struct {
     uint8_t action_identifier; // 0x67
     uint8_t action_type; // 0x09
     uint16_t player_id;
     uint16_t unknown; // 0 or 1
     // … zeroes
   }
   ```

heinezen:
Oh, just when I thought i clicked every button in the game.. sigh.

* can confirm `0x05`, `0x0a`, `0x0d` and `0x0e`
* I just let a game run through with multiple AIs and never saw `0x1b`, but `0x0b` which looks like this:

```
struct {
  uint8_t action_identifier; // 0x67
  uint8_t action_type; // 0x0b
  uint8_t player_number;
  // ... zero
  uint8_t player_number2;
  // … zeroes
  uint8_t policy; // was 0x14 for allies and 0x41 for enemies once at the start of the game
  // ... zeroes
}
```
I guess this is used to set the AI policy how to deal with a certain player? This action also repeats every 13 seconds in multiplayer games, with *policy* being zero. Needs more testing, although the AI is not test-friendly.

* Also never seen `0x09` ever in any network dump. Could be a left-over from AoK or another button I've missed.
* Have to note, that all `0x67` actions use the *player_number* and not *player_id*. Btw: is there a reason why you sometimes use `uint16_t` instead of `uint8_t` for the player number?

goto-bus-stop:
Oops sorry, yes I meant 0x0b not 0x1b. As for uint8 and uint16, its just IDA assigning different types based on byte alignment, the actual value is always a single byte.

> Have to note, that all 0x67 actions use the player_number and not player_id. 

Good note, I often mix them up! 😅

goto-bus-stop:
> Also never seen `0x09` ever in any network dump. Could be a left-over from AoK or another button I've missed.

Yeah, I think 0x09 is unused, it's sent after clicking some button ID but I can't find any buttons that have that ID, and the language strings the action uses also do not exist.

goto-bus-stop:
I think in the chat protocol, the "Chat Window Close Byte" and the confusing part at the end are both uninitialized memory. Likely the Separator 0x32 and the following empty byte too. When the game stores the YN intended recipient flags, it allocates 11 bytes. Then it starts counting player IDs from 1 through 8, and sets YN flags in indices 1 through 8 in the buffer.
For  a message to all players, you end up with a buffer like this:
```
{ ?, Y, Y, Y, Y, Y, Y, Y, Y, ?, ?}
```

Also, when transmitting the message, the game tells the network code to send `text_length + 22` bytes. But, the metadata is less than 22 bytes: 1 byte for the player number, 11 for the intended recipients, 4 for the message length, 1 byte for an unknown bool, and 1 byte 0x00 to end the chat message string. That's a total of 18 bytes, leaving 4 bytes of uninitialized memory at the end. That may explain some of the weird patterns you're seeing there. For example, the uppercased start of the message could come from the cheat code check, which uppercases the message before comparing it to cheat codes.

> The last 4 bytes of data start to look like someone purposefully created a buffer overflow and appended them to the message. But why waste 4 bytes of the message for this?

So… this, but it's probably not on purpose :)

doc/reverse_engineering/networking/13-other.md
```diff
@@ -0,0 +1,351 @@
...
+Always has the value `0x1b`.
+
+*:exit*  
+Represents whether this action is a normal Save (`0x00`) or a Save & Exit (`0x01`).
+
+*:player_id*  
+The ID of the player saving the game.
+
+*:data*  
+A 253 byte long data field which cannot be parsed easily. Contains filename and presumably a checksum.
```
goto-bus-stop: This is only the filename and the checksum. The checksum is 4 bytes at the very end. Anything in between the filename and the checksum is uninitialized memory.

heinezen:
> As for uint8 and uint16, its just IDA assigning different types based on byte alignment, the actual value is always a single byte.

Oh thank god, I though I have to go change everything again :D

> I think in the chat protocol, the "Chat Window Close Byte" and the confusing part at the end are both uninitialized memory. Likely the Separator 0x32 and the following empty byte too. When the game stores the YN intended recipient flags, it allocates 11 bytes. Then it starts counting player IDs from 1 through 8, and sets YN flags in indices 1 through 8 in the buffer.

That's right, I have to rework the whole part sooner or later. Back then, I thought this was an actual structured protocol where everything had a purpose and not a memory dump 😄 

> 1 byte for the player number, 11 for the intended recipients, 4 for the message length, 1 byte for an unknown bool, and 1 byte 0x00 to end the chat message string. That's a total of 18 bytes, leaving 4 bytes of uninitialized memory at the end.

Fun Fact: Only the YN flags and the message are actually interpreted. The game doesn't check if *message_length* actually fits to the message, nor does it check if *player_number* fits to the *sender_id* in the header. Messages are not validated, in contrast to the actions which is the reason why message spoofing works so well. I think UserPatch and HD have changed that because the MITM attack doesn't succeed there anymore.

doc/reverse_engineering/networking/13-other.md
heinezen: Uuuh, so it really is that simple. Fixed it.

TheJJ:
anything left todo here (apart from the kevin messages)?

heinezen:
@TheJJ No, it's finished. All remaining kevin messages are the Markdown double whitespace for line breaks.

PR #886, opened 2017-08-20
doc: Update DRS file format.

goto-bus-stop:
Internally, Age of Empires does something like this when reading files
from a DRS archive:

```c
read_drs_file('bina', 50500);
```

Note how it uses a C char with multiple bytes. This is stored on disk as a
little-endian integer, so the 'a' byte ends up at the lowest address,
and the 'b' byte ends up at the highest address. Hence why it appears
to be reversed.

The "magical byte" is simply a padding space ' ' character or the final 'a' of
the "bina" file type.

### DIFF
--- a/doc/media/drs-files.md
+++ b/doc/media/drs-files.md
@@ -39,27 +39,34 @@ Python format: `Struct("< 40s 4s 12s i i")`
 ### DRS Table info
 
 One `drs_table_info` struct is stored for each file type (Possible file
-extensions are `bin`, `slp`, `wav`).
+extensions are `bina`, `slp `, `wav `).
 
 Length  | Type   | Description            | Example
 --------|--------|------------------------|--------
-1 byte  | uint8  | Related with file type | a, 0x61
-3 bytes | string | Reversed extension     | nib (reversed 'bin')
+4 bytes | string | File extension         | 'anib', ' pls'
 4 bytes | int32  | Table offset           | 112, 0x00000070
 4 bytes | int32  | Number of files        | 71, 0x00000047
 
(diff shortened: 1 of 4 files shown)


doc/media/drs-files.md
```diff
@@ -39,27 +39,34 @@ Python format: `Struct("< 40s 4s 12s i i")`
...
-3 bytes | string | Reversed extension     | nib (reversed 'bin')
+4 bytes | string | File extension         | 'anib', ' pls'
 4 bytes | int32  | Table offset           | 112, 0x00000070
 4 bytes | int32  | Number of files        | 71, 0x00000047
 
 ```cpp
 struct drs_table_info {
-	uint8 file_type;
-	char file_extension[3];
+   char file_extension[4];
```
TheJJ: tab here pls

goto-bus-stop: oops, done

goto-bus-stop:
I'll update `drs.py`. The conversion output will change because `.bin` files will become `.bina` files like is already the case in HD Edition

TheJJ:
As we don't use those files I think we should be fine.

goto-bus-stop:
Probably. I did mark the affected components in `changelog.py` anyway, just in case. I can revert that if you prefer but I don't think it'll hurt anyone.

I ran a full conversion with the new code on AoC 1.0c and it seems good

goto-bus-stop:
Removed changelog entry, not sure if github sends notifications for force pushes

TheJJ:
Not it doesn't :)

PR #887, opened 2017-08-20
Support and document debugging with gdbgui

arialwhite:
I figured out how to debug libopenage with gdbgui. It is a browser-based frontend to gdb built with Python and JavaScript.

gdbgui: https://github.com/cs01/gdbgui

This PR:
- Add a .gdbinit (gdb commands) to debug openage and break at main.cpp
- Document how to debug openage and how to do this with gdbgui

gdbgui3

### DIFF
--- a/copying.md
+++ b/copying.md
@@ -105,6 +105,7 @@ _the openage authors_ are:
 | David Carlier               | devnexen                    | devnexen à gmail dawt com                         |
 | Tushar Maheshwari           | tusharpm                    | tushar27192 à gmail dawt com                      |
 | Piotr Szpetkowski           | piotr-szpetkowski           | piotr dawt szpetkowski à pyquest dawt space       |
+| Julian Guillotel            | arialwhite                  | julian dawt gullotel à gmail dawt com             |
 
 If you're a first-time committer, add yourself to the above list. This is not
 just for legal reasons, but also to keep an overview of all those nicknames.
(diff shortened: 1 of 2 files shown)


arialwhite:
Thanks for the review I agree with you and will do some changes.

TheJJ:
In my setup I can debug cython and python code. My setup is mainly this (I wrote large parts of the article): https://wiki.gentoo.org/wiki/Debugging

So I have (relevant for openage):
```
sys-libs/glibc installsources debugsyms
dev-libs/libuv debugsyms installsources
media-libs/libsdl2 installsources debugsyms
media-libs/libepoxy installsources debugsyms
dev-lang/python installsources debugsyms
media-libs/opusfile installsources debugsyms
```

Most importantly, python is built with debug infos and source files are provided.
Python also installs gdb helper functions like `py-bt` and the like, or `py-print`, which you can use to examine stacktraces with python-vm calls (as described in the DebuggingWithGdb wiki page)

I've never used `cython --gdb` so far, but it seems like this would be rather easy to add to the buildsystem.

My debugging setup uses `gdb` and `voltron` and I just launch the `run` program, set a breakpoint or use our `BREAKPOINT` macro, that's it.

arialwhite:
Thanks for your reply, I rebased my branch.

I was not aware of voltron it looks to be a great tool.

PR #891, opened 2017-08-21
buildsystem: add convenience options for using ccache

TheJJ:
Speeds up some compilations if used.

The cache will be in `~/.cache/ccache` by default.

`ccache -s` will show you statistics. which you can `watch -n1` during builds :)


TheJJ:
Nope, the kevin vms are not persistent. We could, in theory, mount a persistent cache folder in the vm to speed up the build process even more.

PR #893, opened 2017-08-24
Use directly libpng to create screenshots.

castilma:
\<rant\> libpng's doc is ugly! the api isn't really documented. there is only a list of all functions and a walkthrough. but I miss an explanation for every function, what it does, what the arguments are and what is returned. \</rant\>
this uses directly libpng to create the screenshot. it strips the alpha channel and does the encoding in another thread. (well, when I'm finished)

note to self: I need to learn how to use shared_ptr and unique_ptr. somehow I get unreadable build errors, when I try to use them...

### DIFF
--- a/doc/build_instructions/arch_linux.md
+++ b/doc/build_instructions/arch_linux.md
@@ -4,7 +4,7 @@
 
 This command should provide required packages for Arch Linux installation:
 
-`sudo pacman -S --needed python python-pillow python-numpy python-pygments cython libepoxy ttf-dejavu freetype2 fontconfig harfbuzz cmake sdl2 sdl2_image opusfile opus-tools python-pylint qt5-declarative qt5-quickcontrols`
+`sudo pacman -S --needed python python-pillow python-numpy python-pygments cython libepoxy libpng ttf-dejavu freetype2 fontconfig harfbuzz cmake sdl2 sdl2_image opusfile opus-tools python-pylint qt5-declarative qt5-quickcontrols`
 
 If you don't have a compiler installed, you can select between these commands to install it:
  - `sudo pacman -S --needed gcc`
--- a/doc/build_instructions/ubuntu_16.04.md
+++ b/doc/build_instructions/ubuntu_16.04.md
@@ -1,5 +1,5 @@
 # Prerequisite steps for Ubuntu users (Ubuntu 16.04)
 
  - `sudo apt-get update`
- - `sudo apt-get install cmake libfreetype6-dev python3-dev python3-pip libepoxy-dev libsdl2-dev libsdl2-image-dev libopusfile-dev libfontconfig1-dev libharfbuzz-dev opus-tools python3-pil python3-numpy python3-pygments qtdeclarative5-dev qml-module-qtquick-controls`
+ - `sudo apt-get install cmake libfreetype6-dev python3-dev python3-pip libepoxy-dev libsdl2-dev libsdl2-image-dev libopusfile-dev libfontconfig1-dev libharfbuzz-dev libpng-dev opus-tools python3-pil python3-numpy python3-pygments qtdeclarative5-dev qml-module-qtquick-controls`
  - `pip3 install cython`
--- a/doc/building.md
+++ b/doc/building.md
@@ -34,6 +34,7 @@ Dependency list:
       A   python imaging library (PIL) -> pillow
     CR    opengl >=2.1
     CR    libepoxy
+    CR    libpng
      R    dejavu font
     CR    freetype2
     CR    fontconfig
--- a/libopenage/screenshot.cpp
+++ b/libopenage/screenshot.cpp
@@ -2,24 +2,27 @@
 
 #include "screenshot.h"
 
+#include <cstdio>
+#include <cstring>
 #include <ctime>
 #include <epoxy/gl.h>
 #include <math.h>
 #include <memory>
+#include <png.h>
 #include <stdlib.h>
-#include <SDL2/SDL.h>
-#include <SDL2/SDL_image.h>
 
-#include "util/strings.h"
 #include "coord/window.h"
+#include "job/job_manager.h"
 #include "log/log.h"
+#include "util/strings.h"
 
 namespace openage {
 
 
-ScreenshotManager::ScreenshotManager()
+ScreenshotManager::ScreenshotManager(job::JobManager *job_mgr)
 	:
-	count{0} {
+	count{0},
+	job_manager{job_mgr} {
 }
 
 
@@ -45,45 +48,85 @@ std::string ScreenshotManager::gen_next_filename() {
 }
 
 
-void ScreenshotManager::save_screenshot() {
-	std::string filename = this->gen_next_filename();
+void ScreenshotManager::save_screenshot(coord::window size) {
+	coord::pixel_t width = size.x,
+	               height = size.y;
 
-	log::log(MSG(info) << "Saving screenshot to " << filename);
+	std::shared_ptr<uint8_t> pxdata(new uint8_t[4*width*height], std::default_delete<uint8_t[]>());
+	glReadPixels(0, 0, width, height, GL_RGBA, GL_UNSIGNED_BYTE, pxdata.get());
 
-	int32_t rmask, gmask, bmask, amask;
-	rmask = 0x000000FF;
-	gmask = 0x0000FF00;
-	bmask = 0x00FF0000;
-	amask = 0xFF000000;
-
-	SDL_Surface *screen = SDL_CreateRGBSurface(
-	SDL_SWSURFACE,
-	this->window_size.x,
-	this->window_size.y,
-	32,
-	rmask, gmask, bmask, amask);
-
-	size_t pxcount = screen->w * screen->h;
-
-	auto pxdata = std::make_unique<uint32_t[]>(pxcount);
-
-	glReadPixels(0, 0,
-	             this->window_size.x, this->window_size.y,
-	             GL_RGBA, GL_UNSIGNED_BYTE, pxdata.get());
-
-	uint32_t *surface_pxls = (uint32_t *)screen->pixels;
-
-	// we need to invert all pixel rows, but leave column order the same.
-	for (ssize_t row = 0; row < screen->h; row++) {
-		ssize_t irow = screen->h - 1 - row;
-		memcpy(surface_pxls + row * screen->w,
-		       pxdata.get() + irow * screen->w, 4 * screen->w);
-	}
-
-	// call sdl_image for saving the screenshot to png
-	IMG_SavePNG(screen, filename.c_str());
-	SDL_FreeSurface(screen);
+	auto encode_function = [this, pxdata, size] () {
+		return this->encode_png(pxdata, size);
+	};
+	this->job_manager->enqueue<bool>(encode_function);
 }
 
+
+bool ScreenshotManager::encode_png(std::shared_ptr<uint8_t> pxdata,
+                                   coord::window size) {
+	std::FILE *fout = NULL;
+	coord::pixel_t width = size.x,
+	               height = size.y;
+	auto warn_fn = [] (png_structp /*png_ptr*/, png_const_charp message) {
+		log::log(MSG(err) << "Creating screenshot failed: libpng error: " << message);
+	};
+	auto err_fn = [] (png_structp png_ptr, png_const_charp message) {
+		log::log(MSG(err) << "Creating screenshot failed: libpng error: " << message);
+		longjmp(png_jmpbuf(png_ptr), 1);
+	};
+
+	png_structp png_ptr = png_create_write_struct(PNG_LIBPNG_VER_STRING,
+	                                              (png_voidp) NULL,
+	                                              err_fn, warn_fn);
+	if (!png_ptr)
+		return false;
+
+	png_infop info_ptr = png_create_info_struct(png_ptr);
+	if (!info_ptr) {
+		png_destroy_write_struct(&png_ptr, (png_infopp) NULL);
+		return false;
+	}
+
+	if (setjmp(png_jmpbuf(png_ptr))) {
+		std::fclose(fout);
+		png_destroy_write_struct(&png_ptr, &info_ptr);
+		return false;
+	}
+
+	std::string filename = this->gen_next_filename();
+	fout = std::fopen(filename.c_str(), "wb");
+	if (fout == NULL) {
+		png_destroy_write_struct(&png_ptr, &info_ptr);
+		log::log(MSG(err) << "Could not open '"<< filename << "': "
+		         << std::string(strerror(errno)));
+		return false;
+	}
+
+	png_init_io(png_ptr, fout);
+
+	png_set_IHDR(png_ptr, info_ptr, width, height, 8, PNG_COLOR_TYPE_RGB,
+	             PNG_INTERLACE_NONE, PNG_COMPRESSION_TYPE_DEFAULT,
+	             PNG_FILTER_TYPE_DEFAULT);
+
+	// Put image row pointer into info_ptr so that we can use the high-level
+	// write interface.
+	std::vector<png_bytep> row_ptrs;
+
+	// Invert rows.
+	row_ptrs.reserve(height);
+	for (int i = 1; i <= height; i++) {
+		row_ptrs.push_back(pxdata.get() + (height - i) * 4 * width);
+	}
+	png_set_rows(png_ptr, info_ptr, &row_ptrs[0]);
+
+	//TODO: print ingame message.
+	log::log(MSG(info) << "Saving screenshot to '" << filename << "'.");
+
+	png_write_png(png_ptr, info_ptr, PNG_TRANSFORM_STRIP_FILLER_AFTER, NULL);
+
+	std::fclose(fout);
+	png_destroy_write_struct(&png_ptr, &info_ptr);
+	return true;
+}
 
 } // openage
--- a/libopenage/screenshot.h
+++ b/libopenage/screenshot.h
@@ -1,25 +1,34 @@
-// Copyright 2014-2016 the openage authors. See copying.md for legal info.
+// Copyright 2014-2017 the openage authors. See copying.md for legal info.
 
 #pragma once
 
 #include <string>
 #include <ctime>
+#include <memory>
 
-#include <SDL2/SDL.h>
 #include "coord/window.h"
 
 namespace openage {
 
+namespace job {
+class JobManager;
+}
+
 class ScreenshotManager {
 public:
-	ScreenshotManager();
+	/**
+	 * Initializes the screenshot manager with the given job manager.
+	 */
+	ScreenshotManager(job::JobManager* job_mgr);
+
 	~ScreenshotManager();
 
-	/** to be called to save a screenshot */
-	void save_screenshot();
+	/** To be called to save a screenshot. */
+	void save_screenshot(coord::window size);
 
-	/** size of the game window, in coord_sdl */
-	coord::window window_size;
+	/** To be called by the job manager. Returns true on success, false otherwise. */
+	bool encode_png(std::shared_ptr<uint8_t> pxdata,
+	                coord::window size);
 
 
 private:
@@ -32,6 +41,9 @@ private:
 
 	/** contains the last time when a screenshot was taken */
 	std::time_t last_time;
+
+	/** the job manager this screenshot manager uses */
+	job::JobManager *job_manager;
 
 };
 
--- a/doc/build_instructions/fedora_20_21.md
+++ b/doc/build_instructions/fedora_20_21.md
@@ -1,3 +1,3 @@
 # Prerequisite steps for Fedora users (Fedora 20, 21)
 
-`sudo yum install cmake gcc-c++ clang SDL2-devel SDL2_image-devel python3-devel python3-numpy python3-pillow libepoxy-devel opus-tools opusfile-devel fontconfig-devel harfbuzz-devel qt5-qtdeclarative-devel qt5-qtquickcontrols`
+`sudo yum install cmake gcc-c++ clang SDL2-devel SDL2_image-devel python3-devel python3-numpy python3-pillow libepoxy-devel libpng-devel opus-tools opusfile-devel fontconfig-devel harfbuzz-devel qt5-qtdeclarative-devel qt5-qtquickcontrols`
--- a/doc/build_instructions/fedora_22.md
+++ b/doc/build_instructions/fedora_22.md
@@ -1,3 +1,3 @@
 # Prerequisite steps for Fedora users (Fedora 22)
 
-`sudo dnf install cmake gcc-c++ clang SDL2-devel SDL2_image-devel python3-Cython python3-devel python3-numpy python3-pillow python3-pygments libepoxy-devel opus-tools opusfile-devel fontconfig-devel harfbuzz-devel qt5-qtdeclarative-devel qt5-qtquickcontrols`
+`sudo dnf install cmake gcc-c++ clang SDL2-devel SDL2_image-devel python3-Cython python3-devel python3-numpy python3-pillow python3-pygments libepoxy-devel libpng-devel opus-tools opusfile-devel fontconfig-devel harfbuzz-devel qt5-qtdeclarative-devel qt5-qtquickcontrols`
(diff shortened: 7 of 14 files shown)


libopenage/screenshot.cpp
```diff
@@ -46,44 +52,89 @@ std::string ScreenshotManager::gen_next_filename() {
 
 
 void ScreenshotManager::save_screenshot() {
+	job::JobManager *job_mgr = this->job_manager;
+
+	uint32_t *pxdata = new uint32_t [this->window_size.x * this->window_size.y];
```
TheJJ:
``` cpp
auto pxdata = std::make_unique<uint32_t[]>(this->window_size.x * this->window_size.y);
```

libopenage/screenshot.cpp
```diff
@@ -46,44 +52,89 @@ std::string ScreenshotManager::gen_next_filename() {
...
+	glReadPixels(0, 0,
+	             this->window_size.x, this->window_size.y,
+	             GL_RGBA, GL_UNSIGNED_BYTE, pxdata);
+
+	encode_png(pxdata);
```
TheJJ: Depending if you want to transfer ownership, either do:

``` cpp
encode_png(std::move(pxdata));
```

to transfer the memory ownership (funcion arg needs to be a `std::unique_ptr<uint32_t[]> &&`

or

``` cpp
encode_png(pxdata.get());
```

The `.get()` returns an unowned raw-pointer to the allocated memory.

It may be better to transfer ownership and then write the data to the file in the `encode_png` function.

Alternative architectures would be that the `encode_png` function creates some memory chunk that is returned and can then be written to a file, or used as texture, etc.

libopenage/screenshot.cpp
```diff
@@ -46,44 +52,89 @@ std::string ScreenshotManager::gen_next_filename() {
...
+//	job_mgr->enqueue<void>(&encode_png);
+//asdf:	clean unneccessary includes/variables
+}
+
+void ScreenshotManager::encode_png(uint32_t* pxdata) {
+	uint32_t **rows = nullptr;
```
TheJJ: If libpng only works with rows, the better choice here would be a `std::vector<uint32_t*> row_ptrs`. 

``` cpp
row_ptrs.reserve(this->window_size.y);
for (size_t i = 0; i < this->window_size.y; i++) {
    row.push_back(&pxdata[i * this->window_size.x]);
}
// as vector is guaranteed to be a continuous memory block:
png_set_rows(png_ptr, info_ptr, static_cast<png_bytepp>(&rows[0]));
```

I guess png can do this internally though, and you can just give it the pxdata buffer and dimensions.

libopenage/screenshot.cpp
castilma: thanks. can you explain, why it needs a reference to a reference?

libopenage/screenshot.cpp
castilma: what's the difference between static_cast and ()-cast?
```
libopenage/screenshot.cpp:131:70: error: invalid static_cast from type ‘unsigned int**’ to type 'png_bytepp {aka unsigned char**}’
  png_set_rows(png_ptr, info_ptr, static_cast<png_bytepp>(&row_ptrs[0]));
                                                                      ^```

File: libopenage/screenshot.cpp (hunk 2 above)
Comment (TheJJ) [reply libopenage/screenshot.cpp]: That is called ["rvalue reference"](http://en.cppreference.com/w/cpp/language/reference). It prefers to bind to rvalues and extens the lifetime of a temporary object. The `move` is needed to make the `unique_ptr` an rvalue ("remove the name"), and we need to do that so the object is not copied to the function argument. Thus, the object is "transferred" into the function.

File: libopenage/screenshot.cpp (hunk 3 above)
Comment (TheJJ) [reply libopenage/screenshot.cpp]: See the `Explanation 1)`: http://en.cppreference.com/w/cpp/language/explicit_cast

The cast fails because the typecheck for `static_cast` detects something invalid here, as an int to char cast is platform dependent (byteorder).

Comment (castilma) [reply libopenage/screenshot.cpp]: so, what's the fix? and since it's an **int to **char cast, the byteorder shouldn't be a problem?

File: libopenage/screenshot.cpp [hunk 4]
```diff
@@ -46,44 +52,87 @@ std::string ScreenshotManager::gen_next_filename() {
...
 void ScreenshotManager::save_screenshot() {
+	coord::pixel_t width = this->window_size.x,
+	               height = this->window_size.y;
+
+	auto pxdata = std::make_unique<uint32_t[]>(width * height);
+	glReadPixels(0, 0, width, height, GL_RGBA, GL_UNSIGNED_BYTE, pxdata.get());
+
+	encode_png(std::move(pxdata), width, height);
+
+	auto enc_no_args = std::bind(encode_png, std::move(pxdata), width, height);
```
TheJJ: This is invalid. the `pxdata` ownage was transferred, the object is now invalid. I think you just have to remove the above function and replace the `std::bind` with:

```
auto encode_func = [pixels=std::move(pxdata), width, height] () mutable {
    encode_png(std::move(pixels), width, height);
};
```

This method should be documented in the jobmanager howto ;)

libopenage/screenshot.cpp
TheJJ: Hm, I'm not sure if the buffer shouldn't be a `char` array from the beginning. otherwise the int byte order may be flipped on platforms. Just reinterpreting it as something different doesn't solve the problem.

castilma: I don't understand exactly what could be flipped. I understood from [this](http://en.cppreference.com/w/cpp/language/reinterpret_cast) under 'type aliasing' in the last bullet point that it's ok. not sure though, if this refers to our situation. I will change it though.

libopenage/screenshot.cpp
castilma: still doesn't work. does encode_png need to be static?:
```
/home/mcd/.repo/ap/openage/libopenage/screenshot.cpp: In lambda function:
/home/mcd/.repo/ap/openage/libopenage/screenshot.cpp:64:53: error: ‘this’ was not captured for this lambda function
   return encode_png(std::move(pixels), width, height);
                                                     ^
/home/mcd/.repo/ap/openage/libopenage/screenshot.cpp:64:53: error: cannot call member function ‘bool openage::ScreenshotManager::encode_png(std::unique_ptr<unsigned char []>&&, openage::coord::pixel_t, openage::coord::pixel_t)’ without object
/home/mcd/.repo/ap/openage/libopenage/screenshot.cpp: In member function ‘void openage::ScreenshotManager::save_screenshot()’:
/home/mcd/.repo/ap/openage/libopenage/screenshot.cpp:66:50: error: no matching function for call to ‘openage::job::JobManager::enqueue(openage::ScreenshotManager::save_screenshot()::<lambda()>&)’
  this->job_manager->enqueue<bool>(encode_function);
                                                  ^
In file included from /home/mcd/.repo/ap/openage/libopenage/screenshot.h:11:0,
                 from /home/mcd/.repo/ap/openage/libopenage/screenshot.cpp:3:
/home/mcd/.repo/ap/openage/libopenage/job/job_manager.h:93:9: note: candidate: template<class T> openage::job::Job<T> openage::job::JobManager::enqueue(openage::job::job_function_t<T>, openage::job::callback_function_t<T>)
  Job<T> enqueue(job_function_t<T> function,
         ^
/home/mcd/.repo/ap/openage/libopenage/job/job_manager.h:93:9: note:   template argument deduction/substitution failed:
/home/mcd/.repo/ap/openage/libopenage/screenshot.cpp:66:50: note:   cannot convert ‘encode_function’ (type ‘openage::ScreenshotManager::save_screenshot()::<lambda()>’) to type ‘openage::job::job_function_t<bool> {aka std::function<bool()>}’
  this->job_manager->enqueue<bool>(encode_function);
                                                  ^
In file included from /home/mcd/.repo/ap/openage/libopenage/screenshot.h:11:0,
                 from /home/mcd/.repo/ap/openage/libopenage/screenshot.cpp:3:
/home/mcd/.repo/ap/openage/libopenage/job/job_manager.h:113:9: note: candidate: template<class T> openage::job::Job<T> openage::job::JobManager::enqueue(openage::job::abortable_function_t<T>, openage::job::callback_function_t<T>)
  Job<T> enqueue(abortable_function_t<T> function,
         ^
/home/mcd/.repo/ap/openage/libopenage/job/job_manager.h:113:9: note:   template argument deduction/substitution failed:
/home/mcd/.repo/ap/openage/libopenage/screenshot.cpp:66:50: note:   cannot convert ‘encode_function’ (type ‘openage::ScreenshotManager::save_screenshot()::<lambda()>’) to type ‘openage::job::abortable_function_t<bool> {aka std::function<bool(std::function<bool()>, std::function<void()>)>}’
  this->job_manager->enqueue<bool>(encode_function);
                                                  ^

```

libopenage/screenshot.cpp
TheJJ: If we first store the pixels in an `int == int32_t` with gl, it will use some byte order:
```
// GL_UNSIGNED_BYTE denotes that the input is unsigned bytes, so the type should be uint8_t[]
// if we pass int32 there, it will still work and written in order of RGBA
// if one of the ints was read, it would be 0xAABBGGRR
glReadPixels(0, 0, width, height, GL_RGBA, GL_UNSIGNED_BYTE, pxdata.get());
```

If we pass ints there and let libpng read the lines as `char`s, the pixels will again be read `rgba`, so my flipping thing was wrong. But the underlying data type is wrong, the memory is accessed only as `uint8_t`, so it should be a `uint8_t`.

libopenage/screenshot.cpp
TheJJ: add `this` inside the `[]`, so you have `... = [this, pixels=std::move(pxdata), width, height] ()...`, and call `this->encode_png`. Or make `encode_png` static, if it doesn't need access to `this`.

castilma: we're getting closer. It's nice gcc tells me that the problem is, that the function is implicitly deleted because else the default definition would be ill-formed. but why would it be ill-formed, gcc?:
```
/home/mcd/.repo/ap/openage/libopenage/screenshot.cpp: In member function ‘void openage::ScreenshotManager::save_screenshot()’:
/home/mcd/.repo/ap/openage/libopenage/screenshot.cpp:64:50: error: use of deleted function ‘openage::ScreenshotManager::save_screenshot()::<lambda()>::<lambda>(const openage::ScreenshotManager::save_screenshot()::<lambda()>&)’
  this->job_manager->enqueue<bool>(encode_function);
                                                  ^
/home/mcd/.repo/ap/openage/libopenage/screenshot.cpp:61:71: note: ‘openage::ScreenshotManager::save_screenshot()::<lambda()>::<lambda>(const openage::ScreenshotManager::save_screenshot()::<lambda()>&)’ is implicitly deleted because the default definition would be ill-formed:
  auto encode_function = [this, pixels=std::move(pxdata), width, height] () mutable {
                                                                       ^
/home/mcd/.repo/ap/openage/libopenage/screenshot.cpp:61:71: error: use of deleted function ‘std::unique_ptr<_Tp [], _Dp>::unique_ptr(const std::unique_ptr<_Tp [], _Dp>&) [with _Tp = unsigned char; _Dp = std::default_delete<unsigned char []>]’
In file included from /usr/include/c++/5/memory:81:0,
                 from /home/mcd/.repo/ap/openage/libopenage/screenshot.h:7,
                 from /home/mcd/.repo/ap/openage/libopenage/screenshot.cpp:3:
/usr/include/c++/5/bits/unique_ptr.h:598:7: note: declared here
       unique_ptr(const unique_ptr&) = delete;
       ^
In file included from /usr/include/c++/5/memory:79:0,
                 from /home/mcd/.repo/ap/openage/libopenage/screenshot.h:7,
                 from /home/mcd/.repo/ap/openage/libopenage/screenshot.cpp:3:
/usr/include/c++/5/functional:2246:7: note:   initializing argument 1 of ‘std::function<_Res(_ArgTypes ...)>::function(_Functor) [with _Functor = openage::ScreenshotManager::save_screenshot()::<lambda()>; <template-parameter-2-2> = void; <template-parameter-2-3> = void; _Res = bool; _ArgTypes = {}]’
       function<_Res(_ArgTypes...)>::
       ^
/usr/include/c++/5/functional: In instantiation of ‘static void std::_Function_base::_Base_manager<_Functor>::_M_clone(std::_Any_data&, const std::_Any_data&, std::false_type) [with _Functor = openage::ScreenshotManager::save_screenshot()::<lambda()>; std::false_type = std::integral_constant<bool, false>]’:
/usr/include/c++/5/functional:1746:16:   required from ‘static bool std::_Function_base::_Base_manager<_Functor>::_M_manager(std::_Any_data&, const std::_Any_data&, std::_Manager_operation) [with _Functor = openage::ScreenshotManager::save_screenshot()::<lambda()>]’
/usr/include/c++/5/functional:2256:19:   required from ‘std::function<_Res(_ArgTypes ...)>::function(_Functor) [with _Functor = openage::ScreenshotManager::save_screenshot()::<lambda()>; <template-parameter-2-2> = void; <template-parameter-2-3> = void; _Res = bool; _ArgTypes = {}]’
/home/mcd/.repo/ap/openage/libopenage/screenshot.cpp:64:50:   required from here
/usr/include/c++/5/functional:1710:34: error: use of deleted function ‘openage::ScreenshotManager::save_screenshot()::<lambda()>::<lambda>(const openage::ScreenshotManager::save_screenshot()::<lambda()>&)’
    __dest._M_access<_Functor*>() =
                                  ^
```

TheJJ: I think the the explaination is that the lambda function's copy constructor is deleted because it captures a unique_ptr. If one could copy the lambda, you would also copy the contained unique_ptr. The solution is to add a `JobManager::enqueue(job_function_t<T> &&)` method, otherwise this lambda will always be copied into the jobmanager, which is wrong.

This would mean some more complication, so I'd just go for a shared_ptr. It behaves exactly the same, except you can copy it.

castilma:
on my machine png_voidp_NULL is defined in png.h. I have `libpng12-dev:amd64                             1.2.54-1ubuntu1`. but I can change it to `(png_voidp)NULL`. what version is on the ci system installed?

castilma:
It's ready.

libopenage/screenshot.cpp
```diff
@@ -46,44 +50,83 @@ std::string ScreenshotManager::gen_next_filename() {
...
-	             GL_RGBA, GL_UNSIGNED_BYTE, pxdata.get());
+	png_infop info_ptr = png_create_info_struct(png_ptr);
+	if (!info_ptr) {
+		fclose(fout);
+		png_destroy_write_struct(&png_ptr, (png_infopp) NULL);
+		return false;
+	}
 
-	uint32_t *surface_pxls = (uint32_t *)screen->pixels;
+	if (setjmp(png_jmpbuf(png_ptr))) {
```
Tomatower: After I recovered from the shock reading this, I read upon the code of libpng - It is actually the only way to go without terminating the whole game...

libopenage/screenshot.cpp
```diff
@@ -46,44 +50,83 @@ std::string ScreenshotManager::gen_next_filename() {
...
-	this->window_size.x,
-	this->window_size.y,
-	32,
-	rmask, gmask, bmask, amask);
 
-	size_t pxcount = screen->w * screen->h;
+bool ScreenshotManager::encode_png(std::shared_ptr<uint8_t> pxdata,
+	                               coord::pixel_t width, coord::pixel_t height) {
+	std::string filename = this->gen_next_filename();
+	FILE *fout = fopen(filename.c_str(), "wb");
```
Tomatower: why opening the fd up here, and not after the initialization of libpng - it would make many `fclose(fout)` obsolete.

libopenage/screenshot.cpp
```diff
@@ -46,44 +50,83 @@ std::string ScreenshotManager::gen_next_filename() {
 
 
 void ScreenshotManager::save_screenshot() {
-	std::string filename = this->gen_next_filename();
+	coord::pixel_t width = this->window_size.x,
+	               height = this->window_size.y;
```
Tomatower: Who is setting these values? They seem to be unset all the way.

Why not just add width and height as parameters to `save_screenshot` (and remove this class member)?

libopenage/screenshot.cpp
castilma: they were inspired by c++ exceptions.

TheJJ: http://thread.gmane.org/gmane.comp.graphics.png.devel/5641
The mentioned callbacks are here: http://www.libpng.org/pub/png/libpng-manual.txt
found under "Setting up callback code".
If we let libpng handle the allocations/frees, this is also not necessary. This really seems strange and we should look into how Pillow or SDL2_Image interacts with libpng.

libopenage/screenshot.h
```diff
@@ -1,23 +1,36 @@
...
 
 #pragma once
 
 #include <string>
 #include <ctime>
+#include <memory>
 
-#include <SDL2/SDL.h>
 #include "coord/window.h"
+//#include "job/job_manager.h"
```
TheJJ: leftover

libopenage/screenshot.cpp
```diff
@@ -46,44 +50,83 @@ std::string ScreenshotManager::gen_next_filename() {
...
+		fclose(fout);
+		png_destroy_write_struct(&png_ptr, &info_ptr);
+		return false;
+	}
+
+	png_init_io(png_ptr, fout);
+
+	png_set_IHDR(png_ptr, info_ptr, width, height, 8, PNG_COLOR_TYPE_RGB,
+	             PNG_INTERLACE_NONE, PNG_COMPRESSION_TYPE_DEFAULT,
+				 PNG_FILTER_TYPE_DEFAULT);
```
TheJJ: indent

libopenage/screenshot.cpp
```diff
@@ -46,44 +50,83 @@ std::string ScreenshotManager::gen_next_filename() {
...
-	SDL_Surface *screen = SDL_CreateRGBSurface(
-	SDL_SWSURFACE,
-	this->window_size.x,
-	this->window_size.y,
-	32,
-	rmask, gmask, bmask, amask);
 
-	size_t pxcount = screen->w * screen->h;
+bool ScreenshotManager::encode_png(std::shared_ptr<uint8_t> pxdata,
+	                               coord::pixel_t width, coord::pixel_t height) {
```
TheJJ: indent

libopenage/screenshot.cpp
```diff
@@ -46,44 +50,83 @@ std::string ScreenshotManager::gen_next_filename() {
...
-	log::log(MSG(info) << "Saving screenshot to " << filename);
+	std::shared_ptr<uint8_t> pxdata(new uint8_t[4*width*height], std::default_delete<uint8_t[]>());
+	glReadPixels(0, 0, width, height, GL_RGBA, GL_UNSIGNED_BYTE, pxdata.get());
 
-	int32_t rmask, gmask, bmask, amask;
-	rmask = 0x000000FF;
-	gmask = 0x0000FF00;
-	bmask = 0x00FF0000;
-	amask = 0xFF000000;
+	auto encode_function = [this, pxdata, width, height] () mutable {
```
TheJJ: i think the mutable is no longer necessary as we don't move a captured element out (and therefore invalidate something from a const function, what the mutable was for to allow anyway)

libopenage/screenshot.cpp
TheJJ: https://hg.libsdl.org/SDL_image/file/6332f9425dcc/IMG_png.c#l388 They also do it, but do different handling for libpng12.

On the CI there is libpng16 (archlinux). libpng12 is very old, so I think we should target >12.

TheJJ:
If possible, this whole libpng-mess could stay offloaded with sdl2 if we just put the encoding and saving to the job manager in another thread. Then sdl will continue to abstract the interface and we don't have to deal with the backends. And who knows, maybe soon there'll be another image compression format which we'll then get for free from sdl?

What do you think?

castilma:
directly using libpng allows us to remove the alpha channel.

TheJJ:
And SDL can't do that? Well then, the work is good so we can then abstract the image encoding/png stuff to a separate class which we then can also use from cython for asset encodings.

libopenage/screenshot.cpp
castilma: it wasn't me. it was like that before I touched `screenshot`!  `engine.cpp:259` sets the value in a callback on `SDL_WINDOWEVENT_RESIZED` the new size is not available (AFAIK) when save_screenshot is called.

libopenage/screenshot.cpp
castilma: so leave it as it is?

castilma:
what's the correct way to update files in `doc/build_instructions`? I can only guess for some systems how the package with png header files is called.

libopenage/screenshot.cpp
Tomatower: I agree that it has been like that - but it is very tightly coupled that way. it would be much nicer to have all screenshot stuff in the engine only at one place if possible.

`engine.cpp:207` has all the information, where it has to be called.

libopenage/screenshot.cpp
castilma: on your first comment: 

> If we let libpng handle the allocations/frees, this is also not necessary.

Did you mean: `If we handle the allocations/frees, this is also not necessary.` ?
this is what the 4th post on gmane says.

castilma:
Wait, I just realized the i removed the feedback for screencaps. Until now, the game froze for half a second when making a screenshot. Now, the game just continues. Is that bad? One could show a line saying screenshot taken or so.

Am 29. August 2017 02:17:51 MESZ schrieb Tomatower <notifications@github.com>:
>Tomatower approved this pull request.
>
>I like it!

doc/build_instructions/arch_linux.md
```diff
@@ -4,7 +4,7 @@
 
 This command should provide required packages for Arch Linux installation:
 
-`sudo pacman -S --needed python python-pillow python-numpy python-pygments cython libepoxy ttf-dejavu freetype2 fontconfig harfbuzz cmake sdl2 sdl2_image opusfile opus-tools python-pylint qt5-declarative qt5-quickcontrols`
+`sudo pacman -S --needed python python-pillow python-numpy python-pygments cython libepoxy libpng15 ttf-dejavu freetype2 fontconfig harfbuzz cmake sdl2 sdl2_image opusfile opus-tools python-pylint qt5-declarative qt5-quickcontrols`
```
TheJJ: should just be [`libpng`](https://www.archlinux.org/packages/extra/x86_64/libpng/)

doc/build_instructions/ubuntu_16.04.md
```diff
@@ -1,5 +1,5 @@
 # Prerequisite steps for Ubuntu users (Ubuntu 16.04)
 
  - `sudo apt-get update`
- - `sudo apt-get install cmake libfreetype6-dev python3-dev python3-pip libepoxy-dev libsdl2-dev libsdl2-image-dev libopusfile-dev libfontconfig1-dev libharfbuzz-dev opus-tools python3-pil python3-numpy python3-pygments qtdeclarative5-dev qml-module-qtquick-controls`
+ - `sudo apt-get install cmake libfreetype6-dev python3-dev python3-pip libepoxy-dev libsdl2-dev libsdl2-image-dev libopusfile-dev libfontconfig1-dev libharfbuzz-dev libpng16-dev opus-tools python3-pil python3-numpy python3-pygments qtdeclarative5-dev qml-module-qtquick-controls`
```
TheJJ: [`libpng-dev`](https://packages.ubuntu.com/xenial/libpng-dev) should be enough

doc/building.md
```diff
@@ -34,6 +34,7 @@ Dependency list:
       A   python imaging library (PIL) -> pillow
     CR    opengl >=2.1
     CR    libepoxy
+	CR    libpng
```
TheJJ: indent pls

libopenage/screenshot.cpp
```diff
@@ -6,20 +6,24 @@
 #include <epoxy/gl.h>
 #include <math.h>
 #include <memory>
+#include <png.h>
+#include <setjmp.h>
 #include <stdlib.h>
-#include <SDL2/SDL.h>
-#include <SDL2/SDL_image.h>
+#include <stdio.h>
+#include <string.h>
```
TheJJ: in c++ it's better to use [`csetjmp`](http://en.cppreference.com/w/cpp/utility/program/setjmp), `cstdio` and `cstring`, because that namespaces things with `std::`. Doesn't matter much, though.

TheJJ:
About the success-feedback: For now you can just print in the terminal or print it on screen.

castilma:
I'd leave the ingame message for the future.

PR #896, opened 2017-08-26
Get more tests to pass on Windows

### LINKED ISSUE #876: fslike test broken for case-insensitive filesystems
Windows filesystems and apparently some mac systems (#874) have case insensitive filesystems.

Our `fslike` system was built for case-sensitive filesystems like Linux has.
So we have a test in `openage/util/fslike/test.py` that assumes the underlying filesystem is case-sensitive: It checks that when just an `a` is created, then an `A` is created and redirected to be the same `a`, and then we test if only `a` exists and not `A`, as `A` was redirected to `a`.

If now the underlying filesystem is case-insensitive, it will report that `A` and `a` exist, which breaks the test.

Solution: test if the underlying filesystem is case sensitive or not and adjust the test to expect different results then.

tusharpm:
[Reverted in favor of #895] ~Fixes #876 - verified on Windows (NTFS).~
A few modifications to assertions in existing tests to accommodate MSVC's behavior.

### DIFF
--- a/openage/cppinterface/exctranslate_tests.pyx
+++ b/openage/cppinterface/exctranslate_tests.pyx
@@ -1,4 +1,4 @@
-# Copyright 2015-2016 the openage authors. See copying.md for legal info.
+# Copyright 2015-2017 the openage authors. See copying.md for legal info.
 
 """
 Testing code for exctranslate.pyx.
@@ -74,19 +74,21 @@ def cpp_to_py(int bounce_count = 0):
     else:
         raise TestError("Expected a CPPException, but method returned.")
 
+    is_openage_error = lambda value: b"openage::error::Error" in value
+
     # now let's see about the detailed contents of excobj.
     assert_value(excobj.args[0], "foo")
-    assert_value(excobj.typename, b"openage::error::Error")
+    assert_value(excobj.typename, None, is_openage_error)
 
     cause = getattr(excobj, "__cause__", None)
 
-    assert_value(type(cause), TestError)
+    assert_value(type(cause), None, lambda value: value in {CPPException, TestError})
 
     causeofcause = getattr(cause, "__cause__", None)
 
     assert_value(type(causeofcause), CPPException)
     assert_value(causeofcause.args[0], "rofl")
-    assert_value(causeofcause.typename, b"openage::error::Error")
+    assert_value(causeofcause.typename, None, is_openage_error)
 
     assert_value(getattr(causeofcause, "__cause__", None), None)
 
--- a/openage/testing/misc_cpp.pyx
+++ b/openage/testing/misc_cpp.pyx
@@ -1,4 +1,4 @@
-# Copyright 2015-2016 the openage authors. See copying.md for legal info.
+# Copyright 2015-2017 the openage authors. See copying.md for legal info.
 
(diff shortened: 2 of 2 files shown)


openage/util/fslike/test.py
```diff
@@ -181,8 +191,9 @@ def test():
     # test the union
     test_union(root_path, root_dir)
 
-    # test the case ignoring dir
-    test_case_ignoring(root_path, root_dir)
+    if is_filesystem_case_sensitive():
+        # test the case ignoring dir
+        test_case_ignoring(root_path, root_dir)
```
TheJJ: We should not skip it completely, but only the part where the assumption of a non-insensitive fs is done. I already addressed this in #895. Your `is_filesystem_case_sensitive` function is better for the detection though, so I integrated it into my pull. Thanks :)

openage/cppinterface/exctranslate_tests.pyx
```diff
@@ -74,19 +74,21 @@ def cpp_to_py(int bounce_count = 0):
...
+
     # now let's see about the detailed contents of excobj.
     assert_value(excobj.args[0], "foo")
-    assert_value(excobj.typename, b"openage::error::Error")
+    assert_value(excobj.typename, None, is_openage_error)
 
     cause = getattr(excobj, "__cause__", None)
 
-    assert_value(type(cause), TestError)
+    assert_value(type(cause), None, lambda value: value in {CPPException, TestError})
```
TheJJ: hmm. why is this needed? and why isn't a `== b"openage::error::Error" enough but we have to compare with `in`?

openage/util/fslike/test.py
tusharpm: Oops. I didn't check your PR.
I'll revert changes related to fs from mine.

openage/cppinterface/exctranslate_tests.pyx
tusharpm: This explicit line (numbered 83 in red) reports:

    openage.testing.testing.TestError: unexpected result: <class 'openage.cppinterface.exctranslate.CPPException'>

And without `is_openage_error`, lines numbered 79 and 89 in red report:

    openage.testing.testing.TestError: unexpected result: b'class openage::error::Error'

I made the lambda to look for a common piece among the compilers' outputs.

TheJJ:
You may rebase now, the insensitivefs pull is merged.

Did you investigate further how we can solve the typeid problem in a crossplatform-compatible way? Do we know how mac behaves?

tusharpm:
I rebased the branch.

I didn't get a chance to figure out a cross-platform solution for typeid (my last suggestion was about patching `util::demangle` function). I don't recall how mac (AppleClang) behaves exactly for type names, but if I had to guess, I'd say pretty much the same as clang on Linux.

Regarding this change: if I understand correctly, demangling is meant for troubleshooting and human-readability. Having the `class` or `struct` prefix shouldn't affect that. Therefore just modifying the test to accommodate this nuance sounds alright to me.
Deferring to your judgement for this.

TheJJ:
I think as long as the class name is in that typeid, it's alright for now. We have to come back to the problem when we someday do check for the stored exception type.

PR #901, opened 2017-08-29
fix scrolling without focus

### LINKED ISSUE #871: Mouse controlled window movement doesnt turn off when window is out of focus
Because of #843 the visible frame will move into (Gaben) oblivion when you Alt-Tab or minimize the window. Trying to find the map again turns out to be very difficult.

Solution:
* Turn mouse controlled window movement off when the game is out of focus
* Define boundaries for the camera so it cannot "escape"

castilma:
fixes #871.
this is my fastest pr so far

### DIFF
--- a/libopenage/engine.cpp
+++ b/libopenage/engine.cpp
@@ -372,24 +372,27 @@ void Engine::loop() {
 			// camera movement speed, in pixels per millisecond
 			// one pixel per millisecond equals 14.3 tiles/second
 			float mov_x = 0.0, mov_y = 0.0, cam_movement_speed_keyboard = 0.5;
+			Uint32 win_flags = SDL_GetWindowFlags(this->window);
+			bool inp_focus = win_flags & SDL_WINDOW_INPUT_FOCUS;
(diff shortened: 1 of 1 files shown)


libopenage/engine.cpp
```diff
@@ -374,24 +374,25 @@ void Engine::loop() {
...
 			// one pixel per millisecond equals 14.3 tiles/second
 			float mov_x = 0.0, mov_y = 0.0, cam_movement_speed_keyboard = 0.5;
+			bool has_focus = SDL_GetWindowFlags(this->window) & SDL_WINDOW_MOUSE_FOCUS;
 
 			input::InputManager &input = this->get_input_manager();
 			using Edge = input::InputManager::Edge;
 
 			if (input.is_down(SDLK_LEFT) or
-			    input.is_mouse_at_edge(Edge::LEFT, coord.window_size.x)) {
+			    input.is_mouse_at_edge(Edge::LEFT, coord.window_size.x) and has_focus) {
```
TheJJ: Do we want to stop scrolling for keyboard presses also? I think that would be good.

In any case, I'd make some parenthesis to make it explicit.

TheJJ:
thanks :)

PR #904, opened 2017-08-30
Buildings limits, Resource capacity and more

MaanooAk:
- **Buildings**
  - **cost** check and deduction
  -  **limits** max at single instance, max ever created
    - Player keeps track for each unit type the numbers:
      - **units have**: active 
      - **units had**: active and dead
      - **units pending**: being made
- Player **Resource capacity**
  - **Storage** attribute
- Dynamic resources (first steps)
  - **Resource**: one resource type (e.g wood)
  - **Resources**: set of Resource
  - **ResourceBoundle**: Resource-Amount pairs (created with a Resources)
- **Resource** **decay** and **gather rate multiplier**

EDIT:

- **Resource generator** attribute
  - Tmp logic in idle action
- **ResourceCost** 
  - It can be **constant** or **dynamic** (= function of player)
  - Used in **UnitType** and **ReseachType**

EDIT:

- **ResourceBundle** operations with bundles from different versions of Resources

### DIFF
--- a/libopenage/gamestate/resource.h
+++ b/libopenage/gamestate/resource.h
@@ -6,16 +6,93 @@
 
 namespace openage {
 
+class ResourceBundle;
+
 /**
- * probably becomes part of the nyan game spec in future
+ * A resource
  */
-enum class game_resource : int {
-	wood,
-	food,
-	gold,
-	stone,
-	RESOURCE_TYPE_COUNT
+class Resource {
+public:
+
+	Resource();
+
+	virtual int id() const = 0;
+
+	virtual std::string name() const = 0;
+
+	// TODO add images and icons
+
 };
+
+class ResourceProducer : public Resource {
+public:
+
+	ResourceProducer(int id, std::string name)
+		:
+		_id{id},
+		_name{name} { }
+
+	int id() const override { return _id; }
+
+	std::string name() const override { return _name; }
+
+private:
+
+	int _id;
+	std::string _name;
+};
+
+/**
+ * All the resources
+ */
+class Resources {
+public:
+
+	Resources();
+
+	virtual unsigned int get_count() const = 0;
+
+	virtual const Resource& get_resource(int id) const = 0;
+
+	ResourceBundle create_bundle() const;
+
+};
+
+class ClassicResources : public Resources {
+public:
+
+	ClassicResources()
+		:
+		wood{0, "wood"},
+		food{1, "food"},
+		gold{2, "gold"},
+		stone{3, "stone"},
+		resources{{0, "wood"}, {1, "food"}, {2, "gold"}, {3, "stone"}}{
+	}
+
+	unsigned int get_count() const override { return 4; }
+
+	const Resource& get_resource(int id) const override { return this->resources[id]; };
+
+private:
+
+	const ResourceProducer wood;
+	const ResourceProducer food;
+	const ResourceProducer gold;
+	const ResourceProducer stone;
+
+	const ResourceProducer resources[4];
+};
+
+// TODO remove, here for backwards compatibility
+enum class game_resource : int {
+	wood = 0,
+	food = 1,
+	gold = 2,
+	stone = 3,
+	RESOURCE_TYPE_COUNT = 4
+};
+
 
 /**
  * A set of amounts of game resources.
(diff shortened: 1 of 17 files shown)


libopenage/gamestate/resource.h
```diff
@@ -6,25 +6,109 @@
...
+class ClassicResources : public Resources {
+public:
+
+	ClassicResources()
+		:
+		wood{0, "wood"},
+		food{1, "food"},
+		gold{2, "gold"},
+		stone{3, "stone"},
+		resources{{0, "wood"}, {1, "food"}, {2, "gold"}, {3, "stone"}}{
```
TheJJ: why double them?

MaanooAk: I will keep only the array, but is this model enough?

```c++
class Resources {
public:
	Resources();
	virtual unsigned int get_count() const = 0;
	virtual const Resource& get_resource(int id) const = 0;
	ResourceBundle create_bundle() const; // not virtual
};
```

MaanooAk:
I'm adding **ResourceCost** and replacing the `ResourceBundle cost` with `ResourceCost cost` in the **UnitType**.

**ResourceCost** can either be **constant** or **dynamic**. One dynamic type is **worforce**.

eg.
```c++
ResourceCost constant = ResourceCost(resource_bundle)
ResourceCost constant = ResourceCost(cost_type::constant, resource_bundle) 
ResourceCost dynamic = ResourceCost(cost_type::workforce, resource_bundle) 
```
The `resource_bunlde` in the **worforce** is used as a multiplier.

eg. The cost of the **Spies** research
```c++
ResourceBundle mutiplier = ResourceBundle();
mutiplier[game_resource::gold] = 200;
ResourceCost spies_cost =  ResourceCost(cost_type::workforce, mutiplier)
```

I will wait for the Reseach pr to be merged in order to do the same for then **ReseachType**.

libopenage/gamestate/resource.h
TheJJ: Can't really tell, we only have to design it so that resources can be added or removed at runtime by the API.

MaanooAk:
>  added or removed at runtime by the API

Do you mean the start of the game or when any patch is applied during the game?

===

Can a patch introduce a new resource? Nyan is capable, but is it allowed?

In that case, the ResourceBundle will have to modified (Resources is ok)

TheJJ:
I think we should target "at the start of the game", but I think there's not much difference to researching a new technology that activates a new resource (e.g. electricity?). So I think we should allow it and try not to block the path for implementing this. Actually doing it has no priority of course, just another constraint.

libopenage/gamestate/resource.h
```diff
@@ -46,6 +46,9 @@ class ResourceProducer : public Resource {
  * All the resources.
  *
  * The the ids of the resources must be inside [0, count).
+ *
+ * The static variables wood, food, gold, stone are the ids of the representing resource.
+ * Any extension of Resources must use this ids as they are an engine dependency (at the moment).
```
MaanooAk: @TheJJ Do you like this? Could also be done for all dependencies on the data to avoid strings ( `Resources::food` instead of `resources.get("openage.Food")` ) ( I hate strings )

MaanooAk: At the end there should only be one eg. `Resource::money` that will be used for the market like stuff and it it will be mapped to gold for the classic data

MaanooAk: Nyan impl:
```python
Resources()
  resources: orderedset(Resource)
  money: Resource
````
If `money` is missing make money the first resource in the set

TheJJ: I don't like strings either but the non-string access should just be an optimization to the string-access (for simplicity and consistency and scriptability). That means that if we should store arbitrary resources by their identifier in a dynamic data structure and relay accesses to it to either this dynamic storage or direct class members. That way the string access can be done anyway (e.g. from pythonscripts) but the engine can access the member directly in the struct memory.

The engine API would just provide a generic `Resource`, and the base game would then instanciate this for `Gold`, `Food`, `Wood`, `Stone`.

Having some definition for which resource is the money is really great, we should do that :)
This means we then set `money = Gold`.

`nyan` can enforce that variables must have values, so `money : Resource` must be set by the base game data mod, otherwise nyan will say "undefined member". We can tweak that behavior of course, if needed.

TheJJ: Additionally, except the money thing, the core engine probably never needs to know what the resources are and then do the direct-member-access. The HUD just displays the resource set, and the engine instanciates resource storage for each player, depending on the contents of the set.

MaanooAk:
You can merge this (I don't have time to continue) there are a lot of fixes among the features.

PR #916, opened 2017-09-27
fix scrolling without focus (again)

heinezen:
Removed the requirement for the mouse to be focussed in the window because it created problems in two scenarios:

1. Mouse scrolling in windowed mode is basically useless because the mouse has to be exactly on an edge for scrolling.
2. Same for a fullscreen multi monitor setup. At least one edge is unfeasible for mouse scrolling.


TheJJ:
Can't we rely on some status flag that reports the window is the foreground window? That does not require the mouse to be on top of it, but we know the window is the one the user wants to control.

heinezen:
Isn't that exactly what `SDL_WINDOW_INPUT_FOCUS` is doing? It behaves like you describe it.

Issue #915, opened 2017-09-26
Update Fedora build instructions


TheJJ:
Our build instructions are for fedora 20, 21 and 22.

The currently supported versions are 25 and 26.

Maybe we manage to unify the document to a `fedora.md` which is hopefully future-proof so we don't need to rename it for each release.

TheJJ:
Done in #917 with 9c151e572e664b750eed16c38c94f4be6c9edba9

PR #917, opened 2017-10-01
Consolidate fedora build instructions into a single document

kev946:
Addresses #915


TheJJ:
What do you think of removing the old versions and to write "Fedora >= 22"?

kev946:
I think that it makes sense to write Fedora >= 22, because they've been unsupported for almost 2 years http://fedoraproject.org/wiki/End_of_life

TheJJ:
Please also update the links to the renamed file, otherwise they'll orphan.

kev946:
oops

Issue #807, opened 2017-04-26
Remember previous asset location


TheJJ:
After a "successful" conversion, we could store the location of the source folder in `root/cfg/asset_location` so that upon the next conversion, that folder will be reused by default.

Implementation should probably be in:
* `openage/convert/main.py`
* and pass the full `root` path to the `conversion_required` in `openage/game/main.py` (currently it's just `root/assets` but we wanna have `root/cfg` there to read and write the file)

Chinmaygoyal:
Is the issue still open?

heinezen:
@Chinmaygoyal Functionality was merged with #921 so I think this can be closed.

PR #921, opened 2017-10-07
Remember previous asset location

kev946:
Addresses #807

### DIFF
--- a/openage/convert/main.py
+++ b/openage/convert/main.py
@@ -87,14 +87,14 @@ def mount_drs_archives(srcdir, game_versions=None):
     return result
 
 
-def mount_input(srcdir=None):
+def mount_input(srcdir=None, root=None):
     """
     Mount the input folders for conversion.
     """
 
     # acquire conversion source directory
     if srcdir is None:
-        srcdir = acquire_conversion_source_dir()
+        srcdir = acquire_conversion_source_dir(root)
 
     game_versions = set(get_game_versions(srcdir))
     if not game_versions:
@@ -140,7 +140,7 @@ def mount_input(srcdir=None):
     return (output, game_versions)
 
 
-def convert_assets(assets, args, srcdir=None):
+def convert_assets(root, args, srcdir=None):
     """
     Perform asset conversion.
 
@@ -156,7 +156,7 @@ def convert_assets(assets, args, srcdir=None):
     conversion experience, then passes them to .driver.convert().
     """
 
-    data_dir, game_versions = mount_input(srcdir)
+    data_dir, game_versions = mount_input(srcdir, root)
 
     if not data_dir:
         return False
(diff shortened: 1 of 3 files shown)


openage/convert/main.py
```diff
@@ -156,6 +156,26 @@ def convert_assets(assets, args, srcdir=None):
...
+                srcdir = CaseIgnoringDirectory(file_obj.read().strip()).root
+                print('Found previous AOE2 install path: ' + srcdir.resolve_native_path().decode('utf-8'))
+                while True:
+                    print("Should we use this directory? [Y/n]")
+                    user_selection = input("> ")
+                    if user_selection.lower() in {"yes", "y", ""}:
+                        break
+                    elif user_selection.lower() in {"no", "n"}:
+                        srcdir = None
+                        break
```
TheJJ: maybe we should add an `else` with a "wtf don't understand"-like message :)

openage/convert/main.py
```diff
@@ -205,6 +225,10 @@ def flag(name):
 
         converted_count += 1
 
+    # save the asset location for next time
+    with asset_location_path.open("w") as file_obj:
+        file_obj.write(data_dir.resolve_native_path().decode('utf-8'))
```
TheJJ: I think we should write it with `wb`, then the decode is not necessary and we can ignore if the native path is valid utf8. (Above, also read with `wb` and try decoding with utf8 like currently done).

openage/convert/main.py
kev946: fixed

openage/convert/main.py
```diff
@@ -156,6 +156,28 @@ def convert_assets(assets, args, srcdir=None):
...
 
+    assets = root["assets"]
+    cfg = root["cfg"]
+
+    asset_location_path = cfg / "asset_location"
+    if srcdir is None:
+        try:
+            with asset_location_path.open("rb") as file_obj:
+                srcdir = CaseIgnoringDirectory(file_obj.read().strip()).root
+                print('Found previous AOE2 install path: ' + srcdir.resolve_native_path().decode('utf-8'))
```
TheJJ: If it is not utf8, this decoding needs error handling i think.

Another improvement might be the test if `srcdir` does still exists, otherwise the question doesn't make much sense :)

kev946: I think I fixed the utf-8 decoding error handling with `errors='replace'`. I also think I made the second improvement, but am not sure.

kev946: `CaseIgnoringDirectory` will raise a `FileNotFoundError` if the directory doesn't exist.

TheJJ:
Are doing alright?

kev946:
yeah, just need a little time

kev946:
I think that this should be good now maybe.

kev946:
I think this should be good finally

TheJJ:
Good :) Please do a quick rebase so we can get rid of the merge commit.

kev946:
I think that it's rebased.

PR #922, opened 2017-10-08
Minimap

sandsmark:
just rebased #559, + added support for using the mouse to move around from the minimap.

### DIFF
--- a/libopenage/gui/minimap.cpp
+++ b/libopenage/gui/minimap.cpp
@@ -67,7 +67,7 @@ Minimap::Minimap(QQuickItem *parent)
 	Q_UNUSED(registration);
 
 	this->setFlag(QQuickItem::ItemHasContents, true);
-    this->setAcceptedMouseButtons(Qt::LeftButton);
+	this->setAcceptedMouseButtons(Qt::LeftButton);
 }
 
 Minimap::~Minimap() {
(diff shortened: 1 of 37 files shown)


libopenage/gui/minimap.cpp
```diff
@@ -293,15 +293,15 @@ std::unique_ptr<QSGNode> Minimap::create_tree() {
...
+		const QMatrix4x4 local_to_world_transform = (turn->matrix() * world_to_background->matrix()).inverted();
+		const QPointF pos = local_to_world_transform.map(mouseEvent->localPos());
 
-        if (Engine *engine = unwrap(game->get_engine())) {
-            engine->get_coord_data()->camgame_phys.ne = pos.x();
-            engine->get_coord_data()->camgame_phys.se = pos.y();
-        }
-    };
+		if (Engine *engine = unwrap(game->get_engine())) {
+			engine->get_coord_data()->camgame_phys.ne = pos.x();
```
ChipmunkV: This executes in the GUI thread and touches values from the gamelogic thread. (these threads are currently all in one because of the font library that should be removed in the future)

Also, a game may have no engine attached yet (or the game may be null itself).

I'd use something like this:

	if (this->game)
		if (EngineLink *engine = game->get_engine()) {
			emit this->game->in_game_logic_thread_blocking([this, engine] {
				...

Or do all the coordinate changes in the `Minimap::updatePaintNode()`.

sandsmark:
fixed that.

but there seems to be more threading fuckery, I get a consistent crash on exit because of unbalanced access to the dbus library. I think it is related to the hacks with how qguiapplication is instantiated and executed, it has a ton of internal entry/atexit() handlers and whatnot, from plugins and all the modules, so e. g. trying to run exit() while it is alive is going to crash. why isn't it just executed as normal in main()?

```
   process 31792: arguments to dbus_connection_unref() were incorrect, assertion "connection->generation == _dbus_current_generation" failed in file dbus-connection.c line 2824.
This is normally a bug in some application using the D-Bus library.
```

ChipmunkV:
> process 31792: arguments to dbus_connection_unref() were incorrect, assertion "connection->generation == _dbus_current_generation" failed in file dbus-connection.c line 2824.
This is normally a bug in some application using the D-Bus library.

That's something SDL2-related. I remember deleting all of the Qt code and running the game - same error happens.

> hacks with how qguiapplication is instantiated and executed ... why isn't it just executed as normal in main()?

That simplifies the "deleting all of the Qt code" part while testing.

There are requests to run headless, so Qt isn't needed in that configuration. The main() doesn't need to directly pass anything to qguiapplication even when there is a GUI. So we can try to avoid cluttering main() if possible.

sandsmark:
I'm too dumb for the coordinate magic, so if anyone wants to pick up this and fix it, feel free

TheJJ:
Oh no! The new coord system should be pretty straight-forward with the same ideas as the old one but with a much cleaner implementation. Would it help you if you came by in our chatroom so I/@mic-e answer your questions? What also might be useful is [this document](https://github.com/SFTtech/openage/blob/master/doc/code/coordinate-systems.md).

PR #924, opened 2017-10-18
[Refactor] util/system.py free_memory() single exit point

PythonicChemist:
I changed the function so that there is a single and clear exit point. I set memory to INF at the start to silence the linter although I would prefer it in the else branch personally.


PythonicChemist:
Sadly I have no access to OSX. psutil would support all three platforms. (psutil.virtualmemory().available would be the equivalent)

I would like to do more of these low hanging fruits if desirable.

TheJJ:
I'm not sure if it would be worth pulling psutil as dependency just for that.
I'll merge your stuff now then.

TheJJ:
Such improvements are probably all over the codebase. Go and find them!

PR #926, opened 2017-10-20
util/strings.py updating to same string formatting method and slight refactor of binstr()

PythonicChemist:
I updated all functions to the same method of string formatting. Also I refactored binstr() slightly by replacing the while loop with a list comprehension.

Is there actually a case when colorize function would be called without a colorcode?

### DIFF
--- a/cfg/.gitignore
+++ b/cfg/.gitignore
@@ -1,0 +1,1 @@
+asset_location
--- a/openage/convert/main.py
+++ b/openage/convert/main.py
@@ -87,14 +87,14 @@ def mount_drs_archives(srcdir, game_versions=None):
     return result
 
 
-def mount_input(srcdir=None):
+def mount_input(srcdir=None, root=None):
     """
     Mount the input folders for conversion.
     """
 
     # acquire conversion source directory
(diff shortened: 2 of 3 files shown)


PythonicChemist:
Should I revert the formatting changes and just update colorize to the %-formatting?

I saw in other parts of the codebase both methods are used. Do you have any rule of thumb when you want to use each?

TheJJ:
We don't have any rule for that, and I think it does not matter which style we choose for simple formatting. When one parameter is repeated, the `.format()` may be better.

Depends on the specific situation I'd say.

In this pull, hard to tell. If in doubt, I'd go for f-strings, but we shouldn't require py3.6 now, so the `.format` is probably the "preferred" method. What do you think?

PythonicChemist:
I think it would be good to have uniform string formatting throughout the codebase to help new contributors easing into the source. Also we could easier/faster transition to f-literals once 3.5 is EOL.

I agree that .format is often redundant, but for understanding the code you can usually skip everything past the string itself. If we agree to not put logic into .format() you never should need to read past the string anyways.

PR #938, opened 2017-11-28
util/iterators.py removed superfluous ListIterator for iter() and refactored denote_last

PythonicChemist:
ListIterator was only used once and reimplemented iter() so I replaced it with iter().

I didn't like the "while True:" loop in denote_last, so I researched a cleaner way of implementing the logic.


TheJJ:
The procedure for the rebase is described here: https://github.com/SFTtech/openage/blob/master/doc/contributing.md#rebasing

PR #939, opened 2017-11-29
Link to nyan library

TheJJ:
initial [nyan](https://github.com/SFTtech/nyan) integration, just build and link the lib for now.


CMakeLists.txt
```diff
@@ -39,6 +42,10 @@ else()
 	project(openage CXX)
 endif()
 
+# C++ standard
+set(openage_cxx_standard 14)
```
tusharpm: This variable and its usage should be implied by the following lines from `HandleCXXOptions.cmake`:
```cmake
set(CMAKE_CXX_STANDARD 14)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
```

buildsystem/DependencyFetch.cmake
```diff
@@ -0,0 +1,66 @@
...
+
+# cmake external project wrapper
+
+
+
+function(fetch_project)
+	set(KW_ARGS
+		NAME
+		LOCATION
+		TAG
```
tusharpm: `LOCATION` and `TAG` seem unused (here and in the `ExternalFetch.cmake.in`). Can remove.
And `NAME` can be made the function's argument, removing the need to call `cmake_parse_arguments` below.

buildsystem/DependencyFetch.cmake
```diff
@@ -0,0 +1,66 @@
...
+		"${PROJ_DL_DIR}/CMakeLists.txt"
+		@ONLY
+	)
+
+	# run cmake to "configure" the external project.
+	# this prepares the download that will be done in the next step.
+	execute_process(COMMAND
+		${CMAKE_COMMAND}
+		"-G${CMAKE_GENERATOR}"
+		"-DCMAKE_MAKE_PROGRAM:FILE=${CMAKE_MAKE_PROGRAM}"
```
tusharpm: As we've established this function just downloads the project, it should be alright to let CMake use the defaults for everything: the generator, and `CMAKE_MAKE_PROGRAM`.

doc/building.md
```diff
@@ -73,8 +74,30 @@ described below for some of the most common ones:
...
+Please [follow the `nyan` build
+instructions](https://github.com/SFTtech/nyan/blob/master/doc/building.md).
+**Only install** `nyan` when packaging it. You don't need to clobber your
+filesystem.
+
+`cmake` looks for `nyan` in the user package registry
+(`~/.cmake/packages/nyan/`) and the system package registry
+(`/usr/lib/cmake/nyan/`) for `nyanConfig.cmake`.
+
+If it cannot be found, you can hint the location of the location of the
```
tusharpm: nit: just one "the location of".

CMakeLists.txt
TheJJ: ah, i changed this so every target has to declare its cxx_standard, so the standard definition in `HandleCXXOptions.cmake` is obsolete.

buildsystem/DependencyFetch.cmake
TheJJ: That's right. The 'download project' should not be needed anywhere. What generator does it use on windows then?

tusharpm: The Windows build for CMake uses the latest available "Visual Studio" generator.
The package from cygwin uses "Unix Makefiles".
I don't remember for MSYS/MinGW package(s) - my guess: "MSYS Makefiles" and "MinGW Makefiles" respectively.

CMakeLists.txt
tusharpm: This approach looks more verbose. Any particular reason to not use CMake's builtin variable(s)?

TheJJ: I wanted a single location where our standard is specified; that variable is then used for cython modules and the libopenage.

tusharpm: All cython modules and libopenage are added using `add_library` and should respect the value in `CMAKE_CXX_STANDARD` automatically.
Using a different variable has the possibility of someone missing the specification for a new target.

TheJJ: You're right. I'll try if it still works, for nyan i had to specify it for the library because the include mechanism does `add_subdirectory` on the nyan repo, so the c++17 has to be specified for the nyan targets separately. Let's see if it still works with the global definition.

buildsystem/DependencyFetch.cmake
TheJJ: What's the problem with passing through customizations?

tusharpm: There's no problem. The comment was just to reduce boilerplate code.

TheJJ: Okay, so what do? Should we keep it or use the defaults? I'd go for passing the configuration, then customizations will propagate and not produce an unexpected and uninfluenceable subproject buildsystem.

CMakeLists.txt
tusharpm: For nyan (assuming the commit removing `set(CMAKE_CXX_STANDARD 17)` is reverted), it should be okay to use  `add_subdirectory`, as CMake keeps that definition limited to the nyan directory scope.

buildsystem/DependencyFetch.cmake
tusharpm: Fair point. We can keep it.
It might help someone later. 👍

CMakeLists.txt
TheJJ: hm i'm not sure what my mistake was, then. I added the global cxx standard setting in openage back in.

TheJJ:
Wohoo it works! `openage` should now be able to use `nyan` in 3 modes:
* as system package (when the distro has a `nyan` package installed)
* as a user package (when a developer wants to work on nyan and openage in parallel, both are just git repos checked out wherever one wants)
* as a subproject (downloaded automatically with `./configure --download-nyan`)

CMakeLists.txt
```diff
@@ -98,6 +118,17 @@ if(ENABLE_CCACHE)
...
 
 
+# clang tidy static analysis
+option(
+	ENABLE_CLANG_TIDY
+	"activate clang tidy messages"
+	OFF
+)
+if(ENABLE_CLANG_TIDY)
+	set(CMAKE_CXX_CLANG_TIDY "clang-tidy;-checks=-*,readability-*")
```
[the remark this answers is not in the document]
TheJJ: I haven't really looked into how this works, but this is a built-in cmake feature. That said we have to investigate if we can place the configuration in an external file. Not sure if we shall do that now.

buildsystem/DependencyFetch.cmake
```diff
@@ -0,0 +1,79 @@
...
+# You need to specify the location with options of
+# ExternalProject_Add, for example:
+#
+# fetch_project(
+#     NAME yourmom
+#     GIT_REPOSITORY https://github.com/nevergonna/giveyouup
+#     GIT_TAG origin/master
+# )
+#
+function(fetch_project)
```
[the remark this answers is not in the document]
TheJJ: How can we pass the remaining arguments (the `GIT_TAG` and repo) to the `ExternalProject_Add`?

CMakeLists.txt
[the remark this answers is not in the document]
tusharpm: I was referring to `-checks=-*,readability-*`. The rest is, as you mentioned, CMake built-in.
I concur, a later change can address that.

libopenage/CMakeLists.txt
```diff
@@ -108,6 +109,91 @@ if(WANT_BACKTRACE)
...
+	"Disable the automatic update of subprojects over the internet"
+	OFF
+)
+
+# check if nyan was used as subproject before (and still exists)
+if(DEFINED NYAN_SUBPROJ_LOCATION AND IS_DIRECTORY "${NYAN_SUBPROJ_LOCATION}")
+	set(USED_NYAN_SUBPROJ TRUE)
+else()
+	set(USED_NYAN_SUBPROJ FALSE)
+	set(NYAN_SUBPROJ_LOCATION "" CACHE INTERNAL
```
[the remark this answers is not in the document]
TheJJ: OIh, that's right! So even if the folder is not there any more, it would be downloaded again. and this hopefully is what the user expects.

I'll remove this then, because if I reset `DOWNLOAD_NYAN`, it will never be downloaded in the first place.

buildsystem/DependencyFetch.cmake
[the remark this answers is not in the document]
tusharpm: The other arguments remain exactly the same as now: `${ARGN}` (through `@PROJ_UNPARSED_ARGUMENTS@`). Note that ARGN doesn't include named (positional) arguments.
Just the first argument to `ExternalProject_Add` changes to `@name@-fetch`.

PR #940, opened 2017-12-03
Debian Sid Build Instructions

chocoladisco:
Not using pip3 and incorporated the dependency in the apt-get line


chocoladisco:
Should fix the issues present in the PR #923

PR #949, opened 2017-12-29
Update README.md

alanyee:
Made explicit HTTPS calls and replace "OS X" with "macOS"


README.md
```diff
@@ -148,10 +148,10 @@ Guidelines:
 
 To prevent accidental violation of one of those guidelines, you should *never*
 
-* [learn git](http://git-scm.com/book/en/Git-Basics)
+* [learn git](https://git-scm.com/book/en/v1/Git-Basics)
```
TheJJ: I'd rather leave this without the version, then it should redirect to the latest version.

TheJJ:
I did the change in #951 (and retained your authorship in the commit).
Thank you!

PR #955, opened 2018-01-01
Require C++ compiler to support C++17 features

tusharpm:
Checked features:
- nested namespaces
- structured bindings


buildsystem/CheckCompilerFeatures.cmake
```diff
@@ -7,3 +7,27 @@ if(HAVE_THREAD_LOCAL_STORAGE)
...
+HAVE_REQUIRED_CXX17_SUPPORT
+)
+if(NOT HAVE_REQUIRED_CXX17_SUPPORT)
+	message("
+
+The compiler doesn't support required C++17 features.
+Please upgrade your compiler to build openage.
+
+")
+	message(FATAL_ERROR "aborting")
```
TheJJ: Should we display the features we require in the error message?

PR #956, opened 2018-01-02
[STALE] snapcraft packaging

gocarlos:


TheJJ:
What I wanted to avoid was the manual build of cmake :) Would it be a problem to require a newer (non-lts) ubuntu that has the required cmake and compiler version by default?

TheJJ:
The `DOWNLOAD_NYAN` option is intended for development only, when packaging, it should not be used. I can't tell how that would be done with snap, if there can be package dependencies?

simonsan:
@gocarlos Could you update this to some recent versions or should/can I take care of this?

gocarlos:
> @gocarlos Could you update this to some recent versions or should/can I take care of this?

sorry, I'm not working on this currently and I'm will not be able to work on it further in the near future...
You are welcome to take on this :D

simonsan:
> > @gocarlos Could you update this to some recent versions or should/can I take care of this?
> 
> sorry, I'm not working on this currently and I'm will not be able to work on it further in the near future...
> You are welcome to take on this :D

Thank you for your feedback. Will continue working on this then.

PR #959, opened 2018-01-10
update Homebrew-Cask installation instruction

nyx:
it seems this process has changed, or I'm missing something.

https://caskroom.github.io


TheJJ:
The sanity check failure is fixed by #960.

PR #936, opened 2017-11-20
Fixed minor typo in main_cpp.pyx

wanderrful:
Don't mind me, just fixing a typo.

PR #962, opened 2018-01-24
Minor fixes

TheJJ:
Some minor fixes.

Closes #936.


TheJJ:
To be honest I have no idea where I got them from, I changed those a long time ago and wanted to un-dirty my git tree.

PR #963, opened 2018-01-25
Make pylint shut up after update to 1.8.1

zuntrax:
This includes and thus closes #938.

* [x] Remove `ListIterator`, builtin function `iter` does the same
* [x] Rewrite `denote_last` to be more concise, disable warning
* [x] Add some explicit `return` statements
* [x] Modify control flow so that pylint understands it
* [x] Silence complexity warning in code we are going to shitcan anyway


zuntrax:
Sanity check fails because of old pylint version, new version understands `stop-iteration-return`. Can be solved through VM upgrade, but this will bring up other issues on master.

TheJJ:
VM was updated to latest archlinux.

PR #966, opened 2018-01-26
Link to Debian build instructions

heinezen:
Small fix in `building.md`.


TheJJ:
If you now have multiple mails in the git commit history, you need to add other addresses to the mailmap file.

heinezen:
@TheJJ Well, it's mostly my inability to properly use `git` as I just discovered that mail addresses can be different for individual repositories :D Should be good now, but I think I killed `kevin` in the process...

zuntrax:
@heinezen nope, wasn't you. mic-e's home router is a piece of shit and randomly closes all forwarded ports every now and then. needed a reset to fix it, as usual.

PR #975, opened 2018-02-16
Coord rewrite

TheJJ:
Finally, here's the working version of openage with the new coordinate system.

Please test if things work as desired.

This is a long awaited dependency for #850.

Closes #662 (which is @mic-e's pull for this).


heinezen:
The Ubuntu instructions need `python3-jinja2` (and the minumum version should be changed to 17.10).

Did only some minor testing and encountered this error

coord-error

The gathering area for resources is offset too much to the left. Happens when harvesting gold, wood, stone. Sheep are affected too I think, but it was hard to see.

TheJJ:
Fixed by removing a workaround that was there due to a bug in the old coord system.

Issue #745, opened 2017-02-18
Code Compliance automatic fixing


Tomatower:
The Code Compliance tester can automatically determine, where spaces are faulty inserted, so a script could easily fix them without having to go through every file manually - at least for a lot of often occurring issues.

castilma:
do you want `make check` to commit the changes, too? or just add them? or just make the changes?

TheJJ:
Hm, as a first step, I'd just add them by some rules we have to provide. This may get horribly complex if we need the AST of Python or C++. After the change, we should write the new file to `/tmp` and then print the diff to the original (`diff original /tmp/fixed`). If that looks good, we can in-place write the file with some `--write` option, or save the patch and `git apply` it.

TheJJ:
I ignored my safety-idea in #976 and update the copyright years directly. The secure modification might be a better choice for other compliance checks, though.

PR #976, opened 2018-02-17
codecompliance: automatic fixing for copyright years

TheJJ:
First step for #745.


buildsystem/codecompliance/__main__.py
```diff
@@ -137,17 +140,49 @@ def main(args):
...
     issues_count = 0
-    for title, text in find_all_issues(args, check_files):
+    for title, text, apply_fix in find_all_issues(args, check_files):
         issues_count += 1
         print("\x1b[33;1mWARNING\x1b[m {}: {}".format(title, text))
 
+        if apply_fix:
+            if args.fix:
+                print("        This will be fixed automatically.")
+                auto_fixes.append(apply_fix)
```
zuntrax: If `args.fix` is `False`, no fixes ever get appended to `auto_fixes`. This does not seem to be your intention, as you are checking for `not args.fix and auto_fixes` below for printing a message about possible fixes. I'd just omit checking `args.fix here` and always add the fix to the list and write "can be fixed". This would also make pylint happy (too many branches).

buildsystem/codecompliance/__main__.py
```diff
@@ -137,17 +140,49 @@ def main(args):
...
+
+    if args.fix and auto_fixes:
+        print("\x1b[33;1mApplying %d "
+              "automatic fixes...\x1b[m" % len(auto_fixes))
+
+        for auto_fix in auto_fixes:
+            print(auto_fix())
+            issues_count -= 1
+
+        print("")
```
zuntrax: `print()` vs `print("")`, see 10 lines above.

PR #982, opened 2018-02-24
Fixed tests on VS2017

schnema123:
Hello.

I was getting two failed tests at windows, because these functions were not implemented. I implemented them for windows. I dont see any problems with linux, since everything I was doing is #ifdef'd away on that platform. Please give feedback.

Greetings.


libopenage/util/compiler.cpp
```diff
@@ -73,8 +140,15 @@ std::string symbol_name(const void *addr, bool require_exact_addr, bool no_pure_
 
 bool is_symbol(const void *addr) {
 #ifdef _MSC_VER
-	// TODO: Get dladdr equivalent for MSVC.
-	return true;
+
+	auto symbol_name_result = symbol_name_win(addr);
+	if (symbol_name_result.initialization_fail)
```
tusharpm: Oh, I see why the `struct` was defined.  However, if that is changed, this check can still be done by elevating `initialized_symbol_handler_successfully` to the anonymous namespace.

libopenage/util/compiler.cpp
```diff
@@ -39,12 +45,73 @@ std::string addr_to_string(const void *addr) {
...
+	symbol_name_result result = {};
+	if (!initialized_symbol_handler_successfully)
+	{
+		result.initialization_fail = true;
+		return result;
+	}
+
+	// The magic of winapi
+	constexpr int name_buffer_size = 1024;
+	char buffer[sizeof(SYMBOL_INFO) + name_buffer_size * sizeof(char)];
```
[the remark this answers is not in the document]
schnema123: Hey. Yes it is true that sizeof(char) is defined as 1. Here is my intention on doing so: If I left out the sizeof part, imo it would not have been 100% clear why I am adding the buffer size to the buffer. I would have left a comment like "Allocate a buffer that can hold the size of the struct + a string of name_buffer_size". But by adding the sizeof(char), I think that it becomes immediatly clear that the thing I'm adding is meant to be a string of that size. Of course I can still change it. :)

schnema123:
Hey, ANSI color codes, which you use in logging, have to be enabled explicitly on windows machines, otherwise the console just prints rubbish. For me, the file I put this in does not seem ideal. Seems to me like too much mixing of platformagnostic with platform dependent code. Maybe it would be nice to have some platform-specific implementation files where we could put these things?

TheJJ:
We had a directory once where we intended to add plattform specific code. This turned out to be a bad idea: The structure of the project would have to be replicated inside that folder again, and it was really hard to decide and design which parts of the code should go in there and which are general enough.

After this did not work out nicely, we dropped the folder and added preprocessor switches at the right locations directly, which was much more simple and elegant to organize and implement. If there are too many switches, the function should be refactored anyways, which may be the case here for the color codes.

tusharpm:
I verified these changes work on my machine. The stack traces from libopenage are now legible. 😄
This can close #480, and consequently even #479.

@schnema123, the indentation and brace style is still different. Please correct that.
When you're ready, please rebase and squash the related commits, and remove the [WIP] prefix.

TheJJ:
Can you figure out why the enum does not work under windows? The implementation is quite straight-forward and doesn't use funny corner-cases. So It should work with msvc.

libopenage/util/misc.h
```diff
@@ -7,6 +7,7 @@
 #include <cstring>
 #include <string>
 #include <vector>
+#include <algorithm>
```
tusharpm: Can move to the top (alphabetical order).

libopenage/util/compiler.cpp
```diff
@@ -63,8 +122,8 @@ std::string symbol_name(const void *addr, bool require_exact_addr, bool no_pure_
 			return demangle(addr_info.dli_sname);
 		} else {
 			return util::sformat("%s+0x%lx",
-			                     demangle(addr_info.dli_sname).c_str(),
-			                     symbol_offset);
+				demangle(addr_info.dli_sname).c_str(),
+				symbol_offset);
```
TheJJ: the indent here was alright :)  tabs to reach the indent level of `return`, then spaces to align to the `(`.

libopenage/util/compiler.cpp
```diff
@@ -39,12 +46,64 @@ std::string addr_to_string(const void *addr) {
...
+		initialized_symbol_handler = true;
+
+		process_handle = GetCurrentProcess();
+		initialized_symbol_handler_successfully = SymInitialize(process_handle, nullptr, TRUE);
+	}
+
+	if (initialized_symbol_handler_successfully) {
+		// The magic of winapi
+		constexpr int name_buffer_size = 1024;
+		char buffer[sizeof(SYMBOL_INFO) + name_buffer_size * sizeof(char)];
```
TheJJ: the modern style here would be to use a `std::array<char>`.

Issue #616, opened 2016-09-28
Add quit button


TheJJ:
One should be able to quit the game by pressing an exit button. This shall be done in the menu mode.

vijeth-aradhya:
@TheJJ Hey, I'm new to open source and I want to start contributing. Could you please help me go about this issue?

TheJJ:
First, I can recommend reading our README and the [contributing intro](https://github.com/SFTtech/openage/blob/master/doc/contributing.md).

Then, try building the game and afterwards try to find the code section where the current menu is generated (maybe just search for the texts you see in clickable buttons).

When you found it, play around and google a bit to add a new button at the right place.
Finally, make the button do the right thing by figuring out how other buttons do their thing (just how, not what they do).

Then, in theory, you're done :)

marcobarbosa16:
@TheJJ Hey, I'm found just the Creation Mode, Editor Mode and Action Mode. There is a Menu Mode?

voidestructor:
If there is no one assigned to this issue, I would like to work on it. Can you assign it to me ?

jazzpi:
@voidestructor There's two open PRs: #620 and #621, but neither of them really work... You might want to ask the authors of those two first :wink:

TakesTheBiscuit:
@TheJJ / @ChipmunkV - upon quit we need to instruct all player units and buildings to stop moving/producing etc? or would we have the server do this in future?
so
```
quit
-> client disconnects and gracefully exits
-> server will stop all units and buildings owned by this player
-> other players continue like normal
```

heinezen:
@TakesTheBiscuit All the event related to the units and buildings will have to be canceled too. This probably will be the servers job because we can't always ensure that a player disconnects cleanly (e.g. when his game crashes).

TheJJ:
For this issue, and because we have no server part, I think its enough to just add a button that then calls a function where we can do more complicated things in the future. For now, it can just call `Engine::stop()`.

TheJJ:
Done in #990. We will have to improve this greatly to make it a 'disconnect from game' button instead of a engine killer.

PR #990, opened 2018-03-25
Add quit button

samuelgrigolato:
This is a third attempt to #616. It builds upon #620, #621 and this specific @TheJJ's comment:

> For this issue, and because we have no server part, I think its enough to just add a button that then calls a function where we can do more complicated things in the future. For now, it can just call Engine::stop().

### DIFF
--- a/copying.md
+++ b/copying.md
@@ -109,6 +109,7 @@ _the openage authors_ are:
 | Arne Sellmann               | PythonicChemist             | arne dawt sellmann à gmx dawt de                  |
 | Rafael X. Morales Georgi    | chocoladisco                | chocoladisco à gmail dawt com                     |
 | Marcel Schneider            | schnema123                  | marcelschneider5 à outlook dawt de                |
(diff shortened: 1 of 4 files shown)


copying.md
```diff
@@ -109,6 +109,7 @@ _the openage authors_ are:
 | Arne Sellmann               | PythonicChemist             | arne dawt sellmann à gmx dawt de                  |
 | Rafael X. Morales Georgi    | chocoladisco                | chocoladisco à gmail dawt com                     |
 | Marcel Schneider            | schnema123                  | marcelschneider5 à outlook dawt de                |
+| Samuel Grigolato            | samuelgrigolato             | samuel dawt grigolato à gmail dawt com                |
```
heinezen: The rightmost border of the table has to line up

samuelgrigolato: Fixed!

heinezen:
works for me and should for others as well :)

PR #991, opened 2018-03-26
doc: update Fedora build prerequisites for Fedora 27

AndersenJ:
Not sure what to do about nyan; there certainly isn't a fedora package for it, so right now I'm just going to assume Fedora users can follow the instructions given when configure is run.


AndersenJ:
Deleting this pull request; I'll make a new one that makes a clearer distinguishment between openage and nyan dependencies.

heinezen:
> Deleting this pull request; I'll make a new one that makes a clearer distinguishment between openage and nyan dependencies.

You don't necessarily need to do that. If a PR is changing in scope, you can just change the title and update the description.

Issue #994, opened 2018-04-12
cmake is unable to find PYTHON_LIBRARIES on macOS


Mrwerdo:
The problem: `cmake` is unable to find `PYTHON_LIBRARIES`
```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
PYTHON_LIBRARIES
    linked by target "cmTC_c3988" in directory /Users/mrwerdo/Developer/openage/.bin/clang++-debug-Oauto-sanitize-none/CMakeFiles/CMakeTmp

CMake Error at buildsystem/modules/FindPython.cmake:181 (try_compile):
  Failed to configure test project build system.
Call Stack (most recent call first):
  buildsystem/HandlePythonOptions.cmake:6 (find_package)
  CMakeLists.txt:141 (include)
```
Python 3.6.4 is installed using homebrew.

On lines 166-169 of buildsystem/modules/FindPython.cmake, the variables considered are:
```
PYTHON_INCLUDE_DIRS="/usr/local/opt/python/Frameworks/Python.framework/Versions/3.6/include/python3.6m"
PYTHON_LIBRARY_DIR="/usr/local/opt/python/Frameworks/Python.framework/Versions/3.6/lib"
PYTHON_LIBPL="/usr/local/opt/python/Frameworks/Python.framework/Versions/3.6/lib/python3.6/config-3.6m-darwin"
PYTHON_LIBRARY_NAME="Python.framework/Versions/3.6/Python"
```

Replacing `PYTHON_LIBRARY_NAME` with `Python` results in `find_library` finding a system package (`/System/Library/Frameworks/Python.framework`, which has out of date versions of python: `2.3`, `2.5`, `2.6` and `2.7`). This appears to work, but later on it fails linking - missing various python functions.

Finally, **running `./configure -- "-DPYTHON_LIBRARIES=$(python3-config --prefix)/lib/libpython3.6.dylib"` actually works**. `python3-config --prefix` expands to `/usr/local/opt/python/Frameworks/Python.framework/Versions/3.6`.

Perhaps `find_library` is failing to set `PYTHON_LIBRARIES` because `PYTHON_LIBRARY_DIR` has `lib` attached to the end of it?

Mrwerdo:
It turns out after a lot of digging around, that:

[...]

Issue #852, opened 2017-05-28
macOS event dispatching broken


TheJJ:
#828 figured out that for a successful mac run the event dispatching for mac needs to be changed.
This has to be fixed properly.

Birch-san:
Useful references…

- The [magic commit](https://github.com/ChipmunkV/openage/commit/d577f5681bf61e09cd7d8175661be2df389b9b4d).

- https://github.com/SFTtech/openage/pull/828#issuecomment-299688716:

> that magic partially disables input handling, so it needs some work: https://github.com/SFTtech/openage/issues/819#issuecomment-298831187

- https://github.com/SFTtech/openage/issues/819#issuecomment-298831187:

> [Birch-san] @ChipmunkV, I know you've said that the commit was removed because it was unexplainable. But where did you get the idea to do that? Did you read something somewhere?
>
> [ChipmunkV] @Birch-san, I knew that event loops in Qt use event dispatchers that have OS-specific implementation. Then you just disable the input handling part and try.
>
> I think that the issue is easily fixable by implementing and installing a custom dispatcher with [`QCoreApplication::setEventDispatcher()`](http://doc.qt.io/qt-5/qcoreapplication.html#setEventDispatcher).

Birch-san:
I actually had a stab at ChipmunkV's idea (installing a custom dispatcher):

https://github.com/SFTtech/openage/compare/master...Birch-san:set-event-dispatcher

I have no idea what I'm doing. At the time of writing: my custom event dispatcher compiles, and openage runs, but we still have a white screen followed by a black screen.  
I also haven't checked whether my event dispatcher is installed correctly. It needs to be installed _before_ you init the QCoreApplication. I _assume_ I've done that, but haven't actually verified.

A good reference for how to build a custom event dispatcher is here (MIT License, peper0@GitHub):

https://github.com/peper0/qtasio/blob/master/qasioeventdispatcher.cpp
https://github.com/peper0/qtasio/blob/master/qasioeventdispatcher.h

Note: peper0's code includes private headers such as:

```cpp
#include "private/qabstracteventdispatcher_p.h"
#include "private/qtimerinfo_unix_p.h"
```

We're not supposed to do that. It has the following message:

> This file is not part of the Qt API.  It exists purely as an
> implementation detail.  This header file may change from version to
> version without notice, or even be removed.
> 
> We mean it.

[...]

PR #995, opened 2018-04-13
macOS touch ups

Mrwerdo:
There are a few things in this pull request:

1. This reapplies the magic commit from #852 to making rendering work again.
2. Qt elements were flickering... they no longer are.
3. Python libraries wasn't being detected automatically. The solution in the commit is terrible but works. This is trying to solve this issue: #994 
4. Fixes a crash when closing the game window using the red x button (again on macs). A deconstructor had this instruction in its method body: [ud2](https://hjlebbink.github.io/x86doc/html/UD2.html)

### DIFF
--- a/buildsystem/modules/FindPython.cmake
+++ b/buildsystem/modules/FindPython.cmake
@@ -166,9 +166,12 @@ foreach(PYTHON ${PYTHON_INTERPRETERS})
 		py_get_config_var(LIBDIR PYTHON_LIBRARY_DIR)
 		py_get_config_var(LIBPL PYTHON_LIBPL)
 		py_get_config_var(LDLIBRARY PYTHON_LIBRARY_NAME)
-		find_library(PYTHON_LIBRARIES ${PYTHON_LIBRARY_NAME}
+		py_get_config_var(VERSION PYTHON_VERSION)
+
+		find_library(PYTHON_LIBRARIES "${PYTHON_LIBRARY_NAME}" "libpython${PYTHON_VERSION}.dylib"
 			PATHS "${PYTHON_LIBRARY_DIR}" "${PYTHON_LIBPL}"
 		)
+
 	endif()
 
 	# there's a static_assert that tests the Python version.
--- a/buildsystem/modules/FindPython.cmake
+++ b/buildsystem/modules/FindPython.cmake
@@ -1,4 +1,4 @@
-# Copyright 2015-2017 the openage authors. See copying.md for legal info.
+# Copyright 2015-2018 the openage authors. See copying.md for legal info.
 
 # Find Python
 # ~~~~~~~~~~~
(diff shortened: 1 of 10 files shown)


buildsystem/modules/FindPython.cmake
```diff
@@ -169,6 +170,19 @@ foreach(PYTHON ${PYTHON_INTERPRETERS})
...
+        # On macOS, the above was not working for homebrew installations.
+        # Insetad we inqurie `python3-config --prefix` for the installation,
+        # and find the location of the libraries.
+        if (PYTHON_LIBRARIES STREQUAL "PYTHON_LIBRARIES-NOTFOUND")
+            py_exec(
+                "from distutils.sysconfig import PREFIX; print(PREFIX)"
+                PYTHON_CONFIG_DIR
+            )
+            # TODO: Don't hardcode version numbers
+            set(PYTHON_LIBRARIES "${PYTHON_CONFIG_DIR}/lib/libpython3.6.dylib")
```
TheJJ: Please fix the indentation of this block (with tabs).
Why is the hardcoding neccessary?
I would suggest to investigate why `${PYTHON_LIBRARY_NAME}` is not found in `PYTHON_LIBRARY_DIR` where python distutils sysconfig tells us about.

 (also, typos in `insetad`, `inqurie`).

libopenage/game_control.h
```diff
@@ -56,6 +56,7 @@ class OutputModeSignals : public QObject {
 class OutputMode : public input::InputContext {
 public:
 	explicit OutputMode(qtsdl::GuiItemLink *gui_link);
+	virtual ~OutputMode() = 0;
```
TheJJ: setting it as a pure virtual function is not right. => remove the `= 0`

buildsystem/modules/FindPython.cmake
```diff
@@ -169,6 +170,19 @@ foreach(PYTHON ${PYTHON_INTERPRETERS})
...
+        endif()
+        message("!!! Using ${PYTHON_LIBRARIES}")
```
TheJJ: this is likely a leftover.

buildsystem/modules/FindPython.cmake
```diff
@@ -116,6 +116,7 @@ else()
 		"/usr/local/Frameworks/Python.framework/Versions/*/bin/python*"
 		"~/Library/Frameworks/Python.framework/Versions/*/bin/python*"
 		"/System/Library/Frameworks/Python.framework/Versions/*/bin/python*"
+		"/usr/local/opt/python/Frameworks/Versions/*/bin/python*"
```
TheJJ: That line should be the only thing needed to find python and its libs correctly, I hope.

Mrwerdo: Turns out this is pointless - both `/usr/local/Frameworks/Python.framework/Versions` and `/usr/local/opt/python/Frameworks/Versions` symlink to the same place.

Mrwerdo:
What's holding this up? Were you going to merge this?

TheJJ:
Ah, good. Please do a rebase in order to get rid of the merge commits. In this step you should also meld "Removed erroneous space" into "Ignore macOS .DS_store" and the copyright update commit into the commit that actually changed the contents. I.e. please clean up the history, then we can merge.

PR #996, opened 2018-04-20
fslike: add readwrite and append modes for opening files

TheJJ:
should be the missing piece for #749.

### DIFF
--- a/openage/util/fslike/test.py
+++ b/openage/util/fslike/test.py
@@ -1,4 +1,4 @@
-# Copyright 2017-2017 the openage authors. See copying.md for legal info.
+# Copyright 2017-2018 the openage authors. See copying.md for legal info.
 """
 Tests for the filesystem-like abstraction.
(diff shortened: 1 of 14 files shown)


openage/util/fslike/test.py
```diff
@@ -154,15 +154,42 @@ def test_case_ignoring(root_path, root_dir):
...
+        assert_value(fil.writable(), True)
+        fil.write(b"just")
 
     # check if changes went to known-name file
     with root_path["lemme_in"].open("rb") as fil:
-        assert_value(fil.read(), b"changedit")
+        assert_value(fil.read(), b"just")
+
+    # append some data
+    with root_path["lemme_in"].open("ab") as fil:
```
mic-e: Putting the various file-opening into `test_case_ignoring()` seems wrong, they should have their own method.

TheJJ:
done.

PR #1002, opened 2018-04-23
doc: number of projectiles per garrisoned unit

heinezen:
Info from ZeroEmpires [video and comments](https://www.youtube.com/watch?v=tCBlWAkanPY).


TheJJ:
I think the ci server got a connection reset while your build was running, so to trigger it again pls amend your commit and push again.

PR #1004, opened 2018-04-29
buildsystem: suppress useless cython warnings

zuntrax:
Cython always emits a warning when calling `cythonize`, no matter if the mentioned line is there or not:
```
Please put "# distutils: language=c++" in your .pyx or .pxd file(s)
```
They always bothered me, now they are gone.

### DIFF
--- a/buildsystem/cythonize.py
+++ b/buildsystem/cythonize.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 #
-# Copyright 2015-2017 the openage authors. See copying.md for legal info.
+# Copyright 2015-2018 the openage authors. See copying.md for legal info.
 
 """
 Runs Cython on all modules that were listed via add_cython_module.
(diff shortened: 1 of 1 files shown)


buildsystem/cythonize.py
```diff
@@ -11,6 +11,49 @@
...
+        self.buf += data
+        lines = self.buf.split('\n')
+        for line in lines[:-1]:
+            if line not in self.filtered:
+                self.stream.write(line + '\n')
+        self.buf = lines[-1]
+
+    def flush(self):
+        """ Flushes output stream """
+        self.stream.flush()
```
TheJJ: i think we can [remove this method](https://docs.python.org/3/reference/datamodel.html#object.__getattr__)

buildsystem/cythonize.py
```diff
@@ -11,6 +11,49 @@
...
+
+    def close(self):
+        """
+        Writes buffer to output stream so incomplete lines at the end don't get
+        lost. Make sure this gets called, e.g. with contextlib.closing.
```
TheJJ: we could add a `__exit__` (and `__enter__`) function, so that no contextlib is needed but we can regularly do `with CythonFilter() as filter:`

zuntrax: I don't really like how `close` doesn't really close the filter, so implementing a context manager is probably better.

PR #1014, opened 2018-05-05
[WIP] buildsystem: enable cython cache

zuntrax:
This should speed up our builds a bit more.


zuntrax:
Errr nope. Maybe later.

PR #1021, opened 2018-05-17
Modding API draft (data)

heinezen:
Visualizes the possible structure of the AoE2 game data after it is converted to nyan. I chose to use UML logic for the diagram, but even if you don't know UML it should be quite understandable. Objects that are pink can likely be used for AoE1/SWGB too and are candidates for engine API objects.

Related:
* Modding API draft (scripting) (#1076)
* Modding API draft (UI) (#1077)

To view the UML source file use [UMLet](https://github.com/umlet/umlet).

Drafts are [on the blog](https://blog.openage.sft.mx) and on https://pad.stusta.de/p/openage-mod-api


heinezen:
Alright, from my perspective this is finished. Feel free to review it because I'm sure I forgot at least one or two things. The diagram is still missing a definition of HUD elements, but I think we can make a separate UML representation for them.

PR #1022, opened 2018-05-17
This is the new WIP renderer

surajhanchinal:
This is the new renderer I have been working on which is meant to be independent from the game state system and is more efficient than the current renderer. This is based on Vtec234's new-renderer branch which provided the OpenGL framework for this renderer. The build instructions are the same as the main game but there might be a Eigen3 dependency. That can be easily resolved by
```
sudo apt-get install libeigen3-dev
```
the code currently resides as a demo in and can be run by
```
./run  test -d renderer.batch_test.batch_demo 0
```
The main code that is being executed is in `/libopenage/renderer/batch_test.cpp`. You can go ahead and change what you want to display on the screen in there. So the basic unit for displaying anything on the screen is the `opengl::Sprite_2` class.

Here is a brief documentation on how to use this 
So you create an object with the constructor. It takes x,y,w,h,r,g,b,a in that order as arguments where x and y are positions and w and h are the width and height.  r,g,b,a are the colors . This renders a square with the given parameters on the screen. Then there are 2 helper functions for the `Sprite_2.set_texture(int id, bool metafile)`. The id is chooses the texture to be used. The ids are decided by the number of the textures in the `assets/converted/graphics` folder. For example, an elephant moving texture has an id 805 since its filename is `805.slp.png`. The second argument is basically choosing if you want to use texture description file which contains subtexture coordinates.

The next function is `set_subtex(int)` which chooses the subtexture from the texture and this function just takes the index of the subtexture as input.

Once the sprite is created, it needs to be submitted to the renderer in order for it to be shown on the screen. this is done by the `renderer->submit()` function which takes in the sprite as an argument. The submit must be placed between `renderer->begin()` and `renderer->end()` and before `renderer->render()`.

There are various commented parts of the code and you can uncomment them and run and experiment.

Please test this renderer and report the performance you are getting and any suggestions. The frame rate can be seen in the terminal that you run the command. :)

Edit:
You can also render terrain but I've coded the renderer to use aoe2hd textures. If you do have the game, then copy the `terrain` folder from the game installation and place it in `/assets/` folder. And then for convinience and for indexing please rename the textures in `/assets/terrain/textures` from `1.png` to whatever `x.png` that you might want to use. Now to access these textures you would use another helper function of `opengl::Sprite_2` named `set_terrain(int)`. If you want to use `x.png`, you set the argument of the function `set_terrain()` as `6000+x`. I agree it's a little hackish but since the game asset system is being rewritten I did not want to integrate with the old asset manager, hence this problem.


TheJJ:
I fear this branch is quite messed up, your `first commit` seems to commit the whole repo state again and thus is very large (>20MB).

surajhanchinal:
I am planning to merge this with Vtec's renderer as soon as possible so that there is just a single pull request for the renderer. That should reduce the confusion.

TheJJ:
Ah, great! It would be nice indeed if Vtec could merge your improvements into his branch so he maintains the main "new renderer" pull request.

TheJJ:
I'll close this in favor of #850, where the code of this pull went into anyway.

Issue #1001, opened 2018-04-22
Lazy log message formatting


TheJJ:
Our logging system does not support lazy message formatting, like the python `logging` module.

``` python
from log import info

# current version: always performs string formatting
var = "with formatting"
info("some message %s" % var)

# new version: should only format the string if the message is actually emitted
var = "lazyness"
info("everything is better™ with ", var)
```

The functions are in `openage/log/__init__.py` and then relay the messages to `libopenage`.
The formatting of messages should thus happen (either in C++ or the python part, not sure) **only if** the log level of the message is higher than the minimum log level of a sink that gets the message. 

This is an optimization, so no real priority :wink:

kev946:
Are we just trying to do lazy formatting on the messages sent from the Python side?

TheJJ:
We can do both, in C++ and Python. The C++ `operator <<` could be abused for that, and we would have to store references to the message parts, to merge them later if the message has to be formatted.

We could add that C++ (which controls the loglevel) notifies Python of the current level, so messages are not even transported C++ if the message is not important enough.

TheJJ:
I added a possible approach section above, with the main idea to create a backend for our C++ logger for the python `logging` module.

kev946:
For the Python side, maybe we could create a `CppHandler` class derived from [`Handler`](https://docs.python.org/3/library/logging.html#logging.Handler). We override [`setLevel`](https://docs.python.org/3/library/logging.html#logging.Handler.setLevel) and [`emit`](https://docs.python.org/3/library/logging.html#logging.Handler.emit) to call into the C++ side. Does this sound good?

TheJJ:
Yes, except that we have to call `setLevel` from the C++ side so the C++-level is set in the python world.

PR #1025, opened 2018-06-04
[WIP] Lazy Log Message Formatting

kev946:
Addresses #1001


kev946:
These are only the Python side changes. I am still trying to figure out how to do the C++ side changes.

openage/__init__.py
```diff
@@ -40,3 +42,14 @@
...
+# setup logging
+# do not overwrite any of the predefined levels
+# https://docs.python.org/3/library/logging.html#logging-levels
+logging.addLevelName(1, "MIN")
+logging.addLevelName(5, "SPAM")
+logging.addLevelName(51, "MAX")
+
+logging.setLoggerClass(CppLogger)
+logger = logging.getLogger('openage')
+logger.addHandler(CppHandler())
```
TheJJ: Can we do the setup in the `log` python module?
So we could just import `log.setup` (or some better name) here and then call `log_setup()`, because then the logging code is kept in the right module (if the `logging.getLogger` needs to be passed, we can just give it to `log_setup` as argument).

openage/cabextract/cab.py
```diff
@@ -355,8 +356,7 @@ def read_file_headers(self, cab):
         # seek to the correct position
         if cab.tell() != self.header.coffFiles:
             cab.seek(self.header.coffFiles)
-            dbg("cabfile has nonstandard format: seek to header.coffFiles " +
-                "was required")
+            logging.debug("cabfile has nonstandard format: seek to header.coffFiles was required")
```
TheJJ: `logger`

openage/cabextract/cab.py
```diff
@@ -255,7 +256,7 @@ def __init__(self, cab):
             header.next_cab = try_decode(read_nullterminated_string(cab))
             header.next_disk = try_decode(read_nullterminated_string(cab))
 
-        dbg(header)
+        logging.debug(header)
```
TheJJ: `logger`

openage/convert/blendomatic.py
```diff
@@ -88,7 +89,7 @@ def __init__(self, idx, data_file, tile_count, header):
         self.pxcount = header[0]
         # tile_flags = header[1:]  # TODO what do they do?
 
-        dbg("blending mode %d tiles have %d pixels" % (idx, self.pxcount))
+        logging.debug("blending mode %d tiles have %d pixels", idx, self.pxcount)
```
TheJJ: `logger`

openage/convert/slp_converter_pool.py
```diff
@@ -115,7 +114,7 @@ def convert(self, slpdata, custom_cutter=None):
         self.idle.put((inqueue, outqueue))
 
         if isinstance(result, BaseException):
-            err("exception in worker process: %s" % result)
+            logging.error("exception in worker process: %s", result)
```
TheJJ: `logger`

openage/log/log.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class CppHandler(Handler):
+    cpp_levels = {}
+
+    def __init__(self, level=logging.NOTSET):
+        super().__init__(level)
+
+    def setLevel(self, level):
+        cpp_level = self.cpp_levels.get(level, None)
```
TheJJ: how is `self.cpp_levels` ever assigned a value?

kev946:
`logger = logging.getLogger('openage')` gets a reference to the global 'openage' logger object. It has the `CppHandler` added to it which is what calls into the C++ side for actual logging. Adding that line to each module that does logging is the recommended way. But if that is redundant, we could create a wrapper like:

```python
# openage/log/__init__.py
import logging

logger = logging.getLogger('openage')

def info(msg, *args, **kwargs):
    logger.info(msg, args, kwargs)

# more log functions below
...
```

TheJJ:
I think that wouldn't be too bad, then we could keep the current log function invocations (but could still use the standard logging module). So we could add such wrapper functions.

Are messages from imported modules (numpy, pillow, whatever) also logged to the correct logger which sends to C++ then? Or does that only work with the `openage` logger?

kev946:
Only the 'openage' logger will relay mesages to C++.

kev946:
The only way to do this is to set the `CppLogger` as the root logger. Now if external libraries do `logger = logging.getLogger()`, then they will get a reference to the `CppLogger` by default.

TheJJ:
I just peeked in [the documentation](https://docs.python.org/3.6/library/logging.html#logging.debug) and it seems that when we add the `CppLogger`-`Handler` to the root logger with then also `logging.info`, `.error`, ... invocations use the root logger. So your `ROOT_LOGGER.` is probably not necessary.

I think that's a good solution you took now :)

kev946:
Not sure if there is a great way to do this that is not a hack. Basically the `MessageBuilder` base class `StringFormatter` would need its `operator <<`s to know about the current `msg.lvl` (in`MessageBuilder`) and then call into the logging system to check if any of the `LogSink`s have `msg.lvl >= sink->loglevel`. If that is false, then the `operator <<`s just return and do nothing.

TheJJ:
Indeed. I don't think we can prevent the `MessageBuilder` to be created, because the arguments are all evaluated first before the log function is called.

That the `MessageBuilder` can foresee where it will be logged and check the level is a problem indeed, we'd have to store the "minimum required level" all registered `LogSink`s currently have and use this value to let `operator <<` of `MessageBuilder` do nothing, just like you said.

PR #1029, opened 2018-06-23
Lazy Log Message Formatting (C++ Side)

kev946:
Addresses #1001


kev946:
Should be good now. We can use `std::list::remove`.

TheJJ:
Great! The next step would be having better filtering, that is if one suspects a bug in a specific subsystem, only messages of that system should be emitted more verbosely.
We can do this similar to pythons `logging`, the messages are already tracking their origins, which we can use to set filters for.

Issue #715, opened 2017-01-22
Provide support for building openage under a virtualenv


samkellett:
Bug to document issues uncovered when building under a virtualenv.

The first issue found from my attempt is that virtualenv ships it's own site.py which doesn't have getsitepackages defined. This appears to be a known issue for python 2.7 (see pypa/virtualenv#355) -- I can't see much online about this still being the case in python 3.6 but it definitely seems to be for me.

TensorFlow had this problem and fixed it by using the inspect module instead of site, [see here](https://github.com/tensorflow/tensorflow/pull/702) for details.

samkellett:
Another potential alternative to getsitepackages: http://stackoverflow.com/a/29062571/649140

TheJJ:
Fixed in #1031

PR #1031, opened 2018-06-29
Added work around getsitepackages Issue #715

akshit-sharma:
Support for building openage under a virtualenv. Issue #715

### DIFF
--- a/buildsystem/HandlePythonOptions.cmake
+++ b/buildsystem/HandlePythonOptions.cmake
@@ -1,4 +1,4 @@
-# Copyright 2015-2017 the openage authors. See copying.md for legal info.
+# Copyright 2015-2018 the openage authors. See copying.md for legal info.
 
 # finds the python interpreter, install destination and extension flags.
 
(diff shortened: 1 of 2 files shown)


buildsystem/HandlePythonOptions.cmake
```diff
@@ -29,7 +29,7 @@ if(NOT CMAKE_PY_INSTALL_PREFIX)
 	if(MSVC)
 		set(CMAKE_PY_INSTALL_PREFIX "python")
 	else()
-		py_exec("import site, os; print(os.path.normpath(site.getsitepackages()[0]))" PREFIX)
+		py_exec("import os; print(os.path.dirname(os.__file__) + '/site-packages')" PREFIX)
```
TheJJ: This breaks the installation for distribution packaging because there the destination path must be the sitepackages directory. What is the intention of this fix, or rather: Why doesn't it work with a virtualenv currently? To solve this correctly, we'll need a more complete solution.

TheJJ: Please excuse me, you are right, i did not realize the `os.__file__`'s effect.

Maybe we can improve it further to not depend on this "os.py is one folder above site packages":
https://stackoverflow.com/a/122340/1348535
so we would have:
```python
from distutils.sysconfig import get_python_lib; print(get_python_lib())
```

akshit-sharma:
```from distutils.sysconfig import get_python_lib; print(get_python_lib())``` also works for me. 

I have updated my pull request.

TheJJ:
Thank you!

PR #1042, opened 2018-08-04
libopenage: add missing include directives

tusharpm:
Add missing include directives that caused compile errors locally.
This should not affect other platforms.

### DIFF
--- a/libopenage/event/eventstore.h
+++ b/libopenage/event/eventstore.h
@@ -6,6 +6,7 @@
 #include "../util/misc.h"
 
 #include <memory>
+#include <unordered_map>
 
 
 namespace openage::event {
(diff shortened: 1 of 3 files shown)


TheJJ:
I really wonder why those include bugs are so different from machine to machine.. Is the standard library so different on other platforms so that headers are transitively included on Linux-GCC but on others they're not?

tusharpm:
A reasonable library implementation should provide (mostly) portable headers.
I'm not sure if Kevin is testing the build with both `libstdc++` and `libc++` though.

TheJJ:
No it's not, both runs are with `libstdc++`, `clang` isn't told to use `libc++`. Not sure if it would be worth it when you're such a good bug hunter :D

Issue #1051, opened 2018-09-29
std::unary_function removed in C++17


Fish201809:
I tried to compile the source code in vs2017, But something went wrong.
The error is as follows. Why did this error occur?

“unary_function”: not the member of “std”

heinezen:
This is a function removed in C++17.

https://en.cppreference.com/w/cpp/utility/functional/unary_function

Seems like `gcc` and `clang` haven't removed it yet, but VS2017 did. That's why it slipped under our radar.

liuping1997:
Maybe u need this patch code?
```
#if _MSC_VER >= 1910
namespace std{
template <class Arg, class Result>
struct unary_function
{
    typedef Arg argument_type;
    typedef Result result_type;
};

template <class Arg1, class Arg2, class Result>
struct binary_function
{
    typedef Arg1 first_argument_type;
    typedef Arg2 second_argument_type;
    typedef Result result_type;
};
};
#endif
```

mic-e:
No. MSVC is right, we should not work around it. openage is a C++17 project, so it should not use `unary_function`.

Instead, we need to change `eventqueue.h` to use the more modern `std::function` instead of `std::unary_function`.

PR #1056, opened 2018-10-11
Remove std::unary_function usage

kev946:
Addresses #1051


kev946:
I'm not sure if this is right, but I was basing the removal off of this [stack overflow comment](https://stackoverflow.com/questions/609937/what-is-the-benefit-of-inheriting-from-stdbinary-function-or-stdunary-funct#comment20070913_609937).

TheJJ:
Thanks! I think it's safe to just remove the definition.

PR #1072, opened 2018-12-07
doc: fix table formatting in terrain.md.

goto-bus-stop:
Github's new parser understands the spaces at the start of these table rows to be a code block. Moving them to the start of the line fixes the issue; I added `---:` to the header column to right-align these numbers in the rendered document.


TheJJ:
Thank you :)

PR #1078, opened 2018-12-24
renderer: add support for OpenGL uniform blocks in shader_program

Vtec234:
This adds partial support for uniform buffers in the renderer. Namely, the OpenGL backend can parse and extract all information out of shaders containing uniform blocks, but I didn't add any way to actually set the block contents. This is intentional, because I'm not sure what the best interface for that would be, so let's leave it until it's needed.


libopenage/renderer/opengl/shader_program.cpp
```diff
@@ -74,16 +73,111 @@ GlShaderProgram::GlShaderProgram(const std::vector<resources::ShaderSource> &src
...
 	size_t unif_count = val;
 	glGetProgramiv(handle, GL_ACTIVE_UNIFORM_MAX_LENGTH, &val);
-	size_t unif_maxlen = val;
+	max_name_len = std::max(size_t(val), max_name_len);
+	glGetProgramiv(handle, GL_ACTIVE_UNIFORM_BLOCKS, &val);
+	size_t unif_block_count = val;
+	glGetProgramiv(handle, GL_ACTIVE_UNIFORM_BLOCK_MAX_NAME_LENGTH, &val);
+	max_name_len = std::max(size_t(val), max_name_len);
+
+	std::vector<char> name(max_name_len);
```
tusharpm: minor: can use `std::string` and avoid a copy below.

Vtec234: I don't think I can use `string`, because its implementation keeps a separate length, while this vector is modified by c-style GL functions which just write into it and add a null-terminator at the end. If the written name is shorter, the null-terminated string will have a different length than that stored by `string`, which doesn't sound correct.

tusharpm: The string can have the same semantics as this vector (null-initialized with max-length).
It's alright. I don't think the optimization is worth the confusion in this case.

PR #1082, opened 2019-01-26
buildsystem: use GNU gold linker by default

TheJJ:
This speeds up the linking part because [gold is faster](https://en.wikipedia.org/wiki/Gold_(linker)).

This patch also adds support splitting DWARF information into separate files.


janisozaur:
Why not lld though?

TheJJ:
llvm's ld? sure, add support for it :) I didn't know it until just now.

PR #1085, opened 2019-01-31
Fix windows build

### LINKED ISSUE #1062: Two build errors in win10 vs2017
error 1:
    3>c:\program files (x86)\microsoft visual `g:\workstation\current\openage\libopenage\log\logsink.cpp(66): error C2039: 'min': is not a member of 'std'`

The error was resolved when I tried to include `#include <algorithm>`

error 2:
  `  LINK : fatal error LNK1104: cannot open file 'pthread.lib'`

I try to run the command [`vcpkg install pthread`], But mistakes remain, I'm not sure how to correct the cmake file or something else

tusharpm:
Summary:
- Improve path handling in `buildsystem/pxdgen.py` that fails to match using `str.startswith`.
- Improve thread library target to use instead of `pthread`. Fixes #1062.
- Decorate the DLL interfaces used by the cython extensions with `OAAPI`, which otherwise cause link errors.

### DIFF
--- a/buildsystem/pxdgen.py
+++ b/buildsystem/pxdgen.py
@@ -1,4 +1,4 @@
-# Copyright 2015-2018 the openage authors. See copying.md for legal info.
+# Copyright 2015-2019 the openage authors. See copying.md for legal info.
 
 """
 Auto-generates PXD files from annotated C++ headers.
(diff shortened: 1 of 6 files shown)


TheJJ:
Maybe we can add a sanity check to enforce `OAAPI` for functions that are pxd-annotated, I think that would help forgetting those (as long as we don't have windows ci).

tusharpm:
I'll try to patch something up in codecompliance.

TheJJ:
awesome! :grinning:

PR #1087, opened 2019-02-03
codegen: take the realpath of generated file paths

TheJJ:
this is then safe against symlinks


openage/codegen/codegen.py
```diff
@@ -166,8 +166,11 @@ def codegen(mode, input_dir, output_dir):
             err("unknown codegen mode: %s", mode)
             sys.exit(1)
 
-    generated = {os.path.abspath(path).decode() for path in generated}
-    depends = {os.path.abspath(path) for path in get_codegen_depends(wrapper)}
+    def path_sanitize(path):
+        return os.path.realpath(os.path.abspath(path))
```
mic-e: `os.path.abspath` is not needed here

PR #1089, opened 2019-02-26
buildsystem: use run.py from the build directory

tusharpm:
- disallow running `run.py.in` directly
- fix `cleancython` target
- skip creating `__init__.p{xd,y}` in the top-level binary directory


TheJJ:
thanks :)

PR #1103, opened 2019-04-27
Updating some code using C++17

tusharpm:
- constexpr_map: complete rewrite using `std::array` and relaxed `constexpr` functions.
- make special members for the coordinate base classes use `default`.


TheJJ:
thx :)

Issue #468, opened 2016-01-12
Consistently use openage::util::hash_combine()


lyle-nel:
I have tracked down 3 places that use a different hash implementation than that defined in `libopenage/util/hash.h`. However, both implementations seem to serve the same purpose, that is, combine two hashes.

For the sake of consistency it might be good to reuse that function instead of having two implementations.

The places in question are:

https://github.com/SFTtech/openage/blob/master/libopenage/coord/phys3.h#L72-L74
https://github.com/SFTtech/openage/blob/master/libopenage/coord/tile.h#L75-L79
https://github.com/SFTtech/openage/blob/master/libopenage/pathfinding/path.h#L207-L209

okovko:
Boost does not advise using its hash function as a general purpose hash function

http://www.boost.org/doc/libs/1_49_0/doc/html/hash/rationale.html

TheJJ:
`if you do need a higher quality hash function, then neither the standard hash function or boost::hash are appropriate`, but we just wanna have a high probablility of non-colliding hashes for any sort of container.

For any crypto needs, we will go with `libsodium` and of couse not with those non-cryptographic hashes.

okovko:
As far as integrals are hashed, the output of hash is simply the input you gave it anyways.

The Boost hash_combine() should serve that purpose (high probability of non-colliding hashes) for integral pairs. 0x9E3779B9 is 2 ^ 32 / phi; a 32 bit binary expansion of an irrational number provides 32 uniform bits with no simple correlation. The xor will generate distinct hashes for similar inputs because it smudges the leading and trailing bits of hash2 + inv_phi_32 by adding the shifts of hash1 + inv_phi_32.

Note that Boost's hash_combine() expands to

hash1 + inv_phi_32    ^    hash2 + inv_phi_32 + ( ( hash1 + inv_phi_32 ) << 6 ) + ( ( hash1 + inv_phi_32 ) >> 2 )

Whereas the openage::util::hash_combine() expands to

hash1    ^    hash2 + inv_phi_32 + ( hash1 << 6 ) + ( hash1 >> 2 )

Here, hash1 has little effect on the output of the xor, and there's generally less bit shuffling. I would think there will be more collisions relative to the Boost hash_combine(), and the distribution of pairs -> hashes will be whacky? I could be missing a specific reason it's written this way.

[...]

PR #1110, opened 2019-05-25
[WIP] Consistently use openage::util::hash_combine()

simonsan:
Fix for #468, Early WIP pull request for CI.


libopenage/coord/tile.h
```diff
@@ -1,11 +1,13 @@
...
+// Copyright 2016-2019 the openage authors. See copying.md for legal info.
 
 #pragma once
 
 #include "declarations.h"
 #include "coord_nese.gen.h"
 #include "coord_neseup.gen.h"
 #include "../util/misc.h"
+#include "../util/hash.h"
+#include <typeindex>
```
TheJJ: when we introduce the helper function, we don't need the typeindex header anymore here.

libopenage/coord/phys.h
```diff
@@ -74,11 +77,11 @@ namespace std {
 template<>
 struct hash<openage::coord::phys3> {
 	size_t operator ()(const openage::coord::phys3 &pos) const {
-		return (
-			openage::util::rol<size_t, 2>(hash<openage::coord::phys_t>{}(pos.ne)) ^
-			openage::util::rol<size_t, 1>(hash<openage::coord::phys_t>{}(pos.se)) ^
-			                              hash<openage::coord::phys_t>{}(pos.up)
-		);
+		size_t hash = std::hash<std::type_index>()(std::type_index(typeid(openage::coord::phys3)));
```
TheJJ: as we invoke this function multiple times over the codebase and it seems quite verbose, we could add a helper function to `../util/hash.h` that looks like:
``` cpp
template <typename T>
type_hash() {
    return std::hash<std::type_index>()(std::type_index(typeid(T)));
}
```

libopenage/coord/phys.h
```diff
@@ -1,13 +1,16 @@
...
 
 #include "../util/misc.h"
+#include "../util/hash.h"
 
 #include "declarations.h"
 #include "coord_nese.gen.h"
 #include "coord_neseup.gen.h"
 
+#include <typeindex>
+
```
TheJJ: when the helper function is added, the typeindex header is in `util/hash.h` and no longer needed here.

libopenage/pathfinding/path.h
```diff
@@ -1,16 +1,19 @@
-// Copyright 2014-2018 the openage authors. See copying.md for legal info.
+// Copyright 2014-2019 the openage authors. See copying.md for legal info.
 
 #pragma once
 
 #include <functional>
 #include <memory>
 #include <unordered_map>
 #include <vector>
+#include <typeindex>
```
TheJJ: also no longer needed with the helper

libopenage/util/hash.cpp
```diff
@@ -11,7 +11,7 @@ namespace util {
 
 size_t hash_combine(size_t hash1, size_t hash2) {
 	// Taken from http://www.boost.org/doc/libs/1_55_0/doc/html/hash/reference.html#boost.hash_combine
-	return hash1 ^ (hash2 + 0x9e3779b9 + ((hash1 << 6) + (hash1 >> 2)));
+	return ( hash1 + 0x9e3779b9 ) ^ ( hash2 + 0x9e3779b9 + ( ( hash1 + 0x9e3779b9 ) << 6 ) + ( ( hash1 + 0x9e3779b9 ) >> 2 ) );
```
TheJJ: in the new version, the implementation for `size_t` is still the same, what you inserted here is the int32 one:
* https://www.boost.org/doc/libs/1_70_0/doc/html/hash/reference.html#boost.hash_combine
* https://www.boost.org/doc/libs/1_70_0/boost/container_hash/hash.hpp

The correct one is this (which is the one we had before)`:
``` cpp
        template <typename SizeT>
        inline void hash_combine_impl(SizeT& seed, SizeT value)
        {
            seed ^= value + 0x9e3779b9 + (seed<<6) + (seed>>2);
        }
```

simonsan: changed it back

libopenage/coord/phys.h
simonsan: Done, juchu!

PR #1114, opened 2019-05-27
Changed TODO to \todo for doxygen Todolist export

simonsan:
kT


TheJJ:
we decided not to do this. instead, we want to filter the files with `sed` and replace our TODOs with ones that doxygen can detect.

PR #1123, opened 2019-06-15
[WIP] Add DLLs to EmbedWinDependencies

simonsan:
`Qt5Widgets.dll` was missing in the installer.

The installer still misses some Qt-files though. Still diagnosing which ones.

EDIT: Somehow it copies `Qt5Widgets` now, but not qtquickcontrolsplugin.dll and qwindows.dll

For qwindows.dll I had to manually make a subfolder in openage and put qwindows.dll inside, then it is starting up. But still missing the following deps:
`QtQuick 2.4`
`QtQuick.Controls 1.1`
`QtQuick.Layouts 1.1`
`QtQuick.Controls.Styles`

As tu already said we should use [qt-deploy.ps1](https://github.com/Microsoft/vcpkg/blob/master/ports/qt5-base/qtdeploy.ps1) and [applocal.ps1](https://github.com/Microsoft/vcpkg/blob/master/scripts/buildsystems/msbuild/applocal.ps1) to deploy the subfolder structure of Qt that we need on the target device.

### DIFF
--- a/buildsystem/scripts/EmbedWinDependencies.cmake
+++ b/buildsystem/scripts/EmbedWinDependencies.cmake
@@ -1,4 +1,4 @@
-# Copyright 2017-2017 the openage authors. See copying.md for legal info.
+# Copyright 2017-2019 the openage authors. See copying.md for legal info.
 
 if(NOT forward_variables)
 	message(FATAL_ERROR "CMake configuration variables not available. Please include(ForwardVariables.cmake)")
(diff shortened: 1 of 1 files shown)


simonsan:
Update: We now use `applocal.ps1` directly to resolve the Qt dependencies. The structure in the installer is a bit off. Applocal copies all the dependencies, but into `.\bin`and from there into a subfolder structure like `.\bin\plugins\platforms`. So we need to figure out how to let the right paths be known. One way could be a `qt.conf` at the right place, if there is no other idea.

PR #1124, opened 2019-06-18
Added windows installer reference to various locations

simonsan:
I've referenced the windows installer in various locations to bring more attention to it for people that want to play around with openage while not wanting to build it for themselves.


simonsan:
> thx. maybe we could merge the windows information in the main README, as we now have two blocks.

Yep, I thought about that, too. But didn't want to make "Dependencies, Building, Running, Installing" out of it? Or how and where would you imagine it?

PR #1126, opened 2019-06-18
[WIP] Using applocal.ps1 for resolving and deploying Qt dependencies

simonsan:


simonsan:
Check #1127, could fix issue with my git.

PR #1127, opened 2019-06-18
Using applocal.ps1 for resolving and deploying Qt dependencies

simonsan:


buildsystem/scripts/EmbedWinDependencies.cmake
```diff
@@ -37,19 +52,47 @@ function(resolve binary)
...
+if((DEFINED POWERSHELL_COMMAND) AND (DEFINED APPLOCAL_SCRIPT))
+	foreach(file ${CMAKE_INSTALL_MANIFEST_FILES})
+		if(file MATCHES "\\.dll$")
+				# DEBUG
+				# message(STATUS "Use this command manually for diagnosing: ${POWERSHELL_COMMAND} ${APPLOCAL_SCRIPT} -targetBinary ${file} -installedDir ${vcpkg_dir}/bin -tlogFile ${BUILDSYSTEM_DIR}/applocal.log -copiedFilesLog ${BUILDSYSTEM_DIR}/applocal_copied_files.log!")
+
+				execute_process(COMMAND "${POWERSHELL_COMMAND}" "${APPLOCAL_SCRIPT}"
+						-targetBinary "${file}"
+						-installedDir "${vcpkg_dir}/bin"
+						-tlogFile "${BUILDSYSTEM_DIR}/applocal.log"
```
tusharpm: We should avoid writing to the source directory, even for packaging (`BUILDSYSTEM_DIR` being a subdirectory).
Maybe we can add a designated directory for the packaging logs, or write to the binary directory.

Same for the few lines below.

buildsystem/scripts/EmbedWinDependencies.cmake
```diff
@@ -37,19 +52,47 @@ function(resolve binary)
...
+						-copiedFilesLog "${BUILDSYSTEM_DIR}/applocal_copied_files.log"
+						ERROR_VARIABLE "${BUILDSYSTEM_DIR}/applocal_error.log"
```
tusharpm: `ERROR_FILE`.

simonsan: done

buildsystem/scripts/EmbedWinDependencies.cmake
simonsan: done

buildsystem/scripts/EmbedWinDependencies.cmake
```diff
@@ -11,6 +11,21 @@ if(NOT dumpbin)
...
+set(APPLOCAL_SCRIPT "${vcpkg_dir}/../../scripts/buildsystems/msbuild/applocal.ps1")
+set(POWERSHELL_COMMAND "C:/Windows/System32/WindowsPowerShell/v1.0/powershell.exe")
+
+# Check if applocal exists
+if(NOT EXISTS ${APPLOCAL_SCRIPT})
+	unset(APPLOCAL_SCRIPT)
+endif()
+
+# Check if powershell exists
+if(NOT EXISTS ${POWERSHELL_COMMAND})
```
tusharpm: Micro-optimization: The two checks can be moved in the `if` below, instead of unsetting the variable here and checking for `DEFINED` later.

buildsystem/scripts/EmbedWinDependencies.cmake
```diff
@@ -11,6 +11,21 @@ if(NOT dumpbin)
 	return()
 endif()
 
+# Paths for applocal and powershell
+set(APPLOCAL_SCRIPT "${vcpkg_dir}/../../scripts/buildsystems/msbuild/applocal.ps1")
+set(POWERSHELL_COMMAND "C:/Windows/System32/WindowsPowerShell/v1.0/powershell.exe")
```
tusharpm: I'm not certain about the hard-coded absolute path.
Can we use `powershell.exe` and let `execute_process` find the location?

simonsan: I'm not sure if it is in everyone's PATH variable, you can chose if you want to use old commandline or powershell as standard. Could be that it is not found, at least on my production system (not the vm) that is the case. :-S

tusharpm: It's alright. This works for now. You can use `$ENV{SYSTEMROOT}/System32/WindowsPowerShell/v1.0/powershell.exe` to avoid the case of Windows installed in a different location.

simonsan:
Squashed commits, ready to merge.

PR #1128, opened 2019-06-19
Updated windows build instructions

simonsan:
Replaced Visual Studio with the lighter buildtools,
added dependencies for nyan,
added flag for downloading nyan automatically,
removed comment for non-working packaging as it's working now

### DIFF
--- a/doc/build_instructions/windows_msvc.md
+++ b/doc/build_instructions/windows_msvc.md
@@ -8,9 +8,10 @@ __NOTE:__ We also have an installer for Win10 (x64), if you just want to play ar
 ## Setting up the build environment
  You will need to download and install the following manually.
  Those who already have the latest stable versions of these programs can skip this:
- - [Visual Studio 2017 Community edition](https://www.visualstudio.com/downloads/)
-   - With the "Desktop development with C++" workload.
-   - With the "Windows 10 SDK". Choose the latest version listed.
+ - [Visual Studio Buildtools](https://download.visualstudio.microsoft.com/download/pr/10413969-2070-40ea-a0ca-30f10ec01d1d/414d8e358a8c44dc56cdac752576b402/vs_buildtools.exe)
+   - With the "Visual C++ Buildtools" workload.
+_NOTE:_ If you are searching for an IDE for development you can get an overview [here](https://en.wikipedia.org/wiki/Comparison_of_integrated_development_environments#C/C++), we've also written some [instructions for developing with different IDEs](/doc/ide.md).
+
  - [Python 3](https://www.python.org/downloads/windows/)
    - With the "pip" option enabled. We use `pip` to install other dependencies.
    - With the "Precompile standard library" option enabled.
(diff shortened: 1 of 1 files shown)


doc/build_instructions/windows_msvc.md
```diff
@@ -8,9 +8,8 @@ __NOTE:__ We also have an installer for Win10 (x64), if you just want to play ar
 ## Setting up the build environment
  You will need to download and install the following manually.
  Those who already have the latest stable versions of these programs can skip this:
- - [Visual Studio 2017 Community edition](https://www.visualstudio.com/downloads/)
-   - With the "Desktop development with C++" workload.
-   - With the "Windows 10 SDK". Choose the latest version listed.
+ - [Visual Studio Buildtools](https://download.visualstudio.microsoft.com/download/pr/10413969-2070-40ea-a0ca-30f10ec01d1d/414d8e358a8c44dc56cdac752576b402/vs_buildtools.exe)
```
tusharpm: BuildTools might not be a fair replacement for the IDE for developers familiar with it.
I think mentioning it as an alternative would be better.

simonsan: I understand what you mean, but we also don't want to promote a Microsoft IDE I think, as there are many other IDEs to use. I added a link to the IDE comparison at Wikipedia, so everyone can chose their own IDE and we don't give kind of a recommendation anymore. For building I think the `build tools` should be sufficient though, what do you think?

tusharpm: The disadvantage of mentioning this (not as an alternative) is that it becomes non-obvious for VS IDE users that this is optional. In other words, while installing build tools is sufficient, it isn't actually necessary (like "need to download and install" two lines above implies) for them.
Furthermore, "VS build tools" is also Microsoft proprietary software. I don't see mentioning it here as promoting a proprietary product as much as streamlining a common workflow.
Documentation being a subjective topic, I'll let the larger audience decide the best way to write this.

simonsan: `_NOTE:_ If you already have Visual Studio installed you don't need the build tools anymore.`, we could write something like this. The problem is, it isn't true in every case because you need to have the `Desktop development with C++` workload and the `Windows 10 SDK` installed, which is maybe not true for everyone. So basically installing `build tools` will also be useful in this case I guess. 
Furthermore both use the `vs_installer.exe`, if you want to install `Visual Studio` afterwards you could do it with this installer being already downloaded directly. 
I think besides the `build tools` being also proprietary software, they fit more the need of this instruction for building, running and packaging openage.
Also the `build tools` are smaller in download size and in disk space. For building and running openage you don't even need to start Visual Studio at all. 
But if we want people to get in touch with `Visual Studio` as an IDE and start developing for the project we should probably write an own starters instruction for the workflow with it for development. We could do that together.
I would love to stick to the approach to give future developers a recommendation based on their needs when they ask us - instead of letting them download a huge IDE that is probably also overwhelming for most starters. 

Are you ok with this?

simonsan: @tusharpm I've added a link to [ide.md](/doc/ide.md), where we could give instructions how to develop with Visual Studio.

Issue #1134, opened 2019-06-27
Release process is undocumented


zuntrax:
We are missing documentation on how to release a new Version. This needs to be fixed ASAP, probably while releasing `0.4.0`, which i should have done two weeks ago.

See also #1133

Issue #1133, opened 2019-06-27
Version numbering


simonsan:
We had a small discussion about that after the installer release. The version number didn't increase for a long time and so we were releasing the installer under a four years old release number. I think it is not easy to decide how and when to increase a version number. With a date-based versioning system we wouldn't really need to think about a version number anymore or if we should increase. 

YYYY - Full year 
0M - Zero-padded month

As an example or proposal: `<YYYY>.<0M><Minor>-<Modifier>`

Output:
19.061-beta
19.062-rc1
19.063-rc2

I was taking inspiration from https://calver.org/

So what do you think about a switch to date-based version numbering and if yes, what would be your favourite style then?

heinezen:
> The version number didn't increase for a long time and so we were releasing the installer under a four years old release number.

Using calver would only make this problem more apparent to users (as we would be on 15.X right now). In my opinion, this versioning scheme is much more suited to projects with regular release schedules and where the date has some implication (e.g. for security, support or else). Also, the usage of numbers higher than 0 while we're still not finished could be confusing because it might be interpreted as the engine being feature complete.

I would personally prefer if the version number had at least some meaning and communicate the severity of changes better. In the end, the users are not interested in the version number anyway but when an update occurs, they do care about:

* Do my mods still work?
* Did a new major feature arrive?
* Do I have to adapt my mod to API changes? (modder)
* Did the behavior of an engine function change? (modder)
* Does a new feature break anything? (modder)

These questions are usually not addressed by calver as a change from 19.06 to 19.07 could include major breakages, while the releases from 19.08 to 20.02 might only handle bugfixes.

I'm much more in favor of a *feature-based* versioning using a `major.minor.patch-<tag>` structure and increase the version number in the following events:

* `major` increases for fundamental changes in the engine design's core parts (e.g. a game state rewrite, new renderer)
* `minor` increases if the API is changed in such away that it breaks older functionality, changes its structure or introduces a bunch of new features
* `patch` increases for bugfixes
* `<tag>` for nightly/alpha/beta/rc builds

This would be far better to communicate as any bump of `major` or `minor` indicates that something has changed which is either exciting (new features) or something to be aware of (API changes, engine behavior, potential breakage).

zuntrax:
I see no reason not to use SemVer as we are already tying to do.

We have three problems here:
1. Nobody cared enough to just tag and publish new release in the last three years. (`0.3.1` was tagged some time ago, but not published.)
[...]

PR #1137, opened 2019-06-28
document releasing process

zuntrax:
This is mostly what I did yesterday. Makes sense?

### DIFF
--- a/doc/releasing.md
+++ b/doc/releasing.md
@@ -1,0 +1,15 @@
+# Releasing
+
+* Bump the version number in `openage_version`. Given `0.y.z`, both `0.y.(z+1)` and `0.(y+1).0` are valid. The latter will probably be more common, but which of those you choose doesn't really matter.
+
+  Read up on [Semver](https://semver.org) for details. People often seem to overlook the special rules for initial development versions. Please avoid prerelease versions (`-alpha`, `-beta`, `-rc.1`, ...) for now. We might decide to include build metadata like date or commit hash, that's up for discussion.
+* Truncate `changelog.txt` if it exists. You might want to use this later for writing a release text.
+* Commit, PR, merge, pull as usual.
+* OPTIONAL: Set up [commit signing](https://help.github.com/en/articles/managing-commit-signature-verification), signing tags will make a "verified" badge show up on the release later.
+* Tag the merge commit, something like `git tag -s v0.5.0`. The prefixed `v` is mandatory.
+* Push the tag to GitHub.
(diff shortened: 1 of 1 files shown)


simonsan:
Regarding #1133 #1134

simonsan:
Maybe to add: I would explicitly state some info how and where to give information about the depreciation of features. I don't know if it already plays a role here or just after 1.0.0

doc/releasing.md
```diff
@@ -0,0 +1,15 @@
...
+* Truncate `changelog.txt` if it exists. You might want to use this later for writing a release text.
+* Commit, PR, merge, pull as usual.
+* OPTIONAL: Set up [commit signing](https://help.github.com/en/articles/managing-commit-signature-verification), signing tags will make a "verified" badge show up on the release later.
+* Tag the merge commit, something like `git tag -s v0.5.0`. The prefixed `v` is mandatory.
+* Push the tag to GitHub.
+* Use the GitHub web interface to publish a release from the tag. Make it look nice, see previous release description. If in doubt, save the draft and have someone else proof read it.
+* In the future, kevin will automatically attach build artifacts like installers to the release. For now, you will need to do that manually. Have fun setting up a windows build environment or if you feel like being lazy, go bug someone who already has. Actually, taking care of where to build for windows before publishing the release might not be a bad idea.
+* OPTIONAL: Brag on social media. /r/openage and AoE2 Discord might be the right place to do so.
```
heinezen: Discord --> Discords

doc/releasing.md
```diff
@@ -0,0 +1,15 @@
...
+
+If you need help or are not completely sure what you are doing, ask in matrix and highlight `zuntrax` or someone else who has done it before. Better having someone help you than screwing up a release.
```
heinezen: Link to the Matrix SFTtech chanel would be good :)

simonsan:
https://pillow.readthedocs.io/en/stable/deprecations.html

As an example. I would put that as an extra point prominent into the sphinx doku.

Start could be that we changed minimal Python version to **Python 3.6** with **0.4.0**

zuntrax:
@simonsan Not sure if I would consider updating a dependency a deprecation. I kinda see it for python2, because it's mostly a different language.

janisozaur:
Here's a similar document I wrote on releasing OpenRCT2, perhaps you will find some inspiration in there as well: https://gist.github.com/janisozaur/693b70c5d2c583d6c6f0e7ffbc983720

PR #1139, opened 2019-06-28
Fix 'math' usage in modding API reference guide

heinezen:
We now properly escape underscores and use `\cdot` in formulas.


simonsan:
I tried out to take a look why mathjax is not rendering the formulas, cannot see why right now. Need to debug.

heinezen:
Missing escape is fixed. The TODOs will be explained in a longer tutorial because it is too complicated for the reference sheet. We can link to it when it's done.

PR #1143, opened 2019-08-05
Changelog

heinezen:
Changelog for engine development that happened until our last release (v0.4.0).

The pre-SFTtech/openage time is not as documented as the other releases because the main source of information is the git log (which is huge). For all changes after SFTtech/openage was created, I used PRs as orientation.

- [x] Engine changelog
- [x] Modding API changelog


heinezen:
Done! What a breeze through the project's history :tada: 

There's a kevinception in the sanity check here because kevin complains about an `asdf` in bullet point saying that he now checks for asdf's. Don't know what to do about it.. maybe we should exclude the changelogs from ASDF and TODO checks.

PR #1148, opened 2019-08-27
Enhanced singlefile conversion and support for DE1 formats

heinezen:
Adds more options to the "singlefile" conversion to support newer game versions and SLPs outside of DRS archives. Specifically:

* New option to convert single SLPs
* Support for SLP 3.0, 4.0X, 4.1 (+ documentation)
* Some preparation for DE2

### DIFF
--- a/doc/media/slp-files.md
+++ b/doc/media/slp-files.md
@@ -5,7 +5,7 @@ textures. Like the DRS format, it can also be read sequentially.
 
 ## SLP file format
 
-#### Header (up to version 3.0)
+### Header (up to version 3.0)
 The SLP file starts with a header:
 
 Length   | Type   | Description      | Example
@@ -24,7 +24,7 @@ struct slp_header {
 ```
 Python format: `Struct("< 4s i 24s")`
 
-#### Header (since version 4.0X)
+### Header (since version 4.0X)
 Age of Empires 1: Definitive Edition SLPs (v4.0) use a differently structured header:
 
 Length   | Type    | Description             | Example
@@ -56,8 +56,8 @@ struct slp_header_v4 {
 ```
 Python format: `Struct("< 4s H H H H i i i 8x")`
 
-#### SLP Frame info
-After the header, there are `num_frames` entries of `slp_frame_info`:
+### SLP Frame info
+After the header, there are `num_frames` entries of `slp_frame_info`.
 Every `slp_frame_info` stores meta-information about a single frame (texture)
 within the SLP.
 
@@ -99,7 +99,8 @@ Python format: `Struct("< I I I I i i i i")`
 
 One image of size `width`*`height` has `height` rows, of course.
 
-#### SLP Frame row edge
+
+### SLP Frame row edge
 At `outline_table_offset` (after the `slp_frame_info` structs), an array of
   `slp_frame_row_edge` (of length `height`) structs begins.
 
@@ -155,9 +156,9 @@ The actual command data starts after the end of the command offsets, or:
 ### SLP drawing commands
 
 The image is drawn line by line, a line is finished with the "End of line"
-command (0x0F). A command is a one-byte number (`cmd_byte`), followed 
-by command-specific data with a length (number of pixels) varying 
-depending on the command. The next command immediately follows the 
+command (0x0F). A command is a one-byte number (`cmd_byte`), followed
+by command-specific data with a length (number of pixels) varying
+depending on the command. The next command immediately follows the
 previous command's data.
 
 Commands can also store meta information like the number of following pixels
@@ -299,7 +300,7 @@ us that the row is finished.
 
 SLPs with versions higher than 4.0 store their shadows in separate frames, but
 use the same command syntax. Shadows are drawn with the *lesser draw*,
-*lesser skip* and *fill* commands
+*lesser skip* and *fill* commands.
 
 Command Name             | Byte value        | Pixel Count    | Description
 -------------------------|-------------------|----------------|------------
@@ -334,7 +335,7 @@ third line stores the number of entries, as text.
 The rest is that many `r g b\r\n` lines. Again, `r`, `g` and `b` are stored as
 text (range `0-255`).
 
-Colors from the palette are referenced in the SLP files with an index. 
+Colors from the palette are referenced in the SLP files with an index.
 The index refers to a line in the palette, so line 3=>index 0,
 line 4=>index 1 etc.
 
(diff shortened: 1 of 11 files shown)


openage/convert/slp.pyx
```diff
@@ -49,14 +49,22 @@ cdef struct boundary_def:
...
-
+    color_standard      # standard pixel
+    color_shadow        # shadow pixel
+    color_shadow_v4     # shadow pixel in the 4.0X version
+    color_transparent   # transparent pixel
+    color_player        # non-outline player color pixel
+    color_player_v4     # non-outline player color pixel
+    color_black         # black outline pixel
+    color_special_1     # player color outline pixel
+    color_special_2     # black outline pixel
```
TheJJ: why change the comment position? that obscures the diff a bit.

openage/convert/slp.pyx
```diff
@@ -405,12 +526,12 @@ cdef class SLPFrame:
...
-                # color = ((color & 0b11001100) | 0b00110011)
-
-                # SpecialColor class preserves the calculation of
-                # player*16 + color
                 for _ in range(cpack.count):
+                    # TODO: verify this. might be incorrect.
+                    # color = ((color & 0b11001100) | 0b00110011)
+
+                    # SpecialColor class preserves the calculation of
+                    # player*16 + color
```
TheJJ: i think i removed the `SpecialColor` class some time ago and replaced it with the enum that contains `color_player`. The comment should therefore read "the `color_player` enum later resolves the special case for the player color".

openage/convert/slp.pyx
```diff
@@ -405,12 +526,12 @@ cdef class SLPFrame:
...
                 color = self.get_byte_at(dpos)
 
-                # TODO: verify this. might be incorrect.
-                # color = ((color & 0b11001100) | 0b00110011)
-
-                # SpecialColor class preserves the calculation of
-                # player*16 + color
                 for _ in range(cpack.count):
+                    # TODO: verify this. might be incorrect.
+                    # color = ((color & 0b11001100) | 0b00110011)
```
TheJJ: i think we can remove that. for the main sprites we have the correct palette index already, so this might be for some other encoding case?

openage/convert/slp.pyx
```diff
@@ -500,59 +621,384 @@ cdef class SLPFrame:
...
+                    )
+                )
 
-    def get_picture_data(self, palette, player_number=0):
-        """
-        Convert the palette index matrix to a colored image.
-        """
-        return determine_rgba_matrix(self.pcolor, palette, player_number)
+            # fetch drawing instruction
+            cmd = self.get_byte_at(dpos)
```
TheJJ: we have to make sure this is properly resolved to a c++-only call. we can see that in the cythonized `html` file. otherwise this will really slow down the conversion.

openage/convert/smp.pyx
```diff
@@ -0,0 +1,837 @@
...
+
+
+class SpecialColorValue(Enum):
+    transparent = " "
+
+    def __str__(self):
+        return self.value
+
+    def __repr__(self):
+        return self.value
```
TheJJ: this isn't used, right?

openage/convert/smp.pyx
```diff
@@ -0,0 +1,837 @@
...
+            dpos += 1
+
+        # end of row reached, return the created pixel array.
+        return
+
+
+cdef class SMPOutlineFrame(SMPFrame):
+    """
+    SMPFrame for the outline graphics.
+    """
```
TheJJ: hm, i wonder why they are doing that. in any way, if the outline information is separate from the graphics info, we have to merge them in order to do our alpha-encoding.

openage/convert/smp.pyx
```diff
@@ -0,0 +1,837 @@
...
+            array_data[y, x, 2] = b
+            array_data[y, x, 3] = alpha
+
+    return array_data
+
+cdef (uint8_t,uint8_t) get_palette_info(pixel image_pixel):
+    """
+    returns the palette used for a pixel.
+    """
+    return image_pixel.palette >> 4, image_pixel.palette & 0x03
```
TheJJ: two numbers are returned, i'd welcome a comment what both of the numbers mean and why there are two.

openage/convert/texture.py
```diff
@@ -134,16 +146,36 @@ def __init__(self, input_data, palette=None, custom_cutter=None):
...
+        convert smp to subtexture or subtextures, using a palette.
         """
         # TODO this needs some _serious_ performance work
         # (at least a 10x improvement, 50x would be better).
-        # ideas: remove PIL and use libpng via CPPInterface,
-        #        cythonize parts of SLP.py
+        # ideas: remove PIL and use libpng via CPPInterface
         subtex = TextureImage(
-            frame.get_picture_data(palette, self.player_id),
+            frame.get_picture_data(main_palette, player_palette),
```
TheJJ: here we have to do the merging of the outline info and the graphics i think.
what do we do about the damage graphics? (we should likely discuss that in the chat :)

doc/media/slp-files.md
```diff
@@ -193,13 +241,92 @@ For later drawing, a graphics file needs flags for:
...
+This time, the next 10 bytes are palette indices that belong to the drawing
+command (`0x99 0x35 0xF4 0x6D 0x67 0x6E 0xA5 0x01 0x4D 0x8E`).
+
+Our next command byte is `0x0F`. This is the *end of row* command which tells
+us that the row is finished.
+
+### SLP shadow commands (since version 4.0X)
+
+SLPs with versions higher than 4.0 store their shadows in separate frames, but
+use the same command syntax. Shadows are drawn with the *lesser draw*,
```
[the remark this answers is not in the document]
heinezen: SMPs also store the shadow separately, but they started it with SLP version 4.0

(The calculation for the alpha value is different between SMP and SLP)

openage/convert/smp.pyx
heinezen: Note to myself: the function should return `image_pixel.palette >> 2`

openage/convert/slp.pyx
heinezen: I think Eclipse aligned the comments to tab width (4 spaces). In the git diff it's readable btw, but Github marks every changed line green somehow.

openage/convert/slp.pyx
heinezen: > i think i removed the SpecialColor class some time ago

It's still in `slp.pyx` (Line 24). I guess it should be removed then?

heinezen: Also, I removed the comment entirely because the player color is calculated on the fly in `determine_rgba_matrix(..)`

openage/convert/slp.pyx
heinezen: Wouldn't the function declaration with `cdef` ensure that?

openage/convert/smp.pyx
heinezen: Probably because they wanted a smaller number of commands (?) Or something to do with their expanded rendering. Merging should work fine, since he outlines does not overlay with any other pixels in the main graphic.

PR #1152, opened 2019-09-08
Updating easy tasks link in readme.md

simonsan:
Old link was not suitable anymore, updated it to `good first issues`-label.


heinezen:
good job in spotting that :)

PR #1165, opened 2019-10-03
Update CMakeLists.txt

Radhika-Modi:


TheJJ:
Hi!
Can you elaborate why you made this change? Thanks!

simonsan:
Would rate that as Spam, user spams PR over different repositories. Closing this here as I see no relevancy in this change.

PR #1167, opened 2019-10-05
Add .clang-format to the repository

### LINKED ISSUE #1166: Add .clang-format configuration file
Referring to something mentioned in #313, `clang-format` does have a checking only mode, and is hence suitable for travis.
Again something that would help with #313 as well, `clang-format` can also enforce (and automatically convert) using spaces for indentation instead of tabs.

While you can keep defining regexes for specific cases like spaces around operators, `clang-format` can support a wide variety of such coding standards, and is a one-stop solution to most such problems.

The benefits of `clang-format` - 

1. Autofixing issues as well as detecting them.
2. One stop solution to all current and future code formatting and compliance issues.
3. Is an Extremely powerful formatting engine, so much so that so far I have yet to encounter cases it fails to handle acceptably, including enforcing a spaces-only mode (#313).
4. Not at all a big task to implement, all the formatting options reside in a simple config file.

On a side note, if people in this community want to use it and aren't familiar with this tool, I can implement the above.

_Originally posted by @akhilkedia in https://github.com/SFTtech/openage/issues/837#issuecomment-307636379_

[Clang-Format Homepage](https://clang.llvm.org/docs/ClangFormat.html)

simonsan:
Fixes #1166 

This is a proposal for a **clang-format** file. 
I was using it on `mom.cpp / mom.h` and `game_main.cpp / game_main.h` as an example.


doc/code_style/mom.cpp
```diff
@@ -59,36 +59,32 @@
...
 };
 
 
-
 /**
  * Yet another Valve game. Go on, nothing to see here.
  */
 class HalfLife3 : public ValveGame {
-// vvv The access modifier is at the same indent level as the class definition:
+	// vvv The access modifier is at the same indent level as the class definition:
```
simonsan: Trying to find out where this comes from :-O

doc/code_style/mom.cpp
```diff
@@ -59,36 +59,32 @@
...
+	                  const Engine *   game_engine,
+	                  Internet *       community,
+	                  bool             keep_secret = true)
+	{
 		// ^ insert a empty line here to make it more readable.
 
 		// http://www.abstrusegoose.com/440
-		auto gaben = team.get_human("gaben");
+		auto gaben       = team.get_human("gaben");
 		auto min_quality = gaben.get_quality_requirement(Engine::type);
```
simonsan: This is a good change I think.

MaanooAk: This is an enormous change, this cannot result in good thing in my opinion

simonsan: everything wip! ;-) just trying to find the right clang-format options for mom.cpp

doc/code_style/mom.cpp
```diff
@@ -109,10 +106,7 @@ class HalfLife3 : public ValveGame {
 
 			// You may also start one line later
 			// and then indent with a tab:
-			team.relocate(
-				game_engine->get_team()->get_current_hq_location(),
-				gaben.get_table_location()
-			);
+			team.relocate(game_engine->get_team()->get_current_hq_location(), gaben.get_table_location());
```
MaanooAk: (Just feedback) When something is written in multiple lines usually there is a good reason

doc/code_style/mom.cpp
```diff
@@ -230,17 +229,20 @@ bool YourMom::tick() {
...
+} // namespace mom
+} // namespace parents
 // ^ When closing namespaces,
 //   close all in the same line and write a comment like this.
 //   In C++17 this will only be a single }
 
 
-int main(int argc, char **argv) {
+int
+main(int argc, char **argv) {
```
MaanooAk: (just my opinion) This is objectively disgusting

libopenage/gamestate/game_main.cpp
```diff
@@ -12,23 +12,20 @@
...
-	spec{generator.get_spec()} {
-
+GameMain::GameMain(const Generator &generator) :
+    OptionNode{"GameMain"}, terrain{generator.terrain()}, placed_units{}, spec{generator.get_spec()} {
 	// players
 	this->players.reserve(generator.player_names().size());
 	unsigned int i = 0;
-	for (auto &name : generator.player_names()) {
+	for (auto &name : generator.player_names())
+	{
```
MaanooAk: (last one) also disgusting (java developer here)

doc/code_style/mom.cpp
```diff
@@ -189,7 +177,8 @@ bool YourMom::tick() {
 		}
 		else if (this->stupidity > 100) {
 			this->die();
-		} else { // Both styles are allowed: '} else {' and '} \n else {'
+		}
+		else { // Both styles are allowed: '} else {' and '} \n else {'
```
simonsan: This is ok, because clang-format uses one of two allowed styles

doc/code_style/mom.cpp
```diff
@@ -234,7 +223,8 @@ void YourMom::buy(const Thing & /*item*/) {
 	// TODO: implement
 }
 
-}} // parents::mom
+} // namespace mom
+} // namespace parents
```
simonsan: This is done automatically, so it's fine i think. Also it gives less errors later in `clang-tidy`

doc/code_style/mom.h
```diff
@@ -18,13 +18,13 @@
 
 // Includes are sorted alphabetically,
 // global includes first, then system inclusions:
-#include <cstddef>
-#include <cstdint>
-
```
simonsan: Is there a switch for this in clang-format? Didn't find one. Should we consider changing this in the template?

doc/code_style/mom.h
```diff
@@ -62,7 +62,7 @@ class YourMom : public MassiveEntity {
 	 * Creates your mom.
 	 * @param mass: Initial mass for this mother.
 	 */
-	YourMom(int mass=DEFAULT_MOM_MASS);
```
simonsan: This is wrong in the template

doc/code_style/mom.h
```diff
@@ -84,7 +84,7 @@ class YourMom : public MassiveEntity {
 
 protected:
 	/** Let your mom buy something */
-	void buy(const Thing& item);
```
simonsan: We should decide for either **one form** regarding also the setting ` PointerAlignment`. I like both forms though as they give somehow to different information (e.g. left=space in mem), so would prefer the left alignment there.

doc/code_style/mom.h
```diff
@@ -62,7 +62,7 @@ class YourMom : public MassiveEntity {
 	 * Creates your mom.
 	 * @param mass: Initial mass for this mother.
 	 */
-	YourMom(int mass=DEFAULT_MOM_MASS);
+	YourMom(int mass = DEFAULT_MOM_MASS);
```
TheJJ, line 65: it should be like python does it, with no spaces around the `=` for default values.

doc/code_style/mom.h
```diff
@@ -113,16 +113,16 @@ class YourMom : public MassiveEntity {
...
 };
 
 
 /**
  * Type aliases are done with "using".
  * They can get proper documentation, too!
  */
-using wtf_t = std::unordered_map<std::string,
-                                 std::unique_ptr<Stuff>>;
+using wtf_t = std::unordered_map<std::string, std::unique_ptr<Stuff>>;
```
TheJJ, line 124: clang's style is ok, but we do allow and want to showcase template argument wrapping here.

doc/code_style/mom.h
```diff
@@ -113,16 +113,16 @@ class YourMom : public MassiveEntity {
...
 
-}} // parents::mom
+} // namespace mom
+} // namespace parents
```
TheJJ, line 127: that's an acceptable change, especially multi-part-namespaces are now allowed with c++17 and require a single {} pair.

doc/code_style/mom.h
```diff
@@ -18,13 +18,13 @@
...
 // global includes first, then system inclusions:
-#include <cstddef>
-#include <cstdint>
-
 #include "../magic/unicorn.h"
 #include "rofl.h"
 #include "stuff.h"
 
+#include <cstddef>
+#include <cstdint>
```
TheJJ, line 26: the include order should be preserved

doc/code_style/mom.cpp
```diff
@@ -230,17 +223,18 @@ bool YourMom::tick() {
...
 // ^ When closing namespaces,
 //   close all in the same line and write a comment like this.
 //   In C++17 this will only be a single }
 
 
-int main(int argc, char **argv) {
+int main(int argc, char** argv) {
```
TheJJ, line 237: ptr to the right

doc/code_style/mom.cpp
```diff
@@ -230,17 +223,18 @@ bool YourMom::tick() {
 }
 
 // Unused parameters are commented out:
-void YourMom::buy(const Thing & /*item*/) {
+void YourMom::buy(const Thing& /*item*/) {
```
TheJJ, line 226: like above, the `&` should be attached to the variable name. One or no space between `&` and the comment would be ok.

doc/code_style/mom.cpp
```diff
@@ -171,7 +163,7 @@ void YourMom::print_mass() {
 bool YourMom::tick() {
 	// Note the spaces!
 	//|                | |  |
-	//v                v v  v
+	// v                v v  v
```
TheJJ, line 166: this should stay, but of course that's only important for this style file, probably all other comments should have a space after `//`

doc/code_style/mom.cpp
```diff
@@ -140,11 +136,7 @@ namespace parents {
 namespace mom {
 // <- A namespace does not increase the indentation level!
 
-YourMom::YourMom(int her_mass)
-	:
-	mass{her_mass},
-	number_of_moons{0},
-	stupidity{1} {
+YourMom::YourMom(int her_mass) : mass{her_mass}, number_of_moons{0}, stupidity{1} {
```
TheJJ, line 139: the member initializers should remain on separate lines, complicated constructor calls and lambdas etc will be used for initialization, which becomes totally unreadable once that's only on one line.

doc/code_style/mom.cpp
```diff
@@ -109,10 +107,8 @@ class HalfLife3 : public ValveGame {
 
 			// You may also start one line later
 			// and then indent with a tab:
-			team.relocate(
-				game_engine->get_team()->get_current_hq_location(),
-				gaben.get_table_location()
-			);
+			team.relocate(game_engine->get_team()->get_current_hq_location(),
+			              gaben.get_table_location());
```
TheJJ, line 111: that's also possible, but this should remain as-in since it demonstrates the 2 argument indentation methods that are allowed.

doc/code_style/mom.cpp
```diff
@@ -59,34 +59,32 @@
...
-
 /**
  * Yet another Valve game. Go on, nothing to see here.
  */
 class HalfLife3 : public ValveGame {
-// vvv The access modifier is at the same indent level as the class definition:
+	// vvv The access modifier is at the same indent level as the class definition:
 public:
-	void release_game(const ValveTeam &team,
+	void release_game(const ValveTeam& team,
```
TheJJ, line 75: we place `*` and `&` to the variable name, not the type, since semantically it's part of the name, and not the type.
`int* lol, rofl;` declares lol as `int*` and rofl as `int`, so when written like this: `int *lol, rofl;` this is directly visible.

doc/code_style/mom.cpp
```diff
@@ -59,34 +59,32 @@
...
 	                  uint128_t        passed_time,
-	                  const Engine    *game_engine,
-	                  Internet        *community,
-	                  bool             keep_secret=true) {
+	                  const Engine*    game_engine,
+	                  Internet*        community,
+	                  bool             keep_secret = true) {
```
TheJJ, line 79: like in python for default argument values, there's no spaces around `=`.

this should not be in columns for change-tracking reasons. once the alighment doesn't match, we have to realign all lines even though they did not change. so the spacing should be done with just one space.

doc/code_style/mom.cpp
```diff
@@ -59,34 +59,32 @@
...
 
 		// ^ insert a empty line here to make it more readable.
 
 		// http://www.abstrusegoose.com/440
-		auto gaben = team.get_human("gaben");
+		auto gaben       = team.get_human("gaben");
 		auto min_quality = gaben.get_quality_requirement(Engine::type);
```
TheJJ, line 85: these alignments should not be done since it requires realigning stuff when variable names change for all other aligned lines, see above.

doc/code_style/mom.cpp
```diff
@@ -59,34 +59,32 @@
...
 
-		if (not keep_secret
-		    and (this->story->awesomeness > 9000)
+		if (not keep_secret and (this->story->awesomeness > 9000)
```
TheJJ, line 87: both are ok, but this piece of code should show how we would wrap boolean expressions

doc/code_style/mom.cpp
```diff
@@ -121,7 +117,7 @@ class HalfLife3 : public ValveGame {
 
 private:
 	/** Links to the story of the game. */
-	Story *story;
+	Story* story;
```
TheJJ, line 120: same as above, ptr to the right.

doc/code_style/mom.cpp
TheJJ: this shouldn't matter much since we will likely never comment an access modifier. it's only important for this example file that the indent level is at 0.

doc/code_style/mom.cpp
```diff
@@ -18,10 +18,10 @@
 #include "mom.h"
 
 // System includes follow, sorted alphabetically
-#include <iostream>
+#include <SDL/SDL.h>
 #include <cmath>
 #include <cstdio>
-#include <SDL/SDL.h>
+#include <iostream>
```
TheJJ, line 24: include order should be strictly alphabetical

doc/code_style/mom.h
simonsan: `SpaceBeforeAssignmentOperators: 'false'`

doc/code_style/mom.h
simonsan: I need to recheck, but afair it was clang-tidy that merged these together in the files automatically. I will make another run soon when I try out the Git workflows (want to let them export patch files from `clang-format` and `clang-tidy` for the people to make fixing their code easier). So I think that would be something for clang-tidy later, I wanted to say with it.

doc/code_style/mom.h
simonsan: `IncludeBlocks: 'Preserve'`, what we saw is `Regroup`

doc/code_style/mom.cpp
simonsan: `ConstructorInitializerAllOnOneLineOrOnePerLine: 'false'`
`AllowAllConstructorInitializersOnNextLine: 'false'`

doc/code_style/mom.cpp
simonsan: `SpaceBeforeAssignmentOperators: 'false'`

Need to see if there is a setting for special circumstances for this.

doc/code_style/mom.h
simonsan: Didn't find any other option than the mentioned one? So I think we would need to agree on one of these:

- either having no space before the assignment operators, but after it
- or having spaces before and after

In my opinion `before and after looks` more "clean" (like in symmetric), what do you think?

doc/code_style/mom.h
simonsan: Works out, pressing enter after the comma aligns it with the templates `<`. I think this is because the column limit of `100` probably would chose the first style if lower. Had to manually edit that, fyi.

doc/code_style/mom.h
simonsan: Can confirm, I think what you mean is this one:
https://clang.llvm.org/extra/clang-tidy/checks/modernize-concat-nested-namespaces.html

doc/code_style/mom.cpp
simonsan: Understood, needed to align that manually because the columnlimit of 100 characters was not reached.

doc/code_style/mom.cpp
simonsan: SortIncludes (bool)
    If true, clang-format will sort #includes.
```
    false:                                 true:
    #include "b.h"                 vs.     #include "a.h"
    #include "a.h"                          #include "b.h"
```
This is sadly the only option here. I think clang-format treats `SDL/SDL.h` different because it's not part of the standard library and for that treated as external.

What about if the let clang-format do it this way, but as long as the grouping is preserved. It has also the advantage, that you see directly on top of this group which non-standard header was included. What do you think?

doc/code_style/mom.cpp
simonsan: that was due to me setting column limit to 100 changed it back to 80, didn't see that this was the standard before

simonsan:
Ok, I think we are pretty near to the code style we already have. I will commit a few reformatted example files from the code base to see how it look in action and if we could agree with it as a sample (not a must, but a good basis for a codestyle).

simonsan:
@TheJJ These three sample files are really just what we could get PR'ed in the future from people. It's not that we need to change the style of the complete codebase, but this would be the `minimum agreement` that everyone will be able to adopt easily by executing one command/pull in one style-patch from the CI. 

Personally I think that is pretty good, although not `perfect` for sure. But it's a step forward. Your thoughts?

doc/code_style/mom.h
TheJJ: Assignment operators in general should have spaces before and after, only in the case of default argument values there should be none (like "[Don't use spaces around the = sign when used to indicate a keyword argument, or when used to indicate a default value for an unannotated function parameter](https://www.python.org/dev/peps/pep-0008/)").
If clang-format doesn't support this, maybe we can persuade it to leave default argument values alone?

doc/code_style/mom.cpp
TheJJ: yea that's probably fine, as long as the groups are preserved.

doc/code_style/mom.h
simonsan: We could turn it off I think, but that would mean as well, that the other option spaces before and after for assignments will not be automatically done anymore. We base the `.clang-format` file on the LLVM style, it could be, that it's done already there - would need to try that. I assume we need to decide if we take the approach that spaces for assignments are made automatically, as they are probably much more in the code. And tell people manually to not use spaces. No one will be forced to use the `.clang-format` any way. 

I see `clang-format` as the minimum standard of coding-style this repository should have with what we agree on that people, that are not coding regularly, can PR to us somehow.

simonsan:
I tried a bit around, exported configs and checked what they do different and we are so near our original style already, that I'm pretty unsure if we can get any better with these options currently.
As we set nearly every single possible value ourselves we could probably even establish an own exportable style with this already as in:
`clang-format -style=openage -dump-config > .clang-format`

simonsan:
@TheJJ Rebased as well

simonsan:
(Outdated)

PR #1168, opened 2019-10-05
[WIP] Autodrop and work

MaanooAk:
Added support for resources that are gather by the villagers to go directly to the player instead of the dropsite logic.

Autodrop is enabled per Resource.

Good examples are the Favor from AoM and all resources in AoE 3.

Talking about Favor, I was thinking how temple from AoM would work or Farms from AoE 3.

So, I added shared attribute `workplace` and unshared attribute `workforce` and the corresponding `WorkAction`.

(for testing, to enable the autodrop)
``` diff
ClassicResources::ClassicResources()
	:
	resources{{wood, "wood"},
-	          {food, "food"},
+	          {food, "food", true},
	          {gold, "gold"},
	          {stone, "stone"}} {
}
```

### DIFF
--- a/doc/nyan/api_reference/reference_aux.md
+++ b/doc/nyan/api_reference/reference_aux.md
@@ -1180,6 +1180,7 @@ A set of requirements that unlock the technology for the researching game entity
 Resource(Entity):
     name        : TranslatedString
     max_storage : int
+    autodrop    : bool
 ```
 
 Defines a resource that can be used in the game. Adding a resources will give an amount of 0 of that resource to all players. The current amount of resources can be influenced by the abilities and modifiers of game entities.
@@ -1189,6 +1190,9 @@ The name of the resource as a translatable string.
 
 **max_storage**
 Maximum amount of resources that can be stored at a time by the player.
+
+**autodrop**
+Determines whether a villager does not need a dropsite.
 
 ## aux.resource.ResourceContingent
 
@@ -1220,7 +1224,7 @@ The maximum contingent size.
 ```python
 ResourceAmount(Entity):
     type   : Resource
-    amount : int
+    amount : float
 ```
 
 A fixed amount of a certain resource.
--- a/libopenage/gamestate/resource.cpp
+++ b/libopenage/gamestate/resource.cpp
@@ -1,4 +1,4 @@
-// Copyright 2015-2017 the openage authors. See copying.md for legal info.
+// Copyright 2015-2019 the openage authors. See copying.md for legal info.
 
 #include <string>
 #include <cmath>
@@ -7,6 +7,20 @@
 #include "resource.h"
 
 namespace openage {
+
+ClassicResources::ClassicResources()
+	:
+	resources{{wood, "wood"},
+	          {food, "food"},
+	          {gold, "gold"},
+	          {stone, "stone"}} {
+}
+
+// TODO remove, here for transition period
+const Resource* ClassicResources::to_resource(game_resource& id) {
+	static ClassicResources cr = ClassicResources();
+	return &cr.get_resource((int) id);
+}
 
 ResourceBundle Resources::create_bundle() const {
 	return ResourceBundle(*this);
--- a/libopenage/gamestate/resource.h
+++ b/libopenage/gamestate/resource.h
@@ -1,4 +1,4 @@
-// Copyright 2015-2018 the openage authors. See copying.md for legal info.
+// Copyright 2015-2019 the openage authors. See copying.md for legal info.
 
 #pragma once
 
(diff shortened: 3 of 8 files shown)


doc/nyan/api_reference/reference_aux.md
```diff
@@ -1180,6 +1180,7 @@ A set of requirements that unlock the technology for the researching game entity
 Resource(Entity):
     name        : TranslatedString
     max_storage : int
+    autodrop    : bool
```
heinezen: This should not be in `Resource` since this is a global reference object. 

I would say that we already could do autodrop with existing abilities: If the game entity has both `Gather` and `DropSite` abilities for the same resource, then the resources are automatically added to the resource pool with the `gather_rate` rate. Alternatively, we could introduce a new `Modifier` that allows autodrop.

doc/nyan/api_reference/reference_aux.md
```diff
@@ -1220,7 +1224,7 @@ The maximum contingent size.
 ```python
 ResourceAmount(Entity):
     type   : Resource
-    amount : int
+    amount : float
```
heinezen: Hmmm... when would this be useful? We can the resources as a double internally, but for the API having non-fractional resource amounts is much cleaner I think.

MaanooAk: Mostly when the `ResourceAmount` is not used as an amount but as a multiplier (eg. repair cost per hp) or a rate (eg. villager gather rate). Also for resources introduced by mod, you never know...

doc/nyan/api_reference/reference_aux.md
MaanooAk: > This should not be in Resource since this is a global reference object.

There is no problem being global.

> If the game entity has both Gather and DropSite

This is more complex and also for new resources introduced with nyan patches you cannot add a DropSite to every Gatherer

heinezen: >There is no problem being global.

It is a problem because the resource should not define if it is droppable. Rather, the villager object should define whether it is allowed to autodrop its resources.

> This is more complex and also for new resources introduced with nyan patches you cannot add a DropSite to every Gatherer

If you introduce a new resource, you can also patch `accepts` in `DropSite` to include it. It would be one additional patch. For complexity, it can be integrated in the search for drop sites that a game entity has to do anyway.

MaanooAk: You are right, I reliase now that there two cases

Case **1**: Favor (AoM) is autodrop, it does not depen on the villager.
Case **2**: VillagerWithACartForFood can autodrop Food, but Food is not autodrop.

So, I suggest we do both: There is a autodrop attribute to the resource and also we allow Gather with DropSite.

Agreed? (on yes, I will impl the special case of both Gather and DropSite)

doc/nyan/api_reference/reference_aux.md
heinezen: For rates we have `ResourceRate`.

I think the reason for keeping `ResourceRate` and `ResourceAmount` separate was that `ResouceRate` is tied to a time unit (Resource/time unit) and thus a bit more complex. `ResourceAmount` is supposed to be the less calculation intensive object.

doc/nyan/api_reference/reference_aux.md
heinezen: You should probably come to the [matrix channel](https://riot.im/app/#/room/#sfttech:matrix.org) for further discussion :) I've seen that you also made suggestion for nyan, but used outdated API ideas, so it would be good to get you up-to-date.

> There is a autodrop attribute to the resource and also we allow Gather with DropSite.

I still think that `autodrop` should not belong to `Resource`. Resources do not have to be gathered, they could be accumulated stored entities (relics), provided temporarily (pop space) or created out of thin air (cheats). The `Resource` object should therefore be neutral. Autodropping is better placed in the related gather abilities (`Gather` or `DropSite`), otherwise the member is redundant for a lot of cases.

TheJJ: I agree that autodropping in general is not a property of a resource (type). The autodropping depends on where you want it do be dropped. I think any resource should be autodroppable, as long as there is an entity near enough that accept this resource as autodrop.

simonsan: @TheJJ @heinezen +1, that would also make some AoE3 behaviour possible if we raise the distance to something "infinite". So for example faith can be "gathered" due e.g. fighting all over the map with let's say having build the monastery as a `DropSite`.

TheJJ:
Anything we can assist you with @MaanooAk?

Issue #1054, opened 2018-10-06
Minimise state changes in new renderer


Vtec234:
### Contributing to the new renderer:
The renderer interface is defined in headers, with the main class in `libopenage/renderer/renderer.h`.
A good place to start looking is `libopenage/renderer/tests.cpp`, which presents a sample usage of the renderer. To run the test, after compiling Openage do `./run test -vd renderer.tests.renderer_demo 0`.

### Required knowledge:
You need to be familiar with how OpenGL modifies its internal
 state when shader programs are switched, different buffers bound, draw calls made, etc.

### What to modify:
1. The logic in `libopenage/renderer/opengl/renderer.cpp` should not simply execute renderables one after another, but rather sort them first - by shader program, then by used buffer. This will require removing some `const`s from `RenderPass` and changing the documentation to specify that the order of `Renderable`s may be changed by the renderer.
3. Documentation should be added to every new header and existing documentation updated to reflect changes.

teldosas:
Is this a duplicate of #34 ?

Vtec234:
I think what's happened is that I hijacked that thread to list the things to be done for the new renderer in general, while this issue is specifically only about reordering drawables so as to not do many shader/texture/whatever switches. I do agree the initial post in #34 does overlap a bit.

PR #1171, opened 2019-10-13
First batch of clang-tidy fixes

simonsan:
Where nothing but the namespaces were changed, I discarded the change and will put them into another PR soon, with just these changes.

I looked over them already, but there are some, where it is quite unsure if the fixes change intended behaviour - I will comment on them to find them easier.

PR #1182, opened 2019-10-24
Minimise state changes in renderer

teldosas:
Adresses #1054 

Is this enough?

### DIFF
--- a/libopenage/renderer/opengl/render_pass.cpp
+++ b/libopenage/renderer/opengl/render_pass.cpp
@@ -1,0 +1,26 @@
+// Copyright 2019-2019 the openage authors. See copying.md for legal info.
+
+#include "render_pass.h"
+
+namespace openage::renderer::opengl {
+
+GlRenderPass::GlRenderPass(const std::vector<Renderable>& renderables, RenderTarget const* target)
+:RenderPass(renderables, target), is_optimised(false) {}
+
+std::vector<Renderable> GlRenderPass::get_renderables() const {
+	return this->renderables;
+}
+
+void GlRenderPass::set_renderables(std::vector<Renderable> renderables) {
+	this->renderables = renderables;
+	this->is_optimised = false;
+}
+
+bool GlRenderPass::get_is_optimised() const {
+	return this->is_optimised;
+}
+
+void GlRenderPass::set_is_optimised(bool flag) {
+	this->is_optimised = flag;
+}
+} // openage::renderer::opengl
--- a/libopenage/renderer/opengl/render_pass.h
+++ b/libopenage/renderer/opengl/render_pass.h
@@ -1,0 +1,21 @@
+// Copyright 2019-2019 the openage authors. See copying.md for legal info.
+
+#pragma once
+
+#include "../renderer.h"
+
+namespace openage::renderer::opengl {
+
+class GlRenderPass final : public RenderPass {
+public:
+	GlRenderPass(const std::vector<Renderable>&, RenderTarget const*);
+	void set_renderables(std::vector<Renderable>);
+	std::vector<Renderable> get_renderables() const;
+	void set_is_optimised(bool);
+	bool get_is_optimised() const;
+private:
+	/// Whether the renderables order is optimised
+	bool is_optimised;
+};
+
+} // openage::renderer::opengl
--- a/libopenage/renderer/opengl/renderer.cpp
+++ b/libopenage/renderer/opengl/renderer.cpp
@@ -1,6 +1,8 @@
-// Copyright 2017-2018 the openage authors. See copying.md for legal info.
+// Copyright 2017-2019 the openage authors. See copying.md for legal info.
 
 #include "renderer.h"
+
+#include <algorithm>
 
 #include "../../log/log.h"
 #include "../../error/error.h"
@@ -40,6 +42,10 @@ std::unique_ptr<Geometry> GlRenderer::add_bufferless_quad() {
 	return std::make_unique<GlGeometry>();
 }
 
+std::unique_ptr<RenderPass> GlRenderer::add_render_pass(std::vector<Renderable> renderables, RenderTarget const* target) {
+	return std::make_unique<GlRenderPass>(renderables, target);
+}
+
 std::unique_ptr<RenderTarget> GlRenderer::create_texture_target(std::vector<Texture2d*> textures) {
 	std::vector<const GlTexture2d*> gl_textures;
 	gl_textures.reserve(textures.size());
@@ -72,14 +78,31 @@ resources::Texture2dData GlRenderer::display_into_data() {
 	return img.flip_y();
 }
 
-void GlRenderer::render(RenderPass const& pass) {
-	auto gl_target = dynamic_cast<const GlRenderTarget*>(pass.target);
+
+void GlRenderer::optimise(GlRenderPass* pass) {
+	if (!pass->get_is_optimised()) {
+		auto renderables = pass->get_renderables();
+		std::stable_sort(renderables.begin(), renderables.end(), [](const Renderable& a, const Renderable& b) {
+			GLuint shader_a = dynamic_cast<GlUniformInput const*>(a.unif_in)->program->get_handle();
+			GLuint shader_b = dynamic_cast<GlUniformInput const*>(b.unif_in)->program->get_handle();
+			return shader_a < shader_b;
+		});
+
+		pass->set_renderables(renderables);
+		pass->set_is_optimised(true);
+	}
+}
+
+void GlRenderer::render(RenderPass* pass) {
+	auto gl_target = dynamic_cast<const GlRenderTarget*>(pass->get_target());
 	gl_target->bind_write();
 
 	glClearColor(0.0f, 0.0f, 0.0f, 0.0f);
 	glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
 
-	for (auto obj : pass.renderables) {
+	auto gl_pass = dynamic_cast<GlRenderPass*>(pass);
+	GlRenderer::optimise(gl_pass);
+	for (auto obj : gl_pass->get_renderables()) {
 		if (obj.alpha_blending) {
 			glEnable(GL_BLEND);
 		}
--- a/libopenage/renderer/opengl/renderer.h
+++ b/libopenage/renderer/opengl/renderer.h
@@ -1,4 +1,4 @@
-// Copyright 2017-2018 the openage authors. See copying.md for legal info.
+// Copyright 2017-2019 the openage authors. See copying.md for legal info.
 
 #pragma once
 
@@ -10,6 +10,7 @@
 #include "../renderer.h"
 #include "shader_program.h"
 #include "render_target.h"
+#include "render_pass.h"
 
 
 namespace openage {
@@ -29,18 +30,22 @@ public:
 	std::unique_ptr<Geometry> add_mesh_geometry(resources::MeshData const&) override;
 	std::unique_ptr<Geometry> add_bufferless_quad() override;
 
+	std::unique_ptr<RenderPass> add_render_pass(std::vector<Renderable>, RenderTarget const*) override;
+
 	std::unique_ptr<RenderTarget> create_texture_target(std::vector<Texture2d*>) override;
 	RenderTarget const* get_display_target() override;
 
 	resources::Texture2dData display_into_data() override;
 
-	void render(RenderPass const&) override;
+	void render(RenderPass*) override;
 
 private:
 	/// The GL context.
 	GlContext *gl_context;
 
 	GlRenderTarget display;
+
+	static void optimise(GlRenderPass*);
 };
 
 }}} // openage::renderer::opengl
--- a/libopenage/renderer/renderer.cpp
+++ b/libopenage/renderer/renderer.cpp
@@ -1,0 +1,18 @@
+// Copyright 2019-2019 the openage authors. See copying.md for legal info.
+
+#include "renderer.h"
+
+namespace openage::renderer {
+
+RenderPass::RenderPass(std::vector<Renderable> renderables, RenderTarget const* target)
+:renderables(renderables), target(target) {}
+
+RenderTarget const* RenderPass::get_target() const {
+	return this->target;
+}
+
+void RenderPass::set_target(RenderTarget const* target) {
+	this->target = target;
+}
+
+} // openage::renderer
--- a/libopenage/renderer/renderer.h
+++ b/libopenage/renderer/renderer.h
@@ -1,4 +1,4 @@
-// Copyright 2015-2018 the openage authors. See copying.md for legal info.
+// Copyright 2015-2019 the openage authors. See copying.md for legal info.
 
 #pragma once
 
@@ -60,9 +60,17 @@ struct Renderable {
 };
 
 /// A render pass is a series of draw calls represented by renderables that output into the given render target.
-struct RenderPass {
+class RenderPass {
+protected:
+	RenderPass(std::vector<Renderable>, RenderTarget const*);
 	/// The renderables to parse and possibly execute.
 	std::vector<Renderable> renderables;
+
+public:
+	virtual ~RenderPass() = default;
+	void set_target(RenderTarget const*);
+	RenderTarget const* get_target() const;
+private:
 	/// The render target to write into.
 	RenderTarget const *target;
 };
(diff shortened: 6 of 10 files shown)


simonsan:
As a first-time contributor please add yourself to the copying.md ;-)

libopenage/renderer/opengl/renderer.cpp
```diff
@@ -79,7 +81,21 @@ void GlRenderer::render(RenderPass const& pass) {
 	glClearColor(0.0f, 0.0f, 0.0f, 0.0f);
 	glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
 
-	for (auto obj : pass.renderables) {
+	std::vector<Renderable> renderables = pass.renderables;
+	std::sort(renderables.begin(), renderables.end(), [](const Renderable& a, const Renderable& b) {
+		GLuint shader_a = ((GlUniformInput*) a.unif_in)->program->get_handle();
+		GLuint shader_b = ((GlUniformInput*) b.unif_in)->program->get_handle();
```
TheJJ: `static_cast<>` pls

libopenage/renderer/opengl/geometry.cpp
```diff
@@ -51,6 +51,10 @@ void GlGeometry::update_verts_offset(std::vector<uint8_t> const &verts, size_t o
 	this->mesh->vertices.upload_data(verts.data(), offset, verts.size());
 }
 
+GLuint GlGeometry::get_vao_handle() const {
+	return this->mesh->vao.get_handle();
```
Vtec234: This should check the geometry type and throw an informative exception if it's not `geometry_t::mesh`.

libopenage/renderer/opengl/renderer.cpp
```diff
@@ -79,7 +81,21 @@ void GlRenderer::render(RenderPass const& pass) {
...
+
+		auto geom_a = dynamic_cast<GlGeometry const*>(a.geometry);
+		auto geom_b = dynamic_cast<GlGeometry const*>(b.geometry);
+		GLuint vao_a = (geom_a)? geom_a->get_vao_handle() : 0;
+		GLuint vao_b = (geom_b)? geom_b->get_vao_handle() : 0;
+		if (shader_a == shader_b)
+			return vao_a < vao_b;
```
Vtec234: If I remember my own code correctly (not necessarily true, so please double-check if you can), this will not work, for two reasons:
1. In general, not every `GlGeometry` has an underlying mesh. `geometry_t::bufferless_quad` ones are generated at runtime, as will be other ones (ideally we would have more basic polygons). So the call to `get_vao_handle()`.
2. As it stands now, the constructor `GlGeometry` will make a new vertex array (VAO) every time you call it, even with the same `MeshData`, and there is no way to deduplicate them. So no VAO is shared and the ordering does not help.

I think the first step here would be to introduce deduplication (either automatic via something like hashing) at the `GlGeometry` level.

On the other hand, the ordering by shaders should help. So a PR doing just `shader_a < shader_b` would be useful.

Vtec234: GitHub is refusing to let me edit that. Should say "the call to `get_vao_handle()` can fail".

Vtec234: And "either automatic" -> "hopefully automatic".

teldosas: > introduce deduplication (either automatic via something like hashing) at the GlGeometry level.

does that means that `GlGeometry` would have to keep track of all the `GlMesh`es it creates?

Also, after implementing that, wouldn't the ordering by VAOs start to help automatically?

libopenage/renderer/opengl/geometry.cpp
```diff
@@ -51,6 +51,10 @@ void GlGeometry::update_verts_offset(std::vector<uint8_t> const &verts, size_t o
 	this->mesh->vertices.upload_data(verts.data(), offset, verts.size());
 }
 
+GLuint GlGeometry::get_vao_handle() const {
```
Vtec234: Could you rewrite this to `std::optional<GLuint>`? It should return the VAO if the geometry has one, empty otherwise.

libopenage/renderer/opengl/renderer.cpp
Vtec234: > does that mean that GlGeometry would have to keep track of all the GlMeshes it creates?

Hm no, that doesn't sound right - my bad. It should be the responsibility of `GlRenderer` to do the deduplication, perhaps in `add_mesh_geometry`. This will likely require some plumbing in `GlGeometry` - for example replacing the owned `GlVertexArray` with a reference (`shared_ptr` i guess) to one - but the main functionality should be part of `GlRenderer`, since `GlGeometry` is only meant to manage a single object. It would have to detect 'MeshData' reuse somehow, but then the detection mechanism should be fast enough not to outweight the performance gains (if any.. afaik switching VAOs often has little to no cost, unlike switching shaders).

> Also, after implementing that, wouldn't the ordering by VAOs start to help automatically?

Yeah, sounds like it should.

Vtec234: One more thing. For a long list of renderables, it would be better to sort them in-place within the `RenderPass` object rather than on every frame. Two alternatives: either change the `pass` parameter type to `RenderPass&` and update the documentation to note that the object might change, or add a `Renderer::optimize(RenderPass& pass)` function which does this separately from `render`.

teldosas: > It should be the responsibility of GlRenderer to do the deduplication, perhaps in add_mesh_geometry.

My idea is that we:
* add a hash function to `MeshData`
* link the index buffer with the VAO (it's not linked now)
* and in the place where we optimize the `Renderpass` replace all the renderables that have the same `MeshData` hash and shader handle with the same `Renderable`

Vtec234: At a first glance, this looks like it should work. Note though that in principle if the hash matches, we should still check the `MeshData` for exact equality in case of collision. However, that would be quite slow, so if the hash is long enough (64/128-bit?), I think we can get away with assuming no collisions.

libopenage/renderer/opengl/geometry.cpp
```diff
@@ -51,6 +51,10 @@ void GlGeometry::update_verts_offset(std::vector<uint8_t> const &verts, size_t o
 	this->mesh->vertices.upload_data(verts.data(), offset, verts.size());
 }
 
+std::optional<GLuint> GlGeometry::get_vao_handle() const {
+	return this->mesh->vao.get_handle();
```
Vtec234: This should check if `mesh` is non-empty, and only try to return the VAO handle then. Otherwise, return empty.

libopenage/renderer/renderer.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include "renderer.h"
+
+namespace openage {
+namespace renderer {
+
+RenderPass::RenderPass(std::vector<Renderable> renderables, RenderTarget const* target)
+:renderables(renderables), target(target) {}
+
+std::vector<Renderable> RenderPass::get_renderables() const {
```
Vtec234: I don't think an OO-interface for `RenderPass` is necessary, seeing as all members are exposed with get/set anyway. The plain struct from earlier was sufficient and less code to maintain.

libopenage/renderer/renderer.h
```diff
@@ -60,11 +60,22 @@ struct Renderable {
...
+	void set_is_optimised(bool);
+	bool get_is_optimised() const;
+private:
 	/// The renderables to parse and possibly execute.
 	std::vector<Renderable> renderables;
 	/// The render target to write into.
-	RenderTarget const *target;
+	RenderTarget const* target;
+	/// Whether the renderables order is optimised
+	bool is_optimised;
```
Vtec234: This should be initialized to `false`.

libopenage/renderer/renderer.h
```diff
@@ -106,7 +117,7 @@ class Renderer {
 	virtual resources::Texture2dData display_into_data() = 0;
 
 	/// Executes a render pass.
-	virtual void render(RenderPass const&) = 0;
+	virtual void render(RenderPass&) = 0;
```
Vtec234: Could you please also update the documentation to note that a renderer implementation might modify the render pass?

libopenage/renderer/opengl/renderer.cpp
```diff
@@ -72,14 +74,38 @@ resources::Texture2dData GlRenderer::display_into_data() {
...
+				auto vao_a = (geom_a)? geom_a->get_vao_handle() : 0;
+				auto vao_b = (geom_b)? geom_b->get_vao_handle() : 0;
+				if (shader_a == shader_b)
+					return vao_a.value_or(0) < vao_b.value_or(0);
+				else
+					return shader_a < shader_b;
+			});
+
+			pass.set_renderables(renderables);
+			pass.set_is_optimised(true);
```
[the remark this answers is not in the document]
Vtec234: Hm, I thought about it a bit more and it seems that reordering by VAOs is not actually a correct optimization - sorry about this! The reason is that uniform state of every shader is guaranteed to be preserved across a render pass. So for example, with the following renderables : "{set UNIF_A=1, draw VAO_1}, {set UNIF_A=2, draw VAO_2}, {draw VAO_1}", the last renderable can expect `UNIF_A` to be `2` and reordering would break that. So it seems the only valid thing is to do a `stable_sort` (`std::sort` is unstable) by shader handles, but not do anything with VAOs.

libopenage/renderer/opengl/renderer.cpp
Vtec234: This looks like it should work. However, for now I don't think it's necessary since for one, we can't actually reorder by VAO (again, sorry about the confusion) and two, deduplication can easily be done by users of the renderer API since multiple renderables can share a `Geometry*`.

TheJJ:
Anything you need help with so we can get this upstream @teldosas?

libopenage/renderer/renderer.cpp
teldosas: The main reason I introduced an OO-interface is the `is_optimized` field, cause it wouldn't make sense to explicitly set `is_optimized` to `false` each time you create a renderpass

teldosas:
@TheJJ sorry for the delay. I was a bit busy...

libopenage/renderer/renderer.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+RenderPass::RenderPass(std::vector<Renderable> renderables, RenderTarget const* target)
+:renderables(renderables), target(target), is_optimised(false) {}
+
+std::vector<Renderable> RenderPass::get_renderables() const {
+  return this->renderables;
+}
+
+void RenderPass::set_renderables(std::vector<Renderable> renderables) {
+  this->renderables = renderables;
```
Vtec234: Should this set `is_optimised = false`?

teldosas: I was also thinking about that. It should definitely not set `is_optimised = false` because it's only used to replace the renderables with the ones that are optimized. For some reason `std::sort` refuses to directly modify the member value. I was thinking that maybe we should name it `set_optimized_renderables` to avoid undesired use.

Vtec234: But you set `is_optimised` after setting the renderables, so even if it is initially reset to `false`, it should be fine, no?

If you'd like to make some functionality only available to the renderer but not to users, you could follow the design of other components (e.g. `GlShaderProgram`) in which there is a base abstract class `Renderable` and a derived concrete `GlRenderable` which adds on the renderer-only functionality.

teldosas: >But you set is_optimised after setting the renderables

Yes, you're right

>base abstract class RenderPass and a derived concrete GlRenderPass

Sounds like the right thing to do!
So, I did it. Take a look

Vtec234: Hey! Sorry about the late response, this looks good!

libopenage/renderer/opengl/render_pass.cpp
```diff
@@ -0,0 +1,28 @@
...
+    this->renderables = renderables;
+    this->is_optimised = false;
+}
+
+bool GlRenderPass::get_is_optimised() const {
+    return this->is_optimised;
+}
+
+void GlRenderPass::set_is_optimised(bool flag) {
+    this->is_optimised = flag;
```
TheJJ: tab indentation please

libopenage/renderer/opengl/render_pass.h
```diff
@@ -0,0 +1,23 @@
...
+class GlRenderPass final : public RenderPass {
+public:
+    GlRenderPass(const std::vector<Renderable>&, RenderTarget const*);
+    void set_renderables(std::vector<Renderable>);
+    std::vector<Renderable> get_renderables() const;
+    void set_is_optimised(bool);
+    bool get_is_optimised() const;
+private:
+    /// Whether the renderables order is optimised
+    bool is_optimised;
```
TheJJ: tab indentation please

libopenage/renderer/opengl/renderer.cpp
```diff
@@ -40,6 +42,10 @@ std::unique_ptr<Geometry> GlRenderer::add_bufferless_quad() {
 	return std::make_unique<GlGeometry>();
 }
 
+std::unique_ptr<RenderPass> GlRenderer::add_render_pass(std::vector<Renderable> renderables, RenderTarget const* target) {
+    return std::make_unique<GlRenderPass>(renderables, target);
```
TheJJ: tab

libopenage/renderer/opengl/renderer.cpp
```diff
@@ -72,14 +78,31 @@ resources::Texture2dData GlRenderer::display_into_data() {
...
+			std::stable_sort(renderables.begin(), renderables.end(), [](const Renderable& a, const Renderable& b) {
+				GLuint shader_a = dynamic_cast<GlUniformInput const*>(a.unif_in)->program->get_handle();
+				GLuint shader_b = dynamic_cast<GlUniformInput const*>(b.unif_in)->program->get_handle();
+				return shader_a < shader_b;
+			});
+
+			pass->set_renderables(renderables);
+			pass->set_is_optimised(true);
+		}
+	}
```
TheJJ: dedent by one tab please

libopenage/renderer/opengl/renderer.h
```diff
@@ -29,18 +30,22 @@ class GlRenderer final : public Renderer {
 	std::unique_ptr<Geometry> add_mesh_geometry(resources::MeshData const&) override;
 	std::unique_ptr<Geometry> add_bufferless_quad() override;
 
+    std::unique_ptr<RenderPass> add_render_pass(std::vector<Renderable>, RenderTarget const*) override;
```
TheJJ: tab indent pls

libopenage/renderer/renderer.h
```diff
@@ -60,11 +60,19 @@ struct Renderable {
...
+class RenderPass {
+protected:
+    RenderPass(std::vector<Renderable>, RenderTarget const*);
+    /// The renderables to parse and possibly execute.
+    std::vector<Renderable> renderables;
+
+public:
+    virtual ~RenderPass() = default;
+	void set_target(RenderTarget const*);
+	RenderTarget const* get_target() const;
```
TheJJ: tab indentation pls

libopenage/renderer/renderer.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+RenderPass::RenderPass(std::vector<Renderable> renderables, RenderTarget const* target)
+:renderables(renderables), target(target) {}
+
+RenderTarget const* RenderPass::get_target() const {
+  return this->target;
+}
+
+void RenderPass::set_target(RenderTarget const* target) {
+  this->target = target;
```
TheJJ: tab indentation please

libopenage/renderer/renderer.h
```diff
@@ -60,11 +60,19 @@ struct Renderable {
...
+private:
 	/// The render target to write into.
-	RenderTarget const *target;
+	RenderTarget const* target;
```
TheJJ: our codebase has the convention to align `*` and `&` to the right (because semantically the `*` belongs to the variable, and not the type)

libopenage/renderer/opengl/render_pass.cpp
```diff
@@ -0,0 +1,28 @@
+// Copyright 2017-2019 the openage authors. See copying.md for legal info.
+
+#include "render_pass.h"
+
+namespace openage {
+namespace renderer {
+namespace opengl {
```
TheJJ: c++17 allows nested namespaces like this:
```
namespace openage::renderer::opengl {...}
```

we introduce this over the whole codebase in #1171, but as this is a new file you should do it right away

simonsan: this nested namespace stuff i would do in another PR with format fixing stuff. 1171 is deeper in the codebase and some namespace in the files we already changed things in. so while we were there I also updated the namespaces. but true, we should do it now right away <3

libopenage/renderer/opengl/render_pass.cpp
teldosas: ah I thought make check all checks for all this... will get to it right away

teldosas:
It should be fine now 😊

PR #1188, opened 2019-11-11
Adding favicon for webpages to assets

simonsan:


TheJJ:
I really hope shitty `.ico` files will be replaced by vector graphics very soon.

TheJJ:
https://caniuse.com/#feat=link-icon-svg maybe we should go for those in cases where the icon url is customizable

simonsan:
> https://caniuse.com/#feat=link-icon-svg maybe we should go for those in cases where the icon url is customizable (or at least png favicons https://caniuse.com/#feat=link-icon-png)

Nice! Didn't know png is "already" a thing :-D Sure, we should also go for this.

Issue #1185, opened 2019-11-02
More information on openage --version 


simonsan:
Right now we give the following information (e.g.):

```
openage v0.4.0-69-g9412facd
opengl
MSVC 19.16.27032.1 [/DWIN32 /D_WINDOWS /W3 /GR /EHsc /MP]
Cython 0.29.14
```

For future debugging on the user side we could give more information:
- OpenGL version number
- nyan-lib version number
- libc
- sdl-version 
- Qt-version
- maybe even opus-codec version

Mese96:
As this is labelled as "good first issue", I would like to try it, as soon as I find some time this month. 
For a first time contributor, anything I need to know ? 
(Except reading contributing.md and getting the project to build)

heinezen:
@Mese96 You should be good to go, it only involves fundamental Python knowledge :)

Mese96:
Need to take a closer look where to get version numbers for OpenGL and libc

simonsan:
If you have any questions feel free to ask us via IRC or [Matrix](https://riot.im/app/#/room/#sfttech:matrix.org). ;-)

abowassm:
> As this is labelled as "good first issue", I would like to try it, as soon as I find some time this month.
> For a first time contributor, anything I need to know ?
> (Except reading contributing.md and getting the project to build)

abowassm:
#1136

abowassm:
[Copy of 9097f21.zip](https://github.com/SFTtech/openage/files/3908790/Copy.of.9097f21.zip)

FrancoB4:
Hi, I’d like to contribute and this seems like a good first issue to start with. Is there anything I should know before getting started?

PR #1191, opened 2019-11-13
Add Libc, SDL, Qt and Nyan versions to openage --version

Mese96:
#1185


heinezen:
Is this finished or do you have things left to do? :)

Mese96:
Infos of nyan, sdl and qt are finished. 
Did not work on the other demanded infos.

buildsystem/modules/FindSDL2.cmake
```diff
@@ -132,6 +132,23 @@ if(SDL2_LIBRARY_TEMP)
...
+  string(REGEX REPLACE "^#define[ \t]+SDL_MAJOR_VERSION[ \t]+([0-9]+)$" "\\1" SDL2_VERSION_MAJOR "${SDL2_VERSION_MAJOR_LINE}")
+  string(REGEX REPLACE "^#define[ \t]+SDL_MINOR_VERSION[ \t]+([0-9]+)$" "\\1" SDL2_VERSION_MINOR "${SDL2_VERSION_MINOR_LINE}")
+  string(REGEX REPLACE "^#define[ \t]+SDL_PATCHLEVEL[ \t]+([0-9]+)$" "\\1" SDL2_VERSION_PATCH "${SDL2_VERSION_PATCH_LINE}")
+  set(SDL2_VERSION_STRING ${SDL2_VERSION_MAJOR}.${SDL2_VERSION_MINOR}.${SDL2_VERSION_PATCH})
+  unset(SDL2_VERSION_MAJOR_LINE)
+  unset(SDL2_VERSION_MINOR_LINE)
+  unset(SDL2_VERSION_PATCH_LINE)
+  unset(SDL2_VERSION_MAJOR)
+  unset(SDL2_VERSION_MINOR)
+  unset(SDL2_VERSION_PATCH)
```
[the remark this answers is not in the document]
simonsan: That's a good solution!

Mese96:
This should be enough to satisfy the requested changes. 

But as seen in the commit history i have not **QUITE** figured out how to rebase while having a pull request, leading to me doing changes multiple times.
What is the correct workflow here ?

TheJJ:
Hi! Oh well, that branch now sure looks like a mess :)
For rebasing, we have a [guide here](https://github.com/SFTtech/openage/blob/master/doc/contributing.md#rebasing).

As the branch is now pretty broken, you can either remove all bad commits in the interactive rebase, or alternatively, use `cherry-pick` to only select the good ones for a new branch. up to you. if you need more interactive assistance, come by in our chat-room :smile:

Mese96:
Yeah, fortunately @simonsan was so kind and fixed it for me. 
Need to experiment with rebasing etc. some time. 

You can just fix those whitespace errors. 
But out of curiosity, what kind are we talking about and how do you prevent them ?

TheJJ:
For each of our used languages, we use a consistent indentation scheme: CMake-Code uses tabs, C++ too, Python uses spaces. We have a document that [describes our usage of tabs](https://github.com/SFTtech/openage/blob/master/doc/code_style/tabs_n_spaces.md).

The insertion of the correct whitespace depends on your Editor and mode. If your editor doesn't insert the right stuff automatically, you can make the space types visible (i.e. display a symbol instead of a tab character, ...).

If you want to try to highlight space types and fix the indentation yourself, go for it.

Mese96:
I hope this is enough.

TheJJ:
uh-oh, now the rebase mess is back again :grin:

buildsystem/modules/FindSDL2.cmake
```diff
@@ -132,6 +132,27 @@ if(SDL2_LIBRARY_TEMP)
...
+		LINK_LIBRARIES ${SDL2_LIBRARY}
+		CMAKE_FLAGS "-DINCLUDE_DIRECTORIES=${SDL2_INCLUDE_DIR}" "-DCMAKE_CXX_STANDARD=17"
+		RUN_OUTPUT_VARIABLE SDL2_VERSION_OUTPUT
+	)
+
+	if(SDL_VERSION_TEST)
+		message(STATUS "SDL versions: ${SDL2_VERSION_OUTPUT}")
+		set(SDL2_VERSION_STRING ${SDL2_VERSION_OUTPUT})
+	else()
+		message(FATAL "uhh something failed")
```
janisozaur: That's not nice for cross-compilation. Why can't it use sdl-provided define?

TheJJ: I don't follow, what define do you mean and how can we improve the patch? :)

janisozaur: In cross-compilation you wouldn't usually be able to execute programs, but you should be able to get SDL version (both compiled against and linked to in runtime) using the API exposed in https://github.com/SDL-mirror/SDL/blob/e50629c5a126c82180ccbeb298f4e2360fa3925f/include/SDL_version.h#L60

simonsan: @janisozaur Could you please give an example on how you would do it?

TheJJ: Sorry for the delay :)
I get what you mean, but then we're back to what the pull request originally did: It parsed the header file to extract the values with a regex.

Then I suggested using the built in API function to fetch the version because it seems like the official way to get it. But I understand that for cross-compilation this wouldn't work.

Can't we build the version-fetching-miniprogram with "`HOSTCC`" and then run it? CMake surely supports something like that, and we'd still get the version through the API.

janisozaur: My point still holds: why do you need it in the cmake in the first place? Surely not just for the sake of reporting the patch level of SDL version found during build time, as that has no value to the person compiling… so what for?

TheJJ: We have it in the build system so it can generate strings that contain the version of libraries the game was linked to. And this has to be done at compile time (hence CMake) to store the compiletime linkage versions.
What's your suggestion on how we should do it instead? Parse the header file (I mean if we have to, so be it, we can write helper functions that find a `#define` that can then also be used for other library headers)?

janisozaur: But why? Why can't you just report the version from define?

TheJJ: We extract the version at compile-time so it can be stored in `openage/config.py` and then printed in `openage/__init__.py` as `LONGVERSION`.
So, we store the versions at compile time to python.

The alternative, and that's what you're suggesting, right?, is to do this in C++, and just use the header file and get the value. It does seem like the better approach, and the python code could fetch C++ library versions via an libopenage-api-call and print python library versions directly. Then we'd have unified the approach and there would be no need for a program run at compiletime.

Implementing that is a bit more complicated, but seems the even better aproach, you're right :)

janisozaur: Yes, just compile it in via the define from actual include and if need be extract via runtime query on target platform.

simonsan: @Mese96 Do you feel you can do it this way?

Mese96: Yeah, if this is now the way to go ;)
I need to look how to do "an libopenage-api-call" but otherwise it should be ok. 
If I understand this correctly this also should include a more general approach, useable for all version numbers ?

TheJJ: yes correctly, this is a generic approach where we store the version numbers of C libraries in C++ instead of a python file (which we create through hacks like parsing the header file or compiletime-program-runs).

The approach is the following:
* store the version variables only in c++ in the config.h file (that method is already implemented)
* add a function that can read these version variables and return them as a key-value map
* access the version variables through cython: we could have a `versions.pyx` file that calls to the above function and converts the c++ key value map with all the versions to a python `dict`
* implement the python version printing by calling the cython function that fetches the versions from c++.

An example of the cython-c++ interface is e.g. our `util/path` system or the `compress/lxzd`. Have a look at:
* https://github.com/SFTtech/openage/blob/master/doc/code/pyinterface.md
* libopenage/util/compress/lzxd.h

If you need further help with cython, just ask me in the chatroom, please :)

simonsan:
@Mese96 I merged you the changes I've made to the same files for checking the config version of the CI in use. It is set via an environment variable called `CI_CFG_VERSION` only to let you know for future changes. =)

Mese96:
@simonsan Thanks for the description will look into it at some point next month.

Mese96:
SDL and libc are now queried trough Cython. 
Everything else is still how it was. 
But now seems the macos build broken because there is no "#include <gnu/libc-version.h>"
Does anyone know where to get the version from ?

@TheJJ Seems everything else ok ? 

And last but not least: 
The commit history
Tried the interactive rebase, but got not very far. 
Technically I only need parts of the old commits. And if I would have two, one with the general stuff and one with the new cython method that would be enough. 
I admit that I am a little bit lost.

simonsan:
Hey there, any news on this? :-)

Mese96:
@TheJJ gave me a tutorial how to rebase. 
I think I did that. 
So the only remaining problem I know of is the libc version on MacOS.

simonsan:
@Mese96 I rebased again, please take a look if everything is still as you expected it to be.

> So the only remaining problem I know of is the libc version on MacOS.
You mean that one?
>> But now seems the macos build broken because there is no "#include <gnu/libc-version.h>"

I think that is not so important in this moment. The important thing is, that Kevin is building - and it does.

simonsan:
@Mese96 I check it on my Ubuntu-VM and it works fine. 👍 
This is my output:
```
openage v0.4.0-126-g3b0dd365 [devmode]
inotify, opengl
Clang 6.0.0 []
Cython 0.26.1
CI config version NOT SET
SDL 2.0.8
libc 2.27
```
How do you feel, are you eager to include the other versions as well or do you feel like you want to leave it at that for now? This should not be read like a reproachful question just to know whether we should merge this or wait with it?

Otherwise these here is an overview from #1185 :
- [ ] OpenGL version number
- [ ] nyan-lib version number
- [x] libc
- [x] sdl-version
- [ ] Qt-version
- [ ] opus-codec version
- [ ] version numbers of python determined during compile time 
- [ ] packaged python version on the user side that starts the openage module (resides either in install folder `openage/python/python` with packaging (e.g. win) or probably just in `PATH` on Linux (and Mac?))

Mese96:
@simonsan I checked, looks good. 
But it should print nyan and Qt numbers as well. 
(At least it is doing it here)

> You mean that one?
> 
> > > But now seems the macos build broken because there is no "#include <gnu/libc-version.h>"

Yeah, i mean this one. 
If thats not important, great. Maybe I find some time to look into it next week. 
Regardless I would like to see this merged, with all numbers we have now. 
We can always add new numbers

heinezen:
@Mese96 I agree, we can merge this and add more features later.

heinezen:
:tada:

simonsan:
@Mese96 
> But it should print nyan and Qt numbers as well.

Hmm ok. Well I was using `./run --version` directly after building and not after installing/packaging. Probably because of this?

Mese96:
@simonsan God damn, you are right, its gone.
My brain was afk and in Sunday mode it seams. 
Of course I should rebuild. Too much Unity it seems. :D
I looked after the new changes which uses the c++ part. 
I checked again, and they got lost in your rebase. 
But that is not so bad, it only used numbers already available. 
Its just 3 lines per version number. 
I already restored it, and will just make a new pull request. 
At least I get some practice :smile:

simonsan:
> 
> 
> @simonsan God damn, you are right, its gone.
> My brain was afk and in Sunday mode it seams.
> Of course I should rebuild. Too much Unity it seems. :D
> I looked after the new changes which uses the c++ part.
> I checked again, and they got lost in your rebase.
> But that is not so bad, it only used numbers already available.
> Its just 3 lines per version number.
> I already restored it, and will just make a new pull request.
> At least I get some practice 😄

Ah ok, yeah I'm sorry. I already thought there could have gone something missing because everything was in one commit from the last months and I had to manually decide if it was related or not. Hope it's fine though. 👍

Mese96:
@simonsan 
https://github.com/SFTtech/openage/pull/1227

PR #1193, opened 2019-11-18
SMX file support

heinezen:
Adds support for SMX files introduced in DE2.

- [X] SMX to PNG
- [x] Format documentation

### DIFF
--- a/openage/convert/slp.pyx
+++ b/openage/convert/slp.pyx
@@ -1018,25 +1018,27 @@ cdef numpy.ndarray determine_rgba_matrix(vector[vector[pixel]] &image_matrix,
 
             elif px_type == color_player_v4:
                 r, g, b = p_lookup[px_val]
-                alpha = 255
+                # TODO: Make this 255 with new renderer
+                # mark this pixel as player color
+                alpha = 254
 
             else:
                 if px_type == color_player:
+                    # TODO: Make this 255 with new renderer
                     # mark this pixel as player color
-                    alpha = 255
+                    alpha = 254
 
                 elif px_type == color_special_2 or\
                      px_type == color_black:
-                    # mark this pixel as outline
-                    alpha = 253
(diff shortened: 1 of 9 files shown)


openage/convert/slp.pyx
```diff
@@ -1023,7 +1023,7 @@ cdef numpy.ndarray determine_rgba_matrix(vector[vector[pixel]] &image_matrix,
             else:
                 if px_type == color_player:
                     # mark this pixel as player color
-                    alpha = 254
+                    alpha = 255
```
TheJJ: this should remain 254 as we currently filter for 254 in `libopenage/game_renderer.cpp`:
```
glUniform1f(teamcolor_shader::alpha_marker_var, 254.0/255.0)
```

also, this would require people to reconvert the graphics.

openage/convert/smx.pyx
```diff
@@ -0,0 +1,1145 @@
...
+                        pixel_data.push_back((pixel_data_odd_0 >> 2) | (pixel_data_odd_1 << 6))
+
+                        # Palette section. Described in byte[2] in bits 4-5.
+                        pixel_data.push_back((pixel_data_odd_1 >> 2) & 0x03)
+
+                        # Damage mask 1. Essentially a rotation of (byte[3]byte[4])
+                        # by 6 to the left, then masking with 0x00FF.
+                        pixel_data_odd_2 = self.get_byte_at(dpos_color + 3)
+                        pixel_data_odd_3 = self.get_byte_at(dpos_color + 4)
+                        pixel_data.push_back((pixel_data_odd_0 >> 2) | (pixel_data_odd_1 << 6))
```
sandsmark: hmm, should this be pixel_data_odd_2/3?

heinezen: Yes, thank you! There was also an issue with the masking there which is now fixed.

PR #1157, opened 2019-09-25
Add clang-tidy to buildsystem checks

simonsan:
Fixes #1060 continued from PR #1063 

Regarding #49 

Plan is: 

- [ ] Add a `diagnose` python-module that is making the checks after you call `./configure` (compile-commands.json) and add several other code-checks (clazy, etc.).
- [ ] Add a make target: `make diagnose` for all that checkups
- [ ] Export a (todo) list with warnings and errors for better overview

PR #1196, opened 2019-11-20
[Umbrella] Adding Github workflows

simonsan:
ci:  adding windows and ubuntu workflows,
doc: adding status badges for mac and win to readme.md

For the caching there is still some fine-tuning needed in the future. I'm also waiting for an increase of the limit and possibility to cache `vcpkg/installed`-dir of github-actions (actions/cache#6) and adding vcpkg packages to the virtual env (actions/virtual-environments#11) and that Qt gets added to the images (actions/virtual-environments#8). That overall could lower the win build for around 12 minutes.

The Ubuntu-CI is **not running** inside this main repo, but just outside for any forks of **openage**. People can make future use of this for pre-diagnosing `clang-tidy`-errors or `clang-format`-stuff inside their forked repositories. This functionality (clang-tidy and clang-format) is not added with this PR though and might come in the future.

### Win
- [ ] Investigate after build test output not being forwarded by shell
- [ ] Packaging Nightly builds directly on Github Releases
- [ ] Look into `install prefix` on x64 target being `program files (x86)` => [Around this function](https://gitlab.kitware.com/cmake/cmake/blob/8065a686ddcb54b10fd6758eca24280f3e68926b/Modules/CMakeGenericSystem.cmake#L68-156) (thanks tu) (**can't reproduce for now, later?**)
- [ ] User PWSH interface to [Windows Defender](https://docs.microsoft.com/en-us/windows/security/threat-protection/windows-defender-antivirus/command-line-arguments-windows-defender-antivirus) to scan package before publishing
- [x] Make use of Compiler Cache ([clcache](https://github.com/frerich/clcache/wiki/Integration), [cclash](https://github.com/inorton/cclash))
- [x] Add build matrix for cases ([versions of Python (Upstream Issue)](https://github.com/actions/setup-python/issues/86) or included in [win environments](https://github.com/actions/virtual-environments/issues/786), versions of compilers)
  - [x] Python 
  - [x] MSVC would be a good start into a build matrix => only Visual Studio 2019 on newest Windows builds
  - [x] Architecture is not needed now, but prepared, as we are not building/testing/packaging for 32bit platforms currently 
- [x] Add caching for vcpkg-dir using [run-vcpkg](https://github.com/marketplace/actions/run-vcpkg) ([example](https://github.com/lukka/CppBuildTasks-Validation/blob/master/.github/workflows/hosted-basic-cache-submod_vcpkg.yml) using [run-cmake](https://github.com/marketplace/actions/run-cmake) as well)

### Win-MSYS2
- [ ] MSYS2 could be an easy replacement where we could build on gcc/g++ and clang as well in a build matrix setup (would be easy to cache as well, one folder)
- [ ] Packaging Nightly builds directly on Github Releases
- [ ] Make use of ccache

### Linux
- [ ] New `make` targets
  - [ ] Code Analysis: Get [clang-tidy](https://github.com/StableCoder/cmake-scripts/blob/master/tools.cmake) [**script**](https://github.com/SFTtech/openage/pull/1196/files#diff-d328e269e5afad08dfe2dd3910ec4896) to work and let it upload a fixes-patch as artifact (#1157)
  - [ ] Code Formatting: Get [clang-format](https://github.com/StableCoder/cmake-scripts/blob/master/formatting.cmake) to work and let it upload a fixes-patch as artifact (#1167)
Ideas:  
  - [ ] [Sanitizer builds](https://github.com/StableCoder/cmake-scripts/blob/master/sanitizers.cmake)?
  - [ ] [Code coverage reports](https://github.com/StableCoder/cmake-scripts/blob/master/code-coverage.cmake)?
  - [ ] [VVVerbose builds](https://github.com/StableCoder/cmake-scripts/blob/master/compiler-options.cmake)?
  - [ ] [Dependency Graph](https://github.com/StableCoder/cmake-scripts/blob/master/dependency-graph.cmake)
- [x] Lockfile: Use config for the same system as lockfile so config changes rebuild cache 
- [x] File headers: Add header from copying.md to scripts

### Mac
- [ ] Packaging?
- [x] Add build matrix for cases (versions of Python, versions of compilers)
- [x] Check current workflow for `fonts` problem
- [x] Wait for the 2 GiB individual cache limit and cache brew+llvm, use brew again to install Qt

### DIFF
--- a/.github/workflows/ubuntu-ci.yml
+++ b/.github/workflows/ubuntu-ci.yml
@@ -1,0 +1,65 @@
+name: Ubuntu-CI
+on: [push, pull_request]
+
+jobs:
+  ubuntu-build:
+    if:
+      contains(github.repository, 'SFTtech')==false # CI for forks, lets this job run just outside of the main repo
+    strategy:
+      matrix:
+        cc: [gcc, clang-9]
+    runs-on: ubuntu-latest
+    env:
+      CC: ${{matrix.cc}}
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v1
+      - name: Cache (ccache)
+        uses: actions/cache@v1.0.2
+        id: cache-ccache
+        if: startsWith(runner.os, 'Linux')
+        with:
+          path: ~/.ccache
+          key: ${{ runner.os }}-${{matrix.cc}}-ccache
+          restore-keys: |
+            ${{ runner.os }}-${{matrix.cc}}-ccache
+      - name: Setup apt repositories
+        run: |
+          if [ $CC = clang-9 ]; then
+            sudo add-apt-repository 'deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-9 main' -y
+          fi
+      - name: Update apt repositories
+        run: sudo apt-get update -q
+      - name: Install Dependencies
+        run: |
+          sudo apt-get install -y --no-install-recommends ccache cmake cython3 libeigen3-dev libepoxy-dev libfontconfig1-dev libfreetype6-dev libharfbuzz-dev libogg-dev libopus-dev libopusfile-dev libpng-dev libsdl2-dev libsdl2-image-dev python3-dev python3-jinja2 python3-numpy python3-pil python3-pip python3-pygments pylint3 python3-pycodestyle qml-module-qtquick-controls qtdeclarative5-dev flex make
+      - name: Install C compiler
+        run: |
+          if [ $CC = clang-9 ]; then
+            sudo apt-get install -y --no-install-recommends clang-9 lld-9 libc++-9-dev libc++abi-9-dev clang-tools-9
+            export CXX=clang++-9
+            export CC=clang-9
+          else
+            sudo apt-get install -y --no-install-recommends $CC g++
+          fi
+      - name: Show environment settings
+        shell: bash
+        run: |
+          set -x
+          uname -a
+          uname -s
+          uname -m
+          env
+      - name: Generate build files
+        run: |
+          if [ $CC = clang-9 ]; then
+            ./configure --compiler=clang++-9 --mode=debug --flags="-Werror" --ccache --download-nyan
+          else
+            ./configure --mode=debug --flags="-Werror" --ccache --download-nyan 
+          fi
+      - name: Build
+        run: make -j$(nproc) build
+      - name: Test
+        run: make tests
+      - name: Install
+        run: make install DESTDIR=/tmp/openage
--- a/.github/workflows/windows-ci.yml
+++ b/.github/workflows/windows-ci.yml
@@ -1,0 +1,241 @@
+name: Windows 10
+on: [push, pull_request]
+env:
+  QT_INSTALL_VERSION: '5.14.2'
+  vcpkgResponseFile: '${{ github.workspace }}\.github\workflows-support\response_file.txt'
+  vcpkgGitCommitID: '28ab0b15cd7cdc1004cac9a28a5ecab198c161b9'        # Updated 26/04/2020
+  nyanGitCommitID: '03a996994b87803b5a8ff6ed0e85ab730163aa2d'         # Updated 27/04/2020
+  CI_CFG_VERSION: 'github-win-0.1.0-dev'
+
+jobs:
+  build_windows_msvc:
+    name: MSVC2019 (${{ matrix.arch }}) (${{ matrix.build-type }}) Python ${{ matrix.python-version }}
+    runs-on: windows-2019
+    continue-on-error: false                                          # DEBUG: set to true
+    strategy:
+      fail-fast: false
+      matrix:
+          arch: [x64] #[Win32, x64]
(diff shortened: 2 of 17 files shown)


.github/workflows/windows-ci.yml
```diff
@@ -0,0 +1,124 @@
...
+        uses: actions/checkout@v1
+      - name: Cache
+        uses: actions/cache@v1.0.2
+        id: cache-python-dl
+        with:
+          path: 'Python38-x64-dbg'
+          key: ${{ runner.OS }}-python
+          restore-keys: |
+            ${{ runner.OS }}-python
+            ${{ runner.OS }}-
```
simonsan: The linkes issues have influence on this part.

.github/workflows/ubuntu-ci.yml
```diff
@@ -0,0 +1,51 @@
+name: Ubuntu-CI
+on: [push, pull_request]
+
+jobs:
+  ubuntu-build:
+    if:
+      contains(github.repository, 'SFTtech')==false
```
simonsan: This lets Ubuntu run outside of the main-repo.

.github/workflows/ubuntu-ci.yml
```diff
@@ -0,0 +1,51 @@
...
+          if [ $CC = clang ]; then
+            sudo apt-get install -y --no-install-recommends clang-9
+            export CXX=clang++
+          else
+            sudo apt-get install -y --no-install-recommends $CC g++
+          fi
+      - name: Sanity check
+        run: make checkall # This is skipped as long as there are issues in the main repo.
+        if: 
+          contains(github.event, 'pull_request')
```
simonsan: Sanity check is skipped as long as there are issues in the main repo.

```
************* Module openage.convert.singlefile
E:146,24: Instance of 'PurePath' has no 'open' member (no-member)
************* Module openage.convert.main
E:506, 0: Bad option value 'possibly-unused-variable' (bad-option-value)
W:507,10: Unused variable 'game_versions' (unused-variable)
W:513, 4: Unused variable 'save' (unused-variable)
W:521, 4: Unused variable 'save_slp' (unused-variable)
W:539, 4: Unused variable 'pprint' (unused-variable)
************* Module openage.testing.testing
E: 60, 0: Bad option value 'try-except-raise' (bad-option-value)
************* Module openage.util.fslike.abstract
E: 96, 0: Bad option value 'useless-return' (bad-option-value)
************* Module buildsystem.pxdgen
C:216, 8: Missing function docstring (missing-docstring)
C:220, 8: Missing function docstring (missing-docstring)
C:228, 8: Missing function docstring (missing-docstring)
C:243, 8: Missing function docstring (missing-docstring)
```

.github/workflows/ubuntu-ci.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Generate build files
+        run: |
+          if [ $CC = clang ]; then
+            ./configure --compiler=clang++ --mode=debug --flags="-Werror" --ccache --download-nyan
+          else
+            ./configure --mode=debug --flags="-Werror" --ccache --download-nyan
```
simonsan: I've added Ccache to the Ubuntu and MacOSX builds, for future use within PRs, where we could cache the `build-dir` after the first build and then pull in the newer commits and run configure again, speeding up the whole process. Also for people in their forks something like this might be a nice feature.

.github/workflows/windows-ci.yml
simonsan: Someone had a tip how to get parts of the vcpkg-dir cached. Pack them, move them into the `github_workspace` and then cache them from there and vice versa (you can just cache from inside the workspace/cloning dir). I'll test that out later, although I think that the directory will be **>400 MiB** (that is the individual cache size, 2 GiB is the overall cache size) and therefore we maybe need to find another solution. I asked some stuff in their repos about this, so let's see what caching will bring.

.github/workflows/windows-ci.yml
```diff
@@ -0,0 +1,124 @@
...
+        shell: cmd
+        if: matrix.arch == 'Win32'
+      - name: Download Python (x64)
```
goto-bus-stop: Python 3.8 should be preinstalled, according to the docs:
https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners#python

Does that not work for openage? if so, a comment explaining why might be nice!

simonsan: Good point, sadly the preinstalled Python often doesn't come with Debug libraries and precompiled standard libraries, so basically we need to reinstall that everytime we use CI. Will make a comment on that to explain.

.github/workflows/ubuntu-ci.yml
simonsan: From Pylint 3, Pylint 2.x doesn't show this.

simonsan:
Cache limit for workflows got increases, so we probably could cache the complete Qt and vcpkg dir to make build times smaller.

> per-cache limit to 2GB to match the per-repo limit
Ref.: https://github.com/actions/cache/issues/6#issuecomment-571290282

simonsan:
(Outdated)

PR #1203, opened 2019-11-29
Add CI config version to openage --version

simonsan:
This adds the version of a config file for the ci from an environment variable to version output of openage for later debugging.


TheJJ:
What would that version represent?

simonsan:
> Could the CI echo that version in its execution script? Or do we need the version to be compiled-in?

Both. CI will output that to the console and as well we need it compiled in. With this it would be easier to find problems that originate from compile time problems. Right now we can not see with which config a package was built, so we don't know if a bug was fixed due to changes in the CI config or changes to pulled in versions etc.

TheJJ:
Do you mean in case we distribute the CI-built package? Because if not, the version would never leave the CI system that echoed the version number it built with.

simonsan:
> Do you mean in case we distribute the CI-built package?

Yes, it's mainly for packages that we built with the CI and config changes that happen after distributing it. Also because the software on the CI systems (besides Kevin) could be changed without our immediate knowledge (updates to the base image). To diagnose errors in these moments (build fails, package is broken, etc.) it's easier to recheck the config version for changes. Overall it makes diagnosing very much easier,.

TheJJ:
We should restructure the python version reporting in the way we figured out in #1191: Store all the version information in c++, and query it from python via the cython-api. That way we define the versions only once and can also embed versions like the sdl-compiletime-version conveniently.

Issue #985, opened 2018-03-17
Convert: Flat terrain textures


heinezen:
As mentioned in #720, we want to convert the old SLP terrain files to a flat terrain texture. I've implemented a working prototype for that utilizing the dimetric projection algorithm from the openage modding repository. The resulting python script can be found [here](https://gist.github.com/heinezen/6ec3e95becc26bb3bb794cf7d66f266a). It operates on the output of our current converter.

`merge()` puts the single tiles together into one giant tile. `transform()` projects the result from dimetric to the desired flat perspective. Here you see the output when converting `15006.slp.png`.

![15006_flat_t](https://user-images.githubusercontent.com/6852422/37560012-930d5dd6-2a30-11e8-9504-3067557ec3dd.png)

Textures from SLPs will always have a ratio of 1:1 and a size of `481x481` (compared to `512x512` in HD).

If somebody want to implement a more efficient version of the script and integrate it into the converter, go ahead.

mikadou:
Since terrain textures in HD are 512x512 pixels, I suggest to generate flat terrain textures with the same resolution in the conversion step.

heinezen:
@mikadou We might not need to anymore, since we agreed on doing 3D terrain. The texture will be independent from the tiling and thus can just be painted on the terrain without the need to scale it.

mikadou:
@heinezen Thanks for confirming the consensus on 3d terrain. I fully agree with going in this direction.
My thought about the 512x512 textures was because a few years ago when I last experimented with OpenGL it was recommended to use textures with sizes power of two. Not sure if this is still relevant though.

TheJJ:
For the usual on-GPU memory allocation algorithms I think it still is (because the malloc algorithms likely use `2^x` bucket sizes). With today's huge memory sizes I think we should ignore this for now and look for optimizations later.

simonsan:
Just to save this link:

How to generate 3D terrain from e.g. Google Maps
https://www.youtube.com/watch?v=Bb0HnaYNUx4

charliergi:
Anyone is working on it ? I can try to do it.

heinezen:
@charliergi No, nobody working on it currently. You can do it if you want :)

It would be great if you could translate the code to [cython](https://cython.org/) (for extra speed) and integrate the script directly into the converter (`openage/convert/` in the repo). You can therefore ignore the output logic from the script linked above and instead take a `numpy.ndarry` as input for the image data. The array is returned by our media conversion:

[...]

PR #1204, opened 2019-12-09
draft for convert_terrain

charliergi:
So here I am with my first contribution, as asked in #985 
There is a lot of optimization that must be done, but first I need to understand the logic of the code.
The code don't run because I can't extract properly the special `SLP` data from an `SLP`, and pass it to `determine_rgba_matrix()`

Thanks in advance for your feedback.


openage/convert/singlefile.py
```diff
@@ -256,3 +259,79 @@ def read_smx_file(smx_path, main_palette, player_palette, output_path):
...
+         interfac_file.name, palette_index)
+    palette_file = DRS(
+        interfac_file).root["%s.bina" % palette_index].open("rb")
+
+    info("parsing palette data...")
+    palette = ColorTable(palette_file.read())
+
+    # import here to prevent that the __main__ depends on SLP
+    # just by importing this singlefile.py.
+    from .slp import determine_rgba_matrix, SLP, SLPFrame, FrameInfo
```
heinezen: `determine_rgba_matrix` is a cdef C function and this means it is *not* callable from here because `singlefile.py`. You need to call `get_palette_info` for a frame as you do in line 321 as that will call `determine_rgba_matrix` from cython for you.

openage/convert/singlefile.py
```diff
@@ -256,3 +259,79 @@ def read_smx_file(smx_path, main_palette, player_palette, output_path):
...
+        row+=1
+        count+=1
+
+    #matrix = determine_rgba_matrix(image, None, 0) ???
+    image_new = Image.fromarray(image)
+    info("merging RGBA Matrix")
+    merged_image = Terrain_convert.merge(image_new)
+
+    info("transforming RGBA Matrix")
+    transformed_image = Terrain_convert.transform(merged_image)
```
heinezen: This is the part we would like to optimize well. Instead of creating an image with pillow and then transforming that, you should directly transform the ndarray and then save it as an image.

openage/convert/terrain_convert.pyx
```diff
@@ -0,0 +1,89 @@
...
+
+import math
+import argparse
+from PIL import Image
+from .slp import SLP
+from .texture import Texture
+
+class Terrain_convert:
+
+    def merge(img):
```
heinezen: For efficiency, the merge and transformation should be written as cdef methods and operate on an ndarray instead of pillow images. You can achieve that by defining and writing `merge` and `transform` as cdef methods and call them from a normal Python def method. The def method will be callable from "the outside", i.e. other Python modules. See `slp.pyx`. You can see how it works there with `get_picture_data` and `determine_damage_matrix`.

You also want as much C logic as possible for maximum speed.

openage/convert/singlefile.py
```diff
@@ -61,9 +61,12 @@ def main(args, error):
...
         if not (args.drs and args.palette_index):
             raise Exception("palette-file needs to be specified")
+        
+        # Only for the test of this draft.
+        #read_slp_in_drs_file(args.drs, args.filename, args.palette_index,
+                             #args.output, interfac=args.interfac)
 
-        read_slp_in_drs_file(args.drs, args.filename, args.palette_index,
-                             args.output, interfac=args.interfac)
+        terrain_convert(args.drs, args.filename, args.palette_index, args.output, interfac = args.interfac)
```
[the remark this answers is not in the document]
charliergi: That's done and will be added in the next commit ! I was not used to argparser :) .

TheJJ:
Did you have a chance to look at our feedback, @charliergi? :)

charliergi:
Yes of course ! I've been busy the two last weeks. I've began to improve the code. I'll finish it as soin as possible :)

TheJJ:
Oh no worries, just making sure it's nothing on our end that hinders your fun journey through the code :wink:

openage/convert/singlefile.py
charliergi: Ok so, if I've well analyzed the code, the only hard part to optimize is the alpha_composite one.
Do you know if I can do it like this ?
image

heinezen: I don't think we need `alpha_composite`. The general idea is to merge the individual RGBA arrays instead of the images. The resulting array can also be transformed to flat texture with the algorithm. Image operations other than `save()` are probably unnecessary then :)

openage/convert/terrain_convert.pyx
```diff
@@ -8,66 +8,89 @@ Transforms the dimetric tiles converted from SLP files to cartesian format.
...
+    
+    # Method to simply put the tile on the big merge img. Something is wrong on this method because the result is transparent.
+    # TO FIX.
+    @staticmethod
+    def replace(overlayer, overlayed, overlayer_x, overlayer_y, overlayed_x, overlayed_y, slp_width, slp_height):
+        to_ret = overlayed
+        xr = overlayer_x
+        xd = overlayed_x
+        yr = overlayer_y
+        yd = overlayed_y
```
heinezen: This needs more commenting :D For example, what is this function supposed to do, what are `overlayer` and `overlayed` supposed to be. It is not clear from the code.

Also, it would help if you used more descriptive variable and method names for easier understanding.

openage/convert/terrain_convert.pyx
```diff
@@ -8,66 +8,89 @@ Transforms the dimetric tiles converted from SLP files to cartesian format.
...
+        while xd < overlayed.shape[0]:
+            while yd < overlayed.shape[1]:
+                # if the overlayer is not transparent, then we can push it on the overlayed
+                # without the if, we would overwrite non-transparent pixels with transparent ones.
+                if xr < overlayer.shape[0] and yr < overlayer.shape[1] and overlayer[xr,yr][3]>0:
+                    to_ret[xd,yd] = overlayer[xr,yr]
+                yd+=1
+                yr+=1
```
heinezen: `yd` is never reset, so the inner loop is only executed once.

Also, what is the purpose of the `xr < overlayer.shape[0]` and `yr < overlayer.shape[1]` conditions?

openage/convert/singlefile.py
```diff
@@ -297,45 +297,29 @@ def terrain_convert(drs, slp_path, palette_index, output_path, interfac=None):
...
-        if row % 10 ==0 and row > 0:
-            column+=1
-            row=0
-        image[row][column] = slp_image.main_frames[count].get_picture_data(palette)
-        row+=1
-        count+=1
 
     #matrix = determine_rgba_matrix(image, None, 0) ???
-    image_new = Image.fromarray(image)
+    image_new = texture_tomod.image_data.get_data().transpose(1,0,2)
```
heinezen: why transpose the array here?

openage/convert/terrain_convert.pyx
```diff
@@ -8,66 +8,89 @@ Transforms the dimetric tiles converted from SLP files to cartesian format.
...
         """
 
         # Our merge result will be slightly smaller than the SLP image
-        merge_img = Image.new('RGBA', (962, 481), (0, 0, 0, 0))
-
-        # The size of each tile in the SLP
-        cdef int slp_tile_width = 98
-        cdef int slp_tile_height = 49
+        merge_img = np.empty([962,481,4], dtype=np.ndarray)
+        merge_img.fill(0)
```
heinezen: If you want the array to be filled with zeros from the start, you can call `numpy.zeros(..)` instead of `numpy.empty`.

openage/convert/terrain_convert.pyx
```diff
@@ -8,66 +8,89 @@ Transforms the dimetric tiles converted from SLP files to cartesian format.
...
-        res_x, res_y = img.size
+        res_x, res_y = img.shape[0], img.shape[1]
 
         tr_res_x = (int)((0.5) * res_x)
 
-        tr_img = Image.new('RGBA', (tr_res_x, res_y), (0, 0, 0, 0))
-
+        #tr_img = Image.new('RGBA', (tr_res_x, res_y), (0, 0, 0, 0))
+        tr_img = np.empty([tr_res_x,res_y,4], dtype=np.ndarray)
+        tr_img.fill(0)
```
heinezen: `numpy.zeros` instead of `numpy.empty` :)

openage/convert/singlefile.py
```diff
@@ -297,45 +297,29 @@ def terrain_convert(drs, slp_path, palette_index, output_path, interfac=None):
 
     # import here to prevent that the __main__ depends on SLP
     # just by importing this singlefile.py.
-    from .slp import determine_rgba_matrix, SLP, SLPFrame, FrameInfo
+    from .slp import SLP, SLPFrame, FrameInfo
     from PIL import Image
-    from .texture import Texture, TextureImage
+    from .texture import Texture, TextureImage, merge_frames
     from .terrain_convert import Terrain_convert
     import numpy as np
```
heinezen: Few things here:

* imports for `SLPFrame` and `FrameInfo` are unncessary, since they are not referenced
* `Texture` is already imported at the start of the file. no need to reimport it here.

Running `make checkall` in the project root will find these redundancies for you (and also give code style hints).

openage/convert/terrain_convert.pyx
charliergi: Of course, I'll add specs on the next commit, but basically, this function is used to paste a tile on the final RGBA array of the terrain texture. overlayer are tiles (and overlayed is the merge_img), because you overwrite a subarray of the terrain array (merge_img) with a tile.
I use nested because I haven't found something more optimal. I simply paste pixel per pixel.

openage/convert/terrain_convert.pyx
charliergi: Wow I feel stupid now haha, of course I have to reset them! 
Because the size of the merge_img is a bit smaller than the original SLP terrain texture, I had to make sure I wasn't pasting any pixel out of the merge_img. Basically, these conditions are there to avoid out of bound indexes.

openage/convert/singlefile.py
charliergi: Everytime I use the function Image.fromarray(image_array, "RGBA"), the array is transposed (x and y are swapped). If I remember well my computer architecture courses, this is because it depends on the array representation in memory. ([https://github.com/python-pillow/Pillow/issues/2619#issuecomment-315333880](url)) I don't know if I'm right, but we should be able to identify which array representation is used in memory.

openage/convert/terrain_convert.pyx
charliergi: i'm gonna correct that, i'm not used to numpy.

openage/convert/singlefile.py
charliergi: Yep you're right! I was not focused on the tests before making sure my code had at least a good logic. I'll correct that.

openage/convert/terrain_convert.pyx
heinezen: Ah! However, I still wonder what the inner and outer loop accomplish exactly. Wouldn't the current logic operate on the whole image in the worst case (`xd, yd = 0`) instead of just one tile?

heinezen: Ah! However, I still wonder what the inner and outer loop accomplish exactly. Wouldn't the current logic operate on the whole image in the worst case (`xd, yd = 0`) instead of just one tile?

openage/convert/singlefile.py
heinezen: Are unit SLPs transposed too? This bug was fixed some time ago and we've been using `Image.fromarray()` to save the textures without problems for a while.

heinezen:
I would still recommend to have the merge begin even sooner with a custom terrain packer in `merge_frames` of the `texture.py` module. Right now, the array/image operations are the following:

1. All tile arrays are appended/merged into a single array (`texture_tomod.image_data`) by `merge_frames` in `texture.py`
2. You then merge the tiles a second time by extracting the tiles again and overlaying them in another merge array
3. The result is transformed into the "flat" representation of the texture.

Appending the tiles in Step 1 is basically redundant right now since the tiles are extracted again anyway. You should aim to integrate Step 2 into Step 1 by having the tiles be merged directly into the first single array (`texture_tomod.image_data`). This can be done via a new packer specifically for terrains.

charliergi:
I'll continue when I'll have time. Thanks for the feedback :)

openage/convert/terrain_convert.pyx
charliergi: That's what I corrected, thanks to your advices. It's now merging the tiles but the coordinates are wrong. I have to correct them.

heinezen:
Sure, take your time. If it merges then you're already halfway done :)

heinezen:
@charliergi Any assistance needed here? :)

PR #1206, opened 2019-12-18
renderer: use shared_ptrs for renderer resources

TheJJ:
now object ownership is clear and simpler

### DIFF
--- a/libopenage/renderer/CMakeLists.txt
+++ b/libopenage/renderer/CMakeLists.txt
@@ -1,13 +1,13 @@
 add_sources(libopenage
-  color.cpp
-  geometry.cpp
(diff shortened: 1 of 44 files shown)


Vtec234:
I still prefer the `unique_ptr` <-> owner and `object*` <-> borrowed reference model, but if you think this is better then LGTM!

TheJJ:
I did this in order to retain the memory alive in the lambda function. That was way easier by capturing the sharedptr there instead of preserving ownership in any other way, and transferring the ownershipt into the lambda body isn't possible I think.

In the long run i think this is the easier approach, and if we do run into performance problems someday we can still optimize those paths, I hope :)

PR #1224, opened 2020-03-24
Correctly display SLP version string

heinezen:
The SLP version is now treated as a string with ASCII decoding when using verbose output.

I also fixed the wrong types for hotspots brought up by #1222.


TheJJ:
thx!

PR #1234, opened 2020-04-23
[WIP] Updating to Python3.cmake

simonsan:
Early testing for pulling in CMake 3.12 as Minimum dependency and using Python3.cmake instead of our own `FindPython` implementation


simonsan:
Superseded by #1235

Closing

PR #1242, opened 2020-04-29
Change the copyright year of copying.md

duanqn:
We are in the year of 2020 now.


heinezen:
Good catch! Thank you :)

PR #1249, opened 2020-05-02
Workflow for code analysis

simonsan:
Early stage, aims for:

- [ ] giving feedback if bugs are found
- [ ] optimize that patch-formatting
- [x] publish fixes (patches) for found errors based on settings in `.clang-tidy`
- [x] publish fixes (patches) based on settings in `.clang-format`
- [ ] use it just on the recent pushed commit, not over all the code base
  - [ ] include switch, we want to check the code base and fix stuff from time to time 
- [x] [LINK_WHAT_YOU_USE](https://cmake.org/cmake/help/latest/prop_tgt/LINK_WHAT_YOU_USE.html)
- [x] make use of [clazy (Qt related compiler warnings)](https://github.com/KDE/clazy)
- [X] [cpp-lint](https://github.com/cpplint/cpplint) (not needed => Clang-Format)

**Buildsystem-Flags**
Q: Does it make sense to create a `diagnose` module under buildsystem for `make diagnose` or would it be enough to document the following switches and create some CMake file around it or so for exporting fixes (incl. Clang-Tidy that is already existing)
- [ ] [Include-what-you-use](https://cmake.org/cmake/help/v3.12/prop_tgt/LANG_INCLUDE_WHAT_YOU_USE.html)
- [ ] [cppcheck](https://cmake.org/cmake/help/v3.12/prop_tgt/LANG_CPPCHECK.html)
- [ ] [clang-tidy](https://cmake.org/cmake/help/v3.12/prop_tgt/LANG_CLANG_TIDY.html)

Future:
- [ ] use result of #1157 and integrate it into the workflow

ref.: #1196


simonsan:
(Outdated)

PR #1250, opened 2020-05-02
Fix Clang-10 warnings & fix configure script

simonsan:
buildsystem: Fix CXX aliases for different clang-versions

Clang-10 bugfix: error: implicit conversion from 'uint64_t' (aka 'unsigned long') to 'double' changes value from 18446744073709551615 to 18446744073709551616

Clang-10 bugfix: operator '?:' has lower precedence than '|'; '|' will be evaluated first

### DIFF
--- a/configure
+++ b/configure
@@ -129,6 +129,9 @@ def get_compiler(args, parser):
         # map alias -> actual compiler
         aliases = {
             "clang": "clang++",
+            "clang-8": "clang++-8",
+            "clang-9": "clang++-9",
+            "clang-10": "clang++-10",
             "gcc": "g++",
         }
 
--- a/libopenage/log/file_logsink.cpp
+++ b/libopenage/log/file_logsink.cpp
@@ -1,4 +1,4 @@
-// Copyright 2015-2016 the openage authors. See copying.md for legal info.
+// Copyright 2015-2020 the openage authors. See copying.md for legal info.
 
 #include "file_logsink.h"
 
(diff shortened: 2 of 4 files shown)


configure
```diff
@@ -129,6 +129,9 @@ def get_compiler(args, parser):
         # map alias -> actual compiler
         aliases = {
             "clang": "clang++",
+            "clang-8": "clang++-8",
+            "clang-9": "clang++-9",
+            "clang-10": "clang++-10",
```
simonsan, line 134: This should be just a small fix, we should probably implement something that splits `compiler` into `compiler-name` and `compiler-version` => `if it's clang` and then append the version to `clang++`. 

Also it seems strange to me, that CC is not set by the configure script and I would need to use something like this:
`CC=clang-9 ./configure --compiler=clang-9 --mode=release --flags="-Werror" --ccache --download-nyan`

To get a special clang-version to work. Is that intended behaviour or should we also set `CC` within the `configure`-script?

simonsan: > Also it seems strange to me, that CC is not set by the configure script

Or is this intended behaviour? I see **kevin** is also building on arch-clang with **gcc** for `C`? 🤔

TheJJ: openage doesn't need any C compiler, only a C++ one. that's why we only set `CXX` and not `CC`.

And yes, the compiler-name-and-version parser would be the real fix. But, if we implement it, the version should actually be respected, since that way you may test multiple compiler versions easily.

PR #1255, opened 2020-05-04
Updating the simple build script

simonsan:
Started updating this, not sure if we want to keep it. PRing for feedback @TheJJ 

If yes, how do we want to make the `codegen` part? In a loop over a bash array?


TheJJ:
i've never seen anyone actually use this, maybe it's the best to just remove it. we should invest the energy into optimizing the real buildsystem and should transform the simple script into the buildsystem documentation, so one can better understand the steps the buildsystem performs.

PR #1260, opened 2020-05-06
Make ConcurrentQueue support both copy-constructible and move-constructible elements

duanqn:
Thought this might be useful...
Test with `make test` -- 1 issue (my email is not in copying.md)

### DIFF
--- a/libopenage/datastructure/concurrent_queue.h
+++ b/libopenage/datastructure/concurrent_queue.h
@@ -1,4 +1,4 @@
-// Copyright 2015-2017 the openage authors. See copying.md for legal info.
+// Copyright 2015-2020 the openage authors. See copying.md for legal info.
 
 #pragma once
 
@@ -6,10 +6,9 @@
 #include <mutex>
 #include <queue>
 #include <thread>
+#include <type_traits>
 
-namespace openage {
-namespace datastructure {
-
+namespace openage::datastructure {
 /**
  * A threadsafe queue.
  * Wraps the std::queue with a mutex.
@@ -26,7 +25,7 @@ class ConcurrentQueue {
 public:
 	/** Removes all elements from the queue. */
 	void clear() {
-		std::lock_guard<mutex_t> lock{this->mutex};
+		std::scoped_lock lock{this->mutex};
 		while (!this->queue.empty()) {
 			this->queue.pop();
 		}
@@ -34,7 +33,7 @@ public:
 
 	/** Returns whether the queue is empty. */
 	bool empty() {
-		std::lock_guard<mutex_t> lock{this->mutex};
+		std::scoped_lock lock{this->mutex};
 		return this->queue.empty();
 	}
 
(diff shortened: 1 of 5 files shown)


libopenage/datastructure/tests.cpp
```diff
@@ -192,4 +191,57 @@ void constexpr_map() {
 	TESTEQUALS(cmap.get(42), 9001);
 }
 
-}}} // openage::datastructure::tests
+std::mutex CopyMove::s_mutex = std::mutex();
+int CopyMove::s_accumulatedConstructionCounter = 0;
```
TheJJ, line 195: I think you can define the classes in the `cpp` file, since they're used nowhere else except this test.

libopenage/datastructure/concurrent_queue.h
```diff
@@ -47,25 +46,46 @@ class ConcurrentQueue {
...
+	T& pop([[maybe_unused]] typename std::enable_if_t<!std::is_move_constructible_v<U> and std::is_copy_constructible_v<U>>* t = NULL) {
+		static_assert(sizeof...(None) == 0, "User-specified template arguments are prohibited.");
 		std::unique_lock<mutex_t> lock{this->mutex};
-		while (this->queue.empty()) {
-			this->elements_available.wait(lock);
-		}
-		auto &item = this->queue.front();
+		auto &item = front();
 		this->queue.pop();
 		return item;
```
TheJJ, line 56: this is a memory error, we have to return a `T`, and not a `T&`. so `auto item = this->front();` and `return item;`

libopenage/datastructure/concurrent_queue.h
```diff
@@ -47,25 +46,46 @@ class ConcurrentQueue {
...
 
-	/** Appends the given item to the queue. */
-	void push(const T &item) {
+	/** Moves the front item in the queue and removes it from the queue. */
+	template<typename... None, typename U = T>
+	T &&pop([[maybe_unused]] typename std::enable_if_t<std::is_move_constructible_v<U>>* t = NULL) {
+		static_assert(sizeof...(None) == 0, "User-specified template arguments are prohibited.");
+		std::unique_lock<mutex_t> lock{this->mutex};
+		T&& ret = std::move(front());
+		this->queue.pop();
```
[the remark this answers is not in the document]
TheJJ: Oh, [`std::queue::front()`](https://en.cppreference.com/w/cpp/container/queue/front) only returns references, so we'll always have to do a copy?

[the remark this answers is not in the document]
duanqn: I applied a fix. Strangely my tests didn't catch anything.

libopenage/datastructure/concurrent_queue.h
```diff
@@ -47,25 +46,48 @@ class ConcurrentQueue {
...
+	/** Copies the front item in the queue and removes it from the queue. */
+	template<typename... None, typename U = T>
+	T pop([[maybe_unused]] typename std::enable_if_t<!std::is_move_constructible_v<U> and std::is_copy_constructible_v<U>>* t = NULL) {
+		static_assert(sizeof...(None) == 0, "User-specified template arguments are prohibited.");
+		std::scoped_lock lock{this->mutex};
+		T ret = this->front();
+		this->queue.pop();
+
+		// Explicitly call the copy constructor since T cannot be move-constructed
+		return T(ret);
```
TheJJ, line 58: if you write `return ret`, c++ will select the best possible return way automatically (move or copy), we don't need to explicitly copy.

libopenage/datastructure/concurrent_queue.h
```diff
@@ -47,25 +46,48 @@ class ConcurrentQueue {
...
+	}
+
+	/** Moves the front item in the queue and removes it from the queue. */
+	template<typename... None, typename U = T>
+	T pop([[maybe_unused]] typename std::enable_if_t<std::is_move_constructible_v<U>>* t = NULL) {
```
TheJJ, line 63: pls replace `NULL` with `nullptr` :)

libopenage/datastructure/tests.cpp
```diff
@@ -192,4 +191,146 @@ void constexpr_map() {
 	TESTEQUALS(cmap.get(42), 9001);
 }
 
-}}} // openage::datastructure::tests
+/**
+ * A simple class that can be move-constructed but not copy-constructed
+ */
+class MoveOnly{
+	private:
+	int m_data;
```
TheJJ, line 199: we place the access labels at the same indent level as the `class` statement.
our member names are (at least until now) never named `m_`, since we always access them with `this->`.
for consistency reasons it would be nice if you removed the prefix (and access the data in `get()` with `this->data` then).

libopenage/datastructure/tests.cpp
```diff
@@ -192,4 +191,146 @@ void constexpr_map() {
...
+
+	int get() const {
+		return this->m_data;
+	}
+
+	static int getAccumulatedConstructionCounter() {
+		return CopyMove::s_accumulatedConstructionCounter;
+	}
+
+	static void resetAccumulatedConstructionCounter() {
```
TheJJ, line 278: we use `snake_case` for all member methods

libopenage/datastructure/tests.cpp
duanqn: OK what about static variables then? Should I also remove the `s_` ?

libopenage/datastructure/tests.cpp
duanqn: Fixed. Thanks.

libopenage/datastructure/concurrent_queue.h
duanqn: `return ret` does not work on GCC 7.5.0 . Will try if it works for a later version.

libopenage/datastructure/concurrent_queue.h
duanqn: Fixed. Thanks.

libopenage/datastructure/tests.cpp
duanqn: Fixed. Thanks.

libopenage/datastructure/concurrent_queue.h
duanqn: I tried to build with GCC 10.1.0, `return ret;` works here but there are several compile errors in other files.

libopenage/datastructure/tests.cpp
TheJJ: I'd say yes, but we don't have a strict rule there yet.

libopenage/datastructure/concurrent_queue.h
TheJJ: I tried this in godbolt now: https://godbolt.org/z/7_42Uy and the C++-Standard makes the difference. with `std=c++17` it works accross all versions, apparently (the copy-selection when move was deleted). `std=c++11` and `c++14` fail with use of deleted function.

since the buildsystem uses 17, i wonder why it fails for you :)

duanqn: Which version did you try on? On godbolt it also fails with GCC 7.5.0

TheJJ: Only with gcc >7... and with =7 it fails. Oh dear. I think we can leave the explicit copy then, but with a `TODO` to change it once we raise our minimum compiler version.

heinezen:
Thank you :)

Issue #1263, opened 2020-05-07
Proposal: Version bump of dependencies


simonsan:
The possible new version are the versions coming with **Ubuntu 20.04 LTS**. I marked the one's where I think we could bump the version and moved them to the very top.

Really happy I personally would be about another version bump of **CMake**, as that would enable us to work further on the Python3 stuff and actually use new methods for library linking (e.g. https://cmake.org/cmake/help/v3.13/command/add_link_options.html ) etc. The last bump I checked for Ubuntu 18.04 LTS dependencies.

If there are some advantages of bumping the **compiler versions** and **Cython** there @TheJJ has problably the most knowledge. 

For Qt I feel we should also bump, as I'm building currently on `5.14.2` in the CI, nevertheless, we should stick with the LTS version that is 5.12.x, which is also the last more open LTS I think ([5.15.x LTS will be only for commercial customers afaik](https://www.qt.io/blog/qt-offering-changes-2020)). For Qt we should also update QtQuickcontrols to Version 2 as Version one is throwing deprecated notes at us.

Probably also NumPy and Cython could get a version bump, I'm not very familiar with them both but I was seeing the deprecated warnings of an old NumPy-API lately - not sure, if that would change with the version bump though.

Dependency | Current Version | Possible New Version 
------------ | -------------| ---------------- 
**cmake** | >=3.12 | **3.16.3** :heavy_check_mark:
**Qt5** | >=5.9 | **5.12.8**
**QtQuickControls** | 1 | **2**
**python** | >=3.6 | **3.8.2**
**numpy** | / | **1.16.5**
**cython** | >=0.25 | **0.29.14**
gcc | >=7 | 9.3.0
clang | >=5 | 10.0
python imaging library (PIL) | / | 6.2.1
opengl | >=3.3 | (~3.14)
libepoxy | / | 1.5.4
libpng16 | / | 1.6.37
fonts-dejavu | 2.37 | 2.37
eigen | >=3 | 3.3.7
freetype2 | / | 2.10.1
fontconfig | / | 2.13.1
harfbuzz | >=1.0.0 | 2.6.4
ncurses | / | 6.2
jinja2 | / | 2.10.1
sdl2 | / | 2.0.10
sdl2_image | / | 2.0.5
opusfile | / | 0.9
opus | / | 1.3.1
ogg | / | 1.3.4
Vulkan | / | 1.2.131.2
[...]

PR #1264, opened 2020-05-07
Bump CMake version to 3.16, buildsystem update to Python search

### LINKED ISSUE #1246: Optional possibility to hardcode a Python directory
With the update to **Python3.cmake** [`-DPYTHON_DIR`](https://github.com/SFTtech/openage/blob/master/buildsystem/modules/FindPython.cmake#L23) is used as a `HINT` to search for a suitable Python3 version.

Sometimes just hinting towards a location is not enough and you would like to **hardcode** that and let the process fail miserably if nothing is found.

To keep the **hinting** behaviour of `-DPYTHON_DIR=<path>` we should add `-DFORCE_PYTHON_DIR=TRUE` that will switch from `hinting` => `enforcing` the search in `PYTHON_DIR` and nowhere else.

The whole processing in packaged CMake runs in an extra file from `Python3.cmake` kept in [`FindPython/Support.cmake`](https://gitlab.kitware.com/cmake/cmake/-/blob/v3.12.4/Modules/FindPython/Support.cmake) that we would need to adopt to `buildsystem/modules/FindPython/Support.cmake` to let it be used by CMake and overwrite default behaviour.

Be aware, that we are limited to the 3.12.x version and functionality of CMake as this is our minimum version requirement, there are newer version of the files and It would make sense to test how far we can update that with the usage of our CMake 3.12.x minimum requirement. In the newer version there is a [`find virtual environment`](https://gitlab.kitware.com/cmake/cmake/-/blob/v3.15.4/Modules/FindPython3.cmake#L161) handling included that could be interesting for example.

>``Python3_FIND_VIRTUALENV``
> This variable defines the handling of virtual environments. It is meaningfull
  only when a virtual environment is active (i.e. the ``activate`` script has
  been evaluated). In this case, it takes precedence over
  ``Python3_FIND_REGISTRY`` and ``CMAKE_FIND_FRAMEWORK`` variables.

simonsan:
buildsystem: Integrate with new Python3.cmake functions (NumPy)

fixes #1246
ref.: #1263


CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 # Copyright 2013-2020 the openage authors. See copying.md for legal info.
 
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.16)
```
TheJJ, line 3: can you annotate why we need that version now ("finding numpy with the findpython3 module" or something)

buildsystem/HandlePythonOptions.cmake
```diff
@@ -1,16 +1,27 @@
...
 # the Python version number requirement is in modules/FindPython_test.cpp
 find_package(Python ${PYTHON_MIN_VERSION} REQUIRED)
+
+# NumPy check
+message(CHECK_START "Looking for NumPy")
+if(NUMPY_FOUND)
+	message(CHECK_PASS "NumPy - found!")
+else()
+	message(FATAL_ERROR "NumPy - not found!")
+endif()
```
TheJJ, line 14: these messages are at the wrong place i think, they should be within our findpython module. but even there, i think that is already printed during the `findpackage` above? if not, i'd remove the `looking for numpy` message, since we're not really looking for numpy there. if necesary (i.e. if the cmake-findpython-module doesn't abort on its own) we can keep the fatal error message printing in our `findpython.cmake`.

TheJJ:
now, this depends on the kevin container update, which will happen soon™.

simonsan:
Waiting for #1268

TheJJ:
Can you rebase pls?

Issue #1244, opened 2020-04-30
Increase verbosity of buildsystem documentation


simonsan:
As I'm finding my way around the buildsystem in the last weeks I feel that we could add some more information to the documentation. Some first thoughts (more added later):

I think we should create an own folder `doc/buildsystem` and explain the different parts in detail. For example, how the `find_package`-System works, where to find the modules. How to add a new one or how to refactor a module. Maybe even some basic `call/caller-graphs` to find the way more easy or to see where optimisation could be needed. I think that could be a long-term profit for everyone. For new people coming in to understand it and for people using it every day to find their way around faster.

In addition to that, as we will probably need it more often, we should document processes like how to create a new `Makefile`-target and how to deal within CMake with these. What to do, if you want to add a new feature within `openage` and/or `libopenage`. How is the buildsystem involved with the c++/python interface etc.

Our documentation for the buildsystem is here right now:
https://github.com/SFTtech/openage/blob/master/doc/buildsystem.md

simonsan:
`cmake .. --graphviz=out.dot`
`dot -Tpng out.dot -o out.png`

Looks currently like this :-P 

![out](https://user-images.githubusercontent.com/14062932/80929032-2cdea680-8da9-11ea-9d74-f49118fcb17c.png)

simonsan:
Not the buildsystem, but [c++ dependencies visualized](https://github.com/tomtom-international/cpp-dependencies) as well:

`cpp-dependencies.exe -dir libopenage --recursive --graph dependencies.dot`
`dot -Tpng dependencies.dot > dependencies.png`

![dependencies](https://user-images.githubusercontent.com/14062932/81338555-20aa6000-90ad-11ea-8927-ab2a3a24c15c.png)

PR #1266, opened 2020-05-08
Buildsystem update and documentation

simonsan:
Starting with creating/reworking the buildsystem documentation (#1244) as I'm hanging around quite much in these catacombs the last days. :-) Will also rework some processes in there and trying to update modules with the corresponding official ones if any.


simonsan:
Waiting for #1268

simonsan:
(Outdated)

PR #1283, opened 2020-08-22
Remove pillow from dependencies

heinezen:
pillow is not required anymore. We now use libpng for conversion :)


TheJJ (approved):

heinezen:
@TheJJ It also needs to be removed from the Github macOS workflow, but it wouldn't let me do that because I have no "workflow scope" (?) or something like that. so jj, pls fix :)

PR #1291, opened 2020-08-24
Read number of SWGB blend modes from .dat file

heinezen:
The blendomatic reader now uses an entry from the .dat file for the number of blend modes it has to read. Previously it used the information in the header which defined an incorrect size. The .dat file entry is unique to SWGB and its purpose was unknown before. According to @withmorten, the blend mode count that is read is probably hardcoded and not read from the .dat, so we cannot test whether the entry is the actual blend mode count. But if it works... ¯\\\_(ツ)_/¯


TheJJ (approved):
k thx

Issue #1297, opened 2020-08-25
Use bicubic filtering to reverse dimetric projection


MayeulC:
Related: #720, #985.

This is the current implementation:

https://github.com/SFTtech/openage/blob/439d224aac94c4d6114e8d0405abf7fd17db924d/openage/convert/processor/export/texture_merge.py#L165-L173

As [written in the chat](https://matrix.to/#/!iPmyRMGnDgSMPGqgaX:matrix.org/$pV5WoNsiNvmLPufTV41LN3xMHgbkwO63Va0tKa2az7c?via=matrix.org&via=sft.lol&via=mayeul.net), I see several issues with this implementation:
- homemade, inefficient matrix multiplication algorithm (while the images are quite small, I'd still leave that to [optimized algorithms](https://en.wikipedia.org/wiki/Matrix_multiplication#Computational_complexity)
- No antialiasing. The code takes values from nearest neighbours instead of interpolating them depending on their proximity (bicubic, bilinear, etc)
- A permutation index is reconstructed from scratch each time the function is called.

There are several libraries that can implement this in a more efficient way, I chose pillow based on the fact that it's used elsewhere. It doesn't have the greatest choice of antialiasing algorithms nor the most efficient implementation, but that's probably enough. Example code based on the [same stackoverflow question](https://gamedev.stackexchange.com/questions/16746/what-is-the-name-of-perspective-of-age-of-empires-ii) and [this tutorial](https://stackabuse.com/affine-image-transformations-in-python-with-numpy-pillow-and-opencv/), though I have ignored a few things, like the matrix inversion.

I suggest something similar to the following code:

```python
import numpy as np
from PIL import Image
import matplotlib.pyplot as plt

file = "45c0011e-e279-11e6-9ed5-30f00391cc87.png"

img = Image.open(file)

def rot_mat(angle_rad):
    return np.array([[np.cos(angle_rad),-np.sin(angle_rad), 0],
                     [np.sin(angle_rad),np.cos(angle_rad),  0],
                     [                0,               0,   1]])

def aoe2_transf(scale=1.118):
    return np.array([[  1,  -1,     0],
                     [0.5, 0.5, scale],
                     [  0,   0,     0]])

# unsure, intuitively deduced from the above
def aoe2_inv_transf(scale=1.118):
    return np.array([[ 0.5, 1,       0],
                     [-0.5, 1, 1/scale],
                     [   0, 0,       0]])


def trans_mat(x,y):
[...]

PR #1298, opened 2020-08-25
Revert "Remove pillow from dependencies"

TheJJ:
Reverts SFTtech/openage#1283

hehe, we still have `PIL` code, which we either have to remove, or keep pillow as dependency. I think having `PIL.Image` is quite handy.


MayeulC:
could be useful for #1297 :upside_down_face:

PR #1324, opened 2020-10-15
Add manifest file creator for a modpack

### LINKED ISSUE #1287: Create manifest file for modpacks
**Required skills**: Python

**Difficulty**: Easy

Modpacks should contain a manifest file which lists all files in the modpack and also stores the hashes of these files. Manifests can be used for an integrity check when the modpack loads. Additionally, manifest files can be signed cryptographically by us which will be used as an indication that a modpack is *safe* (= it doesn't contain malware).

The structure of the file should be very simple. One line (`hash $name`) defines what hash to use, all other lines contain a relative filepath and the hash value of that file, e.g.:

```
# openage autogenerated modpack integrity check
hash SHA3-256
bc821027ed1b666d8653760 relative/path/to/file
ea4e89ce46f0f8c4dad2344 another/file
...
```

SHA3-256 or SHA3-512 should suffice as hash algorithms.

It's important that the order of entries in the file is the same every time and for every platform, otherwise it will be difficult to distribute signatures for locally converted. Using something like breadth-first search should ensure this property.

drs-11:
Tries to fix #1287.

Creates a manifest file for the exported modpack. 
Sorry couldn't carry out tests because my builds were failing because of some local environment conflicts.


drs-11:
Okay looks like everything failed

heinezen:
@drs-11 You only failed the sanity checks (aka code styling) :D Click on `Details` to see whats wrong

I'm going to review your code now.

openage/convert/processor/export/modpack_manifest.py
```diff
@@ -0,0 +1,109 @@
...
+        info_table = {"info": {}}
+        info_table["info"].update({"hash": self.HASHING_FUNCTION})
+
+        output_dict.update(info_table)
+
+        hash_values_table = {'hash-values': {}}
+        for file in self._bfs_directory(exportdir):
+            hash_val = self._get_file_hash(file)
+            relative_path = os.path.relpath(str(file), str(exportdir))
+            hash_values_table['hash-values'].update({hash_val: relative_path})
```
heinezen, line 44: We would like to separate the calculation of the hashes from saving the hashes. This means that the `ManifestFile` does not do the BFS search and hash calculation, but is instead initialized with an empty list that stores the hash values and a method that adds a tuple `(hash, filepath)` to the list. The function doing the BFS search then just appends a hash value using the method.

Instances of `DataDefinition` should generally only get the data they dump and not be involved in the generation of said data. The reason for this is that, as a result, data can be generated and added from multiple sources. `ManifestFile` should be a storage for data that is exported. It does not need to know where the data is from or how it was generated.

openage/convert/processor/export/modpack_manifest.py
```diff
@@ -0,0 +1,109 @@
...
+                             type(exportdir))
+
+        output_dir = exportdir.joinpath(self.targetdir)
+        output_content = self.dump(exportdir)
+
+        # generate human-readable file
+        with output_dir[self.filename].open('wb') as outfile:
+            outfile.write(output_content.encode('utf-8'))
+
+    def _bfs_directory(self, root):
```
heinezen, line 72: This function should be in a processor and not be part of the `ManifestFile` class (see my comment above).

openage/convert/processor/export/modpack_manifest.py
```diff
@@ -0,0 +1,109 @@
...
+            for directory in dirs:
+                for item in directory.iterdir():
+                    item_path = directory.joinpath(item)
+                    if item_path.is_dir():
+                        next_level.append(item_path)
+                    else:
+                        yield item_path
+            dirs = next_level
+
+    def _get_file_hash(self, file_path):
```
heinezen, line 92: same for this function

heinezen (changes requested):

heinezen:
The code looks good so far :)

I have a few requests regarding our code structure and the separation of data generation and data storage.

openage/convert/processor/export/modpack_manifest.py
drs-11: Ok yeah understood. So `_bfs_directory` and `_get_file_hash` should be in a separate processor and `ManifestFile` should use a method which actually uses both of them to generate a `(hash, filepath)` tuple. I get that right?

But then after separating the functions should the `ManifestFile` still be in the processor subdirectory? As I made it an instance of `DataDefinition` should it be now placed at `entity_object` subdirectory?

heinezen: > `ManifestFile` should use a method which actually uses both of them to generate a (hash, filepath) tuple. I get that right?

`ManifestFile` should have a method that takes a `(hash, filepath)` and puts them into a list. The processor should generate them using `_bfs_directory` and `_get_file_hash` and pass them to `ManifestFile` by using the method.

> But then after separating the functions should the ManifestFile still be in the processor subdirectory?

`ManifestFile` goes into `openage/convert/entity_object/export/formats` where all the other format classes are. The `_bfs_directory` and `_get_file_hash` functions should stay in this file.

openage/convert/processor/export/generate_hash.py
```diff
@@ -0,0 +1,82 @@
...
+                break
+            hashf.update(data)
+
+    return hashf.hexdigest()
+
+
+def generate_hashes(modpack, exportdir):
+    """
+    Generate hashes for all the items in a
+    given modpack and adds to the manifest
```
heinezen, line 65: [...] adds **them** [...]

openage/convert/processor/export/generate_hash.py
```diff
@@ -0,0 +1,82 @@
...
+    instance.
+
+    :param modpack: The target modpack.
+    :type modpack: ..dataformats.modpack.Modpack
+    param exportdir: Directory wheere modpacks are stored.
```
heinezen, line 70: Missing `:` before `param`

wheere --> where

openage/convert/processor/export/generate_hash.py
```diff
@@ -0,0 +1,82 @@
+# Copyright 2020-2020 the openage authors. See copying.md for legal info.
```
heinezen, line 1: Small suggestion: Rename this file to `generate_manifest_hashes.py`, so it's more clear what we are doing here :)

openage/convert/processor/export/generate_hash.py
```diff
@@ -0,0 +1,82 @@
...
+Provides functions for traversing a directory and
+generating hash values for all the items inside.
+"""
+
+import hashlib
+import os
+
+
+HASHING_FUNCTION = 'sha3_256'
+BUF_SIZE = 32768
```
heinezen, line 12: For less hardcoding, these should be moved into `generate_hashes(..)` as optional parameters. Then we can change the hashing function and buffer size dynamically if we need to.

I'm thinking of something along the lines of 

```
generate_hashes(modpack, exportdir, hashfunc='sha3_256', bufsize=32768):
```

openage/convert/processor/export/generate_hash.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def bfs_directory(root):
+    """
+    Traverse the given directory in breadth-first way.
```
heinezen, line 17: in breadth-first way --> with breadth-first search

openage/convert/processor/export/generate_hash.py
```diff
@@ -0,0 +1,82 @@
...
+    :param modpack: The target modpack.
+    :type modpack: ..dataformats.modpack.Modpack
+    param exportdir: Directory wheere modpacks are stored.
+    :type exportdir: ...util.fslike.path.Path
+    """
+
+    # set the hashing algorithm in the manifest instance
+    modpack.manifest.set_hashing_func(HASHING_FUNCTION)
+
+    # traverse the directory in breadth-first way and
```
heinezen, line 77: in breadth-first way --> with breadth-first search

heinezen (changes requested):
I've made one last suggestion and found a few typos. We can merge once they are handled :)

heinezen:
@drs-11 Remove the `[WIP]` once you think you're ready

openage/convert/processor/export/generate_manifest_hashes.py
```diff
@@ -0,0 +1,86 @@
...
+
+    # initiate from the root directory
+    dirs = [root]
+
+    while dirs:
+        next_level = []
+
+        for directory in dirs:
+            for item in directory.iterdir():
+                item_path = directory.joinpath(item)
```
heinezen, line 27: This line causes a runtime error. Creating `item_path` fails because `item` is already a `Path` object.

You do not need `item_path` or the `.joinpath()` command here. Just remove this line and change all references of `item_path` to `item`. Then it runs fine.

heinezen (changes requested):
I've encountered one error during testing that needs to be fixed

heinezen (approved):
LGTM :tada: 

Thanks for contributing :)

drs-11:
Had fun! :D

heinezen:
@drs-11 You can do another quick one if you want :D

drs-11:
@heinezen Absolutely! I was checking out other ones. Will let you know in the issue thread.

PR #1326, opened 2020-10-24
make --devmode and --no-devmode mutually exclusive through argparse

mmtj:
As newcomer to this project I don't fully understand the need for having two separate devmode overrides, so I didn't touch that logic.

Nevertheless, it seems to me that manually enforcing mutual exclusion of `--devmode` and `--no-devmode` might be error prone and could be handled entirely within argparse.

Downside of this change is that we lose custom error string here, but in this case it IMO shouldn't matter that much as these two options are opposite of each other.


heinezen (approved):
I didn't even know about this feature :D

LGTM :)

TheJJ:
The reason both flags exist is simple: the devmode is autodetected by trying to import a non-installed python module.
Devmode is therefore "am I running in the development source tree". Devmode is used for changing the search directories for assets, for example.

One may want to force-enable (e.g. if running the system-installed version) or force-disable (e.g. if running in the source tree) the devmode.

Issue #1290, opened 2020-08-23
Set PNG compression optimization level via CLI


heinezen:
**Required skills**: Python, maybe Cython

**Difficulty**: Easy

Our converter PNG export module supports 3 built-in compression levels

* *none*
* *default* libpng compression
* *optimized* compression

However, this functionality is currently unused as the converter is hardcoded to always use the *default* compression setting. We should implement a way to set the desired compression level with a CLI flag. The CLI flag should be added to the main `convert` module creating the modpacks as well as the `convert-file` submodule that can convert standalone files.

**Further reading**:

* [Texture export code](https://github.com/SFTtech/openage/blob/3863e5e76e1c55b19ee49c65b14f97d318870423/openage/convert/entity_object/export/texture.py#L124)
* [Converter architecture](https://github.com/SFTtech/openage/blob/master/doc/code/converter/architecture_overview.md)
* [Converter workflow](https://github.com/SFTtech/openage/blob/master/doc/code/converter/workflow.md)

dev-patel2104:
@heinezen can i work on these one ?

TheJJ:
No need to ask for permission, just do it and submit a pull request :)

heinezen:
@dev-patel2104 You can :)

The most relevant code snippets are here:

https://github.com/SFTtech/openage/blob/dc3980146f814ed42e6aa1c92bcba18621386c8b/openage/convert/service/export/png/png_create.pyx#L57



and here:

https://github.com/SFTtech/openage/blob/dc3980146f814ed42e6aa1c92bcba18621386c8b/openage/convert/entity_object/export/texture.py#L124

janisozaur:
I would recommend you migrate to webp. I tested this recently on a few projects and there is pretty much 20% of bytes saved over lossless, fully optimized PNGs across the board.

heinezen:
@janisozaur Interesting. Do you have more info about the decoding speed?

janisozaur:
No, sorry. I haven't checked.

janisozaur:
According to webp devs themselves, it's pretty much better across the board: https://developers.google.com/speed/webp/docs/webp_lossless_alpha_study (see table 3 for decoding)

heinezen:
@janisozaur If it has better decoding speed that's a big plus :) Even better would be a GPU implementation, but I do not see anything about that yet.

I might implement experimental support for it in the converter. Should not be too hard to do.

heinezen:
Closed via #1327

Issue #1311, opened 2020-09-07
Use processor for saving media export requests


heinezen:
**Required skills**: Python

**Difficulty**: Hard

**Motivation**

During the data conversion, the converter generates so-called *export requests* (`MediaExportRequest`) for media files. Currently, export requests store the filepaths for the source file and the to-be-generated target file as well as a `save()` function. `save()` is used to read the source file, convert it to a media format readable by openage and write the result to a file. For every media type, there is a subclass of `MediaExportRequest` with its own `save()` implementation.

There are problems with handling `save()` in the `MediaExportRequest` implementation: There can be a number of intermediary conversion steps between reading the source file and writing the target file, e.g. texture packing for graphics files. In the worst case, this can lead to a convoluted chain of operations with a workflow that is very hard to understand.  For example, it can result in `MediaExportRequest`'s `save()` function, calling the `save()` function of an intermediary format, which also calls `save()` on a subformat... We also run into the risk of creating circular dependencies and spaghetti code when extending the implementation.

**Task**

To adress these issues, the `save()` functionality should be implemented in a [dedicated processor](https://github.com/SFTtech/openage/blob/master/doc/code/converter/architecture_overview.md#processor) rather than in `MediaExportRequest` or any intermediary subformat. Instead of a convoluted chain, the processor function should have a linear workflow. All intermediary conversion steps should be initiated by the processor function.

The workflow should look like this:

1. Acquire source file path from `MediaExportRequest`
2. Read source file (using a [service](https://github.com/SFTtech/openage/blob/master/doc/code/converter/architecture_overview.md#service))
3. Handle any intermediary steps in linear order, e.g. texture packing or PNG creation
4. Write the end result to file

`MediaExportRequest` would be reduced to a storage format for information used by the processor.

**Further reading**:

* [MediaExportRequest code](https://github.com/SFTtech/openage/blob/master/openage/convert/entity_object/export/media_export_request.py)
* [Converter architecture](https://github.com/SFTtech/openage/blob/master/doc/code/converter/architecture_overview.md)
* [Converter workflow](https://github.com/SFTtech/openage/blob/master/doc/code/converter/workflow.md)

PR #1327, opened 2020-10-25
set PNG compression optimization level via CLI

mmtj:
This should implement #1290

Once you are satisfied with the code, feel free to squash fixups or let me know and I'll clean it up before merging.

### DIFF
--- a/openage/convert/entity_object/export/media_export_request.py
+++ b/openage/convert/entity_object/export/media_export_request.py
@@ -105,7 +105,7 @@ class GraphicsMediaExportRequest(MediaExportRequest):
     def get_type(self):
         return MediaType.GRAPHICS
 
-    def save(self, sourcedir, exportdir, palettes, *args, **kwargs):
+    def save(self, sourcedir, exportdir, palettes, compression_level, *args, **kwargs):
         source_file = sourcedir[self.get_type().value, self.source_filename]
 
         try:
@@ -132,7 +132,11 @@ class GraphicsMediaExportRequest(MediaExportRequest):
             image = SMX(media_file.read())
 
         texture = Texture(image, palettes)
-        metadata = texture.save(exportdir.joinpath(self.targetdir), self.target_filename)
+        metadata = texture.save(
+            exportdir.joinpath(self.targetdir),
+            self.target_filename,
+            compression_level
+        )
         metadata = {self.target_filename: metadata}
 
         self.set_changed()
@@ -148,7 +152,9 @@ class TerrainMediaExportRequest(MediaExportRequest):
     def get_type(self):
         return MediaType.TERRAIN
 
-    def save(self, sourcedir, exportdir, palettes, game_version, *args, **kwargs):
+    # pylint: disable=too-many-arguments
+    def save(self, sourcedir, exportdir, palettes, game_version, compression_level,
+             *args, **kwargs):
         source_file = sourcedir[self.get_type().value, self.source_filename]
         media_file = source_file.open("rb")
 
@@ -167,6 +173,7 @@ class TerrainMediaExportRequest(MediaExportRequest):
         else:
             texture = Texture(image, palettes)
         texture.save(exportdir.joinpath(self.targetdir), self.target_filename)
+        texture.save(exportdir.joinpath(self.targetdir), self.target_filename, compression_level)
 
 
 class SoundMediaExportRequest(MediaExportRequest):
--- a/openage/convert/entity_object/export/texture.py
+++ b/openage/convert/entity_object/export/texture.py
@@ -135,6 +135,15 @@ class Texture(genie_structure.GenieStructure):
                                       - 3 = maximum possible compression; slowdown is 256x
         :type compression_level: int
         """
+        from ...service.export.png import png_create
+
+        COMPRESSION_LEVELS = {
+            0: png_create.CompressionMethod.COMPR_NONE,
+            1: png_create.CompressionMethod.COMPR_DEFAULT,
+            2: png_create.CompressionMethod.COMPR_GREEDY,
+            3: png_create.CompressionMethod.COMPR_AGGRESSIVE,
+        }
+
         if not isinstance(targetdir, Path):
             raise ValueError("util.fslike Path expected as targetdir")
         if not isinstance(filename, str):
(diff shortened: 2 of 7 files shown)


openage/convert/entity_object/export/texture.py
```diff
@@ -135,6 +135,15 @@ def save(self, targetdir, filename, compression_level=1):
...
         :type compression_level: int
         """
+        from ...service.export.png import png_create
+
+        COMPRESSION_LEVELS = {
+            0: png_create.CompressionMethod.COMPR_NONE,
+            1: png_create.CompressionMethod.COMPR_DEFAULT,
+            2: png_create.CompressionMethod.COMPR_GREEDY,
+            3: png_create.CompressionMethod.COMPR_AGGRESSIVE,
+        }
```
mmtj, line 145: I'd rather see this at class level, but it would require to make top-level import of png_create.

I'm not that familiar with codebase yet, so I don't know if it is desired or not.

heinezen: We do not want Cython modules to be imported at top level. That can mess with the builds.

heinezen (changes requested):
This also needs integration for the modpack media export starting in

https://github.com/SFTtech/openage/blob/master/openage/convert/processor/export/modpack_exporter.py#L57

This is where the bulk of graphics files are created.

openage/convert/entity_object/export/texture.py
mmtj: Alright then, thanks for explanation.

mmtj:
> This also needs integration for the modpack media export starting in
> 
> https://github.com/SFTtech/openage/blob/master/openage/convert/processor/export/modpack_exporter.py#L57
> 
> This is where the bulk of graphics files are created :)

Thanks, I updated modpack_exporter.

Is it just media files or does it apply to metadata and manifest as well?

heinezen:
@mmtj It just needs to be forwarded to media file exports. Metadata and manifest files are uncompressed plaintext files.

openage/convert/processor/export/modpack_exporter.py
```diff
@@ -65,7 +65,7 @@ def export(modpack, args):
                 kwargs["palettes"] = args.palettes
 
             for request in cur_export_requests:
-                request.save(sourcedir, modpack_dir, **kwargs)
+                request.save(sourcedir, modpack_dir, args.compression_level, **kwargs)
```
heinezen, line 68: This needs to be in `kwargs` rather than in the expicit function call. The reason for this is that `SoundExportRequest` does not need the PNG compression level.

And the media export request's `save()` function must propagate the compression level to the texture's `save()` function, of course.

mmtj: Thanks for the pointers, I was wondering what is the use of kwargs here.

So compression level should apply only on `MediaType.GRAPHICS` and `MediaType.TERRAIN`?

heinezen: > So compression level should apply only on MediaType.GRAPHICS and MediaType.TERRAIN?

Correct

> Thanks for the pointers, I was wondering what is the use of kwargs here.

We are using `kwargs` here because not all media export requests require the same settings. It's a bit of a catch-all parameter. Once https://github.com/SFTtech/openage/issues/1311 is implemented, the could will hopefully be more understandable in that regard :)

mmtj: > We are using kwargs here because not all media export requests require the same settings. It's a bit of a catch-all parameter. Once #1311 is implemented, the could will hopefully be more understandable in that regard :)

With this explanation it is indeed clear, though it is confusing at first :)

> And the media export request's save() function must propagate the compression level to the texture's save() function, of course.

I'm a little lost at where to find particular export request `save()` function.

I guess it should be some of `graphic_export` used in https://github.com/SFTtech/openage/blob/master/openage/convert/processor/conversion/aoc/modpack_subprocessor.py#L124

But at that point I'm not sure where are these `*_export` defined and where should I look.

heinezen: They can all be found here: https://github.com/SFTtech/openage/blob/master/openage/convert/entity_object/export/media_export_request.py

mmtj: Thanks.

Just small question. why do we pass game_version into MediaType.GRAPHICS, if is not used here?

openage/convert/entity_object/export/media_export_request.py
```diff
@@ -149,7 +153,7 @@ class TerrainMediaExportRequest(MediaExportRequest):
     def get_type(self):
         return MediaType.TERRAIN
 
-    def save(self, sourcedir, exportdir, palettes, game_version, *args, **kwargs):
+    def save(self, sourcedir, exportdir, palettes, game_version, compression_level, *args, **kwargs):
```
mmtj, line 156: Pylint will complain twice on this line (line too long, too many arguments 6/5).

I didn't find anything in code style that would address long function declaration. Please advice what is your preferred code style.

Is silencing pylint on this particular line ok?

heinezen: You can split the function declaration into two lines here and suppress the pylint output for `too-many-arguments`.

heinezen:
@mmtj Also: Please give your commits better descriptions, I beg you :D Otherwise it will not be possible to properly track changes.

* If you need to fix a small error in an existing commit, use `git amend` (also in your favourite git GUI)
* If you introduce a new feature, name the commit accordingly, e.g. `Pass compression_level to MediaExportRequest`

TheJJ:
No worries, his `fixup!` commits are alright, they're created by `git commit --fixup` and can be automatically squashed with `git rebase --autosquash` :) I think his intention was to squash them once ready and then to keep only meaningful commits in the end.

mmtj:
@TheJJ You guessed it right. I'd like to squash fixups at the end of review and keep only meaningful commit(s) to keep git history clean.

My intention is to make one commit with all changes related to png compression level, i.e. commit per feature, but I can split it into multiple commit per logical blocks if you like. e.g. "Pass compression_level to MediaExportRequest", "Pass compression_level through main convertor", and so on

mmtj:
@heinezen In my experience, amending commits during PR review is the thing that makes tracking changes and discussion about particular lines of code harder, whereas with fixups, you can still track smaller changes. Although not ideally.

That being said, I'm used to work with gitlab, which IMO handles tracking of commits during review better. But I can see why it could be confusing if you have different workflow. ;-)

heinezen:
@mmtj @TheJJ Aaaah! Shame on me for not knowing git features again :D I was worried that I would have to deal with a lot of weirdly named commits, but your explanation makes sense. As long as there are properly named commits afterwards I have no objections.

Btw, I've tested your code on AoC now and it works without problems :)

openage/convert/processor/export/modpack_exporter.py
heinezen: Oh, we indeed do not have to pass it anymore. Before 3a9bc4f1706feead417428aac285f76461b7468e `Texture` required game_version to find the palette number. This requirement was removed, so putting it in `kwargs` for the GRAPHICS media type is a leftover code snippet. You can remove it.

mmtj:
@heinezen Nice, thanks for confirmation :-)

openage/convert/processor/export/modpack_exporter.py
mmtj: Resolved in 37ceb58fb3daf61c36feddad6a18e0da58ce8526
Please review it again.

heinezen:
@mmtj Ready to review? :)

heinezen (dismissed):

TheJJ (approved):

heinezen:
:tada:

mmtj:
Nice, I guess we can close #1290 now :wink:

Issue #1312, opened 2020-09-07
Store hash-version map files for acquiring version number


heinezen:
**Required skills**: Python

**Difficulty**: Easy

**Motivation**

Age of Empires 2 and other Genie games do not store their version number in a plaintext file or anything else that is easy to read. Therefore, openage uses the hashes of files to determine the version of a game. The idea behind this is that a file from a specific release will always have the same hash value. By calculating the hash for a file and looking up the hash in a dict of hash values and version numbers, we can find the release version the file belongs to.

**Task**

The dicts of hashes and version numbers should ideally be stored in a config file. TOML can be used as a format since we already use that elsewhere in the converter. There should be one config file for every game edition and game expansion.

Example (using TOML):

```toml
# Header
# The version of this config file format
file_version = "1.0"
hash_algo = "SHA3-256"

[empires2dat]
path = "Data/empires2.dat"

    [empires2dat.map]
    "d45b001ee8493c1c61a1" = "1.0"
    "4239b6d45b001ee38128" = "1.2"


[age2_x1]
path = "age2_x1/age2_x1.exe"

    [age2_x1.map]
    "001ee8493c1c61a1d45b" = "1.0"
    "b001ee381284239b6d45" = "1.2"
```

Every file that we want to hash should get its own table with a subtable for the hash-version map. The recommended hashing algorithms are SHA3-256 or SHA-512.

drs-11:
Hey I can do this one too! Seems similar to the manifest file one.

heinezen:
@drs-11 Great :D

In comparison to the manifest file task, the hashes do not need to be generated this time. The file is manually created by the openage team and read in by the converter into `GameFileVersion` instances.

https://github.com/SFTtech/openage/blob/master/openage/convert/value_object/init/game_file_version.py#L10-L14

I can provide the SHA3-256 hashes for several older releases if you don't own them.

drs-11:
Yes please, I do not currently own any of the releases right now.

Ok so the file is need to be created and then `GameEdition` enumeration is needed to be modified to use the map file to enumerate the versions?

heinezen:
The Conqueros:

Version: 1.0c

```
d02ef2f4935c5ef568eaa873442b4b2b547067ed8e3572758bbbbee8  age2_x1.exe
4568f37f309c0a89fa9e84a0c0f6c61b0f394c28adc82c0b3f2203b3  age2_x1_nocd.exe
[...]

PR #1328, opened 2020-10-26
Create auxilary files for game edition/expansion information

### LINKED ISSUE #1314: Store game edition and game expansion info in auxiliary file instead of hardcoding it
**Required skills**: Python

**Difficulty**: Easy

openage currently stores information about a game edition or expansion inside an `Enum`, e.g. media paths, support status, etc. We should store this information in an auxiliary file instead, similar to #1312. That would make updating and correcting game information easier as we no longer have to replace chunks of code.

The recommended approach is to transfer the information from the existing implementation into a simple data definition file format. TOML is prefered because we already use that elsewhere in the converter.

Example:

```toml
# Header
# The version of this config file format
file_version = "1.0"

# Game edition info
game_edition_id = "AOC"
support_status = True

[media_paths]
datfile = [ "data/empires2.dat", ]
graphics = [ "data/graphics.drs", ]
palettes = [ "data/interfac.drs", ]
```

Since we cannot use Enums when we read the game versions from file, we would have to replace identity checks of the game version `Enum` value with an ID check. These checks happen when reading the `.dat` format or in the service that acquires nyan object names. They should be easy to replace.

**Further Reading**

* [GameVersion(Enum) implementation](https://github.com/SFTtech/openage/blob/master/openage/convert/value_object/init/game_version.py)

drs-11:
Fixes #1314 and #1312.

Things done so far:
- created version-hash toml files for game editions and expansions inside `version-hashes` directory
   - every game/expansion has a single toml file
   - remaining games with MD5 hashes were also created(not sure if this was needed to be done)
   - a directory was created to store the files
- created auxiliary files from `GameEdition` and `GameExpansion` enums
   - single files were created for both

@heinezen any changes or additions needs to be done?

Also `AOE1DE` enum in `game_version.py` has same MD5 hashes for both `AoEDE_s.exe` and `empires.dat`. Is that a coincidence?

### DIFF
--- a/openage/convert/main.py
+++ b/openage/convert/main.py
@@ -11,12 +11,13 @@ from ..util.fslike.wrapper import (DirectoryCreator,
 from ..util.strings import format_progress
 from .service.init.conversion_required import conversion_required
 from .service.init.mount_asset_dirs import mount_asset_dirs
+from .service.init.version_detect import create_version_objects
 from .tool.interactive import interactive_browser
 from .tool.subtool.acquire_sourcedir import acquire_conversion_source_dir, wanna_convert
 from .tool.subtool.version_select import get_game_version
 
 
-def convert_assets(assets, args, srcdir=None, prev_source_dir_path=None):
+def convert_assets(assets, cfg, args, srcdir=None, prev_source_dir_path=None):
     """
     Perform asset conversion.
 
@@ -35,8 +36,12 @@ def convert_assets(assets, args, srcdir=None, prev_source_dir_path=None):
     if srcdir is None:
         srcdir = acquire_conversion_source_dir(prev_source_dir_path)
 
+    # Initialize game versions data
+    auxiliary_files_dir = cfg / "converter" / "games"
+    args.avail_game_eds, args.avail_game_exps = create_version_objects(auxiliary_files_dir)
+
     # Acquire game version info
-    game_version = get_game_version(srcdir)
+    game_version = get_game_version(srcdir, args.avail_game_eds, args.avail_game_exps)
 
     # Mount assets into conversion folder
     data_dir = mount_asset_dirs(srcdir, game_version)
@@ -167,8 +172,14 @@ def main(args, error):
     else:
         srcdir = None
 
+    # mount the config folder at "cfg/"
+    from ..cvar.location import get_config_path
+    from ..util.fslike.union import Union
+    root = Union().root
+    root["cfg"].mount(get_config_path(args.cfg_dir))
+
     if args.interactive:
-        interactive_browser(srcdir)
+        interactive_browser(root["cfg"], srcdir)
         return 0
 
     # conversion target
@@ -176,7 +187,7 @@ def main(args, error):
     outdir = get_asset_path(args.output_dir)
 
     if args.force or wanna_convert() or conversion_required(outdir, args):
-        if not convert_assets(outdir, args, srcdir):
+        if not convert_assets(outdir, root["cfg"], args, srcdir):
             return 1
     else:
         print("assets are up to date; no conversion is required.")
--- a/openage/convert/main.py
+++ b/openage/convert/main.py
@@ -17,7 +17,7 @@ from .tool.subtool.acquire_sourcedir import acquire_conversion_source_dir, wanna
 from .tool.subtool.version_select import get_game_version
 
 
-def convert_assets(assets, cfg, args, srcdir=None, prev_source_dir_path=None):
+def convert_assets(assets, args, srcdir=None, prev_source_dir_path=None):
     """
     Perform asset conversion.
 
@@ -37,7 +37,7 @@ def convert_assets(assets, cfg, args, srcdir=None, prev_source_dir_path=None):
         srcdir = acquire_conversion_source_dir(prev_source_dir_path)
 
     # Initialize game versions data
-    auxiliary_files_dir = cfg / "converter" / "games"
+    auxiliary_files_dir = args.cfg_dir / "converter" / "games"
     args.avail_game_eds, args.avail_game_exps = create_version_objects(auxiliary_files_dir)
 
     # Acquire game version info
@@ -176,7 +176,8 @@ def main(args, error):
     from ..cvar.location import get_config_path
     from ..util.fslike.union import Union
     root = Union().root
-    root["cfg"].mount(get_config_path(args.cfg_dir))
+    root["cfg"].mount(get_config_path())
+    args.cfg_dir = root["cfg"]
 
     if args.interactive:
         interactive_browser(root["cfg"], srcdir)
@@ -187,7 +188,7 @@ def main(args, error):
     outdir = get_asset_path(args.output_dir)
 
     if args.force or wanna_convert() or conversion_required(outdir, args):
-        if not convert_assets(outdir, root["cfg"], args, srcdir):
+        if not convert_assets(outdir, args, srcdir):
             return 1
     else:
         print("assets are up to date; no conversion is required.")
--- a/openage/convert/service/conversion/internal_name_lookups.py
+++ b/openage/convert/service/conversion/internal_name_lookups.py
@@ -92,7 +92,7 @@ def get_class_lookups(game_version):
     if game_edition.game_id == "ROR":
         return ror_internal.CLASS_ID_LOOKUPS
 
-    if game_edition in ("AOC", "HDEDITION", "AOE2DE"):
+    if game_edition.game_id in ("AOC", "HDEDITION", "AOE2DE"):
         return aoc_internal.CLASS_ID_LOOKUPS
 
     if game_edition.game_id == "SWGB":
@@ -115,7 +115,7 @@ def get_command_lookups(game_version):
     if game_edition.game_id == "ROR":
         return ror_internal.COMMAND_TYPE_LOOKUPS
 
-    if game_edition in ("AOC", "HDEDITION", "AOE2DE"):
+    if game_edition.game_id in ("AOC", "HDEDITION", "AOE2DE"):
         return aoc_internal.COMMAND_TYPE_LOOKUPS
 
     if game_edition.game_id == "SWGB":
@@ -206,7 +206,7 @@ def get_gather_lookups(game_version):
     if game_edition.game_id == "ROR":
         return ror_internal.GATHER_TASK_LOOKUPS
 
-    if game_edition in ("AOC", "HDEDITION", "AOE2DE"):
+    if game_edition.game_id in ("AOC", "HDEDITION", "AOE2DE"):
         return aoc_internal.GATHER_TASK_LOOKUPS
 
     if game_edition.game_id == "SWGB":
@@ -262,7 +262,7 @@ def get_restock_lookups(game_version):
     if game_edition.game_id == "ROR":
         return None
 
-    if game_edition in ("AOC", "HDEDITION", "AOE2DE"):
+    if game_edition.game_id in ("AOC", "HDEDITION", "AOE2DE"):
         return aoc_internal.RESTOCK_TARGET_LOOKUPS
 
     if game_edition.game_id == "SWGB":
--- a/openage/convert/service/conversion/internal_name_lookups.py
+++ b/openage/convert/service/conversion/internal_name_lookups.py
@@ -11,7 +11,6 @@ import openage.convert.value_object.conversion.hd.fgt.internal_nyan_names as fgt
 import openage.convert.value_object.conversion.hd.raj.internal_nyan_names as raj_internal
 import openage.convert.value_object.conversion.ror.internal_nyan_names as ror_internal
 import openage.convert.value_object.conversion.swgb.internal_nyan_names as swgb_internal
-from ...value_object.init.game_version import GameEdition
 
 
 def get_armor_class_lookups(game_version):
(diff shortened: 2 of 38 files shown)


heinezen:
The TOML files look good :rocket: 

Now there are a few additional changes we have to make:

* I have updated the list of SHA3 hashes in https://github.com/SFTtech/openage/issues/1312#issuecomment-715425363 to include all game versions now. You should be able to exclusively use SHA3 with this.
* All TOML files need to move outside of the converter code into a "user area" where we can easily update them. A good place in the repo would be `cfg/converter/games/`
* Since we will read the game versions from a file instead of hardcoding them, we obviously can't let them be `Enum`s anymore.
    * `GameEdition` and `GameExpansion` need to be turned into regular classes (should be trivial)
    * These objects now have to be created by reading the files with `toml` and initializing them with the data inside. This can happen in a function inside https://github.com/SFTtech/openage/blob/master/openage/convert/service/init/version_detect.py . The function should return 
        1. A list of `GameEdition` objects
        2. A list of `GameExpansions` objects
    * They should get a new member `game_id` which is initialized with `game_edition_id` from your TOML files
    * These objects should be stored in `args.avail_game_eds` and `args.avail_game_exps`, since we need them later in the conversion process
* All functions that previously used the enums have to be adapted. The most important places are:
    1. The [version detection](https://github.com/SFTtech/openage/blob/e6879138514ea982870681517167bfc5d76b1149/openage/convert/service/init/version_detect.py#L11) service. This needs to get passed the games versions stored in `args`
    2. Every file in the [dat format reader](https://github.com/SFTtech/openage/tree/master/openage/convert/value_object/read/media/datfile). Instead of checking for identity, all version checks should be turned in string cmparisons based on `game_id`. As a result

    ```
    if game_version[0] is GameEdition.ROR:
    ```

    should become

    ```
    if game_version[0].game_id == "ROR":
    ```
    3. Other files might reference `GameEdition` or `GameExpansion` and need to be updated like in ii. You have to search the repo to find them

I can provide you a version of AoC for testing if you need it, btw. Write me in the chat for details :)

> Also AOE1DE enum in game_version.py has same MD5 hashes for both AoEDE_s.exe and empires.dat. Is that a coincidence?

Probably a copy-paste error on my behalf :D

drs-11:
@heinezen need a little help in saving the objects to `args`.

I thought calling the `create_version_objects` function in `convert_assets` function would be a good idea but having a little trouble accessing the cfg folder where the auxiliary files are located. I saw somewhere `args.cfg_dir` being used, can I use that?

heinezen:
@drs-11 You can assign `root["cfg"]` directly to `args.cfg_dir`, since `args` is a namespace. For example

```
args.cfg_dir = root["cfg"]
```
somewhere in `convert.main.py`. The occurence of `args.cfg_dir` you saw is probably in another subparser.

openage/convert/value_object/init/game_version.py
```diff
@@ -22,75 +22,92 @@ class Support(enum.Enum):
...
+                      for the converter.
+        """
+        self.game_id = game_id
+        self.flags = flags
+
+        self.support_status = None
+        self.game_file_versions = []
+        self.media_paths = {}
+        self.target_modpacks = None
+        self.expansions = []
```
heinezen, line 44: Game edition/expansion objects should be immutable (i.e. they should not be altered during conversion for obvious reasons) and therefore, all members should be set during object initialization. You can still use the functions you implemented, but they should be private and only be called from `__init__()`. This means you also have to pass it all necessary values as parameters.

In our architecture view, game editions/expansions are [value objects](https://github.com/SFTtech/openage/blob/master/doc/code/converter/architecture_overview.md#value-object).

drs-11:
@heinezen enum changes are almost finished. Have few queries on remaining enums.

File `struct_definition.py` calls `GameEdition.AOC` directly for `target.get_data_format` function. Which doesn't really uses the enum from what I've looked up. Is that right? Then is it safe to change it to string "AOC" ?

File `singlefile.py` uses `GameEdition.AOC` enum for creating DRS objects. I think passing string "AOC" will require a few changes in other places which calls DRS as well. Should I go ahead with this?

openage/convert/main.py
```diff
@@ -175,8 +180,14 @@ def main(args, error):
     from ..assets import get_asset_path
     outdir = get_asset_path(args.output_dir)
 
+    # mount the config folder at "cfg/"
```
drs-11, line 183: Umm not sure I did it right here. 

There's also `interactive_browser` function at line 175 which requires the `args` to have the `GameEdition` and `GameExpansion` objects to be available. I should push it after `convert_assets` function right?

heinezen:
> File struct_definition.py calls GameEdition.AOC directly for target.get_data_format function. Which doesn't really uses the enum from what I've looked up. Is that right? Then is it safe to change it to string "AOC" ?

Yes, you just give it the `game_id` identifier. `struct_definition.py` is a legacy module btw, so you don't need to worry about it much after changing the parameter to string. If the build fails because of that I will fix it myself.

> File singlefile.py uses GameEdition.AOC enum for creating DRS objects. I think passing string "AOC" will require a few changes in other places which calls DRS as well. Should I go ahead with this?

I think you can do that. The init function getting a string is fine for now. I think the only other place that reads the DRS is the main converter in this place: https://github.com/SFTtech/openage/blob/64b8c740ac3cf97131470a15bded56bf6ab97095/openage/convert/service/init/mount_asset_dirs.py#L27 . You should be able to easily change that.

openage/convert/main.py
heinezen: No, the interactive browser is a separate module that can be invoked before conversion. I would recommend you also pass it the cfg dir and replicate the game edition/expansion reading and subsequent version detection to the interactive browser.

openage/convert/main.py
```diff
@@ -35,8 +36,12 @@ def convert_assets(assets, args, srcdir=None, prev_source_dir_path=None):
     if srcdir is None:
         srcdir = acquire_conversion_source_dir(prev_source_dir_path)
 
+    # Initialize game versions data
+    auxiliary_files_dir = cfg / "converter" / "games"
+    args.avail_game_eds, args.avail_game_exps = create_version_objects(auxiliary_files_dir)
+
     # Acquire game version info
-    game_version = get_game_version(srcdir)
+    game_version = get_game_version(srcdir, args)
```
heinezen, line 44: You do not need to pass all of `args` here. Only the list of game editions and the list of game expansions needs to be passed.

openage/convert/value_object/read/media/datfile/sound.py
```diff
@@ -21,13 +20,13 @@ def get_data_format_members(cls, game_version):
...
 
-        if game_version[0] in (GameEdition.AOE1DE, GameEdition.AOE2DE):
+        if game_version[0].game_id in ("AOE1DE", "AOE2DE"):
             data_format.extend([
                 (SKIP, "name_len_debug", StorageType.INT_MEMBER, "uint16_t"),
                 (READ, "name_len", StorageType.INT_MEMBER, "uint16_t"),
                 (READ_GEN, "name", StorageType.STRING_MEMBER, "char[name_len]"),
             ])
-        elif game_version[0] is GameEdition.SWGB:
+        elif game_version[0].game_id is "SWGB":
```
heinezen, line 29: Should be `==`

drs-11:
Getting `ModuleNotFoundError` for `toml` in `version_detect.py`. Hmmm weird.
Not sure why it's happening.

heinezen:
@TheJJ Can you check if kevin installs toml correctly?

TheJJ:
Done, toml is now in the image. But unfortunately this also means pylint 1.6, and a lot of new compliance issues.....

drs-11:
Ok yeah there's tons of linting issues now which weren't there previously. Also putting string instead of object in `struct_definition.py` led to some implemented `get_data_format_members` functions to have errors. I guess just need to modify the `game_id` checking in those functions.

drs-11:
@heinezen I think I forgot to mention that I passed string "AOC" in `GenieStructure` class as well. The inherited classes in their `get_data_format_members` function will need to have their `game_id` checking as a string instead of objects. Should I go ahead?

heinezen:
@drs-11 I will deal with the legacy stuff myself. Its probably to frustrating for you to deal with.

drs-11:
@heinezen alright then. Anything else that needs to be done in this issue?

heinezen:
@drs-11 Have you ticket the box "Allow edits from maintainers" on the right?

drs-11:
@heinezen yup it's ticked.

heinezen:
@drs-11 For some reason, I can't push, so please extract and apply this patch to fix the buildsystem:


[0001-convert-Fix-buildsystem-requirement-for-new-GameVers.zip](https://github.com/SFTtech/openage/files/5475948/0001-convert-Fix-buildsystem-requirement-for-new-GameVers.zip)

heinezen:
It builds again! Now its time for the review

openage/convert/service/conversion/internal_name_lookups.py
```diff
@@ -90,13 +89,13 @@ def get_class_lookups(game_version):
     game_edition = game_version[0]
     # game_expansions = game_version[1]
 
-    if game_edition is GameEdition.ROR:
+    if game_edition.game_id == "ROR":
         return ror_internal.CLASS_ID_LOOKUPS
 
-    if game_edition in (GameEdition.AOC, GameEdition.HDEDITION, GameEdition.AOE2DE):
+    if game_edition in ("AOC", "HDEDITION", "AOE2DE"):
```
heinezen, line 95: Change to: `game_edition.game_id`

openage/convert/service/conversion/internal_name_lookups.py
```diff
@@ -204,13 +203,13 @@ def get_gather_lookups(game_version):
     game_edition = game_version[0]
     # game_expansions = game_version[1]
 
-    if game_edition is GameEdition.ROR:
+    if game_edition.game_id == "ROR":
         return ror_internal.GATHER_TASK_LOOKUPS
 
-    if game_edition in (GameEdition.AOC, GameEdition.HDEDITION, GameEdition.AOE2DE):
+    if game_edition in ("AOC", "HDEDITION", "AOE2DE"):
```
heinezen, line 209: Change to: `game_edition.game_id`

openage/convert/service/conversion/internal_name_lookups.py
```diff
@@ -260,13 +259,13 @@ def get_restock_lookups(game_version):
     game_edition = game_version[0]
     # game_expansions = game_version[1]
 
-    if game_edition is GameEdition.ROR:
+    if game_edition.game_id == "ROR":
         return None
 
-    if game_edition in (GameEdition.AOC, GameEdition.HDEDITION, GameEdition.AOE2DE):
+    if game_edition in ("AOC", "HDEDITION", "AOE2DE"):
```
heinezen, line 265: Change to: `game_edition.game_id`

heinezen (changes requested):
For next time: **Please test your code.** A lot of these bugs can be avoided by doing a test run.

drs-11:
Omg I'm really sorry, I had been thinking CI tests would be enough for this issue. Clearly mistaken. Sorry for all the trouble!

drs-11:
Tested with AoC this time. Sorry for the inconvenience caused.

heinezen:
@drs-11 Don't worry, it was not that troubling :)

The reason I was emphasizing testing is that it becomes very hard to write a good review if the code does not run because

1. I have to get it running by second-guessing the correct code that you wanted to write
2. Debug my own fixes and your code simultaneously

This can be very confusing :D

heinezen (dismissed):
Runs through for

* AoC
* RoR
* SWGB

So it should be fine.

heinezen:
@drs-11 Plz do a rebase, then we can merge :)

drs-11:
> The reason I was emphasizing testing is that it becomes very hard to write a good review

Absolutely, I agree.

> Plz do a rebase

Squash the commits as well then?

heinezen:
@drs-11 No squash, only rebase

drs-11:
@heinezen done

drs-11:
@heinezen done

openage/convert/service/init/version_detect.py
```diff
@@ -47,3 +54,78 @@ def iterate_game_versions(srcdir):
...
+    version_hashes_path = srcdir.joinpath("version-hashes")
+
+    # load toml config files to a dictionary variable
+    with game_edition_path.open() as game_edition_toml:
+        game_edition_dic = toml.loads(game_edition_toml.read())
+    with game_expansion_path.open() as game_expansion_toml:
+        game_expansion_dic = toml.loads(game_expansion_toml.read())
+
+    # create and list GameEdition objects
+    file_version_game_ed = game_edition_dic.pop("file_version")
```
heinezen, line 80: We can fix the sanity_check test here by changing the line to just

```
game_edition_dic.pop("file_version")
```

openage/convert/service/init/version_detect.py
```diff
@@ -47,3 +54,78 @@ def iterate_game_versions(srcdir):
...
+    for game in game_edition_dic:
+        game_obj = create_game_obj(game_edition_dic[game], version_hashes_path)
+        game_edition_list.append(game_obj)
+
+    # create and list GameExpansion objects
+    file_version_game_ex = game_expansion_dic.pop("file_version")
```
heinezen, line 86: We can fix the sanity_check test here by changing the line to just

```
game_expansion_dic.pop("file_version")
```

heinezen:
@TheJJ Any idea why the sanity check reports the same issue multiple times?

heinezen (approved):

heinezen:
Nice :)

drs-11:
_Phew_ finally.
Thanks! Definitely learned a few things doing this issue :D

heinezen:
Definitely, that was two tasks/issues at once :D

Issue #1343, opened 2020-12-18
Enhance output summary of converter run for debugging


heinezen:
**Required skills:** Python

**Difficulty:** Easy

During a converter run, a bunch of intermediary objects are created in the [processor stage](https://github.com/SFTtech/openage/blob/master/doc/code/converter/workflow.md), specifically AoE-like and API-like objects. These are used in the transformation from Genie Engine formats to openage formats (e.g. [nyan](https://github.com/SFTtech/nyan)).

The content and helper functions of the intermediary objects are used to map properties of the objects in the original game to properties in the openage API (e.g. a unit's ability). The motivation for having debug output for these objects is that this will help us detect semantic errors in the converter implementation. Because of the high number of objects in the game data, checking this during a live debug is very hard to do. 

Debug output of a run contains a plaintext dump of the most relevant content and output of helper functions. We store it in a dedicated directory (`assets/converted/debug`) and distinguish between the runs by using timestamps for folder names. The debug output is separated by conversion stage (init, read, processor, export). A lot of stuff is already implemented ([see here](https://github.com/SFTtech/openage/blob/master/openage/convert/service/debug_info.py)), but improving it can't hurt can it?

- [ ] Log the sound files that were requested for export but not found in the export stage [here](https://github.com/SFTtech/openage/blob/28b4b5ac022a6949a1570242d00704fc8775f45d/openage/convert/processor/export/media_exporter.py#L310)
- [ ] Execution time for each stage (read, conversion, export)

**Further reading**:

* [Debug info service implementation](https://github.com/SFTtech/openage/blob/master/openage/convert/service/debug_info.py)
* [Converter code](https://github.com/SFTtech/openage/blob/master/openage/convert/)
* [Converter architecture](https://github.com/SFTtech/openage/blob/master/doc/code/converter/architecture_overview.md)
* [Converter workflow](https://github.com/SFTtech/openage/blob/master/doc/code/converter/workflow.md)

fabiobarkoski:
Some PRs related to this issue were merged. If remain some work to do, can I try to work on this?

heinezen:
@fabiobarkoski Totally. I've updated the issue with a few ideas.

fabiobarkoski:
Thanks. If in some point I need help, is better ask here or on the matrix chat?

heinezen:
@fabiobarkoski Both is fine. The chat is better for back and forth if you have technical issues but whatever you're more comfortable with usually works best :)

PR #1346, opened 2020-12-28
Output summary of converter run for debugging

heinezen:
Implements the current ideas from https://github.com/SFTtech/openage/issues/1343 in a new dubug service.

  * [x] Init
    * [x] CLI args
    * [x] Game version
    * [x] Mounted folders
  * [x] Reader
    * [x] The .dat format that is read (because it depends the game version)
    * [x] Total number of string resources; maybe also list filenames
    * [x] Total number of media files (by type); maybe also list filenames
  * [x] Processor
    * [x] ~~for each `ConverterObject`: ID + helper function return value (where possible)~~ summary of different converter object types
    * [x] for each `ConverterObjectGroup`: ID + members + helper function return value (where possible)
    * [x] nyan object names
  * [x] Export
    * [x] for each modpack:
      * [x] modpack name
      * [x] info file
      * [x] manifest
      * [x] number of data files
      * [x] number of media files by type + list of converted files

Extra improvements:
* ~~Use a data definition format for output (probably toml) instead of a pure plaintext dumps~~
* [x] Document the debug outputs in `/docs`
* [x] Verbosity levels?

### DIFF
--- a/.github/workflows/macosx-ci.yml
+++ b/.github/workflows/macosx-ci.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Brew update
         run: brew update
       - name: Brew install DeJaVu fonts
-        run: brew tap homebrew/cask-fonts && brew cask install font-dejavu
+        run: brew tap homebrew/cask-fonts && brew install --cask font-dejavu
       - name: Install environment helpers with homebrew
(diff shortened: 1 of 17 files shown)


TheJJ (approved):
well done

PR #1347, opened 2020-12-29
change logo to svg and update domain names

TheJJ:
yea


TheJJ:
the transparent background of the logo in github-darkmode doesn't look good. hmm.

heinezen (approved):

PR #1348, opened 2020-12-30
buildsystem: install all files into prefix on Linux

okias:
Prequisite for Linux Flatpak support without ugly hacks.

Signed-off-by: David Heidelberg <david@ixit.cz>


heinezen (approved):

TheJJ (approved):
makes sense!

Issue #1355, opened 2021-01-02
KeyError: 208


LinuxDonald:
```
INFO [py] launching openage v0.4.1-377-g2be30ba76
INFO [py] compiled by GNU 10.2.0
  Do you want to convert assets? [Y/n]
> Y
  Should we call wine to determine an AOE installation? [Y/n]
> n
Could not find any installation directory automatically.
Please enter an AOE2 install path manually.
> /home/schatzi/nvme/Steam/steamapps/common/AoE2DE/
converting from '/home/schatzi/nvme/Steam/steamapps/common/AoE2DE'
INFO [py] Game edition detected:
INFO [py]  * Age of Empires 2: Definitive Edition
INFO [py] converting metadata
INFO [py] [0] palette
INFO [py] [1] empires.dat
INFO [py] Starting conversion...
INFO [py] Extracting Genie data...
INFO [py] Creating API-like objects...
INFO [py] Linking API-like objects...
INFO [py] Generating auxiliary objects...
INFO [py] Creating nyan objects...
Traceback (most recent call last):
  File "/usr/bin/openage", line 15, in <module>
    main()
  File "/usr/lib/python3.9/site-packages/openage/__main__.py", line 132, in main
    return args.entrypoint(args, cli.error)
  File "/usr/lib/python3.9/site-packages/openage/game/main.py", line 71, in main
    used_asset_path = convert_assets(
  File "/usr/lib/python3.9/site-packages/openage/convert/main.py", line 100, in convert_assets
    for current_item in convert(args):
  File "/usr/lib/python3.9/site-packages/openage/convert/tool/driver.py", line 48, in convert
    yield from convert_metadata(args)
  File "/usr/lib/python3.9/site-packages/openage/convert/tool/driver.py", line 96, in convert_metadata
    modpacks = args.converter.convert(gamespec,
  File "/usr/lib/python3.9/site-packages/openage/convert/processor/conversion/de2/processor.py", line 63, in convert
    modpacks = cls._post_processor(dataset)
  File "/usr/lib/python3.9/site-packages/openage/convert/processor/conversion/de2/processor.py", line 153, in _post_processor
    DE2NyanSubprocessor.convert(full_data_set)
  File "/usr/lib/python3.9/site-packages/openage/convert/processor/conversion/de2/nyan_subprocessor.py", line 38, in convert
    cls._process_game_entities(gamedata)
[...]

PR #1356, opened 2021-01-02
Make generated nyan files more readable

heinezen:
Improves readability and formatting of nyan files created by the converter.

- [x] Use statically generated aliases in import tree
- [x] Sort imports by name
- [x] Do not import object references if they are in the same namespace (e.g. as nested objects)
- [x] Define maximum line width for nyan container values and format them using multiple lines
- [x] No longer append `f` to nyan floats
- [x] fix https://github.com/SFTtech/openage/issues/1355
- [x] Fix macOS CI


heinezen:
**Before:**

Before

**After:**

After

TheJJ (approved):
very nice!

PR #1357, opened 2021-01-06
Adding a Dockerfile

kyleVsteger:
Hey all — this PR aims to dockerize compilation and release building. 
The application is configuring and compiling, but when I go to run the application from the command line I get this error: 
```
root@258b6ac02169:/tmp# openage
  File "/usr/local/bin/openage", line 14, in <module>
    from openage.__main__ import main
ModuleNotFoundError: No module named 'openage'
```
Steps to reproduce:
- Clone the repository
- `cd openage`
- `docker build -t openage:ubuntu`
- `docker run -it openage:ubuntu`
- `root@257bcac02168:/tmp# openage`

I was itching to play but couldn't get the application to configure on my machine (2013 MacBook Pro) so I started to tinker with a dockerfile.
My goal is to get the release building and running in an ubuntu container and eventually use Xquartz to forward the video along

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,47 @@
+FROM ubuntu:20.04
+
+ENV DEBIAN_FRONTEND="noninteractive"
+
+RUN apt-get update && apt-get install \
+    cmake \
+    cython3 \
+    flex \
+    git \
+    libeigen3-dev \
+    libepoxy-dev \
+    libfontconfig1-dev \
+    libfreetype6-dev \
+    libharfbuzz-dev \
+    libogg-dev \
+    libopus-dev \
+    libopusfile-dev \
+    libpng-dev \
+    libsdl2-dev \
+    libsdl2-image-dev \
+    make \
+    python3-dev \
+    python3-jinja2 \
+    python3-numpy \
+    python3-pil \
+    python3-pip \
+    python3-pygments \
+    qml-module-qtquick-controls \
+    qtdeclarative5-dev \
+    -y && apt-get clean
+
+RUN pip3 install \ 
+    cython \
+    numpy \
+    jinja2 \
+    lz4 \
+    pillow \ 
+    pygments \
+    toml
+
+WORKDIR /tmp
+COPY . .
+
+
+RUN ./configure --download-nyan --mode=release --prefix=/usr/local && \
+    make && \
+    make install


simonsan:
I'm not sure if it's already possible, didn't follow containerization for a year or two, but how do you want to run UI applications in a Docker container?

TobiasFP:
I love that you want to containerize it, it makes a lot of sense, however, I would advise against also running the application via docker.
One of the main goals is to bring OpenAge to all platforms, Linux, Mac and Windows (and maybe more), but by using xquartz to forward x11 to windows could lead to a slippery slope of not really building for windows going forward. 
Again, love the idea of using docker to build for Linux, and even making a separate windows docker container to build easily for windows, but I would advise against building for linux and then forwarding the game via xquarts from the container, but these are just my two cents.

kyleVsteger:
> One of the main goals is to bring OpenAge to all platforms, Linux, Mac and Windows (and maybe more), but by using xquartz to forward x11 to mac could lead to a slippery slope of not really building for mac going forward.

Thanks for the feedback @TobiasFP —  I can see why you have pause about this technique. I couldn't get the build to run properly on my Mac, but it was trivial enough to get it building in the container. If nothing else, this can serve as a consistent way to build the application. 

Regardless of _what_ gets done with the container/image, the error presented in the PR still stands —  does anyone have feedback on what may be causing that error?

heinezen:
@kyleVsteger It builds correctly, but he install target might be broken.

When doing

```
root@257bcac02168:/tmp# bin/run
```

it should be 

```
root@257bcac02168:/tmp# openage
```

it works.

kyleVsteger:
> @kyleVsteger It builds correctly, but the `install` target might be broken.
> 
> When running
> 
> ```
> root@257bcac02168:/tmp# bin/run
> ```
> 
> it works.

Thanks! Now I'm trying to get assets together. I'm by no means a linux buff... how does one go about getting assets? I see in `building.md` a call out to 'wine being your friend'

heinezen:
The openage assets are converted from the original data and media files. On startup, openage should ask for a game folder. The game version is automatically detected and - if compatible - converted to our modpack format.

Please not that the engine cannot run with the new asset formats that are generated by the engine from the `master` branch right now. [Reasons here](https://blog.openage.dev/new-gamestate-2020.html). But you can build the latest release that still supports the old converted asset format.

Issue #1338, opened 2020-11-24
Speed up optimized PNG compression with pre-computed parameters


heinezen:
**Required skills:** Python, maybe Cython

**Difficulty:** Medium

The openage converter is able to optimize the compression for created PNG files by utilizing a search algorithm inspired by [OptiPNG](http://optipng.sourceforge.net/). To do this, the search algorithm first tries 8 different compression settings by creating PNGs with these settings in-memory. The compression settings that produced the smallest PNG are then used to export the image data into a PNG file. Using the optimized compression usually results in a much smaller file size (~50% of unoptimized size). However, the optimized compression is currently much slower as the resulting PNG has to be created 8 times (instead of 1 time with unoptimized compression) before it can be exported.

The way we can speed this process up is to save the best compression settings for each source image in a configuration file and use them as input during the PNG export phase. As a result, the PNG exporter can skip the 8 trials of the search algorithm and directly generate the "best" PNG. Furthermore, using optimized compression would then be as fast as using unoptimized compression.

To do this, we have to do the following:

1. Save the best compression settings for each source image in a configuration file
2. Load the file for media conversion and forward the compression settings to the PNG exporter
3. Use the settings during PNG export (this is already implemented)

For the configuration file, TOML should be used. The compression settings should be accompanied by information to identify the source image file (i.e. filepath and a SHA-256 hash value).

Example:

```
# Header
# The version of this config file format
file_version = "1.0"
hash_algo = "SHA3-256"

[images]
    [2.slp]
    "filepath" = "graphics/2.slp"
    "hash" = "DEDF4675668799912378178479281"
    "settings" = [4, 5, 6, 7]

    [1080.slp]
    "filepath" = "graphics/1080.slp"
    "hash" = "79244847392374923749273497237"
    "settings" = [4, 5, 6, 7]
```

**Further Reading**

* [PNG exporter code](https://github.com/SFTtech/openage/blob/master/openage/convert/service/export/png/png_create.pyx)
* [Converter Architecture Overview](https://github.com/SFTtech/openage/blob/master/doc/code/converter/architecture_overview.md)
* [Converter Workflow](https://github.com/SFTtech/openage/blob/master/doc/code/converter/workflow.md)

PR #1363, opened 2021-01-16
Speed up media conversion significantly

heinezen:
This issue brings several speed ups and code quality improvements to the code.

- [x] Move `save()` out of `MediaExportRequest` and `DataDefinition` (https://github.com/SFTtech/openage/issues/1311)
  - Saving is now done in `media_exporter` module
  - Texture packing/transformation triggered by `MediaExporter` function depending on the media type
  - Subclasses of `MediaExportRequest` were removed to make using it more simple

- [x] Speed up SMX/SMP/SLP parsing
  - Minimize calls to `self`
  - Use `ctuple` instead of `tuple` for return type
  - Up to 20x faster parsing

- [x] (Optional) media cache for converted graphics (https://github.com/SFTtech/openage/issues/1338)
  - Can be used to always produce the same graphics or save time during media conversion
  - Generated with converter debug info level 6 (warning: it takes a long time)
  - Places the generated file into the `converter/games` cfg folder of the respective game


TheJJ (approved):
\o/ speeeed

PR #1367, opened 2021-01-20
Fix: Turn Cython profiling off

heinezen:
We accidently left this on in https://github.com/SFTtech/openage/pull/1363, so now there is also a compliance check that tests if profiling is deactivated.


TheJJ (approved):
i wonder if there's a `cylint` tool

PR #1373, opened 2021-02-01
DE2: Lords of the West support

heinezen:
... and some other improvements.

- LOTW .dat file support
- Unit/Building line creation was improved, i.e. the code should make more sense now and is less confusing
- Use new `.sprite` format specification


TheJJ (approved):
awesome!

PR #1374, opened 2021-03-03
Integrate new media formats into converter

heinezen:
Adds specs from https://github.com/SFTtech/openage/pull/1318 and https://github.com/SFTtech/openage/pull/1186 to the converter.


TheJJ (approved):
yeey, more nyan conversion

Issue #1306, opened 2020-09-02
open_memstream() unsupported on Windows


heinezen:
We currently rely on the POSIX function `open_memstream()` to get a file pointer for an in-memory buffer during PNG conversion. `libpng`'s standard output path asks for a `FILE *` stream, so we cannot just pass it a `char *`. Our reason for handling the PNG creation in-memory is that we want to

1. let Python handle the file writing and directory creation **and**
2. if PNG optimization is used, not have to write every optimization trial to the filesystem as that's slow

Unfortunately, Windows does not support `open_memstream()` and therefore we have to do something different. I think the easiest way to handle this would be to implement a custom PNG write function that writes to a buffer and pass it to `libpng` with `png_set_write_fn()` ([described here](https://refspecs.linuxbase.org/LSB_3.1.0/LSB-Desktop-generic/LSB-Desktop-generic/libpng12.png.set.write.fn.1.html)).

duanqn:
Maybe this could help?
https://stackoverflow.com/questions/12249610/c-create-file-in-memory

heinezen:
@duanqn Thx, I think I even considered this when initially wrote the png exporter. Though I would ideally not rely on "dirty" hacks, especially if they are only made for one platform. They tend to break a lot more often and in unexpected ways :D 

However, it's better than having this issue unresolved, so if anyone builds a solution around this, that's also OK.

duanqn:
I may be able to create a `open_memstream()` C++ function that works on both Windows and Linux. But I don't know how you can import them like the current `posix.stdio` package. Can you write the glue code between C++ and Python?

heinezen:
@duanqn Sounds great! Of course, I will assist you with the Cython importing. It's mostly an explicit import of functions from a C/C++ header file, as shown here:

https://github.com/SFTtech/openage/blob/afbb624518e360bb96835badfa2c37979ff2e7b8/openage/convert/service/export/png/libpng.pxd#L3-L8

Cython also [has a tutorial for that](https://cython.readthedocs.io/en/latest/src/userguide/external_C_code.html) if you want to try it yourself.

duanqn:
@heinezen I'll start working on that then. Also, I read that Mac OS does not have `open_memstream()` either, but I don't have a Mac device to confirm that.

heinezen:
macOS should have `open_memstream()` as it implements the POSIX standard (at least I would be surprised if it didn't).

duanqn:
[...]

PR #1375, opened 2021-03-17
Converter speedup and Windows works again

heinezen:
Brings several improvements to the speed of media exporting:

* Speedup binpacking with more Cython
* Do terrain tile merging and projection in a single step
* Deactivate functions that were reliant on POSIX (https://github.com/SFTtech/openage/issues/1306)


TheJJ (approved):
:tada:

PR #1066, opened 2018-11-05
[WIP] Core gamestate implementation

TheJJ:
Yea... 

* [x] move old gamestate
* [x] create demo (`openage.main.tests.engine_demo`)
* [x] implement pong
  * [x] calculated with the event engine
  * [x] displayed with the new renderer
  * [x] in a stupid non-event-based-loop (that still processes events but ticks regularily)
  * [ ] define minimal gamestate for displaying entities
  * [ ] configure pong with nyan
  * [ ] define basic nyan api for entity display
  * [ ] use the new renderer to display the basic entity
  * [ ] prepare abilities and actions (e.g. movement) for the basic entity

once that is done (probably separate pull):
* [ ] create basic presenter that is decoupling the simulation from display
* [ ] update the pong simulation to run in a non-stupid loop (that sleeps dynamically)

PR #1378, opened 2021-03-30
Core gamestate

heinezen:
Brings https://github.com/SFTtech/openage/pull/1066 up-to-date.

* [x] move old gamestate
* [x] create demo (`openage.main.tests.engine_demo`)
* [x] implement pong
  * [x] calculated with the event engine
  * [x] displayed with the new renderer
  * [x] in a stupid non-event-based-loop (that still processes events but ticks regularily)
  * [ ] define minimal gamestate for displaying entities
  * [ ] configure pong with nyan
  * [ ] define basic nyan api for entity display
  * [ ] use the new renderer to display the basic entity
  * [ ] prepare abilities and actions (e.g. movement) for the basic entity

once that is done (probably separate pull):
* [ ] create basic presenter that is decoupling the simulation from display
* [ ] update the pong simulation to run in a non-stupid loop (that sleeps dynamically)


TheJJ:
Although lots of things are missing, let's merge this for now :)

PR #1379, opened 2021-04-03
Remove game data code generation

heinezen:
Replaces the gamedata structs that were previously generated by the old converter for the old gamestate with dummy files.

Deprecated struct generation methods were finally removed from the converter.


TheJJ (approved):
step by step!

PR #1382, opened 2021-04-18
Bugfix: Correct the damage_modifier_2 bitmask

nymanjens:
`damage_modifier_2` is 6 bits long, but the bitmask `0x1F` had only 5 bits (contrary to what the comment above it says). This PR fixes that.

This fix is also aligned with the SMX specification (https://github.com/SFTtech/openage/blob/master/doc/media/smx-files.md) in this project. I found this bug and validated the proposed fix when I was implementing my own SMX updater library in Python.


heinezen (changes requested):
Thank you! :)

Just one small thing: You need to add yourself to [copying.md](https://github.com/SFTtech/openage/blob/master/copying.md). Then our CI won't complain anymore.

nymanjens:
> Just one small thing: You need to add yourself to copying.md. Then our CI won't complain anymore.

Done

heinezen (approved):
:tada:

PR #1383, opened 2021-04-18
Explicitly initialize integers created via cdef

nymanjens:
Not doing this led (on my machine) to some weird non-deterministic behavior: In a small number of resources, the palette would vary in the first 4 pixels (4to5 compression)


heinezen (approved):
I smell funny C-compiler legacy bugs :D

Anyway, looks good to me!

PR #1384, opened 2021-04-24
Fix HD Edition's empires_x1 dat file conversion

heinezen:
The converter now uses the correct number of terrains for that file --> 42 like in AoC.

Language file support was also streamlined, so that the data conversion is successful. Media conversion still fails because of the non-existent terrain SLP files.


TheJJ (approved):
:+1:

PR #1385, opened 2021-04-26
Fixing up some Windows build issues

xarianthe:
fixes the namespace issue in certain files, adds .cmd script to copy .pyd files into the correct directories on windows, adds a conditional import for windows


TheJJ:
moving the files to the system include path doesn't seem right. can you persuade your python to find them inside the build directory?

xarianthe:
> 
> 
> moving the files to the system include path doesn't seem right. can you persuade your python to find them inside the build directory?

Unfortunately, I was unable to get python to find that .pyd file. I tried adding it to `PATH`,`PYTHONPATH`, no luck. Having it in `site-packages` is what's working. That's where manually built python packages go into it seems, that's where python searches for them. Possible workaround that might be to create a virtualenv and move it to that location instead of a system folder.

TheJJ:
We now define `SDL_MAIN_HANDLED` (via #1407), so you can remove it from your patch.
Is there a better way than copying files with `copycmd`? also your last hunk which changes the relative module import to some absolute one for windows only seems to have the same root cause - the python search path is somehow wrong.
I assume it works the same on Windows as it does on Linux, so when running openage inside the build dir, why and how exactly does it fail to function?

TheJJ:
i'll close this for now, it has conflicts - we can come back once we want more windows improvements

PR #1386, opened 2021-04-27
Preliminary DE1 support + 32-Bit SLP conversion

heinezen:
Extends the converter with DE1 support

* Activate conversion Age of Empires 1: Definitive Edition
* Support 32-Bit SLP files
* Fix a few bugs in the SLPv4 conversion


TheJJ (approved):
i like

PR #1395, opened 2021-07-11
Adjustments for DE2 50292 update

heinezen:
DE2 made a few chanes to the dat format

* New attribute `hero_glow_graphic_id`
* Berserk heal rate is now stored in previously unused value `rear_attack_modifier`
* A few new shenanigans with techs and civ boni


TheJJ (approved):
yey

PR #1407, opened 2021-08-30
Suppress SDL_main definition

duanqn:
Define SDL_MAIN_HANDLED to suppress definition of SDL_main. Otherwise it would `#define main SDL_main` which breaks so many places. (See https://github.com/duanqn/openage/runs/3463316414?check_suite_focus=true)


duanqn:
I found this is already covered in #1385 by @xarianthe but I'm splitting it to unblock Windows building.

TheJJ:
This totally feels like a hack, the buildsystem as a whole should not be aware of funny flags sdl needs :D I think a better solution might be this:
* define the preprocessor flag with [`INTERFACE_COMPILE_DEFINITIONS`](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_COMPILE_DEFINITIONS.html), i.e. `set_target_properties(SDL2::SDL2 PROPERTIES INTERFACE_COMPILE_DEFINITIONS ...) `
* and we can even get rid of findsdl2, and use the config-package directly. sdl2 ships this nowadays (for me in `/usr/lib64/cmake/SDL2/sdl2-config.cmake`)

what do you think?

duanqn:
I never used `INTERFACE_COMPILE_DEFINITIONS` before but I'll try with that. I'm definitely not a cmake expert... Can I just keep findsdl module and set the `INTERFACE_COMPILE_DEFINITIONS` for SDL2::SDL2 there?

duanqn:
Ummm how does it look now ? @TheJJ

TheJJ:
Yea exactly this :) I don't think we need to request `2.0`.
Does it work for you like this?

TheJJ (approved):

duanqn:
Yeah 2.0 probably can be removed. I'll try removing it.

TheJJ (approved):
perfect!
removing code while retaining all functionality is always a pleasure :)

PR #1409, opened 2021-09-05
Add a function to save individual SMX files

duanqn:
Would be nice to have such a function in interactive mode. Works on my computer :P


heinezen:
@duanqn Nice one :)

heinezen (approved):

PR #1410, opened 2021-09-05
[WIP] Add Ubuntu 21.04 CI pipeline via GitHub Actions

duanqn:
Add an Ubuntu 21.04 CI pipeline which is capable of
1. Test building on Ubuntu 21.04.
2. Export the build environment as a docker image.
3. Export the build artifacts (currently the entire build folder).


duanqn:
Dockerfile was inspired by #1357 .

.github/workflows/ubuntu-21.04.yml
```diff
@@ -0,0 +1,55 @@
+name: Ubuntu 21.04 CI
+
+on: [push, workflow_dispatch]
+
+jobs:
+  build-devenv:
+    runs-on: ubuntu-20.04
```
TheJJ, line 7: 20 or 21? :)

docker/devenv/Dockerfile.ubuntu.2104
```diff
@@ -0,0 +1,43 @@
...
+RUN python3 --version; \
+    python3 -m pip install --upgrade pip; \
+    python3 -m pip install \ 
+    Cython \
+    numpy \
+    jinja2 \
+    lz4 \
+    pillow \ 
+    pygments \
+    toml; \
```
TheJJ, line 42: many if not all of them are regular packages in the default repos

.github/workflows/ubuntu-21.04.yml
```diff
@@ -0,0 +1,55 @@
...
+        name: devenv-image-compressed
+        path: '/tmp/staging/devenv.tar.gz'
+        if-no-files-found: error
+        retention-days: 30
+
+  build:
+    runs-on: ubuntu-20.04
+    needs: build-devenv
+    steps:
+    - uses: actions/checkout@v2
```
TheJJ, line 30: does it always run? since kevin does the checks for pull requests, is there a way to let github-actions only be run in forked child repos?

.github/workflows/ubuntu-21.04.yml
duanqn: Yeah it's a bit tricky here. This workflow runs on 20.04 machines (provided by GitHub, see https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners), but the build environment is a docker image built from Ubuntu 21.04 (because 21.04 ships gcc-11).

.github/workflows/ubuntu-21.04.yml
duanqn: Currently it runs on every push, and can also be manually triggered (`[push, workflow_dispatch]` on line 3).

duanqn: I believe there is a way to trigger it on specific repos, but personally I think it's also beneficial on the main repo. Kevin CI can still serve as PR checks, while this one can be the source of truth for Ubuntu building. With this workflow, we no longer need to answer "can't compile on my Ubuntu" questions. Everyone who wants to contribute can download the docker image and there's no need to setup local dev environment.
For openage users -- I know the game isn't complete yet, but one day we'll get there -- they can download the output of the second stage, which should be a package (.deb ?) that is ready for the end users. However, currently I just exported the build folder as the output for the second stage. There's some future work on packaging.

heinezen: If it's

`on: [push, pull_request]`

it should only run on forked child repos when they make a PR.

docker/devenv/Dockerfile.ubuntu.2104
duanqn: Hmmm what do you mean? We don't need to install them?

PR #1411, opened 2021-09-05
[WIP] Add Windows CI pipelines

duanqn:
Add Windows CI pipelines for Windows Server 2019 & 2022 (preview). The two pipelines should produce the same output.

Ideally, the pipeline should export a package that people can download and use, but currently the build artifact just contains the build folder.


duanqn:
Depends on #1407 and #1408 .

PR #1414, opened 2021-09-17
Add DE2 Dawn of the Dukes support + HD Edition fixes

heinezen:
Introduces a few improvements for the asset converter.

* Allow alternative paths/filenames for required files used in the version detection (This allows us to use the same game definition for Steam and MS Store distributions)
* Check file hashes in version detection (only for debugging)
* Use game edition ID for naming .dat cache files
* **DE2: Dawn of the Dukes support**
* Terrain textures are now referenced by index instead of SLP ID 
    * HD Edition terrain textures are now exported by copying the original file

### DIFF
--- a/cfg/converter/games/de2/version_hashes.toml
+++ b/cfg/converter/games/de2/version_hashes.toml
@@ -1,16 +1,16 @@
 # version hashes of Age of Empires 2: Definitive Edition
 
-file_version = "1.0"
+file_version = "2.0"
 hash_algo = "SHA3-256"
 
 [AoE2DE_s]
-path = "AoE2DE_s.exe"
+paths = ["AoE2DE_s.exe", "AoE2DE.exe"]
 
   [AoE2DE_s.map]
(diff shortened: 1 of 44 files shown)


cfg/converter/games/de2/version_hashes.toml
```diff
@@ -1,16 +1,16 @@
 # version hashes of Age of Empires 2: Definitive Edition
 
-file_version = "1.0"
+file_version = "2.0"
 hash_algo = "SHA3-256"
 
 [AoE2DE_s]
-path = "AoE2DE_s.exe"
+paths = ["AoE2DE_s.exe", "AoE2DE.exe"]
```
duanqn, line 7: Thank you! <3

TheJJ (approved):

cfg/converter/games/de2/version_hashes.toml
heinezen: @duanqn I totally forgot to ask for the hashes of your `AoE2DE.exe` file :D Shouldn't be a problem though because the hash checks are not mandatory.

PR #1415, opened 2021-09-25
Add a Python argparse option to supply additional DLL search paths for Windows 

duanqn:
As stated in https://docs.python.org/3/library/os.html#os.add_dll_directory, Python 3.8+ on Windows uses DLL paths added with `os.add_dll_directory`, and the PATH is not used.

### DIFF
--- a/openage/__main__.py
+++ b/openage/__main__.py
@@ -56,6 +56,10 @@ def main(argv=None):
                             help=("upon throwing an exception a debug break is "
                                   "triggered. this will crash openage if no "
                                   "debugger is present"))
+    global_cli.add_argument(
+        "--add-dll-search-path", action='append', dest='dll_paths',
+        help="(Windows only) provide additional DLL search path")
+
     devmodes = global_cli.add_mutually_exclusive_group()
     devmodes.add_argument("--devmode", action="store_true",
                           help="force-enable development mode")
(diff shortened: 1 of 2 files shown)


heinezen (approved):

openage/__main__.py
```diff
@@ -56,6 +56,10 @@ def main(argv=None):
                             help=("upon throwing an exception a debug break is "
                                   "triggered. this will crash openage if no "
                                   "debugger is present"))
+    global_cli.add_argument(
+        "--add-dll-search-path", action='append', dest='dll_paths',
+        help="(Windows only) provide additional DLL search path")
+
```
heinezen, line 62: If the argument is only needed on Windows, we could add the `if sys.platform == 'win32'` platform check here too.

openage/game/main.py
```diff
@@ -30,6 +30,13 @@ def main(args, error):
     """
     del error  # unused
 
+    win_dll_path_handles = []
+    import sys
```
heinezen, line 34: This can probably be a top-level import now that we use it twice or more ^^

heinezen (changes requested):

heinezen (dismissed):

heinezen:
One last thing: kevin complains about the import order :D

openage/__main__.py
duanqn: Added. Thanks!

heinezen (approved):

PR #1416, opened 2021-09-25
Fix build issues on GCC-11

duanqn:
openage now builds successfully with GCC-11: https://github.com/duanqn/openage/runs/3708468188

Previous build log: https://github.com/duanqn/openage/actions/runs/1234113009


TheJJ (approved):
great!

PR #1417, opened 2021-09-25
Update Windows build instructions

duanqn:
Update Windows build instructions to include the `--add-dll-search-path` option.


heinezen (approved):
Thanks!

PR #1424, opened 2021-10-04
more detailed log for conversion of media files

deepakdinesh1123:


openage/convert/processor/export/media_exporter.py
```diff
@@ -5,6 +5,7 @@
 Converts media requested by export requests to files.
 """
 import os
+import logging
```
heinezen, line 8: openage has its own logging implementation in `openage/log`. You can use that instead of the native Python logger.

You can access the debug loglevel with `from openage.log import dbg` and use that like you would use the native Python `logging.debug()` function.

openage/convert/processor/export/media_exporter.py
```diff
@@ -122,6 +123,12 @@ def _export_blend(export_request, sourcedir, exportdir, blend_mode_count=None):
                 exportdir[export_request.targetdir],
                 f"{export_request.target_filename}{idx}.png"
             )
+        source_format = export_request[export_request.source_filename.rindex("."):]
```
heinezen, line 126: Export request has no operation for `[]`, so `export_request[export_request.source_filename.rindex("."):]` doesn't work. I guess you are trying to get the extension of the filename. There is a quicker way to access it is using the `suffix` property of the source file path stored in `source_file`, see

https://github.com/SFTtech/openage/blob/master/openage/util/fslike/path.py#L265-L271

The same mechanism exists in Python's `pathlib` (https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.suffix)

openage/convert/processor/export/media_exporter.py
```diff
@@ -122,6 +123,12 @@ def _export_blend(export_request, sourcedir, exportdir, blend_mode_count=None):
...
+        source_size = os.path.getsize(source_file)
+        target_size = os.path.getsize(export_request.targetdir+export_request.target_filename)
+        logging.basicConfig(level=logging.DEBUG)
```
heinezen, line 129: You don't need this line if you import the openage logger. It handles the setup for you.

openage/convert/processor/export/media_exporter.py
```diff
@@ -122,6 +123,12 @@ def _export_blend(export_request, sourcedir, exportdir, blend_mode_count=None):
...
+        logging.debug(f"Source File:{export_request.source_filename} ({source_format}, {source_size}) -> Target File:{export_request.target_filename} (PNG, {target_size})")
```
heinezen, line 130: Looks good to me :)

Instead of hardcoding `PNG` as the target format, you could also use the `suffix` property of `target_file` here. This way, the code becomes more future proof if we decide to add another target format.

heinezen (changes requested):
I have made a few remarks for your first code snippet.

deepakdinesh1123:
I have implemented the changes and created a new commit

heinezen (changes requested):
Great, the code looks good now.

Now, we just have to get rid of the build errors. On the Github PR page, you should see a lot of red :x: in the CI view (`kevin ci`). Before we can merge, these tests must succeed. You can see what you need to fix when you click on "Details".

deepakdinesh1123:
After fixing some formatting errors I am getting an error that in Path(<path>).suffix() suffix is not callable

heinezen:
`suffix` is a [property](https://docs.python.org/3/library/functions.html#property). They are addressed like object member, i.e. `Path().suffix` without the `()`.

deepakdinesh1123:
I have added my email to copying.md but I am getting an error

Issue #1376, opened 2021-03-20
Speed up PNG compression trials with in-memory files


heinezen:
The converter has support for optimizing PNG compression like [OptiPNG](http://optipng.sourceforge.net/) by trying to compress a file several times (each with different compression parameters) and choosing the result with the smallest file size. Currently, we initialize each trial by getting a file pointer from `libc.stdio.tmpfile` and writing to that file. `libpng`'s standard output path asks for a FILE * stream. However, using a temporary file is slower than doing everything in-memory.

One solution would be the usage of the more advanced `libpng` API. We can pass `libpng` a custom write function `png_set_write_fn()` ([described here](https://refspecs.linuxbase.org/LSB_3.1.0/LSB-Desktop-generic/LSB-Desktop-generic/libpng12.png.set.write.fn.1.html)) that writes to a buffer. This would have to be done in pure C++ which is then called by Cython code in the converter.

heinezen:
https://github.com/SFTtech/openage/pull/1495 resolved this issue.

PR #1430, opened 2021-10-27
Optimized PNG compression using in-memory 'files'

MavethGH:
An implementation of https://github.com/SFTtech/openage/issues/1376

Feedback of any kind welcome!


heinezen (dismissed):
Looks like it works! The only thing left is that you have to add your alias to the copying.md file.

openage/convert/service/export/png/png_tmp_file.pxd
```diff
@@ -0,0 +1,97 @@
...
+
+    #include "png.h"
+
+    #include <cstdint>
+    #include <cstdlib>
+    #include <cstring>
+
+    struct tmp_file_buffer_state {
+        char *buffer;
+        size_t size;
```
TheJJ, line 18: since we are in c++ world, can you use a `std::vector` (or stream, or unique_ptr, ...) here?

openage/convert/service/export/png/png_create.pyx
```diff
@@ -390,3 +391,69 @@ cdef void write_to_file(numpy.uint8_t[:,:,::1] imagedata,
...
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+cdef void write_to_buffer(numpy.uint8_t[:,:,::1] imagedata,
+                          png_tmp_file.tmp_file_buffer_state bufstate,
+                          int compression_level, int memory_level,
+                          int compression_strategy, int filters,
+                          int width, int height):
+    """
+    Write an image matrix with RGBA color values to a file.
```
TheJJ, line 404: it's not a file, but rather the given bufstate?

openage/convert/service/export/png/png_tmp_file.pxd
```diff
@@ -0,0 +1,97 @@
...
+    """
+    ctypedef struct tmp_file_buffer_state:
+        char* buffer
+        size_t size
+
+    void tmp_file_png_write_fn(libpng.png_structp png_ptr,
+                               libpng.png_bytep data,
+                               libpng.png_size_t)
+    void tmp_file_flush_fn(libpng.png_structp png_ptr)
+    int check_compressed_size(uint8_t ***imagedata,
```
TheJJ, line 91: what's this used for? currently the function is called from nowhere?

TheJJ (changes requested):
and it needs to be wired up to be actually used :)

openage/convert/service/export/png/png_tmp_file.pxd
heinezen: `libpng` lives in C world still, so keeping it C-like is probably for the best.

TheJJ: indeed, but as we need to pass in a memory buffer state when calling this, I'd prefer a c++ variant.

openage/convert/service/export/png/png_tmp_file.pxd
MavethGH: tmp_file_flush_fn isn't meant to be called. It's just there bc I'm not sure if libpng requires the flush callback to be non-null. The only function that's meant to be called by client code is check_compressed_size.

openage/convert/service/export/png/png_tmp_file.pxd
MavethGH: This struct only gets used by the callback function passed to libpng. I chose to do it this way bc the code is a lot clearer without having intermediate steps to ensure interoperability between C and C++. This way is likely also more performant, but that's not a big deal since this function isn't getting called every frame or something.

openage/convert/service/export/png/png_tmp_file.pxd
heinezen: I've removed `check_compressed_size` from the C++ part and implemented it in Cython instead. Works better in the long run.

openage/convert/service/export/png/png_tmp_file.pxd
heinezen: @TheJJ `char*` can be directly converted to a Python bytes array by Cython ([see here](https://docs.cython.org/en/latest/src/tutorial/strings.html#passing-byte-strings)), so I stayed with that.

openage/convert/service/export/png/png_tmp_file.pxd
MavethGH: Fair enough. The C++ implementation wasn't designed to be particularly reusable.

heinezen:
@MavethGH You still need to add your info to copying.md. I've fixed the other stuff missing from the Cython-side implementation :)

heinezen (changes requested):
see above

heinezen:
Do you need assistance with the copying.md authorship entry?

heinezen:
@MavethGH  Do you want to update the PR or should we close?

MavethGH:
Sorry for the delay; I've been busy. I personally don't really care about having my name in copying.md, but then again, copyright law is nonsense to me, so there might be some silly reason that not putting my name on there affects someone other than me. If that's the case, let me know, and I can put my info on there. As for the code, yeah, seems fine to me.

heinezen:
@MavethGH You can use you pseudonym instead of the name. The email is more important because there are some cases where we would need to contact all authors for approval (license changes for example). It's also a way to give credit to contributors.

If you don't care about the credit at all and don't want to be in copying.md, I could set the commit author to myself. You'd just have to give me permission to transfer the authorship to me.

TheJJ:
@MavethGH pling :) I interpret your last comment as you don't care about the copyright, and we can upstream the commits under our name, at least in case you don't object (or we don't hear back in the nearer future).

heinezen:
superceded by #1495

PR #1456, opened 2022-01-25
Fix/mac os build on arm

dlejay:
It should now compile correctly on ARM macOS.

### DIFF
--- a/libopenage/CMakeLists.txt
+++ b/libopenage/CMakeLists.txt
@@ -24,7 +24,7 @@ find_path(FREETYPE_INCLUDE_DIRS freetype/freetype.h HINTS /usr/include/freetype2
 
 # provide apple qt5 location
 if(APPLE)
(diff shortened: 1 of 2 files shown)


libopenage/CMakeLists.txt
```diff
@@ -24,7 +24,7 @@ find_path(FREETYPE_INCLUDE_DIRS freetype/freetype.h HINTS /usr/include/freetype2
 
 # provide apple qt5 location
 if(APPLE)
-	list(APPEND CMAKE_PREFIX_PATH /usr/local/opt/qt5)
+	list(APPEND CMAKE_PREFIX_PATH ${HOMEBREW_PREFIX}/opt/qt5)
 endif()
```
TheJJ, line 28: how is this variable available here already? probably the `brew --prefix` invocation has to be called in this `if(APPLE)` block and not down below?

TheJJ (commented):
thx

dlejay:
There is a mistake, let me pull again.

PR #1462, opened 2022-02-11
Re-implemented how PYTHON_MIN_VERSION_HEX is computed

### LINKED ISSUE #1438: Python 3.10 breaks build
[FindPython.cmake](https://github.com/SFTtech/openage/blob/master/buildsystem/modules/FindPython.cmake) assumes python minor version is single digit.

sarcXD:
* Removed old implementation
* Using Cmake Math function
* Bit shifting python versions to create PYTHON_MIN_VERSION_HEX  

This PR Fixes #1438


heinezen (dismissed):
Thanks, it works :)

I put your comments from the issue in the file comments, so everyone knows why we need this "hack".

We can merge when you put your mail in copying.md (see CI output).

heinezen (approved):

Issue #1431, opened 2021-11-09
Introduce Type hinting (Python)


heinezen:
**Required skills:** Python

**Difficulty:** Easy

The documentation of the Python module could be improved by adding type hinting to the method and member declarations. [Type hinting](https://docs.python.org/3/library/typing.html) is Python's way of "defining" the type of an input variable (since Python is not strictly typed, this is technically just a recommendation). While useful for documentation purposes, it is also used by IDEs to detect errors or suggesting things for auto-completion.

In practice it looks like this (note the typing hints in the method header):

```python
def list_associates(name: str, friends: list[str]: str, enemies: list[str] = None) -> str:
    output = name + 'has friends' + "".join(friends)

    if enemies:
        output += 'and also enemies' + "".join(enemies)

    return output
```

We can add type hints to each module one at a time. There's no need to do everything at once. I'm sure there also might be a lot of bugs that could be found along the way.

**Further Reading**

* [Type hinting Python docs](https://docs.python.org/3/library/typing.html)
* [openage Python module](https://github.com/SFTtech/openage/tree/master/openage)

simonsan:
MokeyType can help with that :)

https://github.com/Instagram/MonkeyType

PR #1464, opened 2022-02-24
Python type hints

heinezen:
Resolves https://github.com/SFTtech/openage/issues/1431

This adds type hints for all Python modules (excluding Cython). It also introduces some smaller changes to the game version detection (by using an immutable dataclass instance instead of a tuple for the game version object).

The version of type hints used here requires bumping the required Python version to Python 3.9.


TheJJ (approved):
o-∑

PR #1467, opened 2022-04-01
301 Wrong content

rdeprera:
Link https://bintray.com/simonsan/openage-packages/openage-linux-releases doing a `301 HTTP` redirect


TheJJ:
thx :)

PR #1477, opened 2022-06-20
Update media_convert.md

icefog123:
replace ori game id with GAME_ID  in SteamCmd.sh, because different version game such as AOC and DEII has its different ids.


TheJJ:
good idea, but we should then write down the appids of the different versions clearly in some table/list/...
can you add that please?

icefog123:
@TheJJ  OK，I will do it. I will close this pull request and open a new one. 
And add something such as below in the same this media_convert.md file. Because I think only this file at the steamCmd
line need  some game id.


Age of Empires(2013): 221380
Age of Empires: Definitive Edition: 1017900
Age of Empires II Rise of the Rajas:  488060
Age of Empires II: Definitive Edition: 813780

PR #1479, opened 2022-06-26
Update media_convert.md

icefog123:
1、use Game_ID  token to replace different real game id
2、add a game version with its Game_ID table


heinezen (approved):
Great :)

PR #1507, opened 2023-04-26
`imHex` pattern for SLP files

heinezen:
Pattern file for the [imHex](https://imhex.werwolv.net/) editor @ https://github.com/WerWolv/ImHex

Works for SLPs up until version 3.0 (i.e. formats used before the DE releases). Not really optimized at the moment, but that's something to solve for later ¯\\\_(ツ)_/¯.

Example


TheJJ (approved):
great!

PR #1513, opened 2023-05-13
Angle Drawing + Modpack Loading

heinezen:
- [x] Render different angles
  - [x] Calculate angle direction from waypoints
  - [x] Mirrored angles
  - [x] Angle curve in `Position` component
- [x] Renderer
  - [x] Uniform buffers
  - [x] Documentation adjustment
  - [x] Handle missing assets
- [x] Refactoring
  - [x] Rename `engine::Engine` to `gamestate::GameSmulation`
  - [x] Move event loop into game simulation loop
  - [x] Give clock its own time loop
  - [x] Cleanup
- [x] Modpack loading

Other:
- Fix texture cache not working
- Fix highDPI scaling
- Use math constants from C++20
- Reorder uniform settings


heinezen:
Resolved by https://github.com/SFTtech/openage/pull/1515

Issue #1377, opened 2021-03-29
New Gamestate: Drafts


heinezen:
Collection of drafts and notes about the new gamestate. The information does not follow a particular structure yet, so some things might not be understanddable right away. Don't hesitate to ask questions in that case.

* Entity: https://hedgedoc.rbg.tum.de/bhIBok2YR8uiFhevqTCWJQ?view
* Components: https://pad.stusta.de/p/openage-components
* Systems: https://pad.stusta.de/p/openage-systems
* Activities: https://pad.stusta.de/p/openage-activity-system
* Game control: https://pad.stusta.de/p/openage-game-control

heinezen:
Managers visualization:

![grafik](https://user-images.githubusercontent.com/6852422/120297026-fc797c80-c2c8-11eb-94ae-532d60e4f9b7.png)

PR #1515, opened 2023-06-12
Gamestate Loop

heinezen:
Implement the core components of the gamestate loop

- [x] Game entity
  - [X] Create game entity from nyan `GameEntity`
  - [X] Create components from nyan `Ability`
  - [X] Initialize internal components in spawner 
     - [x] Position
     - [x] Ownership
     - [x] Command Queue
  - [x] Pass animations to render entity
    - [X] Pass `std::string` instead of `util::Path`
    - [x] Differentiate between absolute and relative paths (see [file referencing](https://github.com/SFTtech/openage/blob/master/doc/media/openage/file_referencing.md))
    - [x] Move current animation path outside of `GameEntity`
- [x] Basic nyan API interface
  - [x] Ability
    - [x] Fetch properties
      - [x] Animations
      - [x] Sounds
      - [x] Diplo Stances
  - [x] Player Setup (civs)
- [x] Components
  - [x] Idle
  - [x] Live
  - [x] Move
  - [x] Turn
- [x] Activity system (https://github.com/SFTtech/openage/issues/1377#issuecomment-851960340)
  - [x] Node graph
  - [x] Systems
    - [x] Idle
    - [x] Move
  - [x] Activity Component
  - [x] Event Canceling


libopenage/gamestate/event/spawn_entity.cpp
```diff
@@ -71,28 +71,28 @@ void SpawnEntityHandler::invoke(openage::event::EventLoop & /* loop */,
...
-	entity_pos->set_angle(time + 3, coord::phys_angle_t::from_int(180));
-	entity_pos->set_angle(time + 4, coord::phys_angle_t::from_int(135));
-	entity_pos->set_angle(time + 5, coord::phys_angle_t::from_int(90));
-	entity_pos->set_angle(time + 6, coord::phys_angle_t::from_int(45));
-	entity_pos->set_angle(time + 7, coord::phys_angle_t::from_int(0));
-	entity_pos->set_angle(time + 8, coord::phys_angle_t::from_int(315));
-
-	auto test_pos = std::make_shared<component::Position>(this->loop, pos, time);
+
+	if (entity->has_component(component::component_t::MOVE)) {
```
JonasBorchelt, line 76: What do you think about this code? 
```
if (entity->has_component(component::component_t::MOVE)) {
    std::vector<coord::phys3_delta> positions = {
        {0, 1, 0},
        {1, 2, 0},
        {2, 2, 0},
        {3, 1, 0},
        {3, 0, 0},
        {2, -1, 0},
        {1, -1, 0},
        {0, 0, 0} // Assuming this is equivalent to pos
    };

    std::vector<int> angles = {
        279,
        225,
        180,
        135,
        90,
        45,
        0,
        315
    };

    for (int i = 0; i < positions.size(); i++) {
        entity_pos->set_position(time + i + 1, pos + positions[i]);
        entity_pos->set_angle(time + i + 1, coord::phys_angle_t::from_int(angles[i]));
    }
}
```

heinezen: Oh hey, thanks for checking out the code :D

Given that the current code is just test code and soon to be removed, it probably doesn't really matter how it looks. Your solution looks more maintainable though, so if this were an engine demo, it would be the better solution.

TheJJ (approved):

PR #1516, opened 2023-06-16
improvement: replace (un)likely macro with C++20 attribute

### LINKED ISSUE #1514: Replace likely macro with C++20 attribute
We currently use gcc's `__builtin_expect()` macro aliased as `unlikely` and `likely` for static branch prediction. Primarily, this optimizes checks for parser errors. This optimizes performance for code paths which are only taken under very specific circumstances.

Example:

```cpp
if (unlikely(not this->value.exists())) {
    throw InternalError{"fetched nonexisting value of member"};
}
```

Since C++20, the standard supports [attributes](https://en.cppreference.com/w/cpp/language/attributes/likely) that do the same thing. Additionally, they are compiler agnostic, so they should also work in clang or msvc.  They should also be more readable since they don't directly wrap the condition.

Basically the code gets changed to this:

```cpp
if (not this->value.exists()) [[unlikely]] {
    throw InternalError{"fetched nonexisting value of member"};
}
```

If you want to discover the codebase or really like mundane tasks, then this is an issue for you :)

----
Also take note of the same issue in the nyan repository: https://github.com/SFTtech/nyan/issues/105

tarunsamanta2k20:
resolves #1514

### DIFF
--- a/libopenage/datastructure/pairing_heap.h
+++ b/libopenage/datastructure/pairing_heap.h
@@ -256,8 +256,8 @@ public:
 		// 1. link root children pairwise, last node may be alone
 		element_t first_pair = nullptr;
 		element_t previous_pair = nullptr;
-
-		while (unlikely(current_sibling != nullptr)) {
+		
+		while (current_sibling != nullptr) [[unlikely]] {
 			element_t link0 = current_sibling;
 			element_t link1 = current_sibling->next_sibling;
 
@@ -627,7 +627,7 @@ protected:
 	 * insert a node into the heap.
 	 */
 	void root_insert(const element_t &node) {
-		if (unlikely(this->root_node == nullptr)) {
+		if (this->root_node == nullptr) [[unlikely]] {
 			this->root_node = node;
 		} else {
 			this->root_node = this->root_node->link_with(node);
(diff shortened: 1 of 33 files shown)


libopenage/datastructure/pairing_heap.h
```diff
@@ -256,8 +256,8 @@ class PairingHeap final {
 		// 1. link root children pairwise, last node may be alone
 		element_t first_pair = nullptr;
 		element_t previous_pair = nullptr;
-
-		while (unlikely(current_sibling != nullptr)) {
+
```
heinezen, line 259: fix trailing whitespace here

heinezen (changes requested):
Looks good so far :)

Now you only need to pass the sanity checks.

* Update the copyright years. You can run `python3 -m buildsystem.codecompliance --all --fix` in the repo to let the buildsystem do this for you.
* Add you name/username/commit email to [copying.md](https://github.com/SFTtech/openage/blob/master/copying.md)

heinezen:
Also, you can remove the macros in line 52 and 55 in the file below since they're not needed anymore:

https://github.com/SFTtech/openage/blob/c3d7394655f7491ea0fa73ac015d007d3420e88b/libopenage/util/compiler.h#L52-L55

heinezen (approved):
Great!

PR #1517, opened 2023-06-18
improvement: replace likely macro with C++20 attribute

derekfrogget:
resolves the second half of https://github.com/SFTtech/openage/issues/1514


heinezen (approved):
Looking good!

Issue #1344, opened 2020-12-18
More informative `Exception`/Error messages in converter


heinezen:
**Required skills:** Python

**Difficulty:** Medium

In AoE2's `.dat` format most associations and assignments of properties are done by IDs (e.g. unit has ability with ID X). The openage converter uses these IDs to lookup the associated openage API property and then map the values from AoE2's `.dat` structure  to the corresponding API object's member values. In short, every property from AoE2 needs to be manually mapped to an openage API property. As such, most of the runtime errors are actually lookup errors (usually Python's `KeyError`) that occur when an AoE2 property was not mapped to an openage API property.

The goal of this task is to make these errors more informative by catching the generic `KeyError` from Python and re-raising it with a better message. For example, we can improve the error message by specifying:
  * Context of the property (i.e. the property is: ability, resource, unit, tech, ...)
  * Data type that is looked up (function, object, string, other ID, ...)
    
**Example**

generic message: 

```
File "openage/convert/processor/conversion/de2/tech_subprocessor.py", line 276, in resource_modify_effect
    upgrade_func = DE2TechSubprocessor.upgrade_resource_funcs[resource_id]
KeyError: 208
```

better message:

```
File "openage/convert/processor/conversion/de2/tech_subprocessor.py", line 276, in resource_modify_effect
    upgrade_func = DE2TechSubprocessor.upgrade_resource_funcs[resource_id]
KeyError: No subprocessor function found for handling upgrade of civ resource: 208
```

Use try-except statement where these lookups could be thrown. Remember to use `raise ... from` to preserve the stack trace of the initial `KeyError`.

**Further reading**:

* [Converter code](https://github.com/SFTtech/openage/blob/master/openage/convert/)
* [Converter architecture](https://github.com/SFTtech/openage/blob/master/doc/code/converter/architecture_overview.md)
* [Converter workflow](https://github.com/SFTtech/openage/blob/master/doc/code/converter/workflow.md)

duanqn:
When you say 

> most of the runtime errors are actually lookup errors

Are those errors code defects, or just because the user does not have the correct AoE2 assets?

heinezen:
Most of them are code defects or missing implementations, e.g.
* For somehing like tech effects, the converter gets the effect ID and looks up a corresponding handler function in a dict. If the ID is not found, the `KeyError` should be caught and raise a `NotImplementedError`.
* KeyErrors when looking up unit ID or tech IDs are almost always the converters fault
[...]

PR #1519, opened 2023-06-30
Fix Github squash resulting in wrong commit mail + `pylint` complaints

heinezen:
Squashing https://github.com/SFTtech/openage/pull/1517 using Github's builtin squash merge resulted in a Github commit mail being used for whatever reason. This means the wrong commit mail also bypassed the CI checks.

Note to self: Only squash manually and don't trust Github.

### DIFF
--- a/buildsystem/codecompliance/__main__.py
+++ b/buildsystem/codecompliance/__main__.py
@@ -29,9 +29,9 @@ def parse_args():
                      help=("check whether all git authors are in copying.md. "
                            "repo must be a git repository."))
     cli.add_argument("--cppstyle", action="store_true",
-                     help=("check the cpp code style"))
+                     help="check the cpp code style")
     cli.add_argument("--cython", action="store_true",
-                     help=("check if cython is turned off"))
+                     help="check if cython is turned off")
     cli.add_argument("--headerguards", action="store_true",
                      help="check all header guards")
     cli.add_argument("--legal", action="store_true",
@@ -40,7 +40,7 @@ def parse_args():
                      help=("check whether files in the repo have the "
                            "correct access bits (-> 0644) "))
     cli.add_argument("--pylint", action="store_true",
-                     help=("run pylint on the python code"))
+                     help="run pylint on the python code")
     cli.add_argument("--pystyle", action="store_true",
                      help=("check whether the python code complies with "
                            "(a selected subset of) pep8."))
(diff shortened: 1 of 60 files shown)


buildsystem/codecompliance/__main__.py
```diff
@@ -140,7 +140,7 @@ def get_changed_files(gitref):
         file_list = subprocess.check_output(invocation)
 
     except subprocess.CalledProcessError as exc:
-        raise Exception(
+        raise SyntaxError(
```
TheJJ, line 143: raising the general is exception is ok to crash the program imo.

etc/pylintrc
```diff
@@ -321,4 +321,4 @@ exclude-protected=_asdict,_fields,_replace,_source,_make
 
 # Exceptions that will emit a warning when being caught. Defaults to
 # "Exception"
-overgeneral-exceptions=Exception
+overgeneral-exceptions=builtins.Exception
```
TheJJ, line 324: i'd favor disabling this - when I throw a general exception it's to crash the program.

openage/codegen/coord.py
```diff
@@ -49,11 +49,11 @@ def format_members(formatstring, join_with=", "):
...
-        template_dict = dict(
-            members=member_list,
-            formatted_members=format_members,
-            camelcase="".join(member.title() for member in member_list),
-        )
+        template_dict = {
+            "members": member_list,
+            "formatted_members": format_members,
+            "camelcase": "".join(member.title() for member in member_list),
+        }
```
TheJJ, line 56: what, why is the functional syntax no longer ok for pylint?

TheJJ (changes requested):
nice! but i wouldn't change all the exception types intending to crash the program.
having them as specific exc types might lure the user to try to catch and handle them.

etc/pylintrc
heinezen: I think it has some merit. The reasoning behind the pylint complaint is that it can result in programmers being incentivised to catch the general `Exception` instead of letting the program crash. Using one of the more specific errors narrows that down and prevents accidental failures in that case. There's also the built-in `RuntimeError` that gets the closest to "please crash".

In the long run, we should prefer implementing our own errors anyway (see https://github.com/SFTtech/openage/issues/1344).

buildsystem/codecompliance/__main__.py
heinezen: I'll change it to `RuntimeError`, see below why.

openage/codegen/coord.py
heinezen: Literals are much faster than the constructors in recent Python versions. 

https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/use-dict-literal.html

TheJJ (approved):

PR #1520, opened 2023-07-12
Small renderer optimizations

heinezen:
Optimize some of the functions that are called very often.


TheJJ (approved):
nice, frametiming as a dedicated class without curves is surely an improvement :)

PR #1548, opened 2023-09-20
Build instructions for Debian 12 issues

mantador:


okias:
I would recommend rather to fix the build-system paths to look in the right place :)

mantador:
Would be nice but honestly I don't know cmake well enough. Also according to an old conversation with @heinezen on the matrix channel, seems to be a Debian-only issue.

doc/building.md
```diff
@@ -121,6 +121,9 @@ Make sure you have all the dependencies installed.
 - Make sure you have everything from the [dependency list](#dependencies)
 - Select the compiler and mode: see `./configure --help`
   - Linux etc: `./configure`
+    - On Debian you might get an error saying that it couldn't find SDL2 library. This happens because the CMAKE prefix and SDL2 path are not set correctly.
+      The solution is to append at the end of the `./configure` command the cmake variables for both the prefix and SDL2 path, like so:
+      - `./configure <your configure opts> -- -DCMAKE_PREFIX_PATH=/usr -DSDL2_DIR=/usr/include/SDL2` (you can use `find` to look for the correct paths)
```
heinezen, line 126: Could you put this into the troubleshooting file instead? :)

https://github.com/SFTtech/openage/blob/master/doc/troubleshooting.md

heinezen (changes requested):

heinezen (approved):

PR #1551, opened 2023-09-23
Add CLI parameter for running without graphics

### LINKED ISSUE #1547: CLI parameter for engine start mode
**Required Skills:** Python, some C++

**Difficulty**: Easy

The engine start mode should be configurable via the CLI. There are only 2 relevant start mode right now:
- full: starts game simulation with graphics
- headless: starts game simulation without graphics (mainly for MP servers)

The parameters should be defined in the Python CLI and then passed through to the engine instantiation.

Tasks:
- [ ] Define parameters in Python for submodules `main` and `game`
- [ ] Pass engine mode to C++ invocation `main_arguments` struct
- [ ] In C++, pass engine mode to `Engine` constructor

tslocum:
Resolves #1547.

### DIFF
--- a/copying.md
+++ b/copying.md
@@ -146,6 +146,7 @@ _the openage authors_ are:
 | Derek Frogget               | FoggyLight                  | fro22003 à byui dawt edu                          |
 | Martin                      | Starman                     | mstarman à seznam dawt cz                         |
 | Zoltán Ács                  | zoli111                     | acszoltan111 à gmail dawt com                     |
+| Trevor Slocum               | tslocum                     | trevor à rocket9labs dawt com                     |
 
 If you're a first-time committer, add yourself to the above list. This is not
 just for legal reasons, but also to keep an overview of all those nicknames.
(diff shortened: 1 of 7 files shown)


TheJJ (commented):
thanks! an improvement idea: there may be more modes in the future, so instead of single flags, i think having a --mode=headless or --mode=full would be more future-proof :)

heinezen (changes requested):
Looks good to me :)

The only thing missing is that you have to add your info to https://github.com/SFTtech/openage/blob/master/copying.md . Otherwise it'd be good to go.

You can try to implement @TheJJ 's suggestion. That would be a bit more tricky but more future proof if we add more modes.

tslocum:
I have added my info.  I agree that expanding this is more future-proof, but will leave that to someone else when the need arises.  Thank you both.

heinezen:
@tslocum Alright :)

Your info doesn't match up with the commit mail, so please change that and we can merge.

heinezen:
Also maybe do a rebase fro master while you are at it :D

tslocum:
Gladly.  I think this is all fixed up now.

heinezen (approved):

PR #1552, opened 2023-09-24
Rename module 'openage' as 'engine', output binary 'openage'

### LINKED ISSUE #1532: Rename `run` binary
We could have a more sensible name for the final binary. Maybe `openage` or `engine` are better names. We just have to make sure the names don't conflict with other files.

tslocum:
Resolves #1532.


TheJJ (commented):
this directory is the name of the main python module, so i don't like to rename it (its what will be installed to /use/lib/python3/site-packages/$name).

when the game is properly installed, the executable name is openage anyway, so this is somthing relevant only for the in-repo build. maybe a simpler solution is to rename "run" to "launcher" or "openage-dev"? having the right py module name outweights the development executable name.

heinezen (commented):
Yeah the `openage` python module shouldn't be renamed and it technically doesn't even contain the core engine, so this change would be more confusing than helpful. I think `engine` for the executable would be fine although we might find an even better name.

(Changes with a lot of impact like this are better discussed in the issue/in chat first btw :) Not because we are against you coming up with ideas, but because there's a chance that you do a lot of work for nothing)

tslocum:
Okay, thanks.  The issue mentioned a few potential names so I thought that had already been decided.  That all makes sense.

heinezen:
@tslocum No problem. Sometimes our issues are just reminders for things we think of during chat discussions, so they are not always thought out as much :D

PR #1559, opened 2023-10-06
Fix an inappropriate test expression to remove a logical short circuit

munahaf:
In file: processor.py, the comparison of Collection length creates a logical short circuit. The length of a collection is always greater than or equal to zero. So testing that a length is less than zero is always false.

I suggested that the Collection length comparison should be done without creating a logical short circuit. 

Sponsorship and Support:

This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.

### DIFF
--- a/copying.md
+++ b/copying.md
@@ -147,6 +147,7 @@ _the openage authors_ are:
 | Martin                      | Starman                     | mstarman à seznam dawt cz                         |
 | Zoltán Ács                  | zoli111                     | acszoltan111 à gmail dawt com                     |
 | Trevor Slocum               | tslocum                     | trevor à rocket9labs dawt com                     |
+| Munawar Hafiz               | munahaf                     | munawar dawt hafiz à gmail dawt com               |
 
 If you're a first-time committer, add yourself to the above list. This is not
 just for legal reasons, but also to keep an overview of all those nicknames.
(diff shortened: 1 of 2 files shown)


heinezen:
Thanks for the fix. Can you do the following before we merge this?

* You need to add your name to the [copying.md](https://github.com/SFTtech/openage/blob/master/copying.md) file
* Can you change the check to `len(uograde_effects) == 0`? While this is considered unpythonic by some, it makes the check more readable.

TheJJ (approved):

TheJJ:
the commit for your authorship isn't pushed to this pull request's branch yet.

heinezen:
Do you need help with any of this?

heinezen:
I've cherry picked your commit and added it to the correct branch. It should be mergeable now.

TheJJ (approved):

Issue #1557, opened 2023-10-03
Asset conversion issue, AttributeError: 'NoneType' object has no attribute 'readable'


FabioLolix:
Hello, using the program build from the latest commit on Arch Linux I wasn't able to convert the assets from 3 steam editions

I had only Age2HD installed, after the first try I performed a file check from steam, then downloaded the other 2 games but didn't work 

I'm missing something?

```
(1) ~/.steam/steam/steamapps/common/Age2HD
(2) ~/.steam/steam/steamapps/common/AoE2DE
(3) ~/.steam/steam/steamapps/common/AoEDE
```

```
$ openage convert --force --source-dir /home/fabio/.steam/steam/steamapps/common/Age2HD
INFO [py] Looking for compatible games to convert...
INFO [py] Compatible game edition detected:
INFO [py]  * Age of Empires 2: HD Edition
INFO [py] converting metadata
INFO [py] using cached wrapper: /tmp/HDEDITION_empires2_x1_p1.dat.pickle
INFO [py] Starting conversion...
INFO [py] Extracting Genie data...
INFO [py] Creating API-like objects...
INFO [py] Linking API-like objects...
INFO [py] Generating auxiliary objects...
INFO [py] Creating nyan objects...
INFO [py] Creating requests for media export...
INFO [py] Starting export...
INFO [py] Dumping info file...
INFO [py] Dumping data files...
INFO [py] Exporting media files...
INFO [py] -- Exporting graphics files...
INFO [py] -- Exporting terrain files...
INFO [py] -- Exporting sound files...
INFO [py] Dumping metadata files...
INFO [py] asset conversion complete; asset version: 7
Traceback (most recent call last):
  File "/usr/sbin/openage", line 20, in <module>
    main()
  File "/usr/lib/python3.11/site-packages/openage/__main__.py", line 171, in main
    return args.entrypoint(args, cli.error)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/openage/convert/main.py", line 263, in main
    convert_assets(outdir, args, srcdir)
  File "/usr/lib/python3.11/site-packages/openage/convert/main.py", line 129, in convert_assets
    with asset_locations_path.open("a") as file_obj:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/openage/util/fslike/path.py", line 133, in open
[...]

PR #1561, opened 2023-10-06
Make asset location cache file write more robust

heinezen:
Fixes https://github.com/SFTtech/openage/issues/1557

* Raise `IOError` if we get an invalid handle on open
* Skip asset location write if the file cannot be accessed

### DIFF
--- a/openage/convert/main.py
+++ b/openage/convert/main.py
@@ -9,6 +9,8 @@ from __future__ import annotations
 from datetime import datetime
 import typing
 
+from ..log import info, warn
+
 from ..util.fslike.directory import CaseIgnoringDirectory
 from ..util.fslike.wrapper import (DirectoryCreator,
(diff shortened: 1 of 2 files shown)


openage/convert/main.py
```diff
@@ -126,11 +126,16 @@ def flag(name):
...
+            with asset_locations_path.open("a") as file_obj:
+                if len(prev_srcdirs) > 0:
+                    file_obj.write("\n")
 
-            file_obj.write(used_asset_path)
+                file_obj.write(used_asset_path)
+
+        except IOError:
+            # cache file cannot be accessed, skip writing
+            pass
```
TheJJ, line 138: such things should produce a log message so one can debug it easier :)

heinezen: good idea ^^

TheJJ (approved):

Issue #1569, opened 2023-10-08
Error launching game, uncaught exception(s)


FabioLolix:
Hello, running `$ openage game --modpacks hd_base` I have 2 different uncaught exception between gcc and clang build

GCC: `std::exception of type std::length_error: basic_string::_M_create`

Clang: `std::exception of type std::bad_alloc: std::bad_alloc`

***
GCC build:
```
$ openage game --modpacks hd_base
INFO [py] launching openage v0.4.1-1528-gaae16d373
INFO [py] compiled by GNU 13.2.1
INFO [py] Modpack directory engine not found
INFO [py] openage nyan API modpack not found
INFO [py] Dumping info file...
INFO [py] Dumping data files...
INFO [py] Found modpack engine
INFO [py] Found modpack hd_base
INFO loading configuration files...
INFO [py] loading config file [Union(<openage.util.fslike.union.Union object at 0x7f5ed94213d0>.root @ (b'cfg',), <openage.util.fslike.union.Union object at 0x7f5ed9423290>.root @ (b'assets',))]:cfg/keybinds.oac...
INFO launching engine with root directoryPath(Union(<openage.util.fslike.union.Union object at 0x7f5ed94213d0>.root @ (b'cfg',), <openage.util.fslike.union.Union object at 0x7f5ed9423290>.root @ (b'assets',)):)
INFO loading configuration files...
INFO [py] loading config file [Union(<openage.util.fslike.union.Union object at 0x7f5ed94213d0>.root @ (b'cfg',), <openage.util.fslike.union.Union object at 0x7f5ed9423290>.root @ (b'assets',))]:cfg/keybinds.oac...
INFO Found modpack: engine
INFO Found modpack: hd_base
INFO Created game simulation
INFO Activated modpack: engine
INFO Activated modpack: hd_base
INFO Created 3 threads (24 available)
INFO [T1] Time loop started
INFO [T2] Presenter: Launching subsystems...
INFO [T2] Presenter: Initializing graphics subsystems...

FATAL: terminate has been called

uncaught exception

std::exception of type std::length_error: basic_string::_M_create

current stack:

Traceback (most recent call last):
  File ?, in ? [0x7f5edc8aa9eb]
  File ?, in ? [0x7f5edb6e1943]
  File ?, in ? [0x7f5ed90829f9]
  File ?, in openage::gamestate::GameSimulation::run()+0x2f [0x7f5ed90ed65f]
[...]

PR #1574, opened 2023-10-10
Fix error on `engine` modpack load

heinezen:
Resolves https://github.com/SFTtech/openage/issues/1569

### DIFF
--- a/openage/convert/service/init/api_export_required.py
+++ b/openage/convert/service/init/api_export_required.py
@@ -16,7 +16,7 @@ if typing.TYPE_CHECKING:
     from openage.util.fslike.union import UnionPath
 
 
-CURRENT_API_VERSION = "0.4.0"
+CURRENT_API_VERSION = "0.5.1"
 
 
 def api_export_required(asset_dir: UnionPath) -> bool:
--- a/openage/convert/tool/api_export.py
+++ b/openage/convert/tool/api_export.py
@@ -33,7 +33,7 @@ def main(args, error):
     del error  # unused
 
     path = Union().root
-    path.mount(Directory(args.dir))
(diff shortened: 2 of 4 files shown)


openage/convert/service/init/api_export_required.py
```diff
@@ -16,7 +16,7 @@
     from openage.util.fslike.union import UnionPath
 
 
-CURRENT_API_VERSION = "0.4.0"
+CURRENT_API_VERSION = "0.5.1"
```
TheJJ, line 19: we should import the global version id here, no?

openage/convert/tool/api_export.py
```diff
@@ -76,7 +76,7 @@ def create_modpack() -> Modpack:
 
     mod_def = modpack.get_info()
 
-    mod_def.set_info("engine", "0.4.0", repo="openage")
+    mod_def.set_info("engine", "0.5.1", versionstr="0.4.0", repo="openage")
```
TheJJ, line 79: here as well. should be a single source of truth?

openage/convert/service/init/api_export_required.py
heinezen: This was more of a test because I thought the API modpack might have been wrongly generated and caused the error in https://github.com/SFTtech/openage/issues/1569 . But this was probably not the issue.

I don't think the modpack version should be bumped with every engine release btw. but maybe if there's a new version of the engine modpack, it should have the same version no as the current engine version.

heinezen: i might close this PR and think of something else to fix the issue

heinezen:
Supreceded by https://github.com/SFTtech/openage/pull/1574

PR #1576, opened 2023-10-10
Allow non-annotated tags from git for setting version

heinezen:
Fixes incorrect version number being displayed in source builds.

### DIFF
--- a/buildsystem/DetectProjectVersion.cmake
+++ b/buildsystem/DetectProjectVersion.cmake
@@ -12,7 +12,7 @@ endif()
 if(IS_DIRECTORY "${CMAKE_SOURCE_DIR}/.git")
 	message(STATUS "Set PROJECT_VERSION from git.")
 	execute_process(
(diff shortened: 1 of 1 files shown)


TheJJ (commented):
https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-describe.html

```
--tags 

Instead of using only the annotated tags, use any tag found in refs/tags namespace. This option enables matching a lightweight (non-annotated) tag. 
```

so it used non-annonated tags before, now only annotated tags?

heinezen:
> so it used non-annonated tags before, now only annotated tags?

No it's the other way around

Before this PR: Only annotated tags were used (resulting in `0.4.1` being used as the version number by the build system since this is the last annotated PR)
With this PR: All tags (annotated or not) are considered, i.e. the correct version `0.5.1` should be used

TheJJ (approved):

Issue #1540, opened 2023-09-17
Temporary file/directory support


heinezen:
**Required Skills:** Python or C++

**Difficulty**: Easy

It would be nice if we could create temporary files/directories in our current file system abstraction API. We currently only use temporary structures in the converter, but they could be useful in other places, e.g. for unzipping archives or on-disk caching.

Creation of temporary files could be implemented in our [Python file API](https://github.com/SFTtech/openage/tree/master/openage/util/fslike) or our [C++ file API](https://github.com/SFTtech/openage/tree/master/libopenage/util/fslike) or in both. The Python-side implementation is more preferrable at the moment, since the converter is written Python and temporary files already have an active use case there.

The easiest way to implement support probably is to add a method that calls the underlying system libraries for temporary file/directory creation:

- Python: `tempfile` module
- C++: `std::filesystem` and `std::tmpfile`

The retrieved path is then wrapped in our `Path`/`Directory` classes and returned by the method. 

Tasks:
- [ ] Create temp file/dir and wrap them in openage file system API
- [ ] Delete temp file/dir explicitely when object is destroyed
- [ ] Use new implementation in code (currently only in converter)
- [ ] (optional) Add method for checking if a file/dir is temporary

**Further Reading**

* [openage Python file API](https://github.com/SFTtech/openage/tree/master/openage/util/fslike) 
* [openage C++ file API](https://github.com/SFTtech/openage/tree/master/libopenage/util/fslike)
* [Python `tempfile` module](https://docs.python.org/3/library/tempfile.html)
* [C++ std::filesystem::temp_directory_path](https://en.cppreference.com/w/cpp/filesystem/temp_directory_path)
* [C++ std::tmpfile](https://en.cppreference.com/w/cpp/io/c/tmpfile)

Ashhar-24:
Hey @heinezen , I'm interested in working with this issue.
I'm new to open source but I've beginner level experience in working with `Python` and `C++` both.
Can you please assign me this issue.

And also can you please elaborate more about the tasks that exactly what I've to do.
Thanks for your cooperation :)

heinezen:
@Ashhar-24 Hey, I assigned you to to the issue nw :)

Most of what we want from the task should be in the issue description already. The general goal is to make temporary file creation/access more easier than it is right now.

For a start, I think you should try out these steps:

[...]

PR #1585, opened 2023-10-19
created Temporary file/directory support #1540

PankajKrana:
# Temporary file/directory support #1540

## Description

This branch adds support for creating temporary files and directories in our current file system abstraction API. The implementation is based on the `tempfile` module in Python and the `std::filesystem` 

## Tasks

- [x] Create temp file/dir and wrap them in openage file system API
- [x] Delete temp file/dir explicitly when object is destroyed
- [x] Use new implementation in code (currently only in converter)
- [ ] Add method for checking if a file/dir is temporary

## Progress

- Created new branch "Temporary file/directory support #1540"
- Implemented creation of temporary files and directories
- Implemented deletion of temporary files and directories
- Updated converter to use new implementation
- Currently working on adding method for checking if a file/dir is temporary

I hope this helps! Let me know if there's anything else I can do for you.

### DIFF
--- a/openage/codegen/Temporary file/directory support.py
+++ b/openage/codegen/Temporary file/directory support.py
@@ -1,0 +1,30 @@
+import tempfile
+import shutil
+import os
+
+class FileSystemAPI:
+    def __init__(self):
+        pass
+
+    def create_temp_file(self):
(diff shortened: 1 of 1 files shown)


PankajKrana:
Some tests fail so I'm going to try to change the code .

Issue #1543, opened 2023-09-17
Fixed point math functions


heinezen:
**Required Skills:** C++

**Difficulty**: Easy

We should add additional math operations to our fixed-point implementation, so that all calculations can be done purely deterministic.

The [`fpm`](https://github.com/MikeLankamp/fpm) library already has these operations and could be used as an inspiration.

Interesting for us:
- [ ] sqrt
- [ ] min
- [ ] max
- [ ] abs
- [ ] hypot
- [ ] atan2

Nice to have:
- [ ] sin
- [ ] cos
- [ ] tan
- [ ] pow
- [ ] log

**Further Reading**

- [openage fixed-point implementation](https://github.com/SFTtech/openage/blob/master/libopenage/util/fixed_point.h)
- [fpm library documentation](https://github.com/MikeLankamp/fpm/blob/master/docs/index.md)

mingjuu:
Can I try to solve this problem?

heinezen:
@mingjuu You can :) 

I'll assign you right now and if you have any questions, feel free to ask here or in chat.

AyiStar:
Hello, I am new to the project and wish I could join you. Can you assign this issue to me so that I can start from here?

heinezen:
@AyiStar sure :)

AyiStar:
@heinezen Thank! I will touch you if I have any questions.

adamresemius:
Hello,

Is there anything left to do here that I can contribute to?

heinezen:
@adamresemius Yes, the unticked boxes in the issue description is what's not implemented at the moment.

Zain-Balkhi:
I think I know how to implement the other functions in fpm. I have pow already on mine.

firednerve:
Hey, I'm a beginner and could try my hand at implementing sqrt (and possibly the others after) if it isn't being worked on right now

heinezen:
@firednerve You are welcome to go for it :)

firednerve:
Thanks! I’ll ask if I have any questions.

anatriaslabella:
Hi! Is this issue still open to complete the "nice to have" operations? I do not see the implementations added and would like to try it out as someone who is new to open source!

PRDeving:
Hi there, sin, cos and tan is already implemented in [this PR](https://github.com/SFTtech/openage/pull/1753/files)

PRDeving:
already created the missing functions and added missing test cases.
[...]

PR #1587, opened 2023-10-20
Add `atan2` to `FixedPoint`

askastitva:
- #1543 
- Added `atan2` as a member function to `FixedPoint`
- Added `atan2` function for `FixedPoint` in std namepace
- Added a test case for `atan2`


heinezen (approved):
Thank you, although this doesn't address the intention of the original issue. The purpose of https://github.com/SFTtech/openage/issues/1543 is to completely remove all floating point calculations in-between, so it's always deterministic. This includes the use of `std::` functions.

heinezen (changes requested):
You need to change the copyright year for `fixed_point_test.cpp` (see the CI output)

heinezen (approved):

PR #1594, opened 2023-10-30
Issue #1540, first two tasks

Ashhar-24:
Completed following two `tasks` of issue #1540 
- Create temp file/dir and wrap them in openage file system API
- Delete temp file/dir explicitely when object is destroyed


openage/util/fslike/path.py
```diff
@@ -330,3 +348,16 @@ def mount(self, pathobj, priority=0) -> NoReturn:
         # pylint: disable=no-self-use,unused-argument
         # TODO: https://github.com/PyCQA/pylint/issues/2329
         raise PermissionError("Do not call mount on Path instances!")
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.delete()
```
heinezen, line 356: Be careful here. If you do this with a regular file (and not a temp file), this will also delete it!

You should either check if this is a tempfile or use `__del__` instead.

heinezen (changes requested):

Ashhar-24 (commented):
Applied the changes as asked. @heinezen

openage/util/fslike/path.py
```diff
@@ -32,7 +33,7 @@ class Path:
     # lower.
     # pylint: disable=too-many-public-methods
 
-    def __init__(self, fsobj, parts=None):
+    def __init__(self, fsobj, parts=None, is_temp=False):
```
heinezen, line 36: This parameter should not be set here since it implies that you could create the Path via `__init__` as a tempfile. This is not what we want (people should use the static method you implemented)

openage/util/fslike/path.py
```diff
@@ -74,6 +77,20 @@ def __repr__(self):
...
+        temp_file = tempfile.NamedTemporaryFile(delete=False)
+        # Wrap the temporary file path in a Path object and return it
+        return Path(temp_file.name)
+
+    @staticmethod
+    def create_temp_dir():
+        # Create a temporary directory using tempfile.mkdtemp
+        temp_dir = tempfile.mkdtemp()
+        # Wrap the temporary directory path in a Path object and return it
+        return Path(temp_dir)
```
heinezen, line 93: you need to set `is_temp` to `True` in this method

heinezen (changes requested):

openage/util/fslike/path.py
Ashhar-24: Changed

heinezen: no, please don't do this via `__init__`. You should set it to true in the `create_temp_file` and `create_temp_dir` functions.

Ashhar-24: like should I pass them into the functions or just add them inside the function?

heinezen: you don't need to pass anything ^^

you need to create the path, then set the member, then return it.

```
path = Path(..)
path.is_temp = True
return path
```

Ashhar-24: oh.... :'(
Changed

heinezen: Hmm... still not right. Maybe my comments are a bit confusing :D

You should return the `path` that you created. Right now, you are still returning a different path.

openage/util/fslike/path.py
```diff
@@ -63,6 +64,8 @@ def __init__(self, fsobj, parts=None):
 
         self.fsobj = fsobj
 
+        self.is_temp = is_temp
```
heinezen, line 67: This should be set to `False` by default

Ashhar-24: image

heinezen (approved):

Issue #1596, opened 2023-11-02
OpenGL error when launching on Wayland


zoli111:
I get this error when I try to launch the demo on Wayland session.
OS: Arch
Compositor: Mutter (I can test with others too)

`./run game --modpacks hd_base`

```
WARN [T2] QEGLPlatformContext: Failed to create context: 3009

FATAL: terminate has been called

uncaught exception

Traceback (most recent call last):
  File ?, in ? [0x7facb48aa9eb]
  File ?, in ? [0x7facb4ae1943]
  File ?, in ? [0x7facb52ea683]
  File ?, in ? [0x7facb52ea65e]
  File ?, in openage::presenter::Presenter::run(bool)+0x6e [0x7facb54060d6]
  File ?, in openage::presenter::Presenter::init_graphics(bool)+0xd8 [0x7facb5404c72]
  File ?, in openage::renderer::Window::create(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned long, unsigned long, bool)+0x4b [0x7facb5430429]
  File ?, in std::shared_ptr<openage::renderer::opengl::GlWindow>::shared_ptr<std::allocator<void>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned long&, unsigned long&, bool&>(std::_Sp_alloc_shared_tag<std::allocator<void> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned long&, unsigned long&, bool&)+0x28 [0x7facb5430f4c]
  File ?, in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::__shared_count<openage::renderer::opengl::GlWindow, std::allocator<void>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned long&, unsigned long&, bool&>(openage::renderer::opengl::GlWindow*&, std::_Sp_alloc_shared_tag<std::allocator<void> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned long&, unsigned long&, bool&)+0x61 [0x7facb5430eb7]
[...]

PR #1598, opened 2023-11-04
Dynamically fetch default FBO

heinezen:
Changes the default FBO ID from being hardcoded to fetching it every frame for the current context. This is necessary for platforms where the default FBO is not fixed to `0` and changes on a frame-by-frame basis, e.g. when using Wayland.

Fixes https://github.com/SFTtech/openage/issues/1596

### DIFF
--- a/libopenage/renderer/opengl/context.cpp
+++ b/libopenage/renderer/opengl/context.cpp
@@ -156,6 +156,10 @@ std::shared_ptr<QOpenGLContext> GlContext::get_raw_context() const {
 	return this->gl_context;
 }
 
+GLuint GlContext::get_default_framebuffer_id() {
+	return this->gl_context->defaultFramebufferObject();
+}
+
 gl_context_spec GlContext::get_specs() const {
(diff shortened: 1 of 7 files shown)


TheJJ (approved):
i wonder why the framebuffer ids of the window may change - if it's because of dpi changes etc then it would have resize events at arbitrary time points, even after getting the current fbuf id. so i'd expect one has to acquire it and release it during draw?

heinezen:
@TheJJ I think this is a side effect of isolation. Looks like the used framebuffer is associated with the window surface and changes to the surface result in new framebuffers being used.

> On some platforms (for instance, iOS) the default framebuffer object depends on the surface being rendered to, and might be different from 0. Thus, instead of calling glBindFramebuffer(0), you should call glBindFramebuffer(ctx->defaultFramebufferObject()) if you want your application to work across different Qt platforms.

https://doc.qt.io/qt-6/qopenglcontext.html#defaultFramebufferObject

PR #1599, opened 2023-11-12
PR templates

heinezen:
Adds pull request templates to the repo.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATES/pull_request_template.md
+++ b/.github/PULL_REQUEST_TEMPLATES/pull_request_template.md
@@ -1,0 +1,13 @@
+### Merge Checklist
+
+<!-- Complete all tasks below to make sure that we can merge your pull request. -->
+<!-- If you are an experienced contributor, you may delete this section ;-) -->
+
+- [ ] I have read the [contribution guide](doc/contributing.md)
+- [ ] I have added my info to [copying.md](copying.md) (only first time contributors)
+- [ ] I have run `make checkall` and fixed all mentioned problems
+
+
+### Description
+
+<!-- Describe the changes this pull request makes to the code base. -->
--- a/.github/PULL_REQUEST_TEMPLATES/release_template.md
+++ b/.github/PULL_REQUEST_TEMPLATES/release_template.md
@@ -1,0 +1,7 @@
+### Checklist
+
+- [ ] Changelog
+    - [ ] Release date
+    - [ ] Version number in title
+    - [ ] Full commit log
+- [ ] Bump version in `openage_version`


.github/PULL_REQUEST_TEMPLATES/pull_request_template.md
```diff
@@ -0,0 +1,19 @@
...
+### Description
+
+<!-- Describe the changes this pull request makes to the code base. -->
+
+
+### Related Issue
+
+<!-- If this pull request is related to an issue, replace LINK with the link to the GitHub issue. -->
+
+Resolves LINK
```
TheJJ, line 19: people will probably ignore this - maybe it should be a comment on how to add related issues. and the relation shouldn't be an extra heading imo.

.github/PULL_REQUEST_TEMPLATES/pull_request_template.md
```diff
@@ -0,0 +1,19 @@
+### Merge Checklist
+
+<!-- Complete all tasks below to make sure that we can merge your pull request. -->
+
+- [ ] I have read the [contribution guide](doc/contributing.md)
+- [ ] I have added my info to [copying.md](copying.md) (only first time contributors)
+- [ ] I have run `make checkall` and fixed all mentioned problems
```
TheJJ, line 7: this will show up in every pull request as a 3-point checklist. not sure we want that. maybe have those in a comment so they will hopefully be read, but they don't show up in every pull request text box?

heinezen: I would say this is the entire point of having templates :D First time contributors should ideally see this every time they view their PR, instead of just on creation (and then promptly forget it). And we can say "plz do the stuff in the checklist" instead of "do this and do that".

Regular contributors like us could just delete the entire thing if they want to anyway...

.github/PULL_REQUEST_TEMPLATES/pull_request_template.md
heinezen: Yeah I think I will remove this entirely and we just do this ourselves when we see the PR.

.github/PULL_REQUEST_TEMPLATES/pull_request_template.md
TheJJ: Then how about adding "If you are an experienced contributor, you may just delete these checkpoints"? :)

heinezen: Ok I'm fine with that!

Done

TheJJ (approved):

PR #1602, opened 2023-11-23
Use block indent for aligning after open brackets

heinezen:
Finnaly let's us auto-format the enclosing bracket on a new line.

Before:

```cpp
this->display_shader->new_mom(
    "mass",
    1337,
    "hide",
    false);
```

After:

```cpp
this->display_shader->new_mom(
    "mass",
    1337,
    "hide",
    false
);
```


TheJJ (approved):
yeey

Issue #1542, opened 2023-09-17
Allow game entity to reference arbitrary data in render updates


heinezen:
In our engine architecture, game entities in the simulation send requests to the renderer to be drawn. The render requests also include updates on what to draw (i.e. which animation) and where (i.e. the entity position). Updates are only send when this info changes.

Currently, we only send information in updates that is always required by the renderer, e.g. the animation path and the entity position. However, we may want to send other information that is only used in specific cases such as the *player ID* when drawing player colors. Furthermore, the renderer also needs to be able to handle this info and transform it into something that can be passed to the GPU.

Tasks:
- [ ] Game simulation needs to be able to lookup which data can be send to the renderer (and in which situations)
- [ ] Renderer can transform optional data into something usable

PR #1603, opened 2023-11-23
Drag Selection

heinezen:
Depends on https://github.com/SFTtech/openage/pull/1550

Adds drag selection to the possible inputs.

- [x] Input System
    - [x] Game controller 
        - [x] Store drag selection start in game controller
        - [x] Create drag selection events with mouse down/up
        - [x] Send drag select events to gamestate
    - [x]  HUD controller
        - [x] Store drag selection start in game controller
        - [x] Convert drag select events to HUD renderer instructions
- [x] Renderer
    - [x] HUD renderer
        - [x] drag selection shader
        - [x] HUD objects
        - [x] render pass
- [x] Gamestate
    - [X] Drag selection event
    - [x] Use `Selectable` ability


doc/code/renderer/level2.md
```diff
@@ -4,21 +4,21 @@ High-level renderer for transforming data from the gamestate to render objects f
...
 
 Every stage has its own subrenderer that manages a `RenderPass` from the level 1 renderer and updates it with `Renderable`s created using update information from the gamestate. Stages also store the vertex and fragment shaders used for drawing the renderable objects.
 
-There are currently 5 stages in the level 2 rendering pipeline:
+There are currently 6 stages in the level 2 rendering pipeline:
 
 1. `SkyboxRenderer`: Draws the background behind the terrain (as a single color).
 1. `TerrainRenderer`: Draws the terrain. Terrains are handled as textured 3D meshes.
 1. `WorldRenderer`: Draws animations and sprites for units/buildings and other 2D ingame objects.
+1. `HudRenderer`: Draws "Head-Up Display" elements like health bars, selection boxes, and others.
```
TheJJ, line 21: maybe the health bars should also be zoomed in, thus be rendered part of the world renderer?

libopenage/gamestate/event/drag_select.cpp
```diff
@@ -0,0 +1,100 @@
...
+	float right = std::max(drag_start.x(), drag_end.x());
+
+	log::log(DBG << "Drag select rectangle (NDC):");
+	log::log(DBG << "\tTop: " << top);
+	log::log(DBG << "\tBottom: " << bottom);
+	log::log(DBG << "\tLeft: " << left);
+	log::log(DBG << "\tRight: " << right);
+
+	std::vector<entity_id_t> selected;
+	for (auto &entity : gstate->get_game_entities()) {
```
TheJJ, line 55: for now ok, but we definitely need an index structure from position -> entity.

libopenage/gamestate/event/drag_select.cpp
```diff
@@ -0,0 +1,100 @@
...
+			// only select entities of the controlled player
+			continue;
+		}
+
+		// Get the position of the entity in the viewport
+		auto pos = std::dynamic_pointer_cast<component::Position>(
+			entity.second->get_component(component::component_t::POSITION));
+		auto current_pos = pos->get_positions().get(time);
+		auto world_pos = current_pos.to_scene3().to_world_space();
+		Eigen::Vector4f clip_pos = cam_matrix * Eigen::Vector4f{world_pos.x(), world_pos.y(), world_pos.z(), 1};
```
TheJJ, line 75: why not use our coord system which has abstractions for converting these coordinate systems?

libopenage/gamestate/event/drag_select.h
```diff
@@ -0,0 +1,46 @@
...
+class Event;
+class State;
+} // namespace event
+
+namespace gamestate::event {
+
+/**
+ * Drag select game entities.
+ */
+class DragSelectHandler : public openage::event::OnceEventHandler {
```
TheJJ, line 26: i'm unsure if unit selections should be done with the event system - they could fit equally well just in the presenter layer? there things like the hud drawing pass take place anyway, and selecting units has nothing to do with the game simulation itself

assets/shaders/hud_drag_select.frag.glsl
```diff
@@ -0,0 +1,10 @@
+#version 330
+
+// Color of the drag rectangle
+uniform vec4 in_col;
+
+layout(location=0) out vec4 out_col;
+
+void main() {
+	out_col = in_col;
```
TheJJ, line 9: we could draw the selection box as a big quad, where we only color the outline when the frag position is less than the desired outline width.

libopenage/input/input_context.cpp
```diff
@@ -15,35 +15,53 @@ const std::string &InputContext::get_id() {
...
 }
 
 void InputContext::bind(const event_class &cl, const input_action act) {
-	this->by_class.emplace(std::make_pair(cl, act));
+	std::vector<input_action> actions{act};
+	this->by_class.emplace(std::make_pair(cl, actions));
+}
+
+void InputContext::bind(const Event &ev, const std::vector<input_action> &&acts) {
+	this->by_event.emplace(std::make_pair(ev, acts));
```
TheJJ, line 53: move acts

libopenage/input/input_context.cpp
```diff
@@ -15,35 +15,53 @@ const std::string &InputContext::get_id() {
...
+}
+
+void InputContext::bind(const event_class &cl, const std::vector<input_action> &&acts) {
+	this->by_class.emplace(std::make_pair(cl, acts));
```
TheJJ, line 57: move acts

libopenage/renderer/stages/hud/render_stage.cpp
```diff
@@ -0,0 +1,128 @@
...
+                               const std::shared_ptr<renderer::camera::Camera> &camera,
+                               const util::Path &shaderdir,
+                               const std::shared_ptr<renderer::resources::AssetManager> &asset_manager,
+                               const std::shared_ptr<time::Clock> clock) :
+	renderer{renderer},
+	camera{camera},
+	asset_manager{asset_manager},
+	drag_object{nullptr},
+	clock{clock} {
+	renderer::opengl::GlContext::check_error();
```
TheJJ, line 30: leftover?

TheJJ (commented):
wooh!

doc/code/renderer/level2.md
heinezen: we could also zoom in in the HUD renderer as long as the shaders there get access to the camera matrices :)

libopenage/gamestate/event/drag_select.cpp
heinezen: true that

libopenage/gamestate/event/drag_select.h
heinezen: The problem is that selection depends a lot on the game simulation properties of entities, e.g.

* is the entity selectable?
* is the entity owned by the player using the drag select rectangle?
* can the entity be put in a selection queue? (buildings cannot for example)

Most of that is not available in the presenter layer, so making it an event makes more sense to me.

assets/shaders/hud_drag_select.frag.glsl
heinezen: That would be nice... for a future PR ;)

libopenage/input/input_context.cpp
heinezen: fixed

libopenage/input/input_context.cpp
heinezen: fixed

libopenage/renderer/stages/hud/render_stage.cpp
heinezen: No, this is done at the start of every render stage initialization (also for world, terrain, etc.).

libopenage/gamestate/event/drag_select.cpp
heinezen: This calculation is not implemented in the coord system right now. And I don't know if it should be, since the float calculations could be problematic (not in this case but for gamestate events). In the future, we probably need to either think about how much of a problem this is for each conversion and implement the conversions using fixed-point math if we have to.

libopenage/gamestate/event/drag_select.h
TheJJ: and if we pass along a property like "selectable" to the presenter as a result of the simulation? selection with the mouse is a ui mechanic, which i think should be (apart from the flag) away from the simulation. i mean it doesn't hurt here but maybe later on we'll encounter a disadvantage or have trouble moving it then?

libopenage/renderer/stages/hud/render_stage.cpp
TheJJ: k, but once we have other active renderer backends we need to geberalize^^

TheJJ (commented):
if we don't move the selection to hud layer now, it's likely still doable later. advantage of simulating it: we easily capture it in recordings. (although tracing the camera pos in presenter has to be done anyway)

libopenage/gamestate/event/drag_select.h
heinezen: I think this would be much harder now than leaving the current implementation intact. We would need to potentially mirror a lot of data and implement filters for deciding what game entity is selectable, since this has to be determined when the selection is made. And then, we might have to rewrite it a lot anyway to support index structures, selection areas, etc.

I would leave it as-is for now and revisit the selection mechanics when we do https://github.com/SFTtech/openage/issues/1542

TheJJ:
has merge conflicts :)
another design point: when units die during being selected, intuitively managing and activating the selected unit's status seems better placed at the hud layer. otherwise the dieing must influence the selectability as simulated unit state, instead of the selectability being a flag & then-visible unit data drawn in the hud

libopenage/gamestate/event/drag_select.h
TheJJ: ok, then let's leave it and hope changing it for an improved method with index structures won't be too complicated :)

TheJJ (dismissed):
ok, let's do it like this for now.

heinezen:
merge conflicts are fixed :)

heinezen:
> another design point: when units die during being selected, intuitively managing and activating the selected unit's status seems better placed at the hud layer. otherwise the dieing must influence the selectability as simulated unit state, instead of the selectability being a flag & then-visible unit data drawn in the hud

Also relevant for the input layer since the selection queue for the controller resides there. This back propagation of the unit state is a bit hard to do across thread boundaries. We can probably manage that via another interface to the presenter though.

TheJJ (approved):

Issue #1544, opened 2023-09-17
Fixed point math constants


heinezen:
**Required Skills:** C++

**Difficulty**: Easy

It would be nice to have common math constants like `pi` defined using our fixed-point implementation, so that they can be used in fixed-point calculations with deterministic outcome. Floating point constants are already implemented [here](https://github.com/SFTtech/openage/blob/master/libopenage/util/math_constants.h).

Tasks:
- [ ] Define existing math constants also as fixed point

**Further Reading**

- [openage math constants](https://github.com/SFTtech/openage/blob/master/libopenage/util/math_constants.h)
- [fpm libary math constants](https://github.com/MikeLankamp/fpm/blob/22bab8ff3abe0acfb00bc20853ed6b74e9959fd7/include/fpm/fixed.hpp#L116)

ArcXeon:
I am interested on working on this issue. Can you please assign it to me?

AyiStar:
@heinezen I opened a WIP PR for adding two fixed point constants, and would be happy to discuss about the implementation and add more constants if everything is ok.

heinezen:
@AyiStar I've looked at it yesterday and no complaints so far :)

PR #1607, opened 2023-12-01
Add math constants corresbonding to `util/math_constants.h` for FixedPoint

AyiStar:
To add pure fixed-point math functions (issue #1543), I found it would be better add some math constants (issue #1544). So here is a WIP PR which adds first two important constants, E and PI. The solution is as follows.

(1) In "fixed_point.h", I add two staic functions which return E and PI, respectively, referrng to the implementation of the fpm library.

(2) The from_fixedpoint() function is fixed to avoid the situation of narrowing warning/error furing compilation when the fractional_bits  of created FixedPoint constant object is smaller than the predefined one, i.e., 61.

(3) By the way, I also amend the comment doc of safeshift function.

(4) In "fixed_point_test.cpp", I add some tests which utilizes the constants defined in "math_constant.h" for comparison. I also test constant under different precisions (see TestTypeShort in the tests).

### DIFF
--- a/libopenage/util/fixed_point.h
+++ b/libopenage/util/fixed_point.h
@@ -170,14 +170,71 @@ public:
 	}
 
 	/**
-	 * Constants
+	 * Math constants represented in FixedPoint
 	 */
+	// naming, definition and value are kept compatible with `math_constants.h`
 	static constexpr FixedPoint e() {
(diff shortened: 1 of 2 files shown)


heinezen (approved):
Pretty nice, thank you!

PR #1608, opened 2023-12-02
nyan data API v0.4.1

heinezen:
Small update that makes activities configurable via the nyan data API.

PR #1609, opened 2023-12-07
Configurable Activities

heinezen:
Makes the activity system configurable with nyan.

- [x] Converter: Configure activities as abilities for game entities
  - [x] Load new nyan data API objects (see https://github.com/SFTtech/openage/pull/1608)
  - [x] Assign default activity to game entity definitions in modpack
  - [x] Export new `engine` modpack version
- [x] Activity system
  - [x] Define condition/event function per branch for `XorGate`/`XorEventGate`
  - [x] Add built-in condition functions
  - [x] Add built-in event primer functions
- [x] Entity creation
  - [x] Get data from activity objects from nyan API
  - [x] Create activity from nyan definition
  - [x] Cache created activities


libopenage/gamestate/api/activity.cpp
```diff
@@ -0,0 +1,116 @@
...
+		std::shared_ptr<nyan::View> db_view = node.get_view();
+		return {db_view->get_object(next->get_name())};
+	}
+	case activity::node_t::START: {
+		auto next = node.get<nyan::ObjectValue>("Start.next");
+		std::shared_ptr<nyan::View> db_view = node.get_view();
+		return {db_view->get_object(next->get_name())};
+	}
+	// 1+ next nodes
+	case activity::node_t::XOR_GATE: {
```
TheJJ, line 51: shouldn't this be part of the xorgate node implementation instead of this central place? might be a suitable application of a std variant and stdvisit.

libopenage/gamestate/api/activity.cpp
```diff
@@ -0,0 +1,116 @@
...
+		return next_nodes;
+	}
+	default:
+		throw Error(MSG(err) << "Unknown activity node type.");
+	}
+}
+
+system::system_id_t APIActivityNode::get_system_id(const nyan::Object &ability_node) {
+	auto ability = ability_node.get<nyan::ObjectValue>("Ability.ability");
+	std::shared_ptr<nyan::View> db_view = ability_node.get_view();
```
TheJJ, line 88: unused

openage/convert/tool/api_export.py
```diff
@@ -76,7 +76,7 @@ def create_modpack() -> Modpack:
 
     mod_def = modpack.get_info()
 
-    mod_def.set_info("engine", "0.4.0", repo="openage")
+    mod_def.set_info("engine", "0.4.1", "0.4.1", repo="openage")
```
TheJJ, line 79: maybe kwarg the duplicate arg

TheJJ (commented):
progress!

libopenage/gamestate/activity/event/command_in_queue.cpp
```diff
@@ -0,0 +1,35 @@
...
+	openage::event::EventHandler::param_map::map_t params{{"next", next_id}}; // move->get_id();
+	auto ev = loop->create_event("game.process_command",
+	                             entity->get_manager(),
+	                             state,
+	                             // event is not executed until a command is available
+	                             std::numeric_limits<time::time_t>::max(),
+	                             params);
+	auto entity_queue = std::dynamic_pointer_cast<component::CommandQueue>(
+		entity->get_component(component::component_t::COMMANDQUEUE));
+	auto &queue = const_cast<curve::Queue<std::shared_ptr<component::command::Command>> &>(entity_queue->get_queue());
```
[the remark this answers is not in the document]
heinezen: I think we can get away with making the getter non-const ^^

libopenage/gamestate/activity/event/command_in_queue.cpp
```diff
@@ -0,0 +1,35 @@
...
+                                                               const std::shared_ptr<gamestate::GameEntity> &entity,
+                                                               const std::shared_ptr<openage::event::EventLoop> &loop,
+                                                               const std::shared_ptr<gamestate::GameState> &state,
+                                                               size_t next_id) {
+	openage::event::EventHandler::param_map::map_t params{{"next", next_id}}; // move->get_id();
+	auto ev = loop->create_event("game.process_command",
+	                             entity->get_manager(),
+	                             state,
+	                             // event is not executed until a command is available
+	                             std::numeric_limits<time::time_t>::max(),
```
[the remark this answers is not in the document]
heinezen: yeah that is a nice idea.

libopenage/curve/queue.h
```diff
@@ -183,24 +184,74 @@ class Queue : public event::EventEntity {
...
+const T Queue<T>::pop_front(const time::time_t &time) {
+	if (this->empty(time)) [[unlikely]] {
+		throw Error{MSG(err) << "Tried accessing front at "
+		                     << time << " but queue is empty."};
+	}
+
+	// search for the first element that is after the given time
+	auto it = this->container.begin();
+	++it;
+	while (it->time() <= time and it != this->container.end()) {
```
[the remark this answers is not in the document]
heinezen: I swapped the search order around, so we start from the end of the queue. Then we probably don't even need an index structure, since in normal gameplay, the front element is more likely to be closer to the end of the container due to how we insert elements:

    for (auto it = this->container.begin(); it != this->container.end(); ++it) {
		if (time < it->time()) {
			insertion_point = this->container.insert(it, queue_wrapper(time, e));
			break;
		}
	}
	if (insertion_point == this->container.end()) {
		insertion_point = this->container.insert(this->container.end(),
		                                         queue_wrapper(time, e));
	}

libopenage/curve/queue.h
```diff
@@ -183,24 +184,74 @@ class Queue : public event::EventEntity {
...
+		++it;
+	}
+
+	auto to = it->time();
+	auto from = time;
+
+	// get the last element inserted before the given time
+	--it;
+	auto val = it->value;
```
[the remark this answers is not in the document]
heinezen: fixed

libopenage/gamestate/api/activity.cpp
heinezen: I don't want to integrate nyan too deep into the gamestate. That's why these helper classes exist as a kind of layer between nyan and openage. I think this also makes it easier to maintain the API when we change it.

openage/convert/tool/api_export.py
heinezen: I expect that these version numbers don't stay equal for too long though. The left version number (i.e. the modpack version) could change when we add new import aliases or reformat the indentation. Changes like these shouldn't affect the nyan API version number on the right.

libopenage/gamestate/activity/event/command_in_queue.cpp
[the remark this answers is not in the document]
TheJJ: yea, something like that

libopenage/gamestate/api/activity.cpp
TheJJ: but isn't the xorgate node class the abstraction layer? so instead of having specific snippets here, having them more local to the specific code using it seems more local.

openage/convert/tool/api_export.py
TheJJ: yea that's why we could name it engineversion=..., modpackversion=... so one sees these are different versions

libopenage/gamestate/api/activity.cpp
heinezen: I don't think this would make more sense though. The function here is only used to traverse the activity graph defined in the modpack, so it doesn't have much to do with activities at runtime. It's only use is to create the runtime activity from the nyan definition. XorGate and the other nodes have their own internal `next` function for traversing the runtime graph.

openage/convert/tool/api_export.py
heinezen: I've now clarified this in code.

TheJJ (approved):

PR #1621, opened 2024-02-11
Added pow function fpm

Zain-Balkhi:
Implemented a potential working version of pow using fixed point math. No way to test because project won't run.

### DIFF
--- a/libopenage/util/fixed_point.h
+++ b/libopenage/util/fixed_point.h
@@ -590,6 +590,19 @@ constexpr double hypot(openage::util::FixedPoint<I, F> x, openage::util::FixedPo
 }
 
 template <typename I, unsigned F>
+constexpr openage::util::FixedPoint<I, F> pow(openage::util::FixedPoint<I, F> base, double exponent) {
+	// Handle special cases: 0^0, 1^x, x^0
+	if (base == 0.0 && exponent == 0.0) {
+		return openage::util::FixedPoint<I, F>::quiet_NaN(); // Undefined; return NaN
+	} else if (base == 1.0 || exponent == 0.0) {
+		return openage::util::FixedPoint<I, F>::from_raw_value(1);
+	}
+	double log_base = std::log(base.get_raw_value());
+	double result = exponent * log_base;
+	return openage::util::FixedPoint<I, F>::from_raw_value(std::exp(result));
+}
+
+template <typename I, unsigned F>
 struct hash<openage::util::FixedPoint<I, F>> {
 	constexpr size_t operator()(const openage::util::FixedPoint<I, F> &n) const {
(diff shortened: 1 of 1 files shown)


libopenage/util/fixed_point.h
```diff
@@ -589,6 +589,19 @@ constexpr double hypot(openage::util::FixedPoint<I, F> x, openage::util::FixedPo
 	return x.hypot(y);
 }
 
+template <typename I, unsigned F>
+constexpr openage::util::FixedPoint<I, F> pow(openage::util::FixedPoint<I, F> base, double exponent) {
+	// Handle special cases: 0^0, 1^x, x^0
+	if (base == 0.0 && exponent == 0.0) {
+		return openage::util::FixedPoint<I, F>::quiet_NaN(); // Undefined; return NaN
+	} else if (base == 1.0 || exponent == 0.0) {
```
heinezen, line 597: why compare to doubles in the if conditions when we could compre to fixed point values instead?

libopenage/util/fixed_point.h
```diff
@@ -589,6 +589,19 @@ constexpr double hypot(openage::util::FixedPoint<I, F> x, openage::util::FixedPo
...
+		return openage::util::FixedPoint<I, F>::from_raw_value(1);
+	}
+	double log_base = std::log(base.get_raw_value());
+	double result = exponent * log_base;
```
heinezen, line 601: The pure fixed-point math functions should avoid any floating point arithmetic. Avoiding the floating point rounding erros is the point of using fp math after all.

cmake-build-debug/.cmake/api/v1/query/cache-v2
heinezen, line 1: this file would need to be removed

cmake-build-debug/.cmake/api/v1/query/cmakeFiles-v1
heinezen, line 1: and this one

cmake-build-debug/.cmake/api/v1/query/codemodel-v2
heinezen, line 1: and this one

cmake-build-debug/.cmake/api/v1/query/toolchains-v1
heinezen, line 1: and this one

heinezen (commented):
Hey thanks for the contribution, but your code still contains floating point math which is not what https://github.com/SFTtech/openage/issues/1543 was about. What we want are fuctions that implement these operations by only using fixed-point arithmetic. 

I don't know if you want to give this another try but we can't really use this code right now.

heinezen:
I'm closing this because this is not what we are looking for.

Issue #1624, opened 2024-02-15
Asset conversion stalls on media conversion


brett-hamilton:
## System Info
- OS: Windows 11
- Python: 3.12.2
---
Following the build instructions from `doc/build_instructions/windows_msvc.md`, when running the first Python command the process stalled out for 30+ minutes before I manually stopped it.

### Command
`py -m openage --add-dll-search-path <path> --add-dll-search-path <path> main`

### Output Before Stopping
```
INFO [py] Looking for compatible games to convert...
INFO [py] Compatible game edition detected:
INFO [py]  * Age of Empires 2: The Conqueror's Trial Version
INFO [py] converting metadata
INFO [py] Finished metadata read (10.50 seconds)
INFO [py] Starting conversion...
INFO [py] Extracting Genie data...
INFO [py] Creating API-like objects...
INFO [py] Linking API-like objects...
INFO [py] Generating auxiliary objects...
INFO [py] Creating nyan objects...
INFO [py] Creating requests for media export...
INFO [py] Finished data conversion (0.61 seconds)
INFO [py] Starting export...
INFO [py] Dumping info file...
INFO [py] Dumping data files...
INFO [py] Exporting media files...
INFO [py] -- Exporting graphics files...
-- Files done:   0/275
```

### Error Output After Stopping
```
INFO [py] -- Exporting graphics files...
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "<openage directory>\build\openage\__main__.py", line 178, in <module>
    sys.exit(main())
             ^^^^^^
  File "<openage directory>\build\openage\__main__.py", line 171, in main
    return args.entrypoint(args, cli.error)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<openage directory>\build\openage\main\main.py", line 82, in main
    convert_assets(asset_path, args)
  File "<openage directory>\build\openage\convert\main.py", line 119, in convert_assets
    convert(args)
  File "<openage directory>\build\openage\convert\tool\driver.py", line 36, in convert
    convert_metadata(args)
  File "<openage directory>\build\openage\convert\tool\driver.py", line 105, in convert_metadata
    ModpackExporter.export(modpack, args)
  File "<openage directory>\build\openage\convert\processor\export\modpack_exporter.py", line 60, in export
[...]

PR #1628, opened 2024-02-23
Always use CWD as DLL search path on Windows

heinezen:
The command working directory is now always used as a search path, even if `--add-dll-search-path` is not specified in the CLI.

This should fix or mitigate issues like https://github.com/SFTtech/openage/issues/1624 .


TheJJ (commented):
phew, this may have unexpected side effects especially when the game is installed and one launches from a cwd where a suitable dll is in. shouldn't this rather be the path wherethe main executable is in?

heinezen:
Didn't thought about that and I see your point. I changed it, so that the path of the directory of `__main__.py` is in is used. Would this be okay?

TheJJ:
guess so, if the dll is in that dir? let's see if it helps :)

TheJJ (approved):

PR #1631, opened 2024-03-29
feat: added flake

akiross:
Hello,
as discussed on matrix few days ago, I'm proposing this PR to add support for nix.

While possibly not idiomatic, this works and adds also support for nyan.

### DIFF
--- a/doc/build_instructions/nix.md
+++ b/doc/build_instructions/nix.md
@@ -1,0 +1,78 @@
+# Building and developing on Nix based systems
+
+The openage repository is a [nix flake](https://nixos.wiki/wiki/Flakes) that
+allow Nix users to easily build, install and develop openage.
+
+To build openage using nix
+
+1. make sure you have nix with flakes enabled, either by permanent system
+   configuration or by using command line flags, as described on
+   [the wiki](https://nixos.wiki/wiki/Flakes);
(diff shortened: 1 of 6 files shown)


heinezen:
Hey thanks a lot,

I just have one small suggestion: Could you add the nix build to [our build instructions list](/doc/building.md#dependency-installation)? That would make the fact that nix is available as an option more obvious. Also, could you link some info on how to maintain these files inside the docs because I assume it's written down somewhere but we don't know nix well enough to know where to look :D

akiross:
Sure I can! I will add a bunch of comments inside the nix files to aid maintaining them.

Oh, just for info: this was tested on Linux (amd64 NixOS) only, I am not sure other platforms (e.g. OSX) are working, since I cannot test them at the moment.

akiross:
I added some build and dev instructions specific for nix. I updated `building.md` to do what "seems intuitive" (list nix among the dependency section), but it's not just that: this solution is a complete build specification, so it would require a different section in that file altogether. But for the people looking for nix, the current doc might be fine.

I also fixed a couple of things related to the code itself and added comments to the main files (flake.nix and openage.nix).

heinezen (approved):

heinezen:
Thanks a lot! This will make it much easier to update and check out.

PR #1635, opened 2024-04-14
Port curve container types to `std::vector`

heinezen:
So far, we have used `std::list` as the underlying type for curve containers. Lists are nice because they offer constant time insertions and erasure from anywhere in the data structure and have stable iterators, i.e. changing the list doesn't invaliate the iterators.

However, these properties may not be as important for our applications and `std::vector` could be better for performance under these assumptions:

- As elements in the curve are sorted by simulation time, insertions will like be at the end of the container in normal gameplay situations
- Random access for reading elements is likely used more than insertion/removal
- Copying the curve container, e.g. when passing the curve over thread boundaries, would put the the contigious memory layout of `std::vector` at an advantage over `std::list`

This PR changes the underlying container type of `KeyframeContainer` and `Queue` to `std::vector` to show that its usage is possible with minimal changes.

### DIFF
--- a/libopenage/curve/keyframe.h
+++ b/libopenage/curve/keyframe.h
@@ -27,16 +27,26 @@ public:
 	 * New, default-constructed element at the given time
 	 */
 	Keyframe(const time::time_t &time) :
-		time{time} {}
+		timestamp{time} {}
 
 	/**
 	 * New element fron time and value
 	 */
 	Keyframe(const time::time_t &time, const T &value) :
-		time{time},
+		timestamp{time},
 		value{value} {}
 
-	const time::time_t time = time::TIME_MIN;
+	const time::time_t &time() const {
+		return this->timestamp;
+	}
+
+	const T &val() const {
+		return this->value;
+	}
+
+private:
+	time::time_t timestamp = time::TIME_MIN;
+
 	T value = T{};
 };
 
--- a/libopenage/curve/keyframe.h
+++ b/libopenage/curve/keyframe.h
@@ -33,21 +33,40 @@ public:
 	 * New element fron time and value
 	 */
 	Keyframe(const time::time_t &time, const T &value) :
-		timestamp{time},
-		value{value} {}
+		value{value},
+		timestamp{time} {}
 
+	/**
+	 * Get the time of this keyframe.
+	 *
+	 * @return Keyframe time.
+	 */
 	const time::time_t &time() const {
 		return this->timestamp;
 	}
 
+	/**
+	 * Get the value of this keyframe.
+	 *
+	 * @return Keyframe value.
+	 */
 	const T &val() const {
 		return this->value;
 	}
 
-private:
-	time::time_t timestamp = time::TIME_MIN;
-
+public:
+	/**
+	 * Value of the keyframe.
+	 *
+	 * Can be modified by the curve if necessary.
+	 */
 	T value = T{};
+
+private:
+	/**
+	 * Time of the keyframe.
+	 */
+	time::time_t timestamp = time::TIME_MIN;
 };
 
 } // namespace openage::curve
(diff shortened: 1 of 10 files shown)


TheJJ:
how about std::deque? especially since we probably can't retain all events for the whole game and need to either serialize/forget them at earlier times

heinezen:
@TheJJ `std::deque` is not contigous which might not be as great for hot code. The faster removal of elements at the front is nice but I don't know how often this would happen in a standard game compared to the other operations.

TheJJ:
we can test/benchmark when it gets relevant, and we can easily swap it hopefully ^^
the deque isn't fully continuous but i think around 100 elements are, so its a mixture of the list and vector as far as i remember. but yea we can vector it for now i guess

heinezen:
As long as we get rid of the list, it should be an improvement either way :D

TheJJ:
then why not try the deque? :) should definitely be an improvement for the regular use-cases

heinezen:
I've created a new branch in https://github.com/SFTtech/openage/pull/1637 to preserve the changes from this PR. We could still switch to vector in the future :)

heinezen:
After a quick detour with https://github.com/SFTtech/openage/pull/1637 I think using vector is the better choice :D Iterator stability for `std::deque` is only guaranteed for insertion/erasure at the front and end, so caching would become more complicated in many situations. In comparison, the std::vector implementation uses indices which only require simple bounds checks to make sure that the cached index ist still in range.

I also found a [C++ benchmark – std::vector VS std::list VS std::deque](https://baptiste-wicht.com/posts/2012/12/cpp-benchmark-vector-list-deque.html) which compares the standard containers performance. Looks like vector is still not that much slower than deque, even for random inserts.

libopenage/curve/keyframe.h
```diff
@@ -27,16 +27,26 @@ class Keyframe {
...
+	const time::time_t &time() const {
+		return this->timestamp;
+	}
+
+	const T &val() const {
+		return this->value;
+	}
+
+private:
+	time::time_t timestamp = time::TIME_MIN;
```
TheJJ, line 48: why not leave the time as a const member? then no function to access it is required.

libopenage/curve/keyframe.h
```diff
@@ -27,16 +27,26 @@ class Keyframe {
...
+
 	T value = T{};
```
TheJJ, line 50: if one wants to change value in-place it should be no problem for the internal curve consistency. so having a function for it should not be necessary, too.

libopenage/curve/keyframe.h
heinezen: If it is `const` then it is not copy assignable/constructible which becomes a problem for working with vector (or deque). I know I can write a new copy/assignment contructor but making them private looks more clean to me.

libopenage/curve/keyframe.h
heinezen: Currently we are not doing that though, but I can make it public anyway.

libopenage/curve/queue.h
```diff
@@ -208,21 +221,24 @@ class Queue : public event::EventEntity {
...
 	/**
 	 * Get the first alive element inserted at t <= time.
 	 *
 	 * @param time The time to get the element at.
 	 *
-	 * @return Iterator to the first alive element or end() if no such element exists.
+	 * @return Index of the first alive element or end() if no such element exists.
 	 */
-	const_iterator first_alive(const time::time_t &time) const;
+	index_t first_alive(const time::time_t &time) const;
```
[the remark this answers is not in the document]
heinezen: Does that mean we just rename `index_t` to `elem_ref`?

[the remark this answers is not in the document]
TheJJ: yea i guess, if you have a better name go for it :)

heinezen:
It's ready to review again!

TheJJ (approved):

PR #1636, opened 2024-04-15
Coordinate types pretty print

heinezen:
Depends on https://github.com/SFTtech/openage/pull/1633

This just flattens the parent hierarchy to make the parent invisible and directly show the coordinate values.

### DIFF
--- a/etc/gdb_pretty/printers.py
+++ b/etc/gdb_pretty/printers.py
@@ -92,8 +92,8 @@ def printer_regex(regex: str):
     return _register_printer
 
 
-@printer_regex('^openage::coord::((phys|scene)2|(chunk|tile))(_delta)?')
-class CoordNeSePrinter:
+@printer_regex('^openage::coord::(camhud|chunk|input|phys|scene|term|tile|viewport)(2|3)?(_delta)?')
+class CoordPrinter:
     """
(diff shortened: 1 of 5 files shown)


TheJJ (approved):
nice!

maybe we can split this into multiple files already

heinezen:
@TheJJ Wouldn't this require switching away from the decorator method? I would like to keep that for now, since it makes adding new printers pretty easy.

TheJJ:
the decorator can be imported like any regular function, no?

heinezen:
@TheJJ Yes, but to use the decorator, the module with the classes would have to be imported too. And somehow, gdb doesn't seem to like that. At least I get errors when I try that for the .gdbinit file but gdb doesn't print the exact issue. I'm not sure how to solve this yet, so having the printers all in one file is the best solution right now.

TheJJ (approved):
maybe leave a note in the cpp code to remember to update the pretty printer when changing it :)

heinezen:
@TheJJ Done :)

TheJJ (dismissed):

heinezen:
Now it should be good to merge, finally

TheJJ (approved):

PR #1637, opened 2024-04-16
Port curve container types to `std::deque`

heinezen:
Same as https://github.com/SFTtech/openage/pull/1635 but with deque.

### DIFF
--- a/libopenage/curve/base_curve.h
+++ b/libopenage/curve/base_curve.h
@@ -204,7 +204,7 @@ void BaseCurve<T>::set_last(const time::time_t &at, const T &value) {
 	auto hint = this->container.last(at, this->last_element);
 
 	// erase max one same-time value
-	if (hint->time == at) {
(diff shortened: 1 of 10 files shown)


heinezen:
Ah there might be a problem with iterator stablity:

`std::deque`:

> insertion and deletion **at either end of a deque** never invalidates pointers or references to the rest of the elements. 

> `erase():` All iterators and references are invalidated, unless the erased elements are at the end or at the beginning of the container, in which case only the iterators and references to the erased elements are invalidated. The [end()](https://en.cppreference.com/w/cpp/container/deque/end) iterator is also invalidated unless the erased elements are at the beginning of the container and the last element is not erased.

`std::list`: 

> Adding, removing and moving the elements within the list or across several lists does not invalidate the iterators or references. An iterator is invalidated only when the corresponding element is deleted. 

So we cannot properly cache with `std::deque` if the erasure happens in the middle of the queue.

PR #1638, opened 2024-04-20
Turn off vsync for renderer stresstest

heinezen:
We didn't gave any vsync settings in the Qt renderer previously, so I added them again. There's also a new `window_settings` struct that can be extended with flags or other types of configuration data.

The stresstest for the renderer now runs with unlimited FPS by default. It didn't really go under 60 FPS anymore, so this makes the frame dropoff a bit more obvious for now.

### DIFF
--- a/libopenage/presenter/presenter.cpp
+++ b/libopenage/presenter/presenter.cpp
@@ -1,4 +1,4 @@
-// Copyright 2019-2023 the openage authors. See copying.md for legal info.
+// Copyright 2019-2024 the openage authors. See copying.md for legal info.
 
 #include "presenter.h"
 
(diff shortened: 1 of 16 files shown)


libopenage/presenter/presenter.cpp
```diff
@@ -93,8 +93,13 @@ std::shared_ptr<qtgui::GuiApplication> Presenter::init_window_system() {
...
 	log::log(INFO << "Presenter: Initializing graphics subsystems...");
 
+	// Start up rendering framework
 	this->gui_app = this->init_window_system();
-	this->window = renderer::Window::create("openage presenter test", 1024, 768, debug);
+
+	// Window and renderer
+	renderer::window_settings settings;
+	settings.debug = debug;
+	this->window = renderer::Window::create("openage presenter test", 1024, 768, settings);
```
TheJJ, line 102: then the resolution should be part of the settings :)

heinezen: great idea :)

heinezen: done

TheJJ (approved):

PR #1639, opened 2024-04-21
Move subtex scaling to shader

heinezen:
Move the scaling of subtextures (i.e. handling of zoom, animation scaling, etc.) to the world shader.

Camera zoom and viewport size are send via the camera's uniform buffer, so these variables are now requested only once per frame. The world object's have to send an additional `float` as a uniform, but no longer have to calculate the scaling themselves which roughly halves the cycles spent on creating these uniform values.

The PR also fixes a bug in the alignment of uniforms in the uniform buffer that was previously not discovered. Uniforms where not properly aligned to a multiple of their size which resulted in the created buffer being too small and using the wrong offsets. This should now be fixed.

### DIFF
--- a/assets/shaders/terrain.vert.glsl
+++ b/assets/shaders/terrain.vert.glsl
@@ -7,9 +7,17 @@ out vec2 tex_pos;
 
 uniform mat4 model;
 
+// camera parameters for transforming the object position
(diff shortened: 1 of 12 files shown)


TheJJ (approved):
yey

PR #1640, opened 2024-04-21
Fix order of processing input contexts

heinezen:
Fixes the order in which input contexts are searched. The intended order was LIFO (like a stack) but the implemented order was FIFO (like a list).

 ~~This was a nice opportunity to use `std::views::reverse` from the `ranges` library in C++20 :D~~ Disregard that because clang is too old :(

I also added a separate callback list for mouse move actions to the window because mixing them with mouse button actions can be annoying during debugging. This lead me to discover an error in the input demo where some keybindings were not properly assigned which is now also fixed.

### DIFF
--- a/libopenage/presenter/presenter.cpp
+++ b/libopenage/presenter/presenter.cpp
@@ -217,6 +217,9 @@ void Presenter::init_input() {
 		this->input_manager->process(ev);
 	});
 	this->window->add_mouse_button_callback([&](const QMouseEvent &ev) {
+		this->input_manager->process(ev);
+	});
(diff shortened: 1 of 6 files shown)


libopenage/presenter/presenter.cpp
```diff
@@ -220,6 +220,10 @@ void Presenter::init_input() {
 		this->input_manager->set_mouse(ev.position().x(), ev.position().y());
```
TheJJ, line 220: button shouldn't update position?

heinezen: fixed

TheJJ (approved):

PR #1642, opened 2024-04-26
Frustum culling

nikhilghosh75:
Implemented Frustum Culling into the renderer. Currently only applies to the world renderer. The camera has a frustum data structure to allow it to cull objects outside the frustum.

An additional stresstest has been created for the renderer to test out frustum culling.

Additional fixes to loading files on windows that have the special \r character are also included in this pull request.


heinezen:
Nice, I'll take a look over the weekend :)

libopenage/renderer/resources/parser/parse_terrain.cpp
```diff
@@ -200,8 +200,8 @@ TerrainInfo parse_terrain_file(const util::Path &file,
 		})};
 
 	for (auto line : lines) {
-		// Skip empty lines and comments
-		if (line.empty() || line.substr(0, 1) == "#") {
+		// Skip empty lines, lines with carriage returns, and comments
+		if (line.empty() || line.substr(0, 1) == "#" || line[0] == '\r') {
```
heinezen, line 204: and here

libopenage/renderer/resources/parser/parse_texture.cpp
```diff
@@ -56,6 +56,11 @@ std::string parse_imagefile(const std::vector<std::string> &args) {
 	// it should result in an error if wrongly used here.
 
 	// Call substr() to get rid of the quotes
+	// If the line ends in a carriage return, remove it as well
+	if (args[1][args[1].size() - 1] == '\r') {
```
heinezen, line 60: and here

libopenage/renderer/resources/parser/parse_texture.cpp
```diff
@@ -179,8 +184,8 @@ Texture2dInfo parse_texture_file(const util::Path &file) {
 		})};
 
 	for (auto line : lines) {
-		// Skip empty lines and comments
-		if (line.empty() || line.substr(0, 1) == "#") {
+		// Skip empty lines, lines with carriage returns, and comments
+		if (line.empty() || line.substr(0, 1) == "#" || line[0] == '\r') {
```
heinezen, line 188: and here

libopenage/renderer/stages/world/render_stage.cpp
```diff
@@ -58,6 +58,10 @@ void WorldRenderStage::update() {
 	std::unique_lock lock{this->mutex};
 	auto current_time = this->clock->get_real_time();
 	for (auto &obj : this->render_objects) {
+		if (!obj->within_camera_frustum(this->camera)) {
+			continue;
```
heinezen, line 62: This will still render the object btw, it just skips the update :D

I'll have to fix this in a separate PR as it is not possile right now to remove specific objects from the render pass.

heinezen:
I only found a few minor issues with this and a have restructering reques. It already looks quite nice. You also have to rebase and resolve the merge conflicts.

copying.md
```diff
@@ -153,6 +153,7 @@ _the openage authors_ are:
 | Astitva Kamble              | askastitva                  | astitvakamble5 à gmail dawt com                   |
 | Haoyang Bi                  | AyiStar                     | ayistar à outlook dawt com                        |
 | Michael Seibt               | RoboSchmied                 | github à roboschmie dawt de                       |
+| Nikhil Ghosh                | NikhilGhosh75               | nghosh606 à gmail dawt com
```
heinezen, line 156: I think the CI fails because you are missing an `|` at the end of the table. Then the mailmap entry is also not necessary.

TheJJ (commented):
can't we calculate a containing circle for each object on creation and when resizing it? then we only have to check one coordinate/distance for each object to cull it.

heinezen:
I'm gonna take this over to fix all problems and make it mergeable.

TheJJ:
> for every animation sprite and not just per object. And then we also have to apply zoom scaling...

how is the object size calculated currently? from the maximum of all its sprites? and the zooming we need anyway? :)

heinezen:
> how is the object size calculated currently? checking the current sprite sizes every frame? then we could do from the maximum of all its sprites instead during load? and the zooming we need anyway? :)

It's done in the shader. The size is a combination of `scalefactor` (from the animation), `size` (from the texture), and `zoom` (from the camera). Calculating `max_size` from all `size`s at load is probably a good idea. This would then become a bounding box that we can multiply with sscalefactor and zoom in the renderer.

heinezen:
Okay the code should now work for 3D frustum culling. I will now add an optimized version for 2D sprites as @TheJJ suggested and a demo for showing off the functionality.

heinezen:
Everything sshould be ready feature-wise now. It's just documentation for the frustum usage & architecture that is missing.

heinezen:
I disabled frustum culling for now, since it's not that performant right now. Currently, it just skips the uniform update, but checking the frustum is much more expensive than performing the uniform update by itself. To make the frustum culling worth our while, we should also exclude the renderables from the draw calls. This would require us to build the list of renderables every frame which can come with its own drawbacks.

libopenage/renderer/camera/camera.cpp
```diff
@@ -25,14 +25,19 @@ Camera::Camera(const std::shared_ptr<Renderer> &renderer,
...
+	frustum_culling{false} {
 	this->look_at_scene(Eigen::Vector3f(0.0f, 0.0f, 0.0f));
 
 	resources::UBOInput view_input{"view", resources::ubo_input_t::M4F32};
 	resources::UBOInput proj_input{"proj", resources::ubo_input_t::M4F32};
 	auto ubo_info = resources::UniformBufferInfo{resources::ubo_layout_t::STD140, {view_input, proj_input}};
 	this->uniform_buffer = renderer->add_uniform_buffer(ubo_info);
 
+	// Make the frustum slightly bigger than the camera's view to ensure objects on the boundary get rendered
+	float real_zoom = 0.7f * this->default_zoom_ratio * this->zoom;
```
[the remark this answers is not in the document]
heinezen: Fixed by making `Frustum2d` with pixel-perfect visisbility checks.

libopenage/renderer/camera/camera.h
```diff
@@ -30,6 +32,9 @@ static const Eigen::Vector3f cam_direction{
 	-1 * (std::sqrt(6.f) / 4),
 };
 
+static const float near_distance = 0.01f;
+static const float far_distance = 100.0f;
+
```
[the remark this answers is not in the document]
heinezen: Fixed by moving all constants to `renderer/camera/definitions.h`

libopenage/renderer/camera/camera.h
```diff
@@ -272,6 +283,18 @@ class Camera {
...
+	 * Is frustum culling enabled? If true, perform frustum culling.
+	 * If false, all frustum checks will return true
+	 */
+	bool frustum_culling;
+
+	/**
+	 * The frustum used to cull objects
+	 * Will be recalculated regardless of whether frustum culling is enabled
+	 */
+	Frustum frustum;
```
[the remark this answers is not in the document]
heinezen: This is implemented now.

libopenage/renderer/demo/stresstest_1.cpp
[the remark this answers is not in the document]
heinezen: Demo 6 is now a frustum culling demo.

TheJJ (approved):

Issue #1644, opened 2024-05-01
Build failing, unable to find Python3 on Windows 10


C3pa:
Hello!

While trying to build the project on Windows 10, I followed the instructions provided at: https://github.com/SFTtech/openage/blob/master/doc/build_instructions/windows_msvc.md#procedure-for-microsoft-windows-users.

I managed to get all the needed dependencies, I got to the following step: https://github.com/SFTtech/openage/blob/master/doc/build_instructions/windows_msvc.md#building-openage
```cmd
cmake -DCMAKE_TOOLCHAIN_FILE=<vcpkg directory>\scripts\buildsystems\vcpkg.cmake ..
```
This step is failing with:
```cmd
-- Could NOT find Python3 (missing: Python3_LIBRARIES Python3_NumPy_INCLUDE_DIRS Development NumPy Development.Module Development.Embed) (found suitable exact version "3.11.6")
CMake Error at C:/Users/Vlasnik/openage/build/CMakeFiles/CMakeTmp/CMakeLists.txt:32 (target_link_libraries):
  Target "cmTC_ff525" links to:

    Python3::Python

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.



CMake Error at buildsystem/modules/FindPython.cmake:81 (try_compile):
  Failed to generate test project build system.
Call Stack (most recent call first):
  C:/vcpkg/scripts/buildsystems/vcpkg.cmake:859 (_find_package)
  buildsystem/HandlePythonOptions.cmake:6 (find_package)
  CMakeLists.txt:176 (include)


-- Configuring incomplete, errors occurred!
```

Do you have any ideas about what I have missed?

heinezen:
Did you install the development version of Python3?

C3pa:
I installed Python 3.12.3 with these checks ticked in the installer:
![image](https://github.com/SFTtech/openage/assets/41503714/d65d48f3-51b5-4302-9520-fffd11141047)
![image](https://github.com/SFTtech/openage/assets/41503714/4a1711cc-29ea-4011-8493-d44cd3297fca)

Then I ran:
```cmd
pip install cython numpy lz4 toml pillow pygments pyreadline3 mako
```

heinezen:
Do you have a Python 3.11 installation? The error above says

```
(found suitable exact version "3.11.6")
```

C3pa:
[...]

PR #1649, opened 2024-05-17
Enhance DLL search on Windows

heinezen:
Fixes the import errors for the compiled libraries and Python modules on Windows by adding better default search paths. If you build openage from source with the documented instructions, there shouldn't be any Python errors after this PR.

The problem in https://github.com/SFTtech/openage/issues/1644 was that the dependent libraries were no longer found automatically since Python 3.8 and you had to use `--add-dll-search-path` to avoid errors. However, it's not always obvious how this argument should be used. Therefore, openage now sets sane defaults to search for:

- `python.dll`
- `openage.dll`
- `nyan.dll`

Furthermore, it fixes the errors in https://github.com/SFTtech/openage/issues/1624 . These happened because the DLLs are only set *per process*, so subprocesses in `multiprocessing` were not aware about the DLL search paths. This problem is now solved by adding a `DllDirectoryManager` object that can be passed around and then used to add/remove DLL search paths.

### DIFF
--- a/openage/util/CMakeLists.txt
+++ b/openage/util/CMakeLists.txt
@@ -3,6 +3,7 @@ add_py_modules(
 	bytequeue.py
 	context.py
 	decorators.py
+	dll.py
 	files.py
 	fsprinting.py
 	hash.py
--- a/openage/util/dll.py
+++ b/openage/util/dll.py
@@ -13,13 +13,22 @@ DEFAULT_PYTHON_DLL_DIR = os.path.dirname(sys.executable)
 
 # openage.dll locations (relative to this file)
 DEFAULT_OPENAGE_DLL_DIRs = [
+    "../../libopenage/Debug",
+    "../../libopenage/Release",
     "../../libopenage/RelWithDebInfo",
+    "../../libopenage/MinSizeRel",
 ]
 
 # nyan.dll locations (relative to this file)
 DEFAULT_NYAN_DLL_DIRS = [
+    "../../../../nyan/build/nyan/Debug",
+    "../../../../nyan/build/nyan/Release",
     "../../../../nyan/build/nyan/RelWithDebInfo",
+    "../../../../nyan/build/nyan/MinSizeRel",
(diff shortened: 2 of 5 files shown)


openage/util/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ add_py_modules(
 	bytequeue.py
 	context.py
 	decorators.py
+    dll.py
```
C3pa, line 6: Looks like this line is space indented, while the rest of the file uses tabs for indentation.

heinezen: fixed

simonsan:
Wow, what a clusterfuck, I remember when we discovered that. Didn't know it's still leaving its trail! Excellent idea with the DllManager, will probably make it much easier to react to Python's weirdness in that regard in the future. 👍🏽

openage/util/dll.py
```diff
@@ -0,0 +1,122 @@
...
+# nyan.dll locations (relative to this file)
+DEFAULT_NYAN_DLL_DIRS = [
+    "../../../../nyan/build/nyan/Debug",
+    "../../../../nyan/build/nyan/Release",
+    "../../../../nyan/build/nyan/RelWithDebInfo",
+    "../../../../nyan/build/nyan/MinSizeRel",
+    "../../nyan-external/bin/nyan/Debug",
+    "../../nyan-external/bin/nyan/Release",
+    "../../nyan-external/bin/nyan/RelWithDebInfo",
+    "../../nyan-external/bin/nyan/MinSizeRel",
```
TheJJ, line 31: maybe use nyan here, since that would be the name next to the openage repo when somebody clones the nyan repo

heinezen: `nyan-external` is where nyan is downloaded when `-DDOWNLOAD_NYAN` is set. git nyan is covered by lines 24-27

TheJJ: ah right. makes sense :)

TheJJ (approved):

PR #1656, opened 2024-06-02
Integrate flow field pathfinder into game simulation

heinezen:
Makes the pathfinder usable from the game simulation.

Depends on https://github.com/SFTtech/openage/pull/1648, https://github.com/SFTtech/openage/pull/1647

Introduces several additions and optimizations to the work started in https://github.com/SFTtech/openage/pull/1620 :

- Flow field documentation
- Gamestate
   - `Map` object for accessing pathfinder
   - use pathfinder in move system
   - new nyan API bindings for `Terrain` objects
- Pathfinder
   - Massive performance increase for flow field calculation
   - LOS propagation across sectors
   - Flow field caching
   - Fixed several bugs


TheJJ (commented):
(ignore pls)

libopenage/gamestate/map.cpp
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	// Set path costs
+	for (size_t chunk_idx = 0; chunk_idx < this->terrain->get_chunks().size(); ++chunk_idx) {
+		auto chunk_terrain = this->terrain->get_chunk(chunk_idx);
+		for (size_t tile_idx = 0; tile_idx < chunk_terrain->get_tiles().size(); ++tile_idx) {
+			auto tile = chunk_terrain->get_tile(tile_idx);
+			if (tile.terrain != std::nullopt) {
+				auto path_costs = api::APITerrain::get_path_costs(*tile.terrain);
+
```
C3pa, line 53: Can you invert the if condition and make it an early `continue` and then reduce the nesting level of the code after:
```c++
if (tile.terrain == std::nulopt) {
    continue;
}
auto path_costs = api::APITerrain::get_path_costs(*tile.terrain);
...
```

heinezen: true, although I just removed that if-branch altogether :D

TheJJ (approved):
i think this is good for now, and an incremental improvement to the already well designed previous code. we surely can improve further and fix bugs that show up in practice :)

Issue #1601, opened 2023-11-18
CMake can't find toml11 on Linux Mint 21.2


C1ffisme:
Followed all the Ubuntu build instructions, but unfortunately, this error prevented my build from being fully completed:

```
CMake Error at /usr/share/cmake/toml11/toml11Config.cmake:36 (include):
  include could not find requested file:

    //lib/x86_64-linux-gnu/cmake/toml11/toml11Targets.cmake
Call Stack (most recent call first):
  libopenage/CMakeLists.txt:53 (find_package)
 ```
 
 All dependencies are installed including libtoml11-dev and python3-toml, so I'm not sure where to go from here.

heinezen:
We had the same issues for some people that used Debian where the toml config was also not found by cmake. They probably used the wrong file for their cmake config. 

I'm not sure if this helps but this is the config that current Ubuntu uses which shouldn't have that problem. Try to make a backup of your old place and replace the files. Maybe that helps.

[toml11.zip](https://github.com/SFTtech/openage/files/13401192/toml11.zip)

C1ffisme:
This worked, but I'd like to note that the `/toml11/` directory in its entirety did not exist until I created it.

TheJJ:
One should open a but at debian if not present already.

patrickziegler:
It does exist already https://bugs.launchpad.net/ubuntu/+source/toml11/+bug/1978418

fabiobarkoski:
> This worked, but I'd like to note that the `/toml11/` directory in its entirety did not exist until I created it.

Same here. Perhaps add a note about mint in the build docs for now would be good. What do you all think?

heinezen:
> > This worked, but I'd like to note that the `/toml11/` directory in its entirety did not exist until I created it.
> 
> Same here. Perhaps add a note about mint in the build docs for now would be good. What do you all think?

Yeah that's a good idea since it doesn't seem to get fixed...

TheJJ:
This is quite outdated, let's hope it has resolved by now. "It's not our bug" anyway 😁

PR #1661, opened 2024-07-25
docs: Add note about linuxmint issue on ubuntu dependencies

fabiobarkoski:
Added a note about the Linux Mint problem discussed in the #1601.

### DIFF
--- a/doc/build_instructions/ubuntu.md
+++ b/doc/build_instructions/ubuntu.md
@@ -15,3 +15,6 @@ from pip:
 ```
 pip3 install cython --break-system-packages
 ```
+
+# Linux Mint Issue
+Linux Mint has a [problem with `toml11`](https://github.com/SFTtech/openage/issues/1601), since CMake can't find it. To solve this, download the [toml11.zip](https://github.com/SFTtech/openage/files/13401192/toml11.zip), after, put the files in the `/usr/lib/x86_64-linux-gnu/cmake/toml11` path. (if the `toml11` directory doesn't exist, create it)


doc/build_instructions/ubuntu.md
```diff
@@ -15,3 +15,6 @@ from pip:
 ```
 pip3 install cython --break-system-packages
 ```
+
+# Linux Mint Issue
+Linux Mint has a [problem with toml11](https://github.com/SFTtech/openage/issues/1601), since CMake can't find it. To solve this, download the [toml11.zip](https://github.com/SFTtech/openage/files/13401192/toml11.zip), after, run the openage and wait it throw the toml11 error, and then put the files in the needed path. (if the toml11 directory doesn't exist, create it)
```
heinezen, line 20: Can you clarify what the path is? It should be the same for all Linux Mint systems I think.

fabiobarkoski: Sure, but at the first time I ran, the `./configure` required the toml11 dir in the `/lib/x86_64-linux-gnu/cmake` and at the second time in the `/usr/lib/x86_64-linux-gnu/cmake`. Put both paths?

heinezen: `/lib` should symlink to `/usr/lib`, so both should be the same. You only need to put the `/usr/lib/...` path in the text then.

fabiobarkoski: Done

heinezen (approved):

PR #1664, opened 2024-07-30
Vectorize uniform inputs

heinezen:
Vectorizes parts of the renderer to increase the speed of the uniform setting and updating.

The changes so far resulted in a ~~40%~~ 70%  increase in FPS for me in stresstest 0.

- Before: 520 FPS
- After: ~~740~~ 880 FPS


libopenage/renderer/opengl/shader_program.cpp
```diff
@@ -338,10 +349,18 @@ void GlShaderProgram::update_uniforms(std::shared_ptr<GlUniformInput> const &uni
...
 	uint8_t const *data = unif_in->update_data.data();
-	for (auto const &pair : unif_in->update_offs) {
-		uint8_t const *ptr = data + pair.second;
-		const auto &unif = this->uniforms.at(pair.first);
+	for (uniform_id_t i = 0; i < this->uniforms.size(); ++i) {
+		const auto &update_off = update_offs[i];
+		if (not update_off.used) {
+			// Uniform value has not been set
+			continue;
+		}
```
heinezen, line 360: This worries me currently. If all uniforms are set, this should be fine as `use` would always be `true` and branch prediction should not be that bad. But if this is only a partial update, then this could get very slow.

heinezen: I've solved this by storing the used IDs in an additional vector. This adds an additional lookup but should avoid pipeline flushes.

It's hard to say what the impact of a mispredicted branch would be, so we could change this back in the future if it turns out to only have a marginal effect.

TheJJ (approved):

PR #1667, opened 2024-08-03
Remove string splits with `\r` from renderer resource parsers

heinezen:
Adds a new dedicated `util::split_newline` function that can handle safe splits.

### DIFF
--- a/libopenage/renderer/resources/parser/parse_blendtable.cpp
+++ b/libopenage/renderer/resources/parser/parse_blendtable.cpp
@@ -1,4 +1,4 @@
-// Copyright 2023-2023 the openage authors. See copying.md for legal info.
+// Copyright 2023-2024 the openage authors. See copying.md for legal info.
 
 #include "parse_blendtable.h"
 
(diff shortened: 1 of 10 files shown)


libopenage/util/strings.h
```diff
@@ -116,5 +116,10 @@ std::vector<std::string> split_escape(const std::string &txt,
                                       char delim,
                                       size_t size_hint = 0);
 
+/**
+ * Newline splitter that works with both \n and \r\n.
+ */
+std::vector<std::string> split_newline(const std::string &txt);
```
TheJJ, line 122: could be returning std::string_views instead :)

libopenage/renderer/resources/parser/parse_blendtable.cpp
```diff
@@ -67,16 +67,16 @@ PatternData parse_pattern(const std::vector<std::string> &args) {
...
-	if (not file.is_file()) [[unlikely]] {
+	if (not path.is_file()) [[unlikely]] {
 		throw Error(MSG(err) << "Reading .bltable file '"
-		                     << file.get_name()
+		                     << path.get_name()
 		                     << "' failed. Reason: File not found");
 	}
 
-	auto content = file.open();
-	auto lines = content.get_lines();
```
TheJJ, line 79: we should improve the get_lines function to improve this for all read files?

heinezen: I've modified it to use the new `util::split_newline` function now.

libopenage/util/strings.h
heinezen: I tried, but it is creating so many problems with existing code that relies on `std::string` which also makes the code harder to read :/ 

For example, `std::stoul` won't accept a string view as an argument, so you have this confusing back and forth conversion between std::string_view and std::string in the renderer parsers...

TheJJ (approved):

Issue #1371, opened 2021-01-22
Speed up converter texture packing with Cython


heinezen:
**Required skills:** Cython

**Difficulty:** Medium

Animation frames from AoE2 graphics files are packed into a texture atlas by the openage converter. We use [bin packing](https://en.wikipedia.org/wiki/Bin_packing_problem) to find the optimal arrangement (= smallest atlas dimensions) of the frames in the atlas. Bin packing becomes computationally intensive if a lot of frames are packed (looking at you Definitive Edition), so we want to optimize the algorithm as much as possible.

The goal of this issue is to speed up the bin packing implemented [here](https://github.com/SFTtech/openage/blob/master/openage/convert/service/export/png/binpack.pyx) using [Cython](http://docs.cython.org/en/latest/src/quickstart/overview.html). The most important methods to cythonize are `pack()` and `factor()`, since they are responsible for the packing logic. Ideally, these functions should contain as little Python overhead as possible.

Additional optimization might be possible such as:

- [ ] Currently `pack()` gets passed a list of frame objects, but technically only *width*, *height* and *index* of a frame are needed.
- [ ] Most of the games have a fixed number of frames per animation (e.g. 50 frames for a unit animation in AoC). These could be hardcoded as lookups into `factor()`
- [ ] [`texture_merge.pyx`](https://github.com/SFTtech/openage/blob/master/openage/convert/processor/export/texture_merge.pyx), the function initiating the binpacking, could also use optimization for merging the individual frames into the atlas using the packer settings.

You can test your changes with the [singlefile media converter](https://github.com/SFTtech/openage/blob/master/doc/convert/convert_single_file.md).

**Further Reading**

* [Binpack code](https://github.com/SFTtech/openage/blob/master/openage/convert/service/export/png/binpack.pyx)
* [Converter architecture](https://github.com/SFTtech/openage/blob/master/doc/code/converter/architecture_overview.md)
* [Converter workflow](https://github.com/SFTtech/openage/blob/master/doc/code/converter/workflow.md)

fabiobarkoski:
Looking in the [`binpack.pyx`](https://github.com/SFTtech/openage/commits/master/openage/convert/service/export/png/binpack.pyx) history, seems that the converter already received some speed up improvement

heinezen:
@fabiobarkoski It has been converted to Cython but none of the points in the ssue have been solved completely afaik.

fabiobarkoski:
[...]

PR #1671, opened 2024-08-07
Speed up converter texture packing with Cython

fabiobarkoski:
Improve texture converter performance with Cython

resolve: #1371

### DIFF
--- a/openage/convert/processor/export/texture_merge.pyx
+++ b/openage/convert/processor/export/texture_merge.pyx
@@ -1,4 +1,4 @@
-# Copyright 2014-2023 the openage authors. See copying.md for legal info.
+# Copyright 2014-2024 the openage authors. See copying.md for legal info.
 #
 # cython: infer_types=True
 # pylint: disable=too-many-locals
@@ -9,14 +9,15 @@ a terrain texture.
 import numpy
 from enum import Enum
 
-from ....log import spam
-from ...entity_object.export.texture import TextureImage
-from ...service.export.png.binpack cimport Packer, DeterministicPacker, RowPacker, ColumnPacker, BinaryTreePacker, BestPacker
-from ...value_object.read.media.hardcoded.texture import (MAX_TEXTURE_DIMENSION, MARGIN,
-                                                          TERRAIN_ASPECT_RATIO)
-
 cimport cython
 cimport numpy
+
+from ....log import spam
+from ...service.export.png.binpack cimport block
+from ...entity_object.export.texture import TextureImage
+from ...service.export.png.binpack cimport DeterministicPacker, RowPacker, ColumnPacker, BinaryTreePacker, BestPacker
+from ...value_object.read.media.hardcoded.texture import (MAX_TEXTURE_DIMENSION, MARGIN,
+                                                          TERRAIN_ASPECT_RATIO)
 
 class PackerType(Enum):
     """
@@ -57,6 +58,7 @@ cdef void cmerge_frames(texture, packer_type=PackerType.BINPACK, cache=None) exc
     :type cache: list
     """
     cdef list frames = texture.frames
+    cdef list blocks = [block(idx, frame.width, frame.height) for idx, frame in enumerate(frames)]
 
     if len(frames) == 0:
         raise ValueError("cannot create texture with empty input frame list")
@@ -64,7 +66,7 @@ cdef void cmerge_frames(texture, packer_type=PackerType.BINPACK, cache=None) exc
     cdef BestPacker packer
 
     if cache:
-        packer = BestPacker([DeterministicPacker(margin=MARGIN,hints=cache)])
+        packer = BestPacker([DeterministicPacker(margin=MARGIN, hints=cache)])
 
     else:
         if packer_type == PackerType.ROW:
@@ -81,7 +83,7 @@ cdef void cmerge_frames(texture, packer_type=PackerType.BINPACK, cache=None) exc
                                  RowPacker(margin=MARGIN),
                                  ColumnPacker(margin=MARGIN)])
 
-    packer.pack(frames)
+    packer.pack(blocks)
 
     cdef int width = packer.width()
     cdef int height = packer.height()
@@ -106,11 +108,11 @@ cdef void cmerge_frames(texture, packer_type=PackerType.BINPACK, cache=None) exc
     cdef int sub_h
 
     cdef list drawn_frames_meta = []
-    for sub_frame in frames:
+    for index, sub_frame in enumerate(frames):
         sub_w = sub_frame.width
         sub_h = sub_frame.height
 
-        pos_x, pos_y = packer.pos(sub_frame)
+        pos_x, pos_y = packer.pos(index)
 
         spam("drawing frame %03d on atlas at %d x %d...",
              len(drawn_frames_meta), pos_x, pos_y)
@@ -143,4 +145,4 @@ cdef void cmerge_frames(texture, packer_type=PackerType.BINPACK, cache=None) exc
     if isinstance(packer, BestPacker):
         # Only generate these values if no custom packer was used
         # TODO: It might make sense to do it anyway for debugging purposes
-        texture.best_packer_hints = packer.get_mapping_hints(frames)
+        texture.best_packer_hints = packer.get_mapping_hints(blocks)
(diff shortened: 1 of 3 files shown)


heinezen (commented):
Hey looks good for now. I think you got the basics of Cython right. What would be your next steps?

openage/convert/service/export/png/binpack.pyx
```diff
@@ -98,9 +99,9 @@ cdef class BestPacker:
 
     cdef void pack(self, list blocks):
         cdef Packer p
+        cdef Packer packer
         for packer in self.packers:
-            p = packer
-            p.pack(blocks)
+            packer.pack(blocks)
```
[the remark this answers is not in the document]
fabiobarkoski: You're right, I'll remove it

openage/convert/service/export/png/binpack.pyx
```diff
@@ -129,10 +130,11 @@ cdef class RowPacker(Packer):
     """
 
     cdef void pack(self, list blocks):
-        self.mapping = {}
-
         cdef unsigned int num_rows
         cdef list rows
+        cdef int y = 0
+
+        self.mapping = {}
```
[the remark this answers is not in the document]
fabiobarkoski: I was putting all the var declaration on top, but I'll change to the old way :)

openage/convert/service/export/png/binpack.pyx
```diff
@@ -160,6 +161,10 @@ cdef class ColumnPacker(Packer):
     """
 
     cdef void pack(self, list blocks):
+        cdef unsigned int num_columns
+        cdef list columns
+        cdef int x = 0
+
```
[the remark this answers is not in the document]
fabiobarkoski: I'll change to the old way too

openage/convert/service/export/png/binpack.pyx
```diff
@@ -1,14 +1,12 @@
 # Copyright 2016-2023 the openage authors. See copying.md for legal info.
 #
-# cython: infer_types=True,profile=False
+# cython: infer_types=True,profile=True
```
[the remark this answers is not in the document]
fabiobarkoski: Ok

fabiobarkoski:
> I think you got the basics of Cython right

Is there anything that I could improve?

> What would be your next steps?

I'll try to make the additional optimizations

heinezen:
@fabiobarkoski I think from a perspective of cythonization of the members and variables, the code is already pretty optimized. We now can only gain more speed with design changes to the code.

I think that this idea from the issue might be worth a try:

> Currently pack() gets passed a list of frame objects, but technically only _width_, _height_ and _index_ of a frame are needed.

The idea here being is that `pack` doesn't get a list of `TextureImage` (i.e. members named `block` in the code) but a list that only contains a tuple of `(width, height)` and then only operates on that. This would also require that we use the block index in `self.mapping` instead of hashing the block itself.

openage/convert/service/export/png/binpack.pyx
```diff
@@ -212,41 +218,41 @@ cdef class BinaryTreePacker(Packer):
         self.mapping = {}
         self.root = NULL
 
-        for block in sorted(blocks, key=maxside_heuristic, reverse=True):
-            self.fit(block)
+        for i, block in enumerate(sorted(blocks, key=maxside_heuristic, reverse=True)):
+            self.fit(i, block)
```
heinezen, line 222: The problem here is that index `i` is the index in the list of the already sorted blocks, not the index of the original list of blocks. You then insert `i` into the mapping in `fit(..)`. When you later call `pos(index)`, it still uses the original block index for the lookup even though `mapping` maps the sorted indices.

This probably also cause your image size being wrong since that depends on `pos(..)`.

heinezen: You should try to preserve the original index during sorting, e.g. by extending the tuple of a block to `(width, height, index)`

heinezen:
I'll start the review now.

heinezen (commented):
It's already pretty good, I just had a few ideas for the code and extra C++ types.

openage/convert/service/export/png/binpack.pxd
```diff
@@ -7,9 +7,11 @@ cdef class Packer:
     cdef dict mapping
 
     cdef void pack(self, list blocks)
```
heinezen, line 9: If you do the changes I suggested and pass a `vector` instead of a list in this function, you can then also pass it by reference:

```suggestion
    cdef void pack(self, vector[block_t] &blocks)
```

This will avoid a copy of the vector and thus save some time.

openage/convert/service/export/png/binpack.pxd
```diff
@@ -35,6 +39,7 @@ cdef class BinaryTreePacker(Packer):
     cdef packer_node *root
 
     cdef void fit(self, block)
```
heinezen, line 41: `block` could be typed, either as a 3-tuple (which you use now) or using the struct type if you use a struct like I suggested.

heinezen: 3-tuple would then be `cdef void fit(self, (unsigned int, unsigned int, unsigned int) block)`

openage/convert/service/export/png/binpack.pyx
```diff
@@ -176,20 +174,20 @@ cdef class ColumnPacker(Packer):
...
-                self.mapping[block] = (x, y)
-                y += block.height + self.margin
+                self.mapping[block[2]] = (x, y)
+                y += block[1] + self.margin
 
-            x += max(block.width for block in column) + self.margin
+            x += max(block[0] for block in column) + self.margin
 
 
 cdef inline (unsigned int, unsigned int, unsigned int, unsigned int) maxside_heuristic(block):
```
heinezen, line 183: Here the `block` argument could also be typed.

openage/convert/service/export/png/binpack.pxd
```diff
@@ -35,6 +39,7 @@ cdef class BinaryTreePacker(Packer):
...
+    cdef tuple get_packer_settings(self)
     cdef packer_node *find_node(self, packer_node *root, unsigned int width, unsigned int height)
```
heinezen, line 43: `find_node`, `split_node`, `grow_node`, `grow_right`, `grow_left` can now be declared `noexcept` since we cythonized everything in them.

openage/convert/processor/export/texture_merge.pyx
```diff
@@ -81,7 +81,7 @@ cdef void cmerge_frames(texture, packer_type=PackerType.BINPACK, cache=None) exc
                                  RowPacker(margin=MARGIN),
                                  ColumnPacker(margin=MARGIN)])
 
-    packer.pack(frames)
+    packer.pack([(frame.width, frame.height, i) for i, frame in enumerate(frames)])
```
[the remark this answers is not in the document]
heinezen: On second thought, let's not do this right now because it's probably really complicated :D

openage/convert/service/export/png/binpack.pxd
heinezen: see https://github.com/SFTtech/openage/pull/1671#discussion_r1864312773

openage/convert/processor/export/texture_merge.pyx
```diff
@@ -143,4 +144,9 @@ cdef void cmerge_frames(texture, packer_type=PackerType.BINPACK, cache=None) exc
     if isinstance(packer, BestPacker):
         # Only generate these values if no custom packer was used
         # TODO: It might make sense to do it anyway for debugging purposes
-        texture.best_packer_hints = packer.get_mapping_hints(frames)
+        texture.best_packer_hints = packer.get_mapping_hints(blocks)
+
+cdef struct block:
+    unsigned int index
+    unsigned int width
+    unsigned int height
```
heinezen, line 152: You should move this to `binpack.pxd`. Then you don't need even need to define `texture_merge.pxd` as you can just import `block` from `binpack.pxd`.

openage/convert/processor/export/texture_merge.pxd
heinezen, line 1: This file is redundant if you define the struct in `binpack.pxd` and import it in `texture_merge.pyx`

heinezen (changes requested):
Nice work, the imports are the last thing that needs changing.

heinezen (approved):
Made a tiny change myself after testing locally. Well done!

Issue #1546, opened 2023-09-17
CLI parameters for window settings


heinezen:
**Required Skills:** Python, some C++

**Difficulty**: Easy

Several window parameters could be set via the CLI on startup. Examples for these parameters include:

- window size (height x width)
- vsync
- fullscreen/borderless/windowed

Most of the parameters are already supported by the `renderer::Window` class. All we have to do is pass them through from the Python CLI.

Tasks:
- [ ] Start parameters have to be defined in Python for start modes `main` and `game`
- [ ] Pass parameters to C++ invocation `main_arguments` struct
- [ ] in C++, pass relevant parameters to presenter/renderer subsystem

**Further Reading**

- [openage renderer documentation](https://github.com/SFTtech/openage/blob/master/doc/code/renderer/README.md)

stoyanov-kaloyan:
Hi! I would like to work on this issue. Can you please assign it to me?

heinezen:
@stoyanov-kaloyan Great :) I've assigned you just now.

RaulGalvez288:
Hi, I want to contribute to this project and I figured this issue would be a great place to start. Can I be assigned to it?

TheJJ:
done - but generally this is not needed to work on something :) usually, just submit the pull request

PR #1672, opened 2024-08-20
Added custom resolution size for main and game

RaulGalvez288:
Partial fix: #1546 

Implemented width and height parameters for main and game start modes.

Still have to add vsync and window modes.

### DIFF
--- a/libopenage/engine/engine.cpp
+++ b/libopenage/engine/engine.cpp
@@ -22,9 +22,9 @@ Engine::Engine(mode mode,
 	running{true},
 	run_mode{mode},
 	root_dir{root_dir},
-	threads{},
+	height{wHeight},
 	width{wWidth},
-	height{wHeight} {
+	threads{} {
 	log::log(INFO
 	         << "launching engine with root directory"
 	         << root_dir
--- a/libopenage/engine/engine.cpp
+++ b/libopenage/engine/engine.cpp
@@ -17,13 +17,13 @@ Engine::Engine(mode mode,
                const util::Path &root_dir,
                const std::vector<std::string> &mods,
                bool debug_graphics,
-			   int wWidth,
-			   int wHeight) :
+			   size_t window_width,
+			   size_t window_height) :
 	running{true},
 	run_mode{mode},
 	root_dir{root_dir},
-	height{wHeight},
-	width{wWidth},
+	height{window_height},
+	width{window_width},
 	threads{} {
 	log::log(INFO
 	         << "launching engine with root directory"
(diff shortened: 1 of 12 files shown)


RaulGalvez288:
I keep getting some warnings for the Copyright year. Would I be able to change this? Wanted to ask in case its legal stuff.

image

libopenage/engine/engine.cpp
```diff
@@ -16,14 +16,20 @@ namespace openage::engine {
 Engine::Engine(mode mode,
                const util::Path &root_dir,
                const std::vector<std::string> &mods,
-               bool debug_graphics) :
+               bool debug_graphics,
+			   int wWidth,
+			   int wHeight) :
```
heinezen, line 21: We use snake case for variable names, so `window_width` or just `width` are better.

heinezen: Also, you can use `size_t` instead of `int` for the type because the width/height shouldn't be negative.

libopenage/engine/engine.cpp
```diff
@@ -56,7 +62,7 @@ Engine::Engine(mode mode,
 	// if presenter is used, run it in a separate thread
 	if (this->run_mode == mode::FULL) {
 		this->threads.emplace_back([&, debug_graphics]() {
```
heinezen, line 64: If you change the captured arguments to `[&, debug_graphics, wWidth, wHeight]`, you don't need to store width and height as members in `Engine`. You can just pass them through in the function.

openage/game/main_cpp.pyx
```diff
@@ -37,6 +37,18 @@ def run_game(args, root_path):
...
+        if args.width is not None:
+            args_cpp.width = args.width
+        else:
+            args_cpp.width = 1024
+
+        # window height
+        if args.height is not None:
+            args_cpp.height = args.height
+        else:
+            args_cpp.height = 768
```
heinezen, line 50: Instead of checking for `None`, you could also just set the `default` argument for the respective args to the default values.

openage/main/main_cpp.pyx
```diff
@@ -37,6 +37,18 @@ def run_game(args, root_path):
...
+        if args.width is not None:
+            args_cpp.width = args.width
+        else:
+            args_cpp.width = 1024
+
+        # window height
+        if args.height is not None:
+            args_cpp.height = args.height
+        else:
+            args_cpp.height = 768
```
heinezen, line 50: same here

heinezen (changes requested):
Hey thank you so much! I can already give you some feedback on the work you've done.

RaulGalvez288:
I did want to ask if changing the engine.cpp and engine.h was alright for this issue? or if I should just use Window::render like #1553

heinezen:
@RaulGalvez288 There is https://github.com/SFTtech/openage/blob/d09adb0901b15deb1b02b257ff437a283f694771/libopenage/renderer/window.h#L27-L38 if you want to use that.

heinezen:
@RaulGalvez288 Are you planning to continue working on this feature?

PR #1673, opened 2024-08-22
Update testing.md

AnujChitari:
## Type of change
Updated the `testing` property in the `testing.md`

## Details
As asked in the issue,  I had to make necessary changes into the `testing`  property of the `doc/code/testing.md` file 

@SFTtech please review this pull request and merge it to solve the issue.


TheJJ (changes requested):
hi, thank you for your submission.

this task was asked in which issue? so we can link them.

i'm sorry but i don't think this is a useful addition - your patch removes documentation of our testing system and replaces it with example code how to start programs with subprocess and use argparse.

this seems to be straight out of a LLM like ChatGPT without considering any of our project's architecture and needs.

PR #1688, opened 2024-09-07
ApplyEffect implementation

heinezen:


paaspaas00:
Hey @heinezen , what's the state of this PR? Is it ready-ish or faraway? Maybe you can add some details in here

doc/code/curves.md
```diff
@@ -292,3 +296,28 @@ Modify operations insert values for a specific point in time.
...
+in the keyframe container.
+
+Furthermore, primitive curves support incremental compression during insertion for the
+`set_insert(t, value)`  and `set_last(t, value)` methods via their `compress` argument.
+If compression is active, `(t, value)` is only inserted when it is not a redundant
+keyframe. `sync(Curve, t)` also supports compression with a flag `compress` passed as
+an argument.
+
+Compression may be used in cases where the size should be kept small, e.g. when the curve
+is tranferred via network or recorded in a replay file. Another application of compression
```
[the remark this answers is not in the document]
heinezen: Thanks I changed it.

libopenage/curve/continuous.h
```diff
@@ -54,20 +61,29 @@ void Continuous<T>::set_last(const time::time_t &at, const T &value) {
...
 }
 
 
-template <typename T>
-void Continuous<T>::set_insert(const time::time_t &t, const T &value) {
+template <KeyframeValueLike T>
+void Continuous<T>::set_insert(const time::time_t &t,
+                               const T &value,
+                               bool /* compress */) {
 	this->set_replace(t, value);
```
TheJJ, line 82: why not pass on compress?

libopenage/curve/interpolated.h
```diff
@@ -33,49 +34,117 @@ class Interpolated : public BaseCurve<T> {
...
 
+template <KeyframeValueLike T>
+void Interpolated<T>::compress(const time::time_t &start) {
+	// Find the last element before the start time
+	auto e = this->container.last_before(start, this->last_element);
+
+	// Store elements that should be kept
+	std::vector<Keyframe<T>> to_keep;
+	auto last_kept = e;
+	for (auto current = e + 1; current < this->container.size() - 1; ++current) {
```
TheJJ, line 105: this has to be size - 2, otherwise the last element is doubled?

libopenage/curve/keyframe_container.h
```diff
@@ -509,40 +597,37 @@ KeyframeContainer<T>::erase_after(KeyframeContainer<T>::elem_ptr last_valid) {
...
 
 
-template <typename T>
+template <KeyframeValueLike T>
 typename KeyframeContainer<T>::elem_ptr
 KeyframeContainer<T>::sync(const KeyframeContainer<T> &other,
                            const time::time_t &start) {
-	// Delete elements after start time
+	// Delete elements from this container after start time
 	elem_ptr at = this->last_before(start, this->container.size());
```
TheJJ, line 613: we could use the default `last_before` which uses the container size anyway.

libopenage/curve/keyframe_container.h
```diff
@@ -509,40 +597,37 @@ KeyframeContainer<T>::erase_after(KeyframeContainer<T>::elem_ptr last_valid) {
...
+	elem_ptr at_other = other.last_before(start, other.size());
+	++at_other; // move one element forward so that at_other.time() >= start
 
 	// Copy all elements from other with time >= start
-	for (size_t i = at_other; i < other.size(); i++) {
-		if (other.get(i).time() >= start) {
-			at = this->insert_after(other.get(i), at);
-		}
-	}
+	this->container.insert(this->container.end(),
```
TheJJ, line 621: this assumes sync always is done on future content only, maybe we should remember for a synced-to curve which point it synced at, so we can also handle changes that happened before the last sync.
is probably nasty to figure out later, but easy to at least error when such a sync happens as we see it now.

libopenage/gamestate/activity/condition/next_command_switch.h
```diff
@@ -0,0 +1,38 @@
...
+class GameEntity;
+class GameState;
+
+namespace activity {
+
+/**
+ * Check the type of the next command in the queue and return its associated key
+ * value.
+ *
+ * The key value is the result of a static_cast of the \p command_t enum value.
```
TheJJ, line 25: why not return the `command_t` value directly then?

libopenage/gamestate/activity/condition/target_in_range.cpp
```diff
@@ -0,0 +1,106 @@
...
+	auto target_entity = state->get_game_entity(target_entity_id);
+	auto target_position = std::dynamic_pointer_cast<gamestate::component::Position>(
+		target_entity->get_component(component::component_t::POSITION));
+	auto target_pos = target_position->get_positions().get(time);
+	log::log(DBG << "Target entity " << target_entity_id << " position: " << target_pos);
+
+	auto distance = (target_pos - current_pos).length();
+	log::log(DBG << "Distance to target entity " << target_entity_id << ": " << distance);
+
+	return (distance >= min_range and distance <= max_range);
```
TheJJ, line 103: this is a really beautiful function to read :heart_eyes:

libopenage/gamestate/activity/types.h
```diff
@@ -13,8 +33,91 @@ enum class node_t {
...
+ */
+using condition_t = std::function<bool(const time::time_t &time,
+                                       const std::shared_ptr<gamestate::GameEntity> &entity,
+                                       const std::shared_ptr<gamestate::GameState> &state,
+                                       const std::shared_ptr<nyan::Object> &api_object)>;
+
+/**
+ * Condition used to determine if an output node is chosen.
+ */
+struct condition {
```
TheJJ, line 79: the names are a bit confusing now, maybe use `struct obj_condition` or rename `condition_t` to something like `condition_fun`?

libopenage/gamestate/api/ability.cpp
```diff
@@ -11,24 +11,54 @@
...
+#include "gamestate/api/util.h"
 
 
 namespace openage::gamestate::api {
 
 bool APIAbility::is_ability(const nyan::Object &obj) {
-	nyan::fqon_t immediate_parent = obj.get_parents()[0];
-	return immediate_parent == "engine.ability.Ability";
+	for (auto &parent : obj.get_parents()) {
+		if (parent == "engine.ability.Ability") {
```
TheJJ: do we want a `obj.has_parent(fqon)` feature?

libopenage/gamestate/api/ability.cpp
```diff
@@ -11,24 +11,54 @@
...
+	return false;
+}
+
+bool APIAbility::check_type(const nyan::Object &ability,
+                            const ability_t &type) {
+	nyan::fqon_t api_parent = get_api_parent(ability);
+	nyan::ValueHolder ability_type = ABILITY_DEFS.get(type);
+
+	std::shared_ptr<nyan::ObjectValue> ability_val = std::dynamic_pointer_cast<nyan::ObjectValue>(
+		ability_type.get_ptr());
```
TheJJ: this could also use a helper function, `ValueHolder.get_obj_val()`?

libopenage/gamestate/api/definitions.h
```diff
@@ -49,58 +244,126 @@ static const auto ABILITY_PROPERTY_DEFS = datastructure::create_const_map<abilit
...
+/**
+ * Maps API command types to engine command types.
+ */
+static const auto COMMAND_LOOKUP = datastructure::create_const_map<nyan::fqon_t, component::command::command_t>(
+	std::pair("engine.util.command.type.Idle",
+              component::command::command_t::IDLE),
+	std::pair("engine.util.command.type.Move",
+              component::command::command_t::MOVE),
+	std::pair("engine.util.command.type.ApplyEffect",
+              component::command::command_t::APPLY_EFFECT));
```
TheJJ, line 367: should we be able to define these lookup aliases within nyan somehow?

libopenage/gamestate/system/apply_effect.cpp
```diff
@@ -0,0 +1,241 @@
...
+		resistor->get_component(component::component_t::POSITION));
+	auto effector_pos = pos_component_effector->get_positions().get(start_time);
+	auto resistor_pos = pos_component_resistor->get_positions().get(start_time);
+	auto effector_angle = pos_component_effector->get_angles().get(start_time);
+
+	auto path_vector = resistor_pos - effector_pos;
+	auto path_angle = path_vector.to_angle();
+
+	if (not turn_speed->is_infinite_positive()) {
+		auto angle_diff = path_angle - effector_angle;
```
TheJJ: `std::abs`?

libopenage/gamestate/system/apply_effect.cpp
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	// properties
+	handle_animated(effector, effect_ability, start_time);
+
+	return total_time;
+}
+
+
+const component::attribute_value_t ApplyEffect::get_applied_discrete_flac(const std::vector<nyan::Object> &effects,
```
TheJJ, line 186: `s/flac/flat`?

libopenage/gamestate/system/move.cpp
```diff
@@ -91,6 +92,38 @@ const time::time_t Move::move_command(const std::shared_ptr<gamestate::GameEntit
...
+			target_entity->get_component(component::component_t::POSITION));
+
+		auto target_pos = position->get_positions().get(start_time);
+
+		return Move::move_default(entity, state, target_pos, start_time);
+	}
+	else if (std::holds_alternative<coord::phys3>(target.value())) {
+		auto target_pos = std::get<coord::phys3>(target.value());
+		return Move::move_default(entity, state, target_pos, start_time);
+	}
```
TheJJ: you can select the variant alternatives without a funny if block using `std::visit` like this:
```cpp
    using value_t = std::variant<int, long, double, std::string>;

    template<class... Ts>
    struct overloaded : Ts... { using Ts::operator()...; };

    std::vector<value_t> vec = {10, 15l, 1.5, "hello"};
    for (auto &v : vec) {
        std::visit(overloaded{
            [](auto arg) { std::cout << arg << ' '; },
            [](double arg) { std::cout << std::fixed << arg << ' '; },
            [](const std::string& arg) { std::cout << std::quoted(arg) << ' '; }
        }, v);
    }
```

libopenage/input/controller/game/controller.cpp
```diff
@@ -146,14 +155,30 @@ void setup_defaults(const std::shared_ptr<BindingContext> &ctx,
...
+				{"entity_ids", controller->get_selected()},
+			};
+		}
+		else {
+			params = {
+				{"type", gamestate::component::command::command_t::APPLY_EFFECT},
+				{"target", target_entity_id},
+				{"entity_ids", controller->get_selected()},
+			};
+		}
```
TheJJ, line 181: this is just awesome

TheJJ (commented):
wow, this is really great :smile:
just a few remarks and questions, but the generalized implementation of effect application looks truly well designed.

libopenage/curve/continuous.h
heinezen: Because `BaseCurve::set_replace` doesn't support `compress` (since there's nothing to compress).

libopenage/curve/interpolated.h
heinezen: why would it be? `current` can only go up to the 2nd last element with the breakout condition. The last element can only be inserted after the for loop.

heinezen: I'm going to add size and element checks to the unit tests though, so that we catch this if it happens.

heinezen: see https://github.com/SFTtech/openage/pull/1688/commits/829af8ea30ca026685613086023f22a263d899aa

libopenage/curve/keyframe_container.h
heinezen: fixed

libopenage/curve/keyframe_container.h
heinezen: Currently that is handled outside of the curve, e.g. the renderer remembers the last time of a sync since the feature is used there the most. I considered implementing sync for time ranges which potentially allows syncing the past too, but it's not worth it yet IMO.

heinezen: But I'm not sure why we would need to error. Can you explain?

libopenage/gamestate/activity/condition/next_command_switch.h
heinezen: Because this function must have the same signature as `switch_function_t` which has return type `int`. This is arguably a bit hacky but it ensures that `XorSwitchGate` works with all enum types that have type int.

heinezen: I could also move the casting to `XorSwitchGate` by using a templated method if that makes more sense?

```cpp
template <typename E>
E evaluate(const time::time_t &time,
           const std::shared_ptr<gamestate::GameEntity> &entity,
           const std::shared_ptr<gamestate::GameState> &state,
           const std::shared_ptr<nyan::Object> &api_object) const {
	return static_cast<E>(
		this->switch_func.function(time, entity, state, api_object));
}
```

libopenage/gamestate/activity/types.h
heinezen: Yeah this got confusing while implementing it too. Technically, the `condition` type is actually meta info for the condition, so maybe `condition_meta_t` would work better...

libopenage/gamestate/api/ability.cpp
heinezen: Yes please.

libopenage/gamestate/api/definitions.h
heinezen: I'm leaning on no, since these are ultimately openage-specific bindings. I'm not sure if it's nyan's responsibility to manage those. How would we do that anyway?

libopenage/gamestate/system/apply_effect.cpp
heinezen: Good idea. We have implemented this for fixed points recently.

libopenage/gamestate/system/apply_effect.cpp
heinezen: FLAC is short for **Fl**at**A**ttribute**C**hange which is openage lingo for changing an attribute value by a fixed amount, e.g. damage, heals, etc. :wink:

libopenage/gamestate/system/move.cpp
heinezen: Cool stuff, didn't know about this.

libopenage/gamestate/api/ability.cpp
heinezen: There already is `Object::extends(..)` which does the same thing :D

I always forget that it exists...

heinezen: https://github.com/SFTtech/openage/pull/1688/commits/6eef46313de0cbfa837809a4a95286614fc1f3fe changes the usage to extends

libopenage/gamestate/api/ability.cpp
heinezen: I added a more sophisticated version of this to https://github.com/SFTtech/nyan/pull/125

libopenage/gamestate/activity/types.h
heinezen: I've renamed the function type to `condition_function_t`, so the difference should be more clear now.

libopenage/curve/keyframe_container.h
TheJJ: yea, currently the sync timing and ranges is done in the renderer transfer - i was thinking if this coordination should maybe become part of the curves itself. maybe introduce a sync-state thing (stored in the renderer) which is required to track the synching state of two curves. but this is just an improvement over the current approach doing/tracking/implementing this "explicitly" in the state->renderer sync.

libopenage/gamestate/activity/condition/next_command_switch.h
TheJJ: maybe cast the other way round, promote the int result to the enum? nevertheless maybe leave comments for future improvements, why ints/enums/... are used there? :)

libopenage/gamestate/api/definitions.h
TheJJ: yea, since this is actually the binding between the byan names and the openage engine implementation

Issue #1683, opened 2024-08-25
Use same syntax for updating uniform inputs and uniform buffer inputs


heinezen:
**Required Skills:** C++, maybe OpenGL

**Difficulty:** Easy

In https://github.com/SFTtech/openage/pull/1664 we switched the structure of OpenGL uniform inputs (`GlUniformInput`) to more performance friendly code. Uniform *buffer* inputs (`GlUniformBufferInput`) were not changed in the PR as the performance gain would have been only marginal. However, it would make sense for code quality purposes to keep the update syntax of both input types roughly the same, so that they can be maintained in the same way.

The tasks mostly consist of porting the changes from https://github.com/SFTtech/openage/pull/1664 to the flow used for updating the uniform buffers. A simple example for uniform buffers that can be used for testing purposes is available in renderer demo 5 which you can access by running this command:

```
./run test -d renderer.tests.renderer_demo 5
```

Tasks:

- [ ] Port the changes to `GlUniformInput` made in https://github.com/SFTtech/openage/pull/1664 to the `GlUniformBufferInput` class
- [ ] Change the update workflow of uniform buffers, so that it aligns with the update workflow of regular uniform inputs
- [ ] (**optional**) Test the performance of the new solution using [callgrind](https://valgrind.org/info/tools.html):

```
valgrind --tool=callgrind ./run test -d renderer.tests.renderer_demo 5
```

**Further Reading**

- [Renderer documentation](https://github.com/SFTtech/openage/tree/master/doc/code/renderer)
- [Uniform input code](https://github.com/SFTtech/openage/blob/master/libopenage/renderer/opengl/uniform_input.h)

Issue #1682, opened 2024-08-25
Boundaries for camera movement


heinezen:
**Required Skills:** C++

**Difficulty:** Medium

Currently, we let the camera move freely in the scene. This can result in the camera accidently moving away from the visible map terrain. It would be better if we could set the boundaries of the camera movement according to the current size of the map, so that the camera can't be moved outside of the play area. This should be implemented for the [`CameraManager`](https://github.com/SFTtech/openage/blob/master/libopenage/renderer/stages/camera/manager.h) and not for the camera class itself!

Tasks:

- [ ] Add 4 boundary values using the `float` type to the [`CameraManager`](https://github.com/SFTtech/openage/blob/master/libopenage/renderer/stages/camera/manager.h) class. For testing purposes, you can hardcode these 4 positions at the start:
    - south west: `12.25f` (min x axis coordinate)
    - south east: `32.25f` (max x axis coordinate)
    - north west: `12.25f` (max z axis corrdinate)
    - north east: `-8.25f` (min z axis coordinate)
- [ ] Check if the camera position is within those boundaries when the camera is moved by the camera manager. If the camera is moved outside of the boundaries, the value should be clamped to the respective min and max values. You can use [`std::clamp`](https://en.cppreference.com/w/cpp/algorithm/clamp) for this purpose.
- [ ] Set the camera boundaries during initialization of the camera manager (you may use hardcoded values).  <br>
       Note that the position the camera *is at* is different from the position the camera *looks at*! When we want to set boundaries, we usually want to set boundaries based on what the camera can look at. However this also means that we have to convert *look-at* coordinates to *is-at* (camera position) first. You can check how the camera dies this by checking out the [`Camera::look_at_scene(..)`](https://github.com/SFTtech/openage/blob/d09adb0901b15deb1b02b257ff437a283f694771/libopenage/renderer/camera/camera.cpp#L99) method.
- [ ] (**optional**; Difficulty: High) Set the camera boundaries from the game simulation using the map size. This requires you to pass the map size across thread boundaries, so you need to understand the C++ thread model.

**Further Reading**

- [Renderer documentation](https://github.com/SFTtech/openage/tree/master/doc/code/renderer)
- [`CameraManager` code](https://github.com/SFTtech/openage/blob/master/libopenage/renderer/stages/camera/manager.h)

yukirine:
Hello, can i help with this issue ?

heinezen:
[...]

Issue #1681, opened 2024-08-25
Optionally check if all uniform values have been set


heinezen:
**Required Skills:** C++ (understanding OpenGL terminology is helpful too)

**Difficulty:** Easy

Whenever we want to change something in the shader for an object, we use a [`UniformInput`](https://github.com/SFTtech/openage/blob/d09adb0901b15deb1b02b257ff437a283f694771/libopenage/renderer/opengl/uniform_input.h#L25) object to store the values. When the object is rendered, the data from the uniform input is passed to the GPU, e.g. via OpenGL. The definitions of the uniform data rypes are stored in the [`ShaderProgram`](https://github.com/SFTtech/openage/blob/master/libopenage/renderer/opengl/shader_program.h) class.

During rendering, the GPU does not strictly require that *all* uniform values have to be set for rendering to happen. However, when a uniform value is not set, e.g. when someone forgets to set pass it to the uniform input, this often leads to rendering errors, or worse: nothing being rendered at all. This is often infuriating to debug as it not always clear from the start whether the value was just forgotten or another error produces the rendering problems. To mitigate these issues, we should add a way to check if all uniform values for a uniform input have been set.

To try out the renderer and see how setting uniform values work, check out renderer demo 4 by running the following command:

```
./run test -d renderer.tests.renderer_demo 4
```

Tasks:

- [ ] Implement an `is_complete(..)` method for [`GlUniformInput`](https://github.com/SFTtech/openage/blob/d09adb0901b15deb1b02b257ff437a283f694771/libopenage/renderer/opengl/uniform_input.h#L25) that checks if all uniform values have been set. You can accomplish this easily by iterating through the `update_offs` vector and testing if the `used` flag has been set.
- [ ] Add `is_complete(..)` checks to the renderer demos in https://github.com/SFTtech/openage/tree/master/libopenage/renderer/demo

**Further Reading**

- [Renderer documentation](https://github.com/SFTtech/openage/tree/master/doc/code/renderer)
- [Uniform input code](https://github.com/SFTtech/openage/blob/master/libopenage/renderer/opengl/uniform_input.h)

AdamYChe:
Hey, I think I'll try this issue out.

heinezen:
@AdamYChe Nice! If you have any problems, contact us in our [chatroom](https://matrix.to/#/#sfttech:matrix.org)

yukirine:
Hey ! can i try this issue too ?

heinezen:
[...]

PR #1690, opened 2024-09-18
Update syntax of GlUniformBuffer to match that of GlUniformInput #1683

EdvinLndh:
Made some similar changes as to what was done on #1683. Hope this is somewhat what you had in mind. 

My first PR ever! =)

### DIFF
--- a/libopenage/renderer/opengl/uniform_buffer.cpp
+++ b/libopenage/renderer/opengl/uniform_buffer.cpp
@@ -16,7 +16,7 @@ namespace openage::renderer::opengl {
 
 GlUniformBuffer::GlUniformBuffer(const std::shared_ptr<GlContext> &context,
                                  size_t size,
-                                 std::unordered_map<std::string, GlInBlockUniform> uniforms,
+                                 std::vector<GlInBlockUniform> uniforms,
                                  GLuint binding_point,
                                  GLenum usage) :
 	GlSimpleObject(context,
@@ -35,7 +35,7 @@ GlUniformBuffer::GlUniformBuffer(const std::shared_ptr<GlContext> &context,
 
 	uniform_id_t unif_id = 0;
 	for (auto &uniform : uniforms) {
-		this->uniforms_by_name.insert(std::make_pair(uniform.first, unif_id));
+		this->uniforms_by_name.insert(std::make_pair(uniform.name, unif_id));
 		unif_id += 1;
 	}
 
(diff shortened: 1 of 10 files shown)


heinezen (changes requested):
Hey thanks a lot! Also great that it runs when I tested it because that's not a given for contributions :D

There's a few things that still need to be added in the PR:

- The changes in the `GlUniformInput` input class should be reflected in the `GlUniformBufferInput` class, i.e. in `uniform_input.h` both classes should look the same apart from the constructor. This is part of the first task in the issue.
- You need to add your info to the table in [COPYING.md](COPYING.md) file in our repo. That's the reason why the CI fails currently.

heinezen:
Nicely done! I'm currently testing your changes locally and will report back later.

libopenage/renderer/opengl/uniform_buffer.cpp
```diff
@@ -51,21 +57,28 @@ void GlUniformBuffer::set_binding_point(GLuint binding_point) {
...
+	const auto &update_offs = glunif_in->update_offs;
 	uint8_t const *data = glunif_in->update_data.data();
-	for (auto const &pair : glunif_in->update_offs) {
-		uint8_t const *ptr = data + pair.second;
-		auto unif_def = this->uniforms[pair.first];
+	for (auto const &pair : this->uniforms_by_name) {
+		auto id = pair.second;
+		auto offset = update_offs[id];
+		uint8_t const *ptr = data + offset.offset;
+		auto unif_def = this->uniforms.find(pair.first)->second;
```
heinezen, line 70: You should not iterate through `this->uniforms_by_name` here and instead use `glunif_in->used_uniforms`. Reasons for this are:

- `UniformBufferInput` can be partial updates where only a few uniforms in the buffer are updated and not all. If you iterate on `this->uniforms_by_name`, this could then read some uninitialized data into the buffer.
- Iterating on a `std::unordered_map` like `this->uniforms_by_name` will be very slow compared to iterating on a `std::vector` (because of their memory layout). This is part of the reason why we want to "vectorize" these parameters: The vectors are much faster in our renderer.

You can check out the update for the regular `UniformInput` in this code sample

https://github.com/SFTtech/openage/blob/master/libopenage/renderer/opengl/shader_program.cpp#L352-365

The syntax should roughly be the same :)

heinezen (changes requested):
Very good! I found one potential problem but once you have fixed that, it should be good to go.

libopenage/renderer/opengl/uniform_buffer.cpp
EdvinLndh: I see, I think one of the reasons I went with that approach was in order to be able to lookup the `GlBlockInUniform` in the _uniforms_ map. When iterating over the id's I'm not really sure how I should retrieve the correct value from the _uniforms_ map. Do you think the _uniforms_ attribute should be kept as a `std::unordered_map`?

heinezen: Yeah I think this might be a good opportunity to change the type of `GlUniformBuffer::uniforms` to `std::vector`. That's already something we do for the uniforms in `shader_program.h` for uniforms outside of blocks:

https://github.com/SFTtech/openage/blob/380d2e9312087b822c98bc38362b60df86753a96/libopenage/renderer/opengl/shader_program.h#L184-L189

The change should be pretty straight-forward I hope. Apart from the type of `uniforms`, you'll have to change the constructor of `GlUniformBuffer` and the `Renderer::add_uniform_buffer` methods:

https://github.com/SFTtech/openage/blob/380d2e9312087b822c98bc38362b60df86753a96/libopenage/renderer/opengl/renderer.cpp#L87

https://github.com/SFTtech/openage/blob/380d2e9312087b822c98bc38362b60df86753a96/libopenage/renderer/opengl/renderer.cpp#L115

heinezen: `shader_data.h` also needs to be changed:

https://github.com/SFTtech/openage/blob/380d2e9312087b822c98bc38362b60df86753a96/libopenage/renderer/opengl/shader_data.h#L78-L81

EdvinLndh: I have now changed from `std::unordered_map` to `std::vector`. I did add a name attribute in the `GlInBlockUniform`. the name is used instead of the the key that was stored with the map previously, mostly debugging and for creating the _uniforms_by_name_ map. I'm not sure if this is how it should be, but I suppose some performance was gained, what do you think?

heinezen: Great, I'll check it out now.

heinezen:
The code already looks good. I'll run a few performance tests now.

heinezen (approved):

heinezen:
Very nice, it works like a charm and is also faster than before :)

heinezen:
@EdvinLndh If you want, there are other renderer issues that should be easily doable now that you worked with part of the code already :smile: 

- https://github.com/SFTtech/openage/issues/1681
- https://github.com/SFTtech/openage/issues/1682

EdvinLndh:
Cool! I'll check them out.

PR #1691, opened 2024-09-25
Renderer: Added camera boundaries to CameraManager and clamping #1682.

EdvinLndh:
Camera boundaries are set using constants for now, will probably look into setting the boundaries dynamically soon.

### DIFF
--- a/libopenage/renderer/camera/camera.cpp
+++ b/libopenage/renderer/camera/camera.cpp
@@ -102,15 +102,37 @@ void Camera::look_at_coord(coord::scene3 coord_pos) {
 	this->look_at_scene(scene_pos);
 }
 
-void Camera::move_to(Eigen::Vector3f scene_pos) {
-	// TODO: Check and set bounds for where the camera can go and check them here
+void Camera::move_to(Eigen::Vector3f scene_pos,
+                     std::optional<std::pair<float, float>> x_bounds,
+                     std::optional<std::pair<float, float>> z_bounds) {
+	// Calculate look at coordinates using same method as in look_at_scene(..)
+	auto y_delta = this->scene_pos[1] - scene_pos[1];
+	auto xz_distance = y_delta * std::numbers::sqrt3;
+
+	float side_length = xz_distance / std::numbers::sqrt2;
+	scene_pos[0] += side_length;
+	scene_pos[2] += side_length;
+
+	// Clamp the x coordinate if x_bounds are provided
+	if (x_bounds.has_value()) {
+		scene_pos[0] = std::clamp(scene_pos[0], x_bounds->first, x_bounds->second);
+	}
+
+	// Clamp the z coordinate if z_bounds are provided
+	if (z_bounds.has_value()) {
+		scene_pos[2] = std::clamp(scene_pos[2], z_bounds->first, z_bounds->second);
+	}
 
 	this->scene_pos = scene_pos;
 	this->moved = true;
 }
 
-void Camera::move_rel(Eigen::Vector3f direction, float delta) {
-	this->move_to(this->scene_pos + (direction * delta));
+void Camera::move_rel(Eigen::Vector3f direction,
+                      std::pair<float, float> x_bounds,
+                      std::pair<float, float> z_bounds,
+                      float delta) {
+
+	this->move_to(this->scene_pos + (direction * delta), x_bounds, z_bounds);
 }
 
 void Camera::set_zoom(float zoom) {
--- a/libopenage/renderer/camera/camera.h
+++ b/libopenage/renderer/camera/camera.h
@@ -15,6 +15,7 @@
 #include "renderer/camera/definitions.h"
 #include "renderer/camera/frustum_2d.h"
 #include "renderer/camera/frustum_3d.h"
+#include <optional>
 
 
 namespace openage::renderer {
@@ -84,7 +85,9 @@ public:
 	 *
 	 * @param scene_pos New 3D position of the camera in the scene.
 	 */
-	void move_to(Eigen::Vector3f scene_pos);
+	void move_to(Eigen::Vector3f scene_pos, std::optional<std::pair<float, float>> x_bounds = std::nullopt,
+											std::optional<std::pair<float, float>> z_bounds = std::nullopt);
+
 
 	/**
 	 * Move the camera position in the direction of a given vector.
@@ -94,7 +97,9 @@ public:
 	 *              value is multiplied with the directional vector before its applied to
 	 *              the positional vector.
 	 */
-	void move_rel(Eigen::Vector3f direction, float delta = 1.0f);
+	void move_rel(Eigen::Vector3f direction, std::pair<float, float> x_bounds,
+											std::pair<float, float> z_bounds,
+											float delta = 1.0f);
 
 	/**
 	 * Set the zoom level of the camera. Values smaller than 1.0f let the
--- a/libopenage/renderer/demo/demo_5.cpp
+++ b/libopenage/renderer/demo/demo_5.cpp
@@ -17,6 +17,7 @@
 #include "renderer/shader_program.h"
 #include "renderer/uniform_buffer.h"
 #include "renderer/uniform_input.h"
+#include "renderer/stages/camera/manager.h"
 
 
 namespace openage::renderer::tests {
@@ -38,6 +39,7 @@ void renderer_demo_5(const util::Path &path) {
 		camera->resize(w, h);
 	});
 
+	camera::CameraManager camera_manager(camera);
 	/* Display the subtextures using the meta information */
 	log::log(INFO << "Loading shaders...");
 
@@ -142,7 +144,7 @@ void renderer_demo_5(const util::Path &path) {
 
 			switch (key) {
 			case Qt::Key_W: { // forward
-				camera->move_rel(Eigen::Vector3f(-1.0f, 0.0f, -1.0f), 0.5f);
+				camera_manager.move_frame(camera::MoveDirection::FORWARD, 0.5f);
 				cam_update = true;
 
 				log::log(INFO << "Camera moved forward.");
@@ -150,13 +152,13 @@ void renderer_demo_5(const util::Path &path) {
 			case Qt::Key_A: { // left
 				// half the speed because the relationship between forward/back and
 				// left/right is 1:2 in our ortho projection.
-				camera->move_rel(Eigen::Vector3f(-1.0f, 0.0f, 1.0f), 0.25f);
+				camera_manager.move_frame(camera::MoveDirection::LEFT, 0.25f);
 				cam_update = true;
 
 				log::log(INFO << "Camera moved left.");
 			} break;
 			case Qt::Key_S: { // back
-				camera->move_rel(Eigen::Vector3f(1.0f, 0.0f, 1.0f), 0.5f);
+				camera_manager.move_frame(camera::MoveDirection::BACKWARD, 0.5f);
 				cam_update = true;
 
 				log::log(INFO << "Camera moved back.");
@@ -164,7 +166,7 @@ void renderer_demo_5(const util::Path &path) {
 			case Qt::Key_D: { // right
 				// half the speed because the relationship between forward/back and
 				// left/right is 1:2 in our ortho projection.
-				camera->move_rel(Eigen::Vector3f(1.0f, 0.0f, -1.0f), 0.25f);
+				camera_manager.move_frame(camera::MoveDirection::RIGHT, 0.25f);
 				cam_update = true;
 
 				log::log(INFO << "Camera moved right.");
--- a/libopenage/renderer/demo/demo_6.cpp
+++ b/libopenage/renderer/demo/demo_6.cpp
@@ -23,6 +23,7 @@
 #include "renderer/shader_program.h"
 #include "renderer/texture.h"
 #include "renderer/uniform_buffer.h"
+#include "renderer/stages/camera/manager.h"
 #include "time/clock.h"
 #include "util/path.h"
 #include "util/vector.h"
@@ -52,16 +53,16 @@ void renderer_demo_6(const util::Path &path) {
 			// move_frame moves the camera in the specified direction in the next drawn frame
 			switch (key) {
 			case Qt::Key_W: { // forward
-				render_mgr.camera->move_rel(Eigen::Vector3f(-1.0f, 0.0f, -1.0f), 0.2f);
+				render_mgr.camera_manager->move_frame(camera::MoveDirection::FORWARD, 0.2f);
 			} break;
 			case Qt::Key_A: { // left
-				render_mgr.camera->move_rel(Eigen::Vector3f(-1.0f, 0.0f, 1.0f), 0.1f);
+				render_mgr.camera_manager->move_frame(camera::MoveDirection::LEFT, 0.1f);
 			} break;
 			case Qt::Key_S: { // back
-				render_mgr.camera->move_rel(Eigen::Vector3f(1.0f, 0.0f, 1.0f), 0.2f);
+				render_mgr.camera_manager->move_frame(camera::MoveDirection::BACKWARD, 0.2f);
 			} break;
 			case Qt::Key_D: { // right
-				render_mgr.camera->move_rel(Eigen::Vector3f(1.0f, 0.0f, -1.0f), 0.1f);
+				render_mgr.camera_manager->move_frame(camera::MoveDirection::RIGHT, 0.1f);
 			} break;
 			default:
 				break;
@@ -438,6 +439,7 @@ void RenderManagerDemo6::create_camera() {
 		1.0f / static_cast<float>(viewport_size[1])};
 	camera_unifs->update("inv_viewport_size", viewport_size_vec);
 	this->camera->get_uniform_buffer()->update_uniforms(camera_unifs);
+	this->camera_manager = std::make_shared<camera::CameraManager>(camera);
 }
 
 void RenderManagerDemo6::create_render_passes() {
--- a/libopenage/renderer/demo/demo_6.h
+++ b/libopenage/renderer/demo/demo_6.h
@@ -19,6 +19,7 @@ class Texture2d;
 
 namespace camera {
 class Camera;
+class CameraManager;
 }
 
 namespace gui {
@@ -71,6 +72,9 @@ public:
 
 	/// Camera
 	std::shared_ptr<camera::Camera> camera;
+
+	/// Camera manager
+	std::shared_ptr<camera::CameraManager> camera_manager;
 
 	/// Render passes
 	std::shared_ptr<renderer::RenderPass> obj_2d_pass;
--- a/libopenage/renderer/stages/camera/manager.cpp
+++ b/libopenage/renderer/stages/camera/manager.cpp
@@ -3,6 +3,7 @@
 #include "manager.h"
 
 #include <eigen3/Eigen/Dense>
+#include <numbers>
 
 #include "renderer/camera/camera.h"
 #include "renderer/uniform_buffer.h"
@@ -21,6 +22,9 @@ CameraManager::CameraManager(const std::shared_ptr<renderer::camera::Camera> &ca
 		camera->get_view_matrix(),
 		"proj",
 		camera->get_projection_matrix());
+
+	x_bounds = std::make_pair(XMIN, XMAX);
+	z_bounds = std::make_pair(ZMIN, ZMAX);
 }
 
 void CameraManager::update() {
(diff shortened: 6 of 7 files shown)


libopenage/renderer/camera/camera.cpp
```diff
@@ -102,15 +102,37 @@ void Camera::look_at_coord(coord::scene3 coord_pos) {
...
+void Camera::move_to(Eigen::Vector3f scene_pos,
+                     std::optional<std::pair<float, float>> x_bounds,
+                     std::optional<std::pair<float, float>> z_bounds) {
+	// Calculate look at coordinates using same method as in look_at_scene(..)
+	auto y_delta = this->scene_pos[1] - scene_pos[1];
+	auto xz_distance = y_delta * std::numbers::sqrt3;
+
+	float side_length = xz_distance / std::numbers::sqrt2;
+	scene_pos[0] += side_length;
+	scene_pos[2] += side_length;
```
heinezen, line 114: `move_to` should not do the look at calculations because it is intended to just move the camera to a position in the scene. Looking at a scene coordinate is already implemented in `look_at_scene`. This might break existing code where `move_to` is currently used.

The changes here might also break `look_at_scene`'s behavior, since the method calls `move_to` but with the already computed camera position. Therefore, the look at calculations would be done a second time on the already translated camera position.

libopenage/renderer/camera/camera.cpp
```diff
@@ -102,15 +102,37 @@ void Camera::look_at_coord(coord::scene3 coord_pos) {
 	this->look_at_scene(scene_pos);
 }
 
-void Camera::move_to(Eigen::Vector3f scene_pos) {
-	// TODO: Check and set bounds for where the camera can go and check them here
+void Camera::move_to(Eigen::Vector3f scene_pos,
+                     std::optional<std::pair<float, float>> x_bounds,
+                     std::optional<std::pair<float, float>> z_bounds) {
```
heinezen, line 107: Instead of using two pairs for the bounds, it probably makes sense to make these parameters a struct and pass that.

libopenage/renderer/camera/camera.cpp
```diff
@@ -102,15 +102,37 @@ void Camera::look_at_coord(coord::scene3 coord_pos) {
...
 
 	this->scene_pos = scene_pos;
 	this->moved = true;
 }
 
-void Camera::move_rel(Eigen::Vector3f direction, float delta) {
-	this->move_to(this->scene_pos + (direction * delta));
+void Camera::move_rel(Eigen::Vector3f direction,
+                      std::pair<float, float> x_bounds,
+                      std::pair<float, float> z_bounds,
```
heinezen, line 132: In `move_to` you made them optional, but here they are non-optional. Is there a reason for this?

libopenage/renderer/camera/camera.h
```diff
@@ -15,6 +15,7 @@
 #include "renderer/camera/definitions.h"
 #include "renderer/camera/frustum_2d.h"
 #include "renderer/camera/frustum_3d.h"
+#include <optional>
```
heinezen, line 18: Small nitpick, but imports of standard libraries should be placed before local includes in our code. See our (arguably a bit hidden) code style examples:

https://github.com/SFTtech/openage/blob/66098cea4d799634d66b0e9b9691fdf715a7118d/doc/code_style/mom.cpp#L17-L32

libopenage/renderer/camera/camera.h
```diff
@@ -84,7 +85,9 @@ class Camera {
 	 *
 	 * @param scene_pos New 3D position of the camera in the scene.
 	 */
-	void move_to(Eigen::Vector3f scene_pos);
+	void move_to(Eigen::Vector3f scene_pos, std::optional<std::pair<float, float>> x_bounds = std::nullopt,
+											std::optional<std::pair<float, float>> z_bounds = std::nullopt);
```
heinezen, line 89: You should document arguments you add to methods in the docstrings above.

heinezen, line 89: I'm not sure if you code formatted this but if you didn't try running `clang-format` over this file. The arguments should either be placed on the same line or in this case on a separate line each.

heinezen, line 89: Instead of using `std::optional` and assigning `std::nullopt` as default value, you could also use regular non-optional floats and set the default value to `std::numeric_limits::min()`/`std::numeric_limits::max()`. Then the clamp operation will work the same, but you won't need to check `has_value` in the code.

libopenage/renderer/camera/camera.h
```diff
@@ -94,7 +97,9 @@ class Camera {
 	 *              value is multiplied with the directional vector before its applied to
 	 *              the positional vector.
 	 */
-	void move_rel(Eigen::Vector3f direction, float delta = 1.0f);
+	void move_rel(Eigen::Vector3f direction, std::pair<float, float> x_bounds,
+											std::pair<float, float> z_bounds,
+											float delta = 1.0f);
```
heinezen, line 102: Same here as above:

- arguments should be documented in the doc string
- formatting should be done for the arguments
- you could assign default values here

libopenage/renderer/demo/demo_5.cpp
```diff
@@ -38,6 +39,7 @@ void renderer_demo_5(const util::Path &path) {
 		camera->resize(w, h);
 	});
 
+	camera::CameraManager camera_manager(camera);
```
heinezen, line 42: The demos in this file should probably not use the camera manager because that introduces more unnecessary complexity to them. We try to show off specific features of the renderer in the code. As `CameraManager` is an interface between the input system and the renderer, it is technically an additional layer that does a bunch of updates which are unnecessary for the demo.

libopenage/renderer/demo/demo_6.cpp
```diff
@@ -438,6 +439,7 @@ void RenderManagerDemo6::create_camera() {
 		1.0f / static_cast<float>(viewport_size[1])};
 	camera_unifs->update("inv_viewport_size", viewport_size_vec);
 	this->camera->get_uniform_buffer()->update_uniforms(camera_unifs);
+	this->camera_manager = std::make_shared<camera::CameraManager>(camera);
```
heinezen, line 442: Same here, using the camera manager is a bit overkill for this demo.

libopenage/renderer/stages/camera/manager.cpp
```diff
@@ -21,6 +22,9 @@ CameraManager::CameraManager(const std::shared_ptr<renderer::camera::Camera> &ca
 		camera->get_view_matrix(),
 		"proj",
 		camera->get_projection_matrix());
+
+	x_bounds = std::make_pair(XMIN, XMAX);
+	z_bounds = std::make_pair(ZMIN, ZMAX);
```
heinezen, line 27: These members should be initialized in the initializer list in the constructor above.

libopenage/renderer/stages/camera/manager.h
```diff
@@ -143,6 +144,17 @@ class CameraManager {
 	 * Uniform buffer input for the camera.
 	 */
 	std::shared_ptr<renderer::UniformBufferInput> uniforms;
+
+	/**
+	 * x and z bounds for the camera.
+	 */
+	std::pair<float, float> x_bounds, z_bounds;
```
heinezen, line 151: Small nitpick but the docstring should explain more about what these bounds are. Also, you should mention that the first value is the min bound and the second value is the max bound. However, if you follow up on the idea to make this a struct, it would probably not be necessary to document the latter.

libopenage/renderer/stages/camera/manager.h
```diff
@@ -143,6 +144,17 @@ class CameraManager {
...
+
+	/**
+	 * Constant values for the camera bounds.
+	 * TODO: Make boundaries dynamic based on map size.
+	 */
+	const float XMIN = 12.25f, XMAX = 32.25f, ZMIN = -8.25f, ZMAX = 12.25f;
```
heinezen, line 157: You could move these to https://github.com/SFTtech/openage/blob/master/libopenage/renderer/camera/definitions.h but since they are meant only for testing, this doesn't really matter.

heinezen (commented):
Looks good so far!

I have a few comments regarding the current status.

heinezen:
@EdvinLndh Are you planning to continue this one soon-ish?

EdvinLndh:
Sorry, I've been out of town for a few weeks, I'll look into it

libopenage/renderer/camera/camera.cpp
```diff
@@ -113,6 +80,15 @@ void Camera::move_rel(Eigen::Vector3f direction, float delta) {
 	this->move_to(this->scene_pos + (direction * delta));
 }
 
+void Camera::move_rel(Eigen::Vector3f direction, float delta, CameraBoundaries camera_boundaries) {
+	auto new_pos = calc_look_at(this->scene_pos + (direction * delta));
+
+	new_pos[0] = std::clamp(new_pos[0], camera_boundaries.x_min, camera_boundaries.x_max);
+	new_pos[2] = std::clamp(new_pos[2], camera_boundaries.z_min, camera_boundaries.z_max);
```
heinezen, line 87: I think it might make sense to move the clamping to `move_to(..)` and pass through the boundaries object. Then we can use the same code for any camera movement, e.g. `look_at_scene(..)`

libopenage/renderer/camera/camera.cpp
```diff
@@ -113,6 +80,15 @@ void Camera::move_rel(Eigen::Vector3f direction, float delta) {
 	this->move_to(this->scene_pos + (direction * delta));
 }
 
+void Camera::move_rel(Eigen::Vector3f direction, float delta, CameraBoundaries camera_boundaries) {
+	auto new_pos = calc_look_at(this->scene_pos + (direction * delta));
```
heinezen, line 84: `move_rel` should not use `calc_look_at` because `move_rel` just moves the camera relative to it's previous position. Calling this would make the camera look at it's own position, which is unnecessary.

libopenage/renderer/camera/camera.h
```diff
@@ -23,6 +25,13 @@ class UniformBuffer;
 
 namespace camera {
 
+/**
+ * Defines constant boundaries for the camera's view in the X and Z axes.
+ */
+struct CameraBoundaries {
+	float x_min, x_max, z_min, z_max;
```
heinezen, line 32: You should document these members. I also prefer them being written out as individual declerations because that is easier to read sometimes, e.g.

```cpp
struct CameraBoundaries {
	float x_min;
	float x_max;
	float z_min;
	float z_max;
};
```

heinezen, line 32: I think it's also a good idea to add `y_min` and `y_max` for the future, even if we don't use them right now.

libopenage/renderer/camera/camera.h
```diff
@@ -96,6 +106,19 @@ class Camera {
...
+	 * Move the camera position in the direction of a given vector taking the
+	 * camera boundaries into account.
+	 *
+	 * @param direction Direction vector. Added to the current position.
+	 * @param delta Delta for controlling the amount by which the camera is moved. The
+	 *              value is multiplied with the directional vector before its applied to
+	 *              the positional vector.
+	 * @param camera_boundaries X and Z boundaries for the camera in the scene.
+	 */
+	void move_rel(Eigen::Vector3f direction, float delta, CameraBoundaries camera_boundaries);
```
heinezen, line 120: Arguments that are/should not be modified can be `const` reference:

```suggestion
	void move_rel(Eigen::Vector3f direction, float delta, const CameraBoundaries &camera_boundaries);
```

libopenage/renderer/camera/camera.h
```diff
@@ -86,6 +95,7 @@ class Camera {
 	 */
 	void move_to(Eigen::Vector3f scene_pos);
 
+
```
heinezen, line 98: redundant newline?

libopenage/renderer/camera/camera.cpp
```diff
@@ -292,8 +268,43 @@ void Camera::init_uniform_buffer(const std::shared_ptr<Renderer> &renderer) {
...
+		target[0] + side_length,
+		this->scene_pos[1], // height unchanged
+		target[2] + side_length);
+}
+
 inline float Camera::get_real_zoom_factor() const {
 	return 0.5f * this->default_zoom_ratio * this->zoom;
 }
 
+
```
heinezen, line 309: redundant newline?

libopenage/renderer/camera/camera.h
heinezen, line 120: Instead of overloading the declaration for `move_rel`, we should probably add `camera_boundaries` to the original `move_rel` implementation and default-initialize it. Then we would always use camera boundaries but the code is a bit simpler. 

The default camera boundaries can be put in `definitions.h` and should use `std::numeric_limits::min()`/`std::numeric_limits::max()` for assignment to the `CameraBoundaries` object. 

```cpp
static constexpr CameraBoundaries DEFAULT_CAM_BOUNDARIES{
    std::numeric_limits::min(),
    std::numeric_limits::max(),
    std::numeric_limits::min(),
    std::numeric_limits::max()
};
```

By default, `std::clamp` would then do practically nothing.

libopenage/renderer/camera/camera.h
```diff
@@ -200,6 +223,7 @@ class Camera {
 	 */
 	const Frustum3d get_frustum_3d() const;
 
+
```
heinezen, line 226: redundant newline?

libopenage/renderer/camera/definitions.h
```diff
@@ -58,4 +58,10 @@ static constexpr float DEFAULT_MAX_ZOOM_OUT = 64.0f;
  */
 static constexpr float DEFAULT_ZOOM_RATIO = 1.0f / 49;
 
+/**
+ * Constant values for the camera bounds.
+ * TODO: Make boundaries dynamic based on map size.
+ */
+static const float X_MIN = 12.25f, X_MAX = 32.25f, Z_MIN = -8.25f, Z_MAX = 12.25f;
```
heinezen, line 65: Making the variable `constexpr` allows the compiler to do some advanced optimizations techniques.

```suggestion
static constexpr float X_MIN = 12.25f, X_MAX = 32.25f, Z_MIN = -8.25f, Z_MAX = 12.25f;
```

libopenage/renderer/demo/demo_6.h
```diff
@@ -19,6 +19,7 @@ class Texture2d;
 
 namespace camera {
 class Camera;
+class CameraManager;
```
heinezen, line 22: This is now unused and can be removed.

libopenage/renderer/demo/demo_6.h
```diff
@@ -72,6 +73,9 @@ class RenderManagerDemo6 {
 	/// Camera
 	std::shared_ptr<camera::Camera> camera;
 
+	/// Camera manager
+	std::shared_ptr<camera::CameraManager> camera_manager;
+
```
heinezen, line 78: Same here

libopenage/renderer/stages/camera/manager.cpp
```diff
@@ -15,7 +15,8 @@ CameraManager::CameraManager(const std::shared_ptr<renderer::camera::Camera> &ca
 	move_motion_directions{static_cast<int>(MoveDirection::NONE)},
 	zoom_motion_direction{static_cast<int>(ZoomDirection::NONE)},
 	move_motion_speed{0.2f},
-	zoom_motion_speed{0.05f} {
+	zoom_motion_speed{0.05f},
+	camera_boundaries({X_MIN, X_MAX, Z_MIN, Z_MAX}) {
```
heinezen, line 19: You can remove the outer parentheses for initialization.

```suggestion
	camera_boundaries{X_MIN, X_MAX, Z_MIN, Z_MAX} {
```

libopenage/renderer/stages/camera/manager.h
```diff
@@ -3,14 +3,17 @@
...
 
+#include "renderer/camera/camera.h"
 
 namespace openage::renderer {
 class UniformBufferInput;
 
 namespace camera {
 
 class Camera;
+struct CameraBoundaries;
```
heinezen, line 16: You don't need the forward declaration, since you include `camera.h` directly.

heinezen (changes requested):

heinezen:
We are slowly getting there :)

libopenage/renderer/camera/camera.h
EdvinLndh: I see, makes sense. 

Do you have any proposition of how I should handle the cyclic dependency that arises when the `definitions.h` file needs to access the definition of the `CameraBoundaries` struct?

One way could be to move the definition of the struct to the `definitions.h` file, but I think it feels slightly weird to define it there.

EdvinLndh:
Yes! Thank you for your feedback and patience with me, I'm learning a lot! 😁

libopenage/renderer/camera/camera.h
heinezen: For plain data storage structs like this (and enums and special types), we usually put them in a `types.h` file in the same folder. You could also put them into a `boundaries.h` file if you think that makes more sense. Then you can import the file in `camera.h` and `manager.h`.

heinezen:
Btw if you think you solved something I mentioned, you can press the "Resolve coonversation" button to mark it as completed. Then I know what you are still working on.

libopenage/renderer/camera/boundaries.h
```diff
@@ -0,0 +1,25 @@
...
+	// The maximum boundary for the camera's X-coordinate.
+	float x_max;
+	// The minimum boundary for the camera's Y-coordinate.
+	float y_min;
+	// The maximum boundary for the camera's Y-coordinate.
+	float y_max;
+	// The minimum boundary for the camera's Z-coordinate.
+	float z_min;
+	// The maximum boundary for the camera's Z-coordinate.
+	float z_max;
```
heinezen, line 22: For docstrings, you should use `///` to distinguish them from regular comments.

libopenage/renderer/camera/camera.cpp
```diff
@@ -71,22 +71,16 @@ void Camera::look_at_coord(coord::scene3 coord_pos) {
 	this->look_at_scene(scene_pos);
 }
 
-void Camera::move_to(Eigen::Vector3f scene_pos) {
+void Camera::move_to(Eigen::Vector3f scene_pos, const CameraBoundaries &camera_boundaries) {
+	scene_pos[0] = std::clamp(scene_pos[0], camera_boundaries.x_min, camera_boundaries.x_max);
+	scene_pos[2] = std::clamp(scene_pos[2], camera_boundaries.z_min, camera_boundaries.z_max);
```
heinezen, line 76: The Y axis should also be clamped here.

libopenage/renderer/camera/definitions.h
```diff
@@ -58,10 +60,18 @@ static constexpr float DEFAULT_MAX_ZOOM_OUT = 64.0f;
...
+	std::numeric_limits<float>::max(),
+	std::numeric_limits<float>::lowest(),
+	std::numeric_limits<float>::max()};
+
 /**
  * Constant values for the camera bounds.
  * TODO: Make boundaries dynamic based on map size.
  */
-static const float X_MIN = 12.25f, X_MAX = 32.25f, Z_MIN = -8.25f, Z_MAX = 12.25f;
+static constexpr float X_MIN = 12.25f, X_MAX = 32.25f, Z_MIN = -8.25f, Z_MAX = 12.25f;
```
heinezen, line 75: I guess we should also define bounds for the Y axis. They should probably be

- min: `0.0f`
- max: `20.0f`

libopenage/renderer/stages/camera/manager.cpp
```diff
@@ -16,7 +16,7 @@ CameraManager::CameraManager(const std::shared_ptr<renderer::camera::Camera> &ca
 	zoom_motion_direction{static_cast<int>(ZoomDirection::NONE)},
 	move_motion_speed{0.2f},
 	zoom_motion_speed{0.05f},
-	camera_boundaries({X_MIN, X_MAX, Z_MIN, Z_MAX}) {
+	camera_boundaries{X_MIN, X_MAX, 0.0f, 0.0f, Z_MIN, Z_MAX} {
```
heinezen, line 19: Initializing to `0.0f` for min/max is not that great, since that will probably move the camera into the terrain. That's why it is initialized to `10.0f` by default.

heinezen (changes requested):

libopenage/renderer/stages/camera/manager.cpp
heinezen: I guess using the suggestion in https://github.com/SFTtech/openage/pull/1691#discussion_r1823620198 would solve this.

heinezen:
I would check this a final time now for merging if that's okay :)

libopenage/renderer/stages/camera/manager.h
heinezen, line 1: `CameraManager` should also have a `set_camera_boundaries` method to change the boundary object. This would be the first step to make the boundaries dynamic.

We also need this in renderer demo 3 because the terrain is smaller in this particular demo (10x10 instead of 20x20). We need to set the boundaries manually after creating the camera manager in line 53: https://github.com/SFTtech/openage/blob/45a7dd55b4e9a3184925d2086a70a5643643df79/libopenage/renderer/demo/demo_3.cpp#L53

heinezen (changes requested):
While testing, I noticed that there is one thing still missing. After that is done, we can finally merge :)

libopenage/renderer/stages/camera/manager.h
EdvinLndh: Okay, sounds very reasonable.

As a heads up, it might take some time until I have an opportunity to work on this again, hope that is okay.

heinezen: Hmm, I think I might do this myself, so we can include this PR in the next release :)

heinezen (dismissed):

heinezen (approved):

heinezen:
It's merged

PR #1692, opened 2024-10-03
Fix thread-safe access for render entities

heinezen:
The current render entity code was not thread-safe because of an oversight in the implementation. When syncing data between the render entity and the objects in the renderer, sometimes references were passed instead of copying the data over, e.g. 

https://github.com/SFTtech/openage/blob/66098cea4d799634d66b0e9b9691fdf715a7118d/libopenage/renderer/stages/world/render_entity.cpp#L63-L67

After returning the reference, the lock is released even though a read has not taken place yet. This results in the simulation thread writing data into the render entity while it is still being read, with the consequence that over- or underreads might happen. I fixed this behavior now by ensuring that locks are present for the full read duration.

Changes in this PR:

- [x] Document which methods of the render entities are thread-safe in the docstrings
-  [x] Add a new method `get_read_lock()` which returns a `std::shared_lock` locking the render entity's mutex. This allows manual control of data reads from the render entity.
-  [x] Add a `renderer::RenderEntity` base class for the shared functionality of the different render entity types in HUD, terrain, and world render stages.
-  [x] Make all data transfers thread-safe

### DIFF
--- a/libopenage/gamestate/game_entity.cpp
+++ b/libopenage/gamestate/game_entity.cpp
@@ -1,4 +1,4 @@
-// Copyright 2022-2023 the openage authors. See copying.md for legal info.
+// Copyright 2022-2024 the openage authors. See copying.md for legal info.
 
 #include "game_entity.h"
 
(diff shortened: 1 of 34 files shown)


TheJJ (approved):
great! less code duplication now, and better locking :)

PR #1699, opened 2024-10-14
Precompute portal node graph

jere8184:


heinezen:
Hey thanks for your contribution. Why did you close this PR immediately after submitting it?

jere8184:
I wasn't happy with the PR as it was merging into master instead of into your branch, please see https://github.com/heinezen/openage/pull/4, hopefully its better. It's far from complete but I would just like to know if i'm going down the right path (pun-intended).

heinezen:
> I wasn't happy with the PR as it was merging into master instead of into your branch, please see [heinezen#4](https://github.com/heinezen/openage/pull/4), hopefully its better. It's far from complete but I would just like to know if i'm going down the right path (pun-intended).

Ah now it makes sense :) I'll check it out!

PR #1702, opened 2024-10-31
Clang format rule: Indent preprocessor directives

heinezen:
Aligns the clang formatting rules for preprocesser directives with what we already use in the files.


TheJJ (approved):
just the macros look slightly worse - the rest is good!

PR #1706, opened 2024-11-07
File system api c++: added support for creating temp. files and directories

alamt22:
Addresses #1540

This is currently a static method in util/path.cpp for creating a temp file, wrapping it in a File object, and returning it. I am planning on doing one for temporary directories later, wrapping it in a Directory object instead. Let me know if you'd like me to wrap both in Path objects instead--I'm not really sure how to do that since Path takes in a fslike shared pointer rather than the file_like one produced by File.

### DIFF
--- a/libopenage/util/file.cpp
+++ b/libopenage/util/file.cpp
@@ -121,5 +121,13 @@ std::ostream &operator<<(std::ostream &stream, const File &file) {
 	return stream;
 }
 
+static File get_temp_file() {
+	std::FILE* tmp_file = std::tmpfile();
+	int temp_fd = fileno(tmp_file);
+	std::string tf_path = "/proc/self/fd/" + std::to_string(temp_fd);
+	mode_t mode = 0777;
+	File file_wrapper = File(tf_path, mode);
+	return file_wrapper;
+}
 
 } // namespace openage::util
--- a/libopenage/util/file.cpp
+++ b/libopenage/util/file.cpp
@@ -16,6 +16,7 @@
 #include "util/filelike/python.h"
 #include "util/path.h"
 #include "util/strings.h"
+#include "util/fslike/directory.h"
 
 
 namespace openage::util {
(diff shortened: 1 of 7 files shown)


alamt22:
@heinezen I think this should be it for the 2 methods. Whenever you're free, let me know if there's anything else I should consider to improve the code.

heinezen:
Thanks, I'll try to review it soon-ish :)

libopenage/util/file.cpp
```diff
@@ -121,5 +121,13 @@ std::ostream &operator<<(std::ostream &stream, const File &file) {
 	return stream;
 }
 
+static File get_temp_file() {
+	std::FILE* tmp_file = std::tmpfile();
+	int temp_fd = fileno(tmp_file);
+	std::string tf_path = "/proc/self/fd/" + std::to_string(temp_fd);
```
heinezen, line 127: Wouldn't this solution only work on Linux?

heinezen: I think a better compromise is to get a temporary directory from `get_temp_directory()` and use `std::tmpnam(nullptr)` to get a filename to place into that. Then we don't have to do a cross-plattform "finding out the filename" solution (yet).

To
heinezen, line 1: I think you committed something on accident here.

libopenage/util/fslike/directory.h
```diff
@@ -47,19 +47,19 @@ class Directory : public FSLike {
...
 	 * basically basepath + "/".join(parts)
 	 */
 	std::string resolve(const Path::parts_t &parts) const;
-
 	std::tuple<struct stat, int> do_stat(const Path::parts_t &parts) const;
-
+	
 	std::string basepath;
 };
-
```
heinezen, line 62: These changes look unecessary.

heinezen (changes requested):

libopenage/util/file.cpp
alamt22: I agree, this would be a good compromise.

alamt22:
Made the requested changes to file.cpp.

heinezen (approved):
Looks good, but you still need to fix the problems indicated by the sanity check.

alamt22:
Fixed the sanity check issues.

heinezen (approved):
I fixed the `To` file issue and I think we can merge here :)

PR #1707, opened 2024-11-07
Paring heap without shared_ptr

### LINKED ISSUE #1675: Implement pairing heap without `shared_ptr`
**Required Skills:** C++

**Difficulty:** Hard

openage provides a [pairing heap implementation](https://github.com/SFTtech/openage/blob/master/libopenage/datastructure/pairing_heap.h) that is mostly used in the A\* algorithm of our pathfinder. Currently, nodes on the heap are connected using `shared_ptr`. However, we have noticed that the creation of `shared_ptr` objects when using the heap is very slow. A variant with raw pointers and without `shared_ptr` may offer better performance benefits.

A previous implementation already used raw pointers. You can see it in commit https://github.com/SFTtech/openage/commit/4ab602a6242448bb82eac5db9eafa96d758c7199

Tasks:

- [ ] Implement the pairing heap without using `shared_ptr` or `enable_shared_from_this`. You can use the implemenation in https://github.com/SFTtech/openage/commit/4ab602a6242448bb82eac5db9eafa96d758c7199 as a reference but be aware that other parts of the code may have changed since then.
- [ ] Use [callgrind](https://valgrind.org/info/tools.html) to compare the speed of the old implementation with your solution that uses precomputed graphs:

```
valgrind --tool=callgrind ./run test -d pathfinding.tests.path_demo 1
```

**Further Reading**

- [pairing heap code](https://github.com/SFTtech/openage/blob/master/libopenage/datastructure/pairing_heap.h)

jere8184:
fixes #1675

### DIFF
--- a/libopenage/datastructure/pairing_heap.h
+++ b/libopenage/datastructure/pairing_heap.h
@@ -38,7 +38,7 @@ class PairingHeap;
 
 
 template <typename T, typename compare = std::less<T>>
-class PairingHeapNode : public std::enable_shared_from_this<PairingHeapNode<T, compare>> {
+class PairingHeapNode {
 public:
 	using this_type = PairingHeapNode<T, compare>;
 
@@ -47,7 +47,6 @@ public:
 	T data;
 	compare cmp;
 
-public:
 	PairingHeapNode(const T &data) :
 		data{data} {}
 
@@ -55,6 +54,10 @@ public:
 		data{std::move(data)} {}
 
 	~PairingHeapNode() = default;
+
+	PairingHeapNode(const this_type &other) = delete;
+
+	this_type &operator=(const this_type &other) = delete;
 
 	/**
 	 * Get contained node data.
@@ -66,14 +69,14 @@ public:
 	/**
 	 * Let this node become a child of the given one.
 	 */
-	void become_child_of(const std::shared_ptr<this_type> &node) {
-		node->add_child(this->shared_from_this());
+	void become_child_of(this_type *const node) {
+		node->add_child(this);
 	}
 
 	/**
 	 * Add the given node as a child to this one.
 	 */
-	void add_child(const std::shared_ptr<this_type> &new_child) {
+	void add_child(this_type *const new_child) {
 		// first child is the most recently attached one
 		// it must not have siblings as they will get lost.
 
@@ -85,7 +88,7 @@ public:
 		}
 
 		this->first_child = new_child;
-		new_child->parent = this->shared_from_this();
+		new_child->parent = this;
 	}
 
 	/**
@@ -93,23 +96,23 @@ public:
 	 * by comparing `this` with `node`.
 	 * The new root is returned, it has the other node as child.
 	 */
-	std::shared_ptr<this_type> link_with(const std::shared_ptr<this_type> &node) {
-		std::shared_ptr<this_type> new_root;
-		std::shared_ptr<this_type> new_child;
+	this_type *link_with(this_type *const node) {
+		this_type *new_root;
+		this_type *new_child;
 
 		if (this->cmp(this->data, node->data)) {
-			new_root = this->shared_from_this();
+			new_root = this;
 			new_child = node;
 		}
 		else {
 			new_root = node;
-			new_child = this->shared_from_this();
+			new_child = this;
 		}
 
 		// children of new root become siblings of new new_child
 		// -> parent of new child = new root
 
-		// this whll be set by the add_child method
+		// this will be set by the add_child method
 		new_child->prev_sibling = nullptr;
 		new_child->next_sibling = nullptr;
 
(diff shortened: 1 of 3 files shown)


jere8184:
for #1675

jere8184:
@heinezen @TheJJ Hopefully this is along the right lines, One issue I've noticed is that unlink_node will leave the user of the function with a node that they will have to then manually delete unless the node is pushed back into the heap at least with my current implementation.

jere8184:
unlink_node is now private. remove_node method added.

libopenage/datastructure/pairing_heap.h
```diff
@@ -607,18 +546,110 @@ class PairingHeap final {
...
+
+
+	/**
+	 * returns the smallest item on the heap and deletes it.
+	 * also known as delete_min.
+	 *                       _________
+	 * Ω(log log n), O(2^(2*√log log n'))
+	 * caller must eventually either add node back to heap or delete it
+	 */
+	element_t pop_node() {
```
heinezen, line 570: It looks like `pop_node()` no longer deletes the node from the pointer, so this may cause issues. Especially if other methods like `unlink_node()` expect that `pop_node()` does deletion.

If you look back at https://github.com/SFTtech/openage/commit/4ab602a6242448bb82eac5db9eafa96d758c7199#diff-84a922def1ded03f0d8849cb1c1d56e758f9ae47a06cb78fc40b193294d40ab9R380 where the shared:ptr implementation was introduced, you can see that there originally was no `pop_node()` (only `pop()`). Maybe it makes sense to merge the methods again.

heinezen (changes requested):
I found a few problems.

libopenage/datastructure/pairing_heap.h
```diff
@@ -176,10 +179,10 @@ class PairingHeapNode : public std::enable_shared_from_this<PairingHeapNode<T, c
 	}
 
 private:
-	std::shared_ptr<this_type> first_child;
-	std::shared_ptr<this_type> prev_sibling;
-	std::shared_ptr<this_type> next_sibling;
-	std::shared_ptr<this_type> parent; // for decrease-key and delete
+	this_type* child = nullptr;
```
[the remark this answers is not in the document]
jere8184: I felt it made more sense as a node can only have one child.

libopenage/datastructure/pairing_heap.h
jere8184: I noticed as well that's, why I made `pop_node()` private, I thought maybe as long as its callers handle the node correctly, it should be fine. I will just merge pop and pop_node.

libopenage/datastructure/pairing_heap.h
```diff
@@ -607,18 +551,112 @@ class PairingHeap final {
...
+		}
+
+
+		// 2. then link remaining trees to the last one, from right to left
+		if (first_pair != nullptr) {
+			this->root_node = first_pair->link_backwards();
+		}
+
+		this->node_count -= 1;
+		size_t erase_result = this->nodes.erase(ret);
```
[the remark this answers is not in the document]
jere8184: This is intended. we now always want to erase the returned node from the set, but like in master, we only throw an error if `OPENAGE_PAIRINGHEAP_DEBUG` is true. Let me know if I'm missing something.

libopenage/datastructure/pairing_heap.h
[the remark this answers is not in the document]
heinezen: In the code yes, but from a conceptual view, each node still has many children. The reason it's done this way is so that the pairing heap can be represented as a binary tree. You can read more about the concept here: https://en.wikipedia.org/wiki/Left-child_right-sibling_binary_tree

So `first_child` (or `left_child`) is more accurate.

libopenage/datastructure/pairing_heap.h
heinezen: In the old implementation, `pop_node` was technically only a delete (more akin to `remove_node` now). It's a bit confusing bcause of the renames. Merging the functions is the right call I think.

libopenage/datastructure/pairing_heap.h
```diff
@@ -631,14 +629,10 @@ class PairingHeap final {
...
 	}
 
-protected:
 	compare cmp;
 	size_t node_count;
 	element_t root_node;
-
-#if OPENAGE_PAIRINGHEAP_DEBUG
 	std::unordered_set<element_t> nodes;
-#endif
```
heinezen, line 641: I wonder if we could keep this set only when `OPENAGE_PAIRINGHEAP_DEBUG` is true, so that the eap becomes even more efficient. Looks like it's only necessary for clearing up the nodes on deletion right now. I suggest we could try this solution:

1. From `clear()`, use `iter_all` to iterate through all the nodes and call `delete` on them.
2. Wrap all interactions of `nodes` with the `#if OPENAGE_PAIRINGHEAP_DEBUG` preprocessor thing again.

heinezen (changes requested):
I have one addtional suggestion and then we can merge.

The rest of the code seems to work out fine :)

libopenage/datastructure/pairing_heap.h
jere8184: alright sound like a plan :)

jere8184:
I think I may have made a small mistake :/

libopenage/datastructure/pairing_heap.h
jere8184: @heinezen  requested changes Implemented at #1713

PR #1708, opened 2024-11-18
Pathfinding: Precomputeportal nodes for A Star

jere8184:


heinezen (approved):
Original review is in https://github.com/heinezen/openage/pull/4

PR #1709, opened 2024-11-18
Fix definition of static functions for creating temp files/dirs

heinezen:
Not sure why the CI checks in https://github.com/SFTtech/openage/pull/1706 didn't catch this but the definition of the functions produced linker errors in my local builds. I have edited the definition, so that this does not happen anymore.

### DIFF
--- a/libopenage/util/file.cpp
+++ b/libopenage/util/file.cpp
@@ -122,12 +122,20 @@ std::ostream &operator<<(std::ostream &stream, const File &file) {
 	return stream;
 }
 
-File File::get_temp_file() {
+File File::get_temp_file(bool executable) {
 	fslike::Directory temp_dir = fslike::Directory::get_temp_directory();
 	std::string file_name = std::tmpnam(nullptr);
 	std::ostringstream dir_path;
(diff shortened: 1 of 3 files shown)


libopenage/util/file.cpp
```diff
@@ -122,13 +122,12 @@ std::ostream &operator<<(std::ostream &stream, const File &file) {
...
 
-static File get_temp_file() {
+File File::get_temp_file() {
 	fslike::Directory temp_dir = fslike::Directory::get_temp_directory();
 	std::string file_name = std::tmpnam(nullptr);
 	std::ostringstream dir_path;
 	temp_dir.repr(dir_path);
-	mode_t mode = 0777;
-	File file_wrapper = File(dir_path.str() + file_name, mode);
+	File file_wrapper = File(dir_path.str() + file_name, 0777);
```
TheJJ, line 130: should not be 777! rather we need a flag if the tmp file should be executable, and have 644 and 755 accordingly.

TheJJ (changes requested):

Issue #1060, opened 2018-10-19
Add clang-tidy to buildsystem checks


Vtec234:
`clang-tidy` can detect lots of C++ programming and style errors, so it would be worthwile to have it included in `make check`. To do so, a Python module invoking it should be added to `buildsystem.codecompliance`.

mcdermott111:
I would be interested in picking up this issue

heinezen:
@mcdermott111 You are very welcome to do that :)

mcdermott111:
What is the best way to decide which clang-tidy errors to be checking for. I think it would be unnecessary to have every single error turned on.  The following are the set of errors I was thinking of including so far: clang-analyzer, bugprone, concurrency, performance and their respective sub-errors.
Here is the list of all clang-tidy errors: https://clang.llvm.org/extra/clang-tidy/checks/list.html

heinezen:
@mcdermott111 I think these are good choices. We can always be more thorough later if we want.

haytham918:
I can do this

PR #1710, opened 2024-11-18
Add and Test Clang-tidy

haytham918:
Addressing the issue in #1060. Added clang-tidy check on files inside "libopenage" directory, focusing on "clang-analyzer", "bugprone", "concurrency", and "performance" issues. @heinezen

### DIFF
--- a/doc/building.md
+++ b/doc/building.md
@@ -57,6 +57,7 @@ Dependency list:
     CR    qt6 >=6.2 (Core, Quick, QuickControls, Multimedia modules)
     CR    toml11
     CR  O vulkan
+       S  clang-tidy
 
       A   An installed version of any of the following (wine is your friend).
           Other versions _might_ work:
--- a/doc/building.md
+++ b/doc/building.md
@@ -39,6 +39,7 @@ Dependency list:
     CR    opengl >=3.3
     CR    libepoxy
     CR    libpng
+       S  clang-tidy
      R    dejavu font
     CR    eigen >=3
     CR    freetype2
@@ -57,7 +58,6 @@ Dependency list:
     CR    qt6 >=6.2 (Core, Quick, QuickControls, Multimedia modules)
     CR    toml11
     CR  O vulkan
-       S  clang-tidy
 
       A   An installed version of any of the following (wine is your friend).
           Other versions _might_ work:
--- a/Makefile
+++ b/Makefile
@@ -128,21 +128,25 @@ mrproperer: mrproper
 checkfast:
 	python3 -m buildsystem.codecompliance --fast
 
-.PHONY: checkall
-checkall:
-	python3 -m buildsystem.codecompliance --all
+.PHONY: checkmerge
+checkmerge:
+	python3 -m buildsystem.codecompliance --merge
 
 .PHONY: checkchanged
 checkchanged:
-	python3 -m buildsystem.codecompliance --all --only-changed-files=origin/master
+	python3 -m buildsystem.codecompliance --merge --only-changed-files=origin/master
 
 .PHONY: checkuncommited
 checkuncommited:
-	python3 -m buildsystem.codecompliance --all --only-changed-files=HEAD
+	python3 -m buildsystem.codecompliance --merge --only-changed-files=HEAD
 
 .PHONY: checkpy
 checkpy:
 	python3 -m buildsystem.codecompliance --pystyle --pylint
+
+.PHONY: checkall
+checkall:
+	python3 -m buildsystem.codecompliance --all
 
 .PHONY: help
 help: $(BUILDDIR)/Makefile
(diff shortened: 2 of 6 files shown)


heinezen:
@TheJJ kevin needs `clang-tidy` to be installed

heinezen:
@haytham918 You should add `clang-tidy` as a sanity check dependency in https://github.com/SFTtech/openage/blob/master/doc/building.md#dependencies

haytham918:
@heinezen Thank you for the notice. I have updated it accordingly.

doc/building.md
```diff
@@ -57,6 +57,7 @@ Dependency list:
     CR    qt6 >=6.2 (Core, Quick, QuickControls, Multimedia modules)
     CR    toml11
     CR  O vulkan
+       S  clang-tidy
```
TheJJ, line 60: please keep the alphabetical sorting :)

haytham918: Hi @TheJJ , I am actually a little confused. Do you mean alphabetical sorting with respect to "S"?

haytham918: Or I guess it's right alphabetical sorting within this:
 R    dejavu font
    CR    eigen >=3
    CR    freetype2
    CR    fontconfig
    CR    harfbuzz
    C   O include-what-you-use
    CR    nyan  (https://github.com/SFTtech/nyan)
    CR  O ncurses
    C     mako
    CR    opusfile
    CRA   opus
    CRA   ogg
       S  pycodestyle
    C     pygments
       S  pylint
    CR    qt6 >=6.2 (Core, Quick, QuickControls, Multimedia modules)
    CR    toml11
    CR  O vulkan

heinezen:
@TheJJ Did you add `clang-tidy` already?

haytham918:
@heinezen @TheJJ , I hope you are doing well with the new year :), any updates on this request?

heinezen:
@haytham918 Happy new year! I'm still waiting for @TheJJ to add `clang-tidy` to the CI machine.

TheJJ:
It's in the image now :)

haytham918:
```
-- The CXX compiler identification is unknown
CMake Error at CMakeLists.txt:34 (project):
  No CMAKE_CXX_COMPILER could be found.

  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.


-- Configuring incomplete, errors occurred!
Error: Process completed with exit code 1.
```
This is what I am seeing from the build output, not sure what exactly is going wrong with the configuration? I don't think I have ever modified the CMake files.

haytham918:
Now clang-tidy seems to be working and generating warnings?

simonsan:
I think the compile command database needs to be added like this: https://stackoverflow.com/questions/63944447/how-to-specify-compilation-database-for-clang-tidy#71826250

`-p ${CMAKE_BUILD_DIR}`

```console
-p <build-path> is used to read a compile command database.

        For example, it can be a CMake build directory in which a file named
        compile_commands.json exists (use -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
        CMake option to get this output). When no build path is specified,
        a search for compile_commands.json will be attempted through all
        parent paths of the first input file . See:
        https://clang.llvm.org/docs/HowToSetupToolingForLLVM.html for an
        example of setting up Clang Tooling on a source tree.
```

TheJJ:
we have the compile command export already. but we need to make clang-tidy to use it, i.e. by integrating it in the buildsystem. it can't run without cmake running before it.

the warnings it currently generates are not helpful since they fail at things like unknown includes, but once it's integrated into the real compile environment, real output is generated :)

mikonse:
Fyi cmake natively supports running clang tidy as part of the normal build.
If the increased compile time is not a problem this is probably the easiest solution.

See https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_CLANG_TIDY.html

TheJJ:
Thx, indeed. And now I remember: We actually added this already:
https://github.com/SFTtech/openage/blob/6597b09b60967a2880b14d02ecff520e97f3544a/configure#L416
https://github.com/SFTtech/openage/blob/6597b09b60967a2880b14d02ecff520e97f3544a/CMakeLists.txt#L152
https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_CLANG_TIDY.html
https://cmake.org/cmake/help/latest/prop_tgt/LANG_CLANG_TIDY.html

heinezen (approved):
So to finally conclude the question of whether we include clang-tidy in the CI checks or how it should be integrated in the sanity check process, I added a compromise solution. Basically, there's now a new make command `checkmerge` which tests the same stuff as before and is now also used by kevin. `checkall` still tests everything, including clang-tidy, but is not used for CI checks. I hope that makes everybody happy :D

Issue #1701, opened 2024-10-17
KeyError during AoE2DE asset conversion on macOS M2


ZzzhHe:
Hi! I'm encountering an issue while trying to run openage with Age of Empires 2: Definitive Edition. After successfully building the project, I initiated the game launch, but the conversion process fails with a `KeyError` related to the `effect_apply_type` during metadata conversion.

I downloaded AoE2DE through the SteamCMD command `app_update 813780 validate` according to the documentation.

Operating system: macOS 14
System architecture: ARM64 (Apple Silicon M2)
C++ Compiler: AppleClang 16.0.0
Python Version: Python 3.11.5

```
❯ ./run main
INFO [py] launching openage v0.5.3-460-ga3df3071
INFO [py] compiled by AppleClang 16.0.0.16000026
INFO [py] running in DEVMODE
INFO [py] openage nyan API modpack is up to date
  Do you want to convert assets? [Y/n]
> Y

openage requires a local game installation for conversion
but no local installation could be found automatically.
  Do you want to download the AoC trial version? [Y/n]
> N
Could not find any installation directory automatically.
Please enter an AOE2 install path manually.
> /Users/Programming/Steam/AoE2
converting from '/Users/Programming/Steam/AoE2'
INFO [py] Looking for compatible games to convert...
INFO [py] Compatible game edition detected:
INFO [py]  * Age of Empires 2: Definitive Edition
INFO [py] converting metadata
Traceback (most recent call last):
  File "run.py", line 20, in init run
    main()
  File "/Users/Programming/GitRepo/openage/.bin/usr-bin-clang++-release-Oauto-sanitize-none/openage/__main__.py", line 168, in main
    return args.entrypoint(args, cli.error)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Programming/GitRepo/openage/.bin/usr-bin-clang++-release-Oauto-sanitize-none/openage/main/main.py", line 84, in main
    convert_assets(asset_path, args)
  File "/Users/Programming/GitRepo/openage/.bin/usr-bin-clang++-release-Oauto-sanitize-none/openage/convert/main.py", line 121, in convert_assets
    convert(args)
  File "/Users/Programming/GitRepo/openage/.bin/usr-bin-clang++-release-Oauto-sanitize-none/openage/convert/tool/driver.py", line 35, in convert
    convert_metadata(args)
[...]

PR #1712, opened 2024-11-24
Convert Battle for Greece DLC

heinezen:
Adds support for the new BfG DLC for DE2.

There's currently a bunch of placeholder names because we can still figure them out later.

Fixes https://github.com/SFTtech/openage/issues/1701

### DIFF
--- a/openage/convert/value_object/read/media/datfile/lookup_dicts.py
+++ b/openage/convert/value_object/read/media/datfile/lookup_dicts.py
@@ -279,6 +279,8 @@ RESOURCE_TYPES = {
     248: "DE2_UNKNOWN_248",
     249: "DE2_UNKNOWN_249",
     250: "DE2_UNKNOWN_250",
+    501: "DE2_UNKNOWN_501",
+    506: "DE2_UNKNOWN_506",
 }
 
 EFFECT_APPLY_TYPE = {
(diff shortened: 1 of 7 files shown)


openage/convert/value_object/read/media/datfile/lookup_dicts.py
```diff
@@ -338,6 +340,10 @@
     102: "TECH_TOGGLE",       # d == research_id
     103: "TECH_TIME_MODIFY",  # a == research_id, if c == 0: d==absval else d==relval
 
+    -54: "UNKNOWN",  # 199: "UNKNOWN",
+    -55: "UNKNOWN",  # 200: "UNKNOWN",
+    -56: "UNKNOWN",  # 201: "UNKNOWN",
```
TheJJ, line 345: -54 sure?

TheJJ (approved):

openage/convert/value_object/read/media/datfile/lookup_dicts.py
heinezen: Oh I forgot about this during experimenting with the values. I think the type is actually unsigned. That's also what genieutils uses for the type:

https://github.com/Tapsa/genieutils/blob/4b454e5b9f26a240078bb6a5c2bd3f514b3dc94b/src/dat/TechageEffect.cpp#L45

PR #1713, opened 2024-11-24
data structure: implement pairing heap without use of shared_ptr

jere8184:
Continuation of #1707 which was accidentally closed when I deleted my fork of openage

### DIFF
--- a/libopenage/datastructure/pairing_heap.h
+++ b/libopenage/datastructure/pairing_heap.h
@@ -405,7 +405,7 @@ public:
 	 */
 	void clear() {
 		auto delete_node = [](element_t node) { delete node; };
-		this->iter_all(delete_node, true);
+		this->iter_all<true>(delete_node);
 		this->root_node = nullptr;
 		this->node_count = 0;
 #if OPENAGE_PAIRINGHEAP_DEBUG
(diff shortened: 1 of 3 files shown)


libopenage/datastructure/pairing_heap.h
```diff
@@ -583,10 +583,9 @@ class PairingHeap final {
 		this->walk_tree(this->root_node, func);
 	}
 
-protected:
+private:
 	void walk_tree(const element_t &root,
 	               const std::function<void(const element_t &)> &func) const {
-		func(root);
```
heinezen, line 589: Ah I didn't thought of that we have top do this for the deletion.

It's probably not a good idea to change the function like this because it effectively reverses the order of walking the tree (`root` will now be the last node that `func` is applied to). I guess the best solution then is to leave this function as-is and not use `iter_all`.

Instead, you can use the code you have written here in `clear()` but replace `func(root)` with `delete root`

heinezen (changes requested):
The sanity checks are not running through and you need top rebase another time.

libopenage/datastructure/pairing_heap.h
jere8184: How about adding a reverse argument to walk tree?

heinezen: Yes, I think that is also a nice option.

I pushed a commit to your branch that additionally makes the reversing a compile-time option, making it faster to execute.

heinezen:
Looks like the amount of calls in the `portal_a_star` function almost halfed. Doesn't make it twice as fast, but it's a good progress. The event queue should also benefit from this change.

heinezen (approved):
I'm going to merge. Well done!

PR #1718, opened 2024-11-28
 Add is_complete() to UniformInput and Implement Checks in Demos 

### LINKED ISSUE #1681: Optionally check if all uniform values have been set

ZzzhHe:
Hi, This is my first PR! Issue resolve #1681 

This PR introduces a virtual is_complete() method to UniformInput, with an override in GlUniformInput to validate that all required uniform values are set. 

I'm currently a bit unsure about the optimal place to call is_complete() in the demos. Here are the options I'm considering:
+ After Creation by `new_uniform_input()`:  Might miss uniforms that are added later using the update() method.
+ After Every `update()`: This could introduce unnecessary overhead, especially if updates occur frequently.
+ Before Rendering: Might not catch updates made inside the rendering loop.

I'm looking forward to any feedback!

### DIFF
--- a/libopenage/renderer/demo/demo_0.cpp
+++ b/libopenage/renderer/demo/demo_0.cpp
@@ -9,6 +9,7 @@
 #include "renderer/resources/mesh_data.h"
 #include "renderer/resources/shader_source.h"
 #include "renderer/shader_program.h"
+#include "renderer/demo/util.h"
 
 
 namespace openage::renderer::tests {
@@ -43,17 +44,16 @@ void renderer_demo_0(const util::Path &path) {
 	auto display_shader = renderer->add_shader({display_vshader_src, display_fshader_src});
 
 	auto quad = renderer->add_mesh_geometry(resources::MeshData::make_quad());
-	auto display_unif = display_shader->create_empty_input();
-	/* Check if all uniform values for uniform inputs have been set */
-	if (!display_unif->is_complete()) {
-		log::log(WARN << "Some Uniform values have not been set.");
-	}
 	Renderable display_stuff{
-		display_unif,
+		display_shader->create_empty_input(),
 		quad,
 		false,
 		false,
 	};
+
+	if (!check_uniform_completeness({display_stuff})) {
+		log::log(WARN << "Uniforms not complete.");
+	}
 
 	auto pass = renderer->add_render_pass({display_stuff}, renderer->get_display_target());
 
--- a/libopenage/renderer/demo/demo_0.cpp
+++ b/libopenage/renderer/demo/demo_0.cpp
@@ -43,8 +43,13 @@ void renderer_demo_0(const util::Path &path) {
 	auto display_shader = renderer->add_shader({display_vshader_src, display_fshader_src});
 
 	auto quad = renderer->add_mesh_geometry(resources::MeshData::make_quad());
+	auto display_unif = display_shader->create_empty_input();
+	/* Check if all uniform values for uniform inputs have been set */
+	if (!display_unif->is_complete()) {
+		log::log(WARN << "Some Uniform values have not been set.");
+	}
 	Renderable display_stuff{
-		display_shader->create_empty_input(),
+		display_unif,
 		quad,
 		false,
 		false,
--- a/libopenage/renderer/demo/demo_0.cpp
+++ b/libopenage/renderer/demo/demo_0.cpp
@@ -2,6 +2,7 @@
 
 #include "demo_0.h"
 
+#include "renderer/demo/util.h"
 #include "renderer/gui/integration/public/gui_application_with_logger.h"
 #include "renderer/opengl/window.h"
 #include "renderer/render_pass.h"
@@ -9,7 +10,6 @@
 #include "renderer/resources/mesh_data.h"
 #include "renderer/resources/shader_source.h"
 #include "renderer/shader_program.h"
-#include "renderer/demo/util.h"
 
 
 namespace openage::renderer::tests {
@@ -51,7 +51,7 @@ void renderer_demo_0(const util::Path &path) {
 		false,
 	};
 
-	if (!check_uniform_completeness({display_stuff})) {
+	if (not check_uniform_completeness({display_stuff})) {
 		log::log(WARN << "Uniforms not complete.");
 	}
 
--- a/libopenage/renderer/demo/util.cpp
+++ b/libopenage/renderer/demo/util.cpp
@@ -1,4 +1,4 @@
-// Copyright 2015-2023 the openage authors. See copying.md for legal info.
+// Copyright 2015-2024 the openage authors. See copying.md for legal info.
 
 #include "util.h"
 #include "renderer/uniform_input.h"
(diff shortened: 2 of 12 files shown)


libopenage/renderer/opengl/uniform_input.cpp
```diff
@@ -50,4 +50,13 @@ GlUniformBufferInput::GlUniformBufferInput(const std::shared_ptr<UniformBuffer>
...
 
+bool GlUniformInput::is_complete() const {
+  for (const auto& uniform : this->update_offs) {
+    if (!uniform.used) {
+      return false;
+    }
+  }
+  return true;
+}
+
```
C3pa, line 61: It looks like this file uses tabs for indenting, while your additions use 2 spaces.

TheJJ (changes requested):
the duplication in each demo/usage to check for the errors should be optimized away, it would be better to check at a place all of them invoke anyway.

libopenage/renderer/demo/demo_0.cpp
```diff
@@ -50,6 +51,10 @@ void renderer_demo_0(const util::Path &path) {
 		false,
 	};
 
+	if (!check_uniform_completeness({display_stuff})) {
```
heinezen, line 54: Instead of `!`, `&&`, and `||` we usually use `not`, `and`, and `or`, since they are more readable :)

libopenage/renderer/demo/demo_0.cpp
```diff
@@ -9,6 +9,7 @@
 #include "renderer/resources/mesh_data.h"
 #include "renderer/resources/shader_source.h"
 #include "renderer/shader_program.h"
+#include "renderer/demo/util.h"
```
heinezen, line 12: Have you added autoformatting with `clang-format` to your editor? The imports should be sorted alphabetically if you use `clang-format`.

libopenage/renderer/demo/util.cpp
```diff
@@ -1,8 +1,22 @@
...
+	bool all_complete = true;
+
+	// Iterate over each renderable object
+	for (const auto &renderable : renderables) {
+		if (renderable.uniform && !renderable.uniform->is_complete()) {
+			all_complete = false;
+		}
+	}
+
+	return all_complete;
```
heinezen, line 19: You don't need `all_complete` if you short circuit in the if clause:

```suggestion
	// Iterate over each renderable object
	for (const auto &renderable : renderables) {
		if (renderable.uniform and not renderable.uniform->is_complete()) {
			return false;
		}
	}

	return true;
```

libopenage/renderer/demo/util.h
```diff
@@ -1,7 +1,9 @@
-// Copyright 2023-2023 the openage authors. See copying.md for legal info.
+// Copyright 2023-2024 the openage authors. See copying.md for legal info.
 
 #pragma once
 
+#include <vector>
+#include "renderer/renderable.h"
```
heinezen, line 6: Small nitpick but you should add a space between external imports from the standard library and internal import from openage.

```suggestion
#include <vector>

#include "renderer/renderable.h"
```

libopenage/renderer/util.cpp
heinezen, line 1: This file did not actually change, so the date should also not be changed.

libopenage/renderer/util.h
heinezen, line 1: same here

heinezen (changes requested):
Very nice! I recommend you use `clang-format` on all files to solve your style issues.

ZzzhHe:
> Very nice! I recommend you use `clang-format` on all files to solve your style issues.

Thanks! I’ve formatted the files with clang-format. Also, I used a rebase to remove the two files that didn’t need changes, so I had to force-push.

ZzzhHe:
BTW, I noticed some code in the demo files is still using `!` instead of `not`. Should I go ahead and change that?

heinezen:
@ZzzhHe yes please :)

ZzzhHe:
> @ZzzhHe yes please :)

done.

TheJJ (changes requested):
i think there's a better way to place the completeness check - in the renderer when a render pass is created.

heinezen:
@TheJJ It should not be done in the renderer itself because we might get a bunch problems. We would no longer be able to make partial uniform updates for shaders. Also, the uniforms may not be initialized when the render pass is created. Uniform initialization should be deferrable for optimization purposes, so that we can create a bunch of render objects without waiting for updates from the gamestate.

I actually just wanted this function for debugging because uninitialized uniform values are really hard to detect :D But they are not necessary for the general workflow I think.

libopenage/renderer/demo/util.cpp
```diff
@@ -1,8 +1,21 @@
...
 
+#include "renderer/uniform_input.h"
+
 
 namespace openage::renderer::tests {
 
+bool check_uniform_completeness(const std::vector<Renderable> &renderables) {
+	// Iterate over each renderable object
+	for (const auto &renderable : renderables) {
+		if (renderable.uniform && not renderable.uniform->is_complete()) {
```
heinezen, line 13:
```suggestion
		if (renderable.uniform and not renderable.uniform->is_complete()) {
```

heinezen (dismissed):

heinezen (approved):

PR #1724, opened 2024-12-03
Array curve

### LINKED ISSUE #1678: Array Curve container
We should add a `curve::Array` type to our curve container classes to record changes to a fixed size sequence over time. In comparison to https://github.com/SFTtech/openage/issues/1677 we can probably optimize a lot more since the size doesn't change over time (i.e. no inserts or erasures).

The new array type should implement these methods:

**Read**

| Method          | Description                                                 |
| --------------- | ----------------------------------------------------------- |
| `get(t, i)`        | Get value of index `i` at time `t`                        |
| `get(t)`        | Get array at time `t`                        |
| `size()`        | Get size of the array                        |
| `frame(t, i)`      | Get the previous keyframe (time and value) of index `i` before or at `t` |
| `next_frame(t, i)` | Get the next keyframe (time and value) of index `i` after `t`            |

There should also be an iterator type to iterate over the array at time `t`.

**Modify**

| Method                  | Description                                                                       |
| ----------------------- | --------------------------------------------------------------------------------- |
| `set_insert(t, i, value)`  | Insert a new keyframe value at time `t` at index `i`                                           |
| `set_last(t, i, value)`    | Insert a new keyframe value at time `t`; delete all keyframes after time `t` at index `i`      |
| `set_replace(t, i, value)` | Insert a new keyframe value at time `t` at index `i`; remove all other keyframes with time `t` at index `i` |

**Copy**

| Method           | Description                                                                                      |
| ---------------- | ------------------------------------------------------------------------------------------------ |
| `sync(Curve, t)` | Replace all keyframes from self after time `t` with keyframes from source `Curve` after time `t` |

jere8184:
Closes #1678

### DIFF
--- a/libopenage/curve/array.h
+++ b/libopenage/curve/array.h
@@ -1,25 +1,34 @@
-// Copyright 2024-2024 the openage authors. See copying.md for legal info.
+// Copyright 2024-2025 the openage authors. See copying.md for legal info.
 
 #pragma once
 
 #include <array>
 
+#include "curve/iterator.h"
 #include "curve/keyframe_container.h"
 #include "event/evententity.h"
 
 
-// remember to update docs
+// ASDF: remember to update docs
 namespace openage {
 namespace curve {
 
 template <typename T, size_t Size>
 class Array : event::EventEntity {
 public:
+	/**
+	 * The underlaying container type.
+	 */
+	using container_t = std::array<KeyframeContainer<T>, Size>;
+
 	Array(const std::shared_ptr<event::EventLoop> &loop,
 	      size_t id,
+	      const T &default_val = T(),
 	      const std::string &idstr = "",
 	      const EventEntity::single_change_notifier &notifier = nullptr) :
-		EventEntity(loop, notifier), _id{id}, _idstr{idstr}, loop{loop} {}
+		EventEntity(loop, notifier),
+		_id{id}, _idstr{idstr}, loop{loop}, container{KeyframeContainer(default_val)} {
+	}
 
 	Array(const Array &) = delete;
 
@@ -37,7 +46,9 @@ public:
 	 */
 	std::array<T, Size> get(const time::time_t &t) const;
 
-	// Get the amount of KeyframeContainers in array curve
+	/**
+	 * Get the amount of KeyframeContainers in array curve
+	 */
 	consteval size_t size() const;
 
 	/**
@@ -107,59 +118,80 @@ public:
 	}
 
 
-	// get a copy to the KeyframeContainer at index
-	KeyframeContainer<T> operator[](size_t index) const {
-		return this->container.at(index);
-	}
-
-	// Array::Iterator is used to iterate over KeyframeContainers contained in a curve at a given time.
+	/**
+	 *  Array::Iterator is used to iterate over KeyframeContainers contained in a curve at a given time.
+	 */
 	class Iterator {
 	public:
 		Iterator(Array<T, Size> *curve, const time::time_t &time = time::TIME_MAX, size_t offset = 0) :
 			curve(curve), time(time), offset(offset) {};
 
-		// returns a copy of the keyframe at the current offset and time
+		/**
+		 *  returns a copy of the keyframe at the current offset and time
+		 */
 		const T operator*() {
 			return this->curve->frame(this->time, this->offset).second;
 		}
 
-		// increments the Iterator to point at the next KeyframeContainer
+		/**
+		 * increments the Iterator to point at the next KeyframeContainer
+		 */
 		void operator++() {
 			this->offset++;
 		}
 
-		// Compare two Iterators by their offset
+		/**
+		 * Compare two Iterators by their offset
+		 */
 		bool operator!=(const Array<T, Size>::Iterator &rhs) const {
 			return this->offset != rhs.offset;
 		}
 
 
 	private:
-		// used to index the Curve::Array pointed to by this iterator
+		/**
+		 * used to index the Curve::Array pointed to by this iterator
+		 */
 		size_t offset;
 
-		// curve::Array that this iterator is iterating over
+		/**
+		 * the curve object that this iterator, iterates over
+		 */
 		Array<T, Size> *curve;
 
-		// time at which this iterator is iterating over
+		/**
+		 * time at which this iterator is iterating over
+		 */
 		time::time_t time;
 	};
 
 
-	// iterator pointing to a keyframe of the first KeyframeContainer in the curve at a given time
+	/**
+	 * iterator pointing to a keyframe of the first KeyframeContainer in the curve at a given time
+	 */
 	Iterator begin(const time::time_t &time = time::TIME_MIN);
 
-	// iterator pointing after the last KeyframeContainer in the curve at a given time
+	/**
+	 *  iterator pointing after the last KeyframeContainer in the curve at a given time
+	 */
 	Iterator end(const time::time_t &time = time::TIME_MIN);
 
 
 private:
+	/**
+	 * get a copy to the KeyframeContainer at index
+	 */
+	KeyframeContainer<T> operator[](size_t index) const {
+		return this->container.at(index);
+	}
+
+
 	std::array<KeyframeContainer<T>, Size> container;
 
 	/**
-	 * hints for KeyframeContainer operations, mutable as hints
-	 * are updated by const read functions.
-	 * This is used to speed up the search for next keyframe to be accessed
+	 * hints for KeyframeContainer operations
+	 * hints is used to speed up the search for next keyframe to be accessed
+	 * mutable as hints are updated by const read-only functions.
 	 */
 	mutable std::array<size_t, Size> last_element = {};
 
--- a/libopenage/curve/array.h
+++ b/libopenage/curve/array.h
@@ -16,10 +16,10 @@ template <typename T, size_t Size>
 class Array : event::EventEntity {
 public:
 	Array(const std::shared_ptr<event::EventLoop> &loop,
-	          size_t id,
-	          const std::string &idstr = "",
-	          const EventEntity::single_change_notifier &notifier = nullptr) :
-		EventEntity(loop, notifier), _id{id}, _idstr{idstr}, loop{loop}{}
+	      size_t id,
+	      const std::string &idstr = "",
+	      const EventEntity::single_change_notifier &notifier = nullptr) :
+		EventEntity(loop, notifier), _id{id}, _idstr{idstr}, loop{loop} {}
 
 	Array(const Array &) = delete;
 
@@ -28,16 +28,16 @@ public:
 	 * Get the last element with elem->time <= time from
 	 * the keyfram container at a given index
 	 */
-	T get(const time::time_t &t, const size_t index) const;
+	T at(const time::time_t &t, const size_t index) const;
 
 
 	/**
 	 * Get an array of the last elements with elem->time <= time from
 	 * all keyfram containers contained within this array curve
 	 */
-	std::array<T, Size> get_all(const time::time_t &t) const;
+	std::array<T, Size> get(const time::time_t &t) const;
 
-
+	// Get the amount of KeyframeContainers in array curve
 	consteval size_t size() const;
 
 	/**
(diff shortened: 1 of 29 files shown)


libopenage/curve/array.h
```diff
@@ -0,0 +1,159 @@
...
+
+
+	Iterator begin(const time::time_t &time = time::TIME_MAX) const;
+
+	Iterator end(const time::time_t &time = time::TIME_MAX) const;
+
+
+private:
+	container_t container;
+	mutable size_t last_hit_index = 0;
```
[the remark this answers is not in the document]
jere8184: it acts as a hint, for KeyframeContainer operations.

jere8184:
is there any chance array curve should be an array of base curves instead of an array of key frame containers?

libopenage/curve/array.h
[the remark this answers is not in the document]
heinezen: But what is the hit index intended to be used for? From what I can see, it's an index into a KeyframeContainer, but there are `Size` KeyframeContainers in the array.

heinezen:
> is there any chance array curve should be an array of base curves instead of an array of key frame containers?

The base curves would have a bunch of duplicate info that we would want to avoid, e.g. they would all have shared pointer to the event loop, their own IDs, and ID strings. Basically, the curve array container should only have this information once.

libopenage/curve/array.h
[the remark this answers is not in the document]
jere8184: array of hints?

jere8184:
> > is there any chance array curve should be an array of base curves instead of an array of key frame containers?
> 
> The base curves would have a bunch of duplicate info that we would want to avoid, e.g. they would all have shared pointer to the event loop, their own IDs, and ID strings. Basically, the curve array container should only have this information once.

would making the base curves store references to these shared fields solve this issue?

heinezen:
> would making the base curves store references to these shared fields solve this issue?

Then the IDs wouldn't be unique which would be a problem for the event system. Since all curves are event entities, duplicating the information would either lead to all contained curves being updated or worse, only one curve being tracked by the event loop. It would also be a waste of space, since we have to store a bunch of redundant data which is what we want to avoid with a dedicated Array class.

Also, `BaseCurve` is an interface and should not be instantiated. It doesn't make sense to use it as a container because `KeyframeContainer` mostly fulfills that jopb already.

libopenage/curve/array.h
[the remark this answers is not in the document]
heinezen: Hmm okay I think I understand now what this is used for. An array of hints (one for each container) would work better, yes.

libopenage/curve/array.h
```diff
@@ -0,0 +1,159 @@
...
+template <typename T, size_t Size>
+void Array<T, Size>::set_replace(const time::time_t &t, const size_t index, T value) {
+	this->container[index].insert_overwrite(Keyframe(t, value), this->last_hit_index);
+}
+
+template <typename T, size_t Size>
+void Array<T, Size>::sync(const Array<T, Size> &other, const time::time_t &start) {
+	for (int i = 0; i < Size; i++) {
+		this->container[i].sync(other, start);
+	}
```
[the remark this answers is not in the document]
jere8184: There was a mistake, here 
I forgot to Index `other` as well.

libopenage/curve/array.h
```diff
@@ -0,0 +1,244 @@
...
+template <typename T, size_t Size>
+std::pair<time::time_t, T> Array<T, Size>::next_frame(const time::time_t &t, const size_t index) const {
+	size_t frmae_index = container[index].last(t, this->last_element[index]);
+	this->last_element[index] = frmae_index;
+	return container[index].get(frmae_index + 1).make_pair();
+}
+
+template <typename T, size_t Size>
+T Array<T, Size>::get(const time::time_t &t, const size_t index) const {
+	return this->frame(t, index).second;
```
heinezen, line 189: You should not call `frame` her because creating the `std::pair` is expensive and might prevent return value optimization. It's better to extract the value directly, evenm if that results in more lines of code.

libopenage/curve/array.h
```diff
@@ -0,0 +1,244 @@
...
+
+template <typename T, size_t Size>
+consteval size_t Array<T, Size>::size() const {
+	return Size;
+}
+
+
+template <typename T, size_t Size>
+void Array<T, Size>::set_insert(const time::time_t &t, const size_t index, T value) {
+	this->last_element[index] = this->container[index].insert_after(Keyframe(t, value), this->last_element[index]);
```
heinezen, line 207: Keyframe object creation should use braced initializers. This will prevent accidental conversion errors.

```suggestion
	this->last_element[index] = this->container[index].insert_after(Keyframe{t, value}, this->last_element[index]);
```

libopenage/curve/array.h
```diff
@@ -0,0 +1,244 @@
...
+template <typename T, size_t Size>
+void Array<T, Size>::set_replace(const time::time_t &t, const size_t index, T value) {
+	this->container[index].insert_overwrite(Keyframe(t, value), this->last_element[index]);
+}
+
+template <typename T, size_t Size>
+void Array<T, Size>::sync(const Array<T, Size> &other, const time::time_t &start) {
+	for (int i = 0; i < Size; i++) {
+		this->container[i].sync(other.container[i], start);
+	}
```
heinezen, line 229: If you are iterating over every element in the container anyway, you should use a for-each loop.

libopenage/curve/keyframe.h
```diff
@@ -54,6 +54,15 @@ class Keyframe {
...
 	}
 
+	/**
+	 * Get the value and timestamp of the keyframe in form of std::pair<time, value>
+	 * @return keyframe pair
+	 */
+	std::pair<time::time_t, T> make_pair() const
+	{
+		return {time(), val()};
+	}
```
heinezen, line 64: You might also use this method in `BaseCurve`'s `frame`/`next_frame` method

libopenage/curve/tests/curve_types.cpp
```diff
@@ -386,6 +387,86 @@ void curve_types() {
...
+		auto it = a.begin(time::time_t(1));
+		TESTEQUALS(*it, 4);
+		++it;
+		TESTEQUALS(*it, 5);
+		++it;
+		TESTEQUALS(*it, 6);
+		++it;
+		TESTEQUALS(*it, 7);
+		++it;
+	}
```
heinezen, line 469: These tests belong in the `libopenage/curve/tests/container.cpp` file.

https://github.com/SFTtech/openage/blob/master/libopenage/curve/tests/container.cpp

libopenage/curve/tests/curve_types.cpp
```diff
@@ -386,6 +387,86 @@ void curve_types() {
 		TESTEQUALS(c.get(1), 0);
 		TESTEQUALS(c.get(5), 0);
 	}
+
+	{ // array
+		auto f = std::make_shared<event::EventLoop>();
+
+		Array<int, 4> a(f,0);
+		a.set_insert(time::time_t(1), 0, 0);
```
heinezen, line 395: The explicit construction of `time::time_t` is unnecessary for calling the functions.

```suggestion
		a.set_insert(1, 0, 0);
```

libopenage/curve/tests/curve_types.cpp
```diff
@@ -386,6 +387,86 @@ void curve_types() {
...
+		a.set_insert(time::time_t(1), 1, 1);
+		a.set_insert(time::time_t(1), 2, 2);
+		a.set_insert(time::time_t(1), 3, 3);
+		//a = [[0:0, 1:0],[0:0, 1:1],[0:0, 1:2],[0:0, 1:3]]
+
+		auto res = a.get_all(time::time_t(1));
+		TESTEQUALS(res[0], 0);
```
heinezen, line 402: You should use `at()` for array access in test code, since it throws errors when the index is not in range. This will save us from unexpected surprises with faulty test code.

```suggestion
		TESTEQUALS(res.at(0), 0);
```

libopenage/curve/tests/curve_types.cpp
```diff
@@ -386,6 +387,86 @@ void curve_types() {
...
+		TESTEQUALS(res[1], 1);
+		TESTEQUALS(res[2], 2);
+		TESTEQUALS(res[3], 3);
+
+		Array<int, 4> other(f,0);
+		other[0].last(999);
+		other[1].last(999);
+		other[2].last(999);
+		other[3].last(999);
```
heinezen, line 411: 1. You are not using `set_last`, so no value is ever inserted.
2. You should not directly access the keyframe container anyway (as mentioned in another comment)

libopenage/curve/tests/curve_types.cpp
```diff
@@ -386,6 +387,86 @@ void curve_types() {
...
+		a.set_insert(time::time_t(2), 0, 15);
+		a.set_insert(time::time_t(2), 0, 20);
+		a.set_replace(time::time_t(2), 0, 25);
+		TESTEQUALS(a.get(time::time_t(2), 0), 25);
+		// a = [[0:0, 1:4, 2:25],[0:0, 1:5],[0:0, 1:6],[0:0, 1:7]]
+
+		a.set_insert(time::time_t(3), 0, 30);
+		a.set_insert(time::time_t(4), 0, 40);
+		a.set_last(time::time_t(3), 0, 35);
+		TESTEQUALS(a.get(time::time_t(3), 0), 35);
```
heinezen, line 443:
```suggestion
		TESTEQUALS(a.get(4, 0), 35);
```

Changing it like this would check if the keyframe `(t=4, v=40)` got deleted.

heinezen (changes requested):

heinezen:
Some additional remarks:

1. You should really try to comment your code more (i.e. docstrings and member descriptions). It gets much harder to review your code if I also have to find out the purpose of your methods/members.
2. You should also document the new array curve type in the docs: https://github.com/SFTtech/openage/blob/master/doc/code/curves.md
3. The code looks very good so far, although I might have a few enhancement ideas after integrating the current review comments ^^

libopenage/curve/array.h
jere8184: other.container[i] still needs to be indexed?

jere8184:
> * You should really try to comment your code more (i.e. docstrings and member descriptions). It gets much harder to review your code if I also have to find out the purpose of your methods/members.
> * You should also document the new array curve type in the docs: https://github.com/SFTtech/openage/blob/master/doc/code/curves.md
> * The code looks very good so far, although I might have a few enhancement ideas after integrating the current review comments ^^

alright, currently implementing the changes, really appreciate all the feedback.

libopenage/curve/array.h
```diff
@@ -0,0 +1,259 @@
...
+	// Array::Iterator is used to iterate over KeyframeContainers contained in a curve at a given time.
+	class Iterator {
+	public:
+		Iterator(Array<T, Size> *curve, const time::time_t &time = time::TIME_MAX, size_t offset = 0) :
+			curve(curve), time(time), offset(offset) {};
+
+		// returns a copy of the keyframe at the current offset and time
+		const T operator*() {
+			return this->curve->frame(this->time, this->offset).second;
+		}
```
heinezen, line 125: 1. This method doesn't return a keyframe, it returns `T`
2. If it only returns `T`, you should use `at()` instead of `frame()` to retrieve the value.

libopenage/curve/array.h
heinezen: ah I didn't see that :see_no_evil:

heinezen (changes requested):
I think I found everything I wanted to comment on :D

libopenage/curve/array.h
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+
+	// get a copy to the KeyframeContainer at index
+	KeyframeContainer<T> operator[](size_t index) const {
+		return this->container.at(index);
+	}
+
+	// Array::Iterator is used to iterate over KeyframeContainers contained in a curve at a given time.
+	class Iterator {
```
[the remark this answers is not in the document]
jere8184: One reason would be that `Array::Iterator` is meant to iterate over a single time, not a range of times. Also  `std::_Array_const_iterator` doesn't have `base` as a member, which is referred to in `iterator.h line:95`.

 But to be honest when the choice was made it was arbitrary.

heinezen:
I'm going to do a few tests today and then see if we can merge :)

libopenage/curve/array.h
```diff
@@ -260,35 +260,50 @@ class Array : event::EventEntity {
...
                                                  const index_t index) const {
-	size_t &hint = this->last_element[index];
+	// find elem_ptr in container to get the last keyframe
+	auto hint = this->last_elements[index];
 	size_t frame_index = this->containers.at(index).last(t, hint);
-	hint = frame_index;
-	return this->containers.at(index).get(frame_index).make_pair();
+
+	// update the hint
+	this->last_elements[index] = frame_index;
```
jere8184, line 275: out of interest what was wrong with the origonal implementation?

heinezen: Nothing in particular. Your code still worked and was fine, in case you wondered about that.

The main motvation for this change was to change the type of the local variable `hint` from `size_t` to `auto` because I got some compiler hints about that. I think these were caused by me changing the element type in the `last_elements` to be `KeyframeContainer<T>::elem_ptr` so that it's more explicit. However, that means that the compiler has to deduce more layers of typedefs and check if they are compatible to `size_t` and that can apparently be a bit wonky when using the member in code, especially in templates. I could have declared `hint` as `KeyframeContainer<T>::elem_ptr` too, but changing the type to `auto` lets the compiler do this automatically and is much shorter to write.

heinezen:
I added a few more tests and fixed the docs.

heinezen (approved):

heinezen:
Thanks again for tackling this one :)

jere8184:
> Thanks again for tackling this one :)

No worries, thanks for the help.

PR #1725, opened 2024-12-04
Add dirty flag and integrate

### LINKED ISSUE #1679: Invalidate flow field cache entry when cost field changes
**Required Skills:** C++

**Difficulty:** Easy

The openage pathfinder uses flow fields to search for a path on the pathfinding grids (please read the [documentation](https://github.com/SFTtech/openage/blob/master/doc/code/pathfinding/README.md) to familiarize yourself with the concepts). To avoid unnecessary computations, fields are cached and reused for similar paths.

Currently, flow fields in the cache never get invalidated or updated when the costs of cells on the grid change, so these changes are never reflected in path requests when using the cached fields. This also means that dynamic changes to the grid are also not possible at the moment which is a huge downsight. A dynamic grid is required for a lot of gameplay features (e.g. placing buildings, change in terrain), so cache invalidation should be implement as soon as possible.

The easiest way to address this issue is to mark sectors/cost fields that have changed as "dirty" and then evict and recompute the flow fields for them when a new path is requested.

To try out the current pathfinder, check out pathfinding demo 1 by running the following command:

```
./run test -d pathfinding.tests.path_demo 1
```

Tasks:

- [ ] Add a way to mark the [`CostField`](https://github.com/SFTtech/openage/blob/master/libopenage/pathfinding/cost_field.h) as dirty. The easiest way to do this is to use a `bool` flag but you can also use the [`openage::time::time_t`](https://github.com/SFTtech/openage/blob/d09adb0901b15deb1b02b257ff437a283f694771/libopenage/time/time.h#L16) type to record the time of the last change.
- [ ] Add a method `is_dirty(..)` to  [`CostField`](https://github.com/SFTtech/openage/blob/master/libopenage/pathfinding/cost_field.h) and [`Sector`](https://github.com/SFTtech/openage/blob/master/libopenage/pathfinding/sector.h) that checks if the cost field or sector are dirty at the specified time.
- [ ] Check if the sector is dirty before looking up a cached field in the [`Integrator`](https://github.com/SFTtech/openage/blob/master/libopenage/pathfinding/integrator.cpp) class. If it is dirty, the cached value should be evicted and recomputed.
- [ ] (**optional**; Difficulty: medium) Implement the cache as a separate class `FieldCache` to make the code a bit more structured.

**Further Reading**

- [Pathfinder documentation](https://github.com/SFTtech/openage/blob/master/doc/code/pathfinding/README.md)
- [`openage::path` pathfinding module](https://github.com/SFTtech/openage/tree/master/libopenage/pathfinding)

dmwever:
fixes #1679. First time working with c++. Been struggling to update a test and getting it to build. Any advice will be helpful. I am using Visual Studio Code.

Please offer much critique. Complain with vigor. Correct with fervor.


dmwever:
A few questions I have for this:
First, how to I integrate the clock with the Map/Terrain/Pathfinding subsystem? I don't just want to start passing the clock through everything, but it seems like, with this task, as well as the Cost Stamp and Record History issues, time will be needed in the pathfinding system. How will this be integrated?

Also, would it be good practice to include non-time based overrides for `CostField set_cost` which simply set the time to `TIME_ZERO` for initialization? I have done this for the `CostField` function that is used by the `Map` class, but only to avoid needing to pass in a time_t from the map.

Lastly, how do I run with debugging using Visual Studio Code? I make changes to a file and run, but it ignores my breakpoints.

libopenage/pathfinding/path.h
```diff
@@ -20,6 +21,8 @@ struct PathRequest {
 	coord::tile start;
 	/// Target position of the path.
 	coord::tile target;
+	/// Time request was made.
```
heinezen, line 24:
```suggestion
	/// Time the request was made.
```

libopenage/pathfinding/path.h
```diff
@@ -34,6 +37,8 @@ struct Path {
 	/// First waypoint is the start position of the path request.
 	/// Last waypoint is the target position of the path request.
 	std::vector<coord::tile> waypoints;
+	/// Time path was created.
```
heinezen, line 40:
```suggestion
	/// Time the path was created.
```

libopenage/pathfinding/integrator.h
```diff
@@ -123,30 +124,11 @@ class Integrator {
...
-	 *
-	 * Key is the portal ID and the sector ID from which the field was entered. Fields that are cached are
-	 * cleared of dynamic flags, i.e. wavefront or LOS flags. These have to be recalculated
-	 * when the field is reused.
-	 */
-	std::unordered_map<std::pair<portal_id_t, sector_id_t>,
-	                   get_return_t,
-	                   pair_hash>
-		field_cache;
+	std::shared_ptr<FieldCache> field_cache;
```
heinezen, line 131: This member can be a unique pointer, since it's not used outside of the class.

heinezen, line 131: This member needs a docstring.

libopenage/pathfinding/sector.h
```diff
@@ -103,6 +104,13 @@ class Sector {
 	 */
 	void connect_exits();
 
+	/**
+	 * Check if the cost field is dirty at the specified time.
+	 * 
+	 * @param time Specified time to check.
+	 */
+	bool is_dirty(time::time_t time);
```
heinezen, line 112: `time` should be `const` and a reference

heinezen, line 112: `is_dirty` should also be `const`

heinezen (changes requested):

heinezen:
> First, how to I integrate the clock with the Map/Terrain/Pathfinding subsystem? I don't just want to start passing the clock through everything, but it seems like, with this task, as well as the Cost Stamp and Record History issues, time will be needed in the pathfinding system. How will this be integrated?

Where would you need to pass the clock? You very likely only need to pass the time value but the clock is not necessary IMO.

> Also, would it be good practice to include non-time based overrides for CostField set_cost which simply set the time to `TIME_ZERO` for initialization? I have done this for the CostField function that is used by the Map class, but only to avoid needing to pass in a time_t from the map.

That would kind of counteract introducing dirty flags, so I'm not sure if this is a good idea. In fact, I'm not really sure why we should just always pass a time. Initialization does not always happen at time zero.

> Lastly, how do I run with debugging using Visual Studio Code? I make changes to a file and run, but it ignores my breakpoints.

It should just work out of the box. What is your start configuration and output?

libopenage/pathfinding/cost_field.cpp
```diff
@@ -33,29 +34,42 @@ cost_t CostField::get_cost(size_t idx) const {
...
-void CostField::set_cost(size_t x, size_t y, cost_t cost) {
+void CostField::set_cost(size_t x, size_t y, cost_t cost, time::time_t &changed) {
 	this->cells[x + y * this->size] = cost;
+	this->changed = changed;
 }
 
 void CostField::set_cost(size_t idx, cost_t cost) {
 	this->cells[idx] = cost;
+	this->changed = time::TIME_ZERO;
+}
```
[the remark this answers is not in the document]
dmwever: See map.cpp, lines 50-52:
```
				auto sector = grid->get_sector(chunk_idx);
				auto cost_field = sector->get_cost_field();
				cost_field->set_cost(tile_idx, path_cost.second);
```
Not sure how else to initialize cost fields with a time stamp without passing some reference to the internal clock into map.cpp. So what I did was leave that method without a time parameter, setting the time to default to the start of the runtime. What route should be taken here to better resolve this?

libopenage/pathfinding/cost_field.cpp
```diff
@@ -33,29 +34,42 @@ cost_t CostField::get_cost(size_t idx) const {
...
 	       "cells vector has wrong size: " << cells.size()
 	                                       << "; expected: "
 	                                       << this->cells.size());
 
 	this->cells = std::move(cells);
+	this->changed = changed;
+}
+
+bool CostField::is_dirty(time::time_t &time) {
+	return time <= this->changed;
```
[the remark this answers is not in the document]
dmwever: That makes sense. I was thinking from the perspective of the cost field. If a time is passed in _before_ the latest change from the cost field's perspective, it would be an old value, e.g. dirty. But we oughtn't care about the cost field's feelings here, correct, only the caller's?

libopenage/pathfinding/cost_field.h
```diff
@@ -95,15 +107,25 @@ class CostField {
...
+	 * @param changed Time cost is set.
 	 */
-	void set_costs(std::vector<cost_t> &&cells);
+	void set_costs(std::vector<cost_t> &&cells, time::time_t &changed);
+
+	/**
+	 * Check if the cost field is dirty at the specified time.
+	 * 
+	 * @param time Specified time to check.
+	 */
```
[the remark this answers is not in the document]
dmwever: I am marking this as `Cost field is dirty if the cost field is accessed before the latest change to the cost field.`, but if we want to change this function to operate from the caller's POV, it will read:
`Cost field is dirty if the cost field is being accessed after the latest change to the cost field.`

dmwever:
> > First, how to I integrate the clock with the Map/Terrain/Pathfinding subsystem? I don't just want to start passing the clock through everything, but it seems like, with this task, as well as the Cost Stamp and Record History issues, time will be needed in the pathfinding system. How will this be integrated?
> 
> Where would you need to pass the clock? You very likely only need to pass the time value but the clock is not necessary IMO.

Perhaps that's a better question. How to get the time (or any time) into the Map subsystem.

libopenage/pathfinding/integrator.cpp
```diff
@@ -140,37 +156,47 @@ Integrator::get_return_t Integrator::get(const std::shared_ptr<CostField> &cost_
...
-	auto cached = this->field_cache.find(cache_key);
-	if (cached != this->field_cache.end()) {
+	if (evict_cache) {
+		this->field_cache->evict(cache_key);
+	}
+	if (this->field_cache->is_cached(cache_key)) {
 		log::log(DBG << "Using cached integration and flow fields for portal " << portal->get_id()
 		             << " from sector " << other_sector_id);
+
+
```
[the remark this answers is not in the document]
dmwever: Hmm. Got a link to help me with that? clang-format is new to me.

libopenage/pathfinding/integrator.cpp
```diff
@@ -36,24 +37,32 @@ std::shared_ptr<IntegrationField> Integrator::integrate(const std::shared_ptr<Co
...
-                                                        bool with_los) {
+                                                        bool with_los,
+											 			bool evict_cache) {
 	auto cache_key = std::make_pair(portal->get_id(), other_sector_id);
-	auto cached = this->field_cache.find(cache_key);
-	if (cached != this->field_cache.end()) {
+	if (evict_cache) {
+		this->field_cache->evict(cache_key);
+	}
+	if (this->field_cache->is_cached(cache_key)) {
```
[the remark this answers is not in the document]
dmwever: Maybe change to an else if?

libopenage/pathfinding/cost_field.cpp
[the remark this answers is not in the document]
dmwever: Let's say two requests are sent to a cost field after it is updated at time T: one at T + 1, and the other at time T + 3. Using (`>=`), the cost field returns dirty if the request happens after the cost field is changed. If both requests are from the integrator, then the integrator will be told that the cost field is dirty at time T + 1, and then again at time T + 3. But the cost field was only updated once. So with this approach, the cost field will be dirty for every request after time T.

Let's say that the cost field is called by the integrator 20 times after it was changed (Let's say 20 commands were given in a sector after a building was placed). That means the cache would be emptied 20 times!

So then is a time_t even the right thing to be using here?

libopenage/pathfinding/cost_field.cpp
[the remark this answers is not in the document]
heinezen: If it's for initialization, you can just pass `time::TIME_ZERO` from the map. I wouldn't worry about how the map does updates yet, since that is a different cattle of fish. However, you can assume for now that the update time will always be passed down the line (from the event that updates the map) and the clock is not necessary. You only need the clock in very few cases where it's necessary to know the current game time.

libopenage/pathfinding/cost_field.cpp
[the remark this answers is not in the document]
heinezen: Ah I see where you are coming from. I think the confusion comes mostly from how I designed the issue, since it leaves out a few things that would make it a "not-so-good-first-issue" :D

So the general idea is that that the integration/flow fields must be recomputed when the cost field becomes dirty. So at any time _before_ it becomes dirty (`t` < `changed`) the cache can still be used, while at any time _after_ it becomes dirty (`t` >= `changed`) the cache should be recomputed (once). I think renaming `changed` to `valid_until` might make more clear that the dirtyness is linked to cache invalidation. 

But you're right that the cache would currently be evicted every time where `t` >= `valid_until`. What's currently missing is a mechanism to clear the "dirty" status from the cost field after the cache was updated. We could clear the dirty status by adding a method that sets `valid_until` to `time::TIME_MAX`. Then `is_dirty` would always fail until another change happens.

libopenage/pathfinding/cost_field.h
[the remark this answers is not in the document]
heinezen: Probably related to my comment in https://github.com/SFTtech/openage/pull/1725#discussion_r1874183595

libopenage/pathfinding/integrator.cpp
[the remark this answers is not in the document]
heinezen: Yes that would make sense

libopenage/pathfinding/integrator.cpp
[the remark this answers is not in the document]
heinezen: Since you are using VSCode, you should be able to just go into your settings and activate **Format On Save**

heinezen:
> Perhaps that's a better question. How to get the time (or any time) into the Map subsystem.

The update time comes from events in the event system that will trigger a map update. However, I wouldn't worry about that right now, since it is not implemented yet.

libopenage/pathfinding/cost_field.cpp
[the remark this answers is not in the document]
dmwever: Nice, an elegant solution that allows us to use time instead of boolean. This will come in handy when we add cost stamps and tracking, as time will already be getting involved with the pathfinding system thanks to this update.

libopenage/pathfinding/cost_field.cpp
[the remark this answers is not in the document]
dmwever: This method could be made private then, if we are keeping it, to ensure that all calls to this class pass in a time value.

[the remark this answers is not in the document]
dmwever: Alternative solution being that we just remove this method, to solve the preceeding review where you are wondering why there is a method with no time parameter.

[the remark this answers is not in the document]
heinezen: The method should also get a time passed, I think.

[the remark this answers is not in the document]
heinezen: Or is there any reason why it shouldn't get one?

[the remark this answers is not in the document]
dmwever: Reason being that there would be an identical method if the time is passed. I will remove this method.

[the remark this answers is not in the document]
heinezen: A I see, then yeah there is no reason for this method.

dmwever:
> It should just work out of the box. What is your start configuration and output?

If you are talking about the configurations in launch.json, it says:
`"configurations": []`

That's probably my problem lol. What should it say?

heinezen:
Something like this should work as a configuration: 

```
{
    "name": "Main",
    "type": "cppdbg",
    "request": "launch",
    "cwd": "${workspaceFolder}/bin",
    "program": "${workspaceFolder}/bin/run",
    "args": [
        "main",
        "--verbose",
        "--gl-debug",
    ],
    "setupCommands": [
        {
            "description": "Enable pretty-printing for gdb",
            "text": "-enable-pretty-printing",
            "ignoreFailures": true
        },
        {
            "text": "source ${workspaceFolder}/etc/openage.gdbinit"
        }
    ],
}
```

dmwever:
> Something like this should work as a configuration:
> 
> ```
> {
>     "name": "Main",
>     "type": "cppdbg",
>     "request": "launch",
>     "cwd": "${workspaceFolder}/bin",
>     "program": "${workspaceFolder}/bin/run",
>     "args": [
>         "main",
>         "--verbose",
>         "--gl-debug",
>     ],
>     "setupCommands": [
>         {
>             "description": "Enable pretty-printing for gdb",
>             "text": "-enable-pretty-printing",
>             "ignoreFailures": true
>         },
>         {
>             "text": "source ${workspaceFolder}/etc/openage.gdbinit"
>         }
>     ],
> }
> ```

So to clarify, I should put this in launch.json? I am used to Visual Studio with C#.

heinezen:
You should put it in the `configurations` array, yes.

dmwever:
> You should put it in the `configurations` array, yes.

<workspace>/bin/run does not exist

heinezen:
Well, you need to build the project before running :D

heinezen:
https://github.com/SFTtech/openage/blob/master/doc/building.md has info on the build procedure.

libopenage/pathfinding/cost_field.h
```diff
@@ -95,15 +99,33 @@ class CostField {
...
+	 * Check if the cost field is dirty at the specified time.
+	 *
+	 * @param time Cost field is dirty if the cost field is accessed after the time given in valid_until.
+	 */
+	bool is_dirty(const time::time_t &time);
+
+	/**
+	 * Cleans the dirty flag by setting it to time_MAX.
+	 */
+	void clean();
```
heinezen, line 116: This method should be called `clear_dirty`

libopenage/pathfinding/cost_field.h
```diff
@@ -95,15 +99,33 @@ class CostField {
...
 private:
 	/**
 	 * Side length of the field.
 	 */
 	size_t size;
 
+	/**
+	 * Time the cost field was last changed.
+	 */
+	time::time_t &valid_until;
```
heinezen, line 127: This member should not be a reference

```suggestion
	time::time_t valid_until;
```

libopenage/pathfinding/integrator.cpp
```diff
@@ -36,24 +37,31 @@ std::shared_ptr<IntegrationField> Integrator::integrate(const std::shared_ptr<Co
...
                                                         const coord::tile_delta &target,
-                                                        bool with_los) {
+                                                        bool with_los,
+                                                        bool evict_cache) {
 	auto cache_key = std::make_pair(portal->get_id(), other_sector_id);
-	auto cached = this->field_cache.find(cache_key);
-	if (cached != this->field_cache.end()) {
+	if (evict_cache) {
+		this->field_cache->evict(cache_key);
+	}
```
heinezen, line 45: Eviction should not be triggered by a parameter but by using a check of `cost_field->is_dirty()`

libopenage/pathfinding/integrator.cpp
```diff
@@ -97,25 +105,32 @@ std::shared_ptr<FlowField> Integrator::build(const std::shared_ptr<IntegrationFi
...
                                              const std::shared_ptr<Portal> &portal,
-                                             bool with_los) {
+                                             bool with_los,
+                                             bool evict_cache) {
 	auto cache_key = std::make_pair(portal->get_id(), other_sector_id);
-	auto cached = this->field_cache.find(cache_key);
-	if (cached != this->field_cache.end()) {
+	if (evict_cache) {
+		this->field_cache->evict(cache_key);
+	}
```
heinezen, line 113: same here

heinezen (changes requested):

dmwever:
> https://github.com/SFTtech/openage/blob/master/doc/building.md has info on the build procedure.

I can't seem to get this process working on windows. I can run using python openage, following the info on this page:
https://github.com/SFTtech/openage/blob/master/doc/build_instructions/windows_msvc.md

But I don't see a bin folder built in openage following the Windows instructions.

jere8184:
> > https://github.com/SFTtech/openage/blob/master/doc/building.md has info on the build procedure.
> 
> I can't seem to get this process working on windows. I can run using python openage, following the info on this page: https://github.com/SFTtech/openage/blob/master/doc/build_instructions/windows_msvc.md
> 
> But I don't see a bin folder built in openage following the Windows instructions.

I can walk you through it in at [matrix ](https://app.element.io/#/room/#sfttech:matrix.org) if you want

libopenage/pathfinding/integrator.cpp
dmwever: The integrate function does not pass in the cost field. I am going to remove the dirty check from the integrate function for now, unless we want to add dirty flags to the integration fields as well.

heinezen: `integrate` receives `const std::shared_ptr<CostField> &cost_field` as its first argument, so you can use it for checking whether the field is dirty.

libopenage/pathfinding/cost_field.h
```diff
@@ -64,25 +65,32 @@ class CostField {
 	 *
 	 * @param pos Coordinates of the cell (relative to field origin).
 	 * @param cost Cost to set.
+	 * @param changed Time at which the cost value is changed.
```
heinezen, line 68: Remember to change the argument name in the docstring too! ;)

```suggestion
	 * @param valid_until Time at which the cost value is changed.
```

libopenage/pathfinding/cost_field.h
```diff
@@ -64,25 +65,32 @@ class CostField {
...
 	 * @param idx Index of the cell.
 	 * @param cost Cost to set.
+	 * @param changed Time at which the cost value is changed.
 	 */
-	void set_cost(size_t idx, cost_t cost);
+
+	inline void set_cost(size_t idx, cost_t cost, const time::time_t &until) {
+		cells[idx] = cost;
+		valid_until = until;
+	}
```
heinezen, line 93: This method should be documented.

libopenage/pathfinding/cost_field.h
```diff
@@ -64,25 +65,32 @@ class CostField {
...
 	 *
 	 * @param idx Index of the cell.
 	 * @param cost Cost to set.
+	 * @param changed Time at which the cost value is changed.
 	 */
-	void set_cost(size_t idx, cost_t cost);
+
+	inline void set_cost(size_t idx, cost_t cost, const time::time_t &until) {
+		cells[idx] = cost;
+		valid_until = until;
```
heinezen, line 92: You should use `this` to access member variables.

```suggestion
		this->cells[idx] = cost;
		this->valid_until = until;
```

libopenage/pathfinding/cost_field.h
```diff
@@ -64,25 +65,32 @@ class CostField {
...
 	/**
 	 * Set the cost at a specified position.
 	 *
 	 * @param idx Index of the cell.
 	 * @param cost Cost to set.
+	 * @param changed Time at which the cost value is changed.
 	 */
-	void set_cost(size_t idx, cost_t cost);
+
+	inline void set_cost(size_t idx, cost_t cost, const time::time_t &until) {
```
heinezen, line 90: The argument should also be called `valid_until` to be in line with the other methods.

```suggestion
	inline void set_cost(size_t idx, cost_t cost, const time::time_t &valid_until) {
```

libopenage/pathfinding/integrator.cpp
dmwever: sorry, I meant build()

libopenage/pathfinding/cost_field.h
```diff
@@ -64,25 +65,32 @@ class CostField {
...
-	void set_cost(const coord::tile_delta &pos, cost_t cost);
+	void set_cost(const coord::tile_delta &pos, cost_t cost, const time::time_t &valid_until);
 
 	/**
 	 * Set the cost at a specified position.
 	 *
 	 * @param x X-coordinate of the cell.
 	 * @param y Y-coordinate of the cell.
 	 * @param cost Cost to set.
+	 * @param changed Time at which the cost value is changed.
```
heinezen, line 78:
```suggestion
	 * @param valid_until Time at which the cost value is changed.
```

libopenage/pathfinding/cost_field.h
```diff
@@ -64,25 +65,32 @@ class CostField {
...
 	 */
-	void set_cost(size_t x, size_t y, cost_t cost);
+	void set_cost(size_t x, size_t y, cost_t cost, const time::time_t &valid_until);
 
 	/**
 	 * Set the cost at a specified position.
 	 *
 	 * @param idx Index of the cell.
 	 * @param cost Cost to set.
+	 * @param changed Time at which the cost value is changed.
```
heinezen, line 87:
```suggestion
	 * @param valid_until Time at which the cost value is changed.
```

libopenage/pathfinding/cost_field.h
```diff
@@ -64,25 +65,32 @@ class CostField {
...
 	 */
-	void set_cost(size_t idx, cost_t cost);
+
+	inline void set_cost(size_t idx, cost_t cost, const time::time_t &valid_until) {
```
heinezen, line 90: You need to remove the newline, otherwise the docstring will not be associated with the method.

```suggestion
	inline void set_cost(size_t idx, cost_t cost, const time::time_t &valid_until) {
```

libopenage/pathfinding/cost_field.h
```diff
@@ -95,15 +103,33 @@ class CostField {
 	 * Set the cost field values.
 	 *
 	 * @param cells Cost field values.
+	 * @param changed Time at which the cost value is changed.
+	 */
+	void set_costs(std::vector<cost_t> &&cells, const time::time_t &changed);
```
heinezen, line 108:
```suggestion
	 * @param valid_until Time at which the cost value is changed.
	 */
	void set_costs(std::vector<cost_t> &&cells, const time::time_t &valid_until);
```

libopenage/pathfinding/cost_field.h
```diff
@@ -95,15 +103,33 @@ class CostField {
...
+
+	/**
+	 * Check if the cost field is dirty at the specified time.
+	 *
+	 * @param time Cost field is dirty if the cost field is accessed after the time given in valid_until.
```
heinezen, line 113: You should always explain what `time` represents/is used for. From your docstring, it's not apparent what a user of this method should pass for `time`.

```suggestion
	 * @param time Time of access to the cost field.
```

libopenage/pathfinding/cost_field.h
```diff
@@ -95,15 +103,33 @@ class CostField {
...
 	 */
-	void set_costs(std::vector<cost_t> &&cells);
+	bool is_dirty(const time::time_t &time);
```
heinezen, line 115: Please also document what the return value does.

libopenage/pathfinding/cost_field.h
```diff
@@ -95,15 +103,33 @@ class CostField {
...
+
+	/**
+	 * Cleans the dirty flag by setting it to time_MAX.
```
heinezen, line 118: The internal mechanism of tracking the dirty status is not really relevant for the user of this method. Therefore, you can remove it from the docstring.

```suggestion
	 * Clear the dirty flag.
```

libopenage/pathfinding/demo/demo_1.cpp
```diff
@@ -27,11 +29,14 @@ namespace openage::path::tests {
 void path_demo_1(const util::Path &path) {
 	auto grid = std::make_shared<Grid>(0, util::Vector2s{4, 3}, 10);
 
+	time::Clock clock = time::Clock();
+	clock.start();
+	const time::time_t time = clock.get_time();
```
heinezen, line 34: You don't need a clock in this demo because you can just pass `0` as the time in all calls.

libopenage/pathfinding/integrator.h
```diff
@@ -56,6 +56,7 @@ class Integrator {
 	 * @param other_sector_id Sector ID of the other side of the portal.
 	 * @param portal Portal.
 	 * @param target Coordinates of the target cell, relative to the integration field origin.
+	 * @param time The time to check is the cached cost field is dirty.
```
heinezen, line 59: This description is not really useful. How should the user of this method know what time value they have to pass? Caching and dirty checks are also not controlled by the user and therefore don't have to be mentioned.

```suggestion
	 * @param time Time of the path request.
```

libopenage/pathfinding/integrator.h
```diff
@@ -114,6 +116,7 @@ class Integrator {
...
```
heinezen, line 119: same here

```suggestion
	 * @param time Time of the path request.
```

heinezen (changes requested):

libopenage/pathfinding/demo/demo_1.cpp
```diff
@@ -25,13 +27,17 @@
 namespace openage::path::tests {
 
 void path_demo_1(const util::Path &path) {
+	auto time_loop = std::make_shared<time::TimeLoop>();
+	time_loop->run();
+	auto clock = time_loop->get_clock();
```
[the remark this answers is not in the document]
dmwever: Picked up later in thread.

libopenage/pathfinding/demo/demo_1.cpp
dmwever: Demo currently seg faults. Still can't debug while testing; I only can use the ./run command found in the [Windows Build Instructions](https://github.com/SFTtech/openage/blob/master/doc/build_instructions/windows_msvc.md) page.

dmwever: Seg faults here:
```
bool FieldCache::is_cached(cache_key_t cache_key) {
	return this->cache.contains(cache_key);
}
```
The cache is the following: 
```
std::unordered_map<cache_key_t,
	                   field_cache_t,
	                   pair_hash>
		cache;
```

It could be the key causing the seg fault or the cache, not sure

dmwever: Or something else a more experienced C++er would know.

I still can't figure out how to debug while testing.

dmwever:
Marking as ready for review, as most of the ground work is laid. There is a seg fault on demo 1 I cannot track using breakpoints, because I cannot get debugging to work with the demo tests using the [Windows Build Instructions](https://github.com/SFTtech/openage/blob/master/doc/build_instructions/windows_msvc.md) page. I assume there is a simple error I am making with c++ unordered maps, but I cannot use breakpoint tools at the moment.

The seg fault occurs when accessing the Field Cache in the is_cached method. Aside from that, we are looking pretty good here.

libopenage/pathfinding/demo/demo_1.cpp
dmwever: Carrying this on in the main thread.

jere8184:
> because I cannot get debugging to work with the demo tests using the

Building via 

> Marking as ready for review, as most of the ground work is laid. There is a seg fault on demo 1 I cannot track using breakpoints, because I cannot get debugging to work with the demo tests using the [Windows Build Instructions](https://github.com/SFTtech/openage/blob/master/doc/build_instructions/windows_msvc.md) page. I assume there is a simple error I am making with c++ unordered maps, but I cannot use breakpoint tools at the moment.
> 
> The seg fault occurs when accessing the Field Cache in the is_cached method. Aside from that, we are looking pretty good here.


cmake --build . --config RelWithDebInfo -- /nologo /m /v:m should produce debug symbols that you can then provide to your debugger, to allow you to use breakpoints , if you want further assistance feel free to reach out on matrix :)

heinezen:
I fixed your segfault and a few other things. The segfault happened because the field cache was not initialized in the `Integrator`'s constructor.

The only thing left to do for you now is to add yourself (and your commit mail) to [copying.md](https://github.com/SFTtech/openage/blob/master/copying.md) and we can merge :)

heinezen (dismissed):

dmwever:
Added to copying.md!

dmwever:
My old email is causing a sanity check error. I added it to .mailmap, but no dice.

dmwever:
> My old email is causing a sanity check error. I added it to .mailmap, but no dice.

Fixed

.mailmap
```diff
@@ -20,4 +20,5 @@ Tobias Feldballe <tobias@osandweb.dk> <tobi.fp@gmail.com>
 Tobias Feldballe <tobias@osandweb.dk> <tf@jumpstory.com>
 Jonas Borchelt <jonas.borchelt@connected.link>
 Derek Frogget <fro22003@byui.edu> <114030121+derekfrogget@users.noreply.github.com>
-Nikhil Ghosh <nghosh606@gmail.com>
\ No newline at end of file
+Nikhil Ghosh <nghosh606@gmail.com>
+David Wever <dmwever@crimson.ua.edu> <56411717+dmwever@users.noreply.github.com>
```
TheJJ, line 24: you don't need this entry if you update and fix your mail/author name in all the commits of this pull request. since they are not merged yet, this can easily be done - the file is for commits that have been merged already :)

heinezen: I think both mails are in the commits. The GitHub one is used when you apply suggestions to PRs in the web interface.

heinezen (approved):
Here we go!

PR #1726, opened 2024-12-04
Add CLI Support for Window Parameters

### LINKED ISSUE #1546: CLI parameters for window settings

ZzzhHe:
Close #1546 

+ Defined window-related startup parameters (size, vsync, mode) in Python for `main` and `game` start modes.
+ Passed these parameters to the `main_arguments` struct by creating a `window_arguments` struct
+ Updated C++ logic to forward parameters along `main.cpp` -> `engine.cpp` -> `presenter.cpp`
+ Implemented Fullscreen, Borderless, Windowed three display model in `libopenage/renderer/opengl/window.cpp`

### DIFF
--- a/libopenage/engine/engine.h
+++ b/libopenage/engine/engine.h
@@ -7,9 +7,9 @@
 #include <thread>
 #include <vector>
 
+#include "renderer/window.h"
 #include "util/path.h"
 
-#include <renderer/window.h>
 
 // TODO: Remove custom jthread definition when clang/libc++ finally supports it
 #if __llvm__
@@ -73,13 +73,11 @@ public:
 	 * @param mode The run mode to use.
 	 * @param root_dir openage root directory.
 	 * @param mods The mods to load.
-	 * @param debug_graphics If true, enable OpenGL debug logging.
-	 * @param window_settings window display setting
+	 * @param window_settings The settings to customize the display window (e.g. size, display mode, vsync).
 	 */
 	Engine(mode mode,
 	       const util::Path &root_dir,
 	       const std::vector<std::string> &mods,
-	       bool debug_graphics = false,
 	       const renderer::window_settings &window_settings = {});
 
 	// engine should not be copied or moved
--- a/libopenage/engine/engine.h
+++ b/libopenage/engine/engine.h
@@ -9,6 +9,8 @@
 
 #include "util/path.h"
 
+#include <renderer/window.h>
+
 // TODO: Remove custom jthread definition when clang/libc++ finally supports it
 #if __llvm__
 	#if !__cpp_lib_jthread
@@ -72,11 +74,13 @@ public:
 	 * @param root_dir openage root directory.
 	 * @param mods The mods to load.
 	 * @param debug_graphics If true, enable OpenGL debug logging.
+	 * @param window_settings window display setting
 	 */
 	Engine(mode mode,
 	       const util::Path &root_dir,
 	       const std::vector<std::string> &mods,
-	       bool debug_graphics = false);
+	       bool debug_graphics = false,
+	       const renderer::window_settings &window_settings = {});
 
 	// engine should not be copied or moved
 	Engine(const Engine &) = delete;
--- a/libopenage/main.cpp
+++ b/libopenage/main.cpp
@@ -44,12 +44,16 @@ int run_game(const main_arguments &args) {
 	else if (args.window_args.mode == "borderless") {
 		wmode = renderer::window_mode::BORDERLESS;
 	}
-	else {
+	else if (args.window_args.mode == "windowed") {
 		wmode = renderer::window_mode::WINDOWED;
 	}
+	else {
+		throw Error(MSG(err) << "Invalid window mode: " << args.window_args.mode);
+	}
 	win_settings.mode = wmode;
+	win_settings.debug = args.gl_debug;
 
-	openage::engine::Engine engine{run_mode, args.root_path, args.mods, args.gl_debug, win_settings};
+	openage::engine::Engine engine{run_mode, args.root_path, args.mods, win_settings};
 
 	engine.loop();
 
--- a/libopenage/main.cpp
+++ b/libopenage/main.cpp
@@ -1,4 +1,4 @@
-// Copyright 2015-2023 the openage authors. See copying.md for legal info.
+// Copyright 2015-2024 the openage authors. See copying.md for legal info.
 
 #include "main.h"
 
@@ -31,7 +31,25 @@ int run_game(const main_arguments &args) {
 		run_mode = openage::engine::Engine::mode::HEADLESS;
 	}
 
-	openage::engine::Engine engine{run_mode, args.root_path, args.mods, args.gl_debug};
+	// convert window arguments to window settings
+	renderer::window_settings win_settings = {};
+	win_settings.width = args.window_args.width;
+	win_settings.height = args.window_args.height;
+	win_settings.vsync = args.window_args.vsync;
+
+	renderer::window_mode wmode;
+	if (args.window_args.mode == "fullscreen") {
+		wmode = renderer::window_mode::FULLSCREEN;
+	}
+	else if (args.window_args.mode == "borderless") {
+		wmode = renderer::window_mode::BORDERLESS;
+	}
+	else {
+		wmode = renderer::window_mode::WINDOWED;
+	}
+	win_settings.mode = wmode;
+
+	openage::engine::Engine engine{run_mode, args.root_path, args.mods, args.gl_debug, win_settings};
 
 	engine.loop();
 
--- a/libopenage/presenter/presenter.h
+++ b/libopenage/presenter/presenter.h
@@ -7,6 +7,9 @@
 
 #include "util/path.h"
 
+#include <renderer/window.h>
+
+
 namespace qtgui {
 class GuiApplication;
 }
@@ -88,8 +91,9 @@ public:
 	 * Start the presenter and initialize subsystems.
 	 *
 	 * @param debug_graphics If true, enable OpenGL debug logging.
+	 * @param window_settings window display setting
 	 */
-	void run(bool debug_graphics = false);
+	void run(bool debug_graphics = false, const renderer::window_settings &window_settings = {});
 
 	/**
 	 * Set the game simulation controlled by this presenter.
@@ -120,7 +124,7 @@ protected:
 	 *     - main renderer
 	 *     - component renderers (Terrain, Game Entities, GUI)
 	 */
-	void init_graphics(bool debug = false);
+	void init_graphics(bool debug = false, const renderer::window_settings &window_settings = {});
 
 	/**
 	 * Initialize the GUI.
--- a/libopenage/presenter/presenter.h
+++ b/libopenage/presenter/presenter.h
@@ -5,9 +5,8 @@
 #include <memory>
 #include <vector>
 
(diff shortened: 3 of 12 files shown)


libopenage/engine/engine.h
```diff
@@ -72,11 +74,13 @@ class Engine {
 	 * @param root_dir openage root directory.
 	 * @param mods The mods to load.
 	 * @param debug_graphics If true, enable OpenGL debug logging.
+	 * @param window_settings window display setting
```
heinezen, line 77: You should describe what the parameter is used for. What you wrote is already obvious from the name.

Also, you should keep the comment style consistent (first letter capitalized, dot at the end).

libopenage/engine/engine.h
```diff
@@ -72,11 +74,13 @@ class Engine {
...
 	 */
 	Engine(mode mode,
 	       const util::Path &root_dir,
 	       const std::vector<std::string> &mods,
-	       bool debug_graphics = false);
+	       bool debug_graphics = false,
+	       const renderer::window_settings &window_settings = {});
```
heinezen, line 83: `debug_graphics` can now be removed, since it is already part of `window_settings`.

libopenage/main.cpp
```diff
@@ -31,7 +31,25 @@ int run_game(const main_arguments &args) {
...
+		wmode = renderer::window_mode::FULLSCREEN;
+	}
+	else if (args.window_args.mode == "borderless") {
+		wmode = renderer::window_mode::BORDERLESS;
+	}
+	else {
+		wmode = renderer::window_mode::WINDOWED;
+	}
+	win_settings.mode = wmode;
+
```
heinezen, line 51: You should initialize `win_settings.debug` with `args.gl_debug`

libopenage/main.cpp
```diff
@@ -31,7 +31,25 @@ int run_game(const main_arguments &args) {
...
+	renderer::window_mode wmode;
+	if (args.window_args.mode == "fullscreen") {
+		wmode = renderer::window_mode::FULLSCREEN;
+	}
+	else if (args.window_args.mode == "borderless") {
+		wmode = renderer::window_mode::BORDERLESS;
+	}
+	else {
+		wmode = renderer::window_mode::WINDOWED;
+	}
```
heinezen, line 49: This should also be an `else if` branch and there should be an additional `else` branch that throws a `openage::error::Error` due to an unrecognized window mode.

libopenage/presenter/presenter.h
```diff
@@ -7,6 +7,9 @@
 
 #include "util/path.h"
 
+#include <renderer/window.h>
```
heinezen, line 10:
```suggestion
#include "renderer/window.h"
#include "util/path.h"
```

libopenage/presenter/presenter.h
```diff
@@ -88,8 +91,9 @@ class Presenter {
 	 * Start the presenter and initialize subsystems.
 	 *
 	 * @param debug_graphics If true, enable OpenGL debug logging.
+	 * @param window_settings window display setting
```
heinezen, line 94: Same for this docstring, see above.

libopenage/presenter/presenter.h
```diff
@@ -88,8 +91,9 @@ class Presenter {
...
 	 */
-	void run(bool debug_graphics = false);
+	void run(bool debug_graphics = false, const renderer::window_settings &window_settings = {});
```
heinezen, line 96: `debug_graphics` should be removed

libopenage/renderer/window.h
```diff
@@ -35,6 +44,8 @@ struct window_settings {
 	bool vsync = true;
 	// If true, enable debug logging for the selected backend.
 	bool debug = false;
+	// Display mode for the window.
+	window_mode mode = window_mode::FULLSCREEN;
```
heinezen, line 48:
```suggestion
	window_mode mode = window_mode::WINDOWED;
```

The default should be windowed as this has been the previous default.

libopenage/presenter/presenter.h
```diff
@@ -120,7 +124,7 @@ class Presenter {
 	 *     - main renderer
 	 *     - component renderers (Terrain, Game Entities, GUI)
 	 */
-	void init_graphics(bool debug = false);
+	void init_graphics(bool debug = false, const renderer::window_settings &window_settings = {});
```
heinezen, line 127: `debug_graphics` should be removed

libopenage/renderer/opengl/window.cpp
```diff
@@ -57,6 +57,21 @@ GlWindow::GlWindow(const std::string &title,
...
+		break;
+	case window_mode::BORDERLESS:
+		this->window->setFlags(this->window->flags() | Qt::FramelessWindowHint);
+		this->window->show();
+		break;
+	case window_mode::WINDOWED:
+	default:
+		this->window->showNormal();
+		break;
+	}
```
heinezen, line 73: You should not use the `show*` methods since they also set the window to visible (which we do not necessarily want). It's better to use the `QWindow::setWindowStates` method.

heinezen, line 73: `default` should throw an error for an unrecognized window mode.

openage/game/main.py
```diff
@@ -36,6 +36,19 @@ def init_subparser(cli: ArgumentParser) -> None:
         help="Check if the assets are up to date"
     )
 
+    cli.add_argument(
+        "--window-size", nargs=2, type=int, default=[1024, 768],
+        metavar=('WIDTH', 'HEIGHT'),
+        help="initial window size in pixels, e.g., --window-size 1024 768")
```
heinezen, line 42:
```suggestion
        help="Initial window size in pixels")
```

You don't need the example because Python's argparse will show enough detail.

openage/game/main.py
```diff
@@ -36,6 +36,19 @@ def init_subparser(cli: ArgumentParser) -> None:
...
+
+    cli.add_argument(
+        "--vsync", action='store_true',
+        help="enable vertical synchronization")
+
+    cli.add_argument(
+        "--window-mode", choices=["fullscreen", "borderless", "windowed"], default="windowed",
+        help="set the window mode: fullscreen, borderless, or windowed (default)")
```
heinezen, line 50:
```suggestion
        help="Set the window mode")
```

same here

openage/game/main.py
```diff
@@ -36,6 +36,19 @@ def init_subparser(cli: ArgumentParser) -> None:
...
     )
 
+    cli.add_argument(
+        "--window-size", nargs=2, type=int, default=[1024, 768],
+        metavar=('WIDTH', 'HEIGHT'),
+        help="initial window size in pixels, e.g., --window-size 1024 768")
+
+    cli.add_argument(
+        "--vsync", action='store_true',
+        help="enable vertical synchronization")
```
heinezen, line 46:
```suggestion
        help="Enable vertical synchronization")
```

openage/main/main.py
```diff
@@ -28,6 +28,19 @@ def init_subparser(cli: ArgumentParser):
         "--modpacks", nargs="+", type=str,
         help="list of modpacks to load")
 
+    cli.add_argument(
+        "--window-size", nargs=2, type=int, default=[1024, 768],
+        metavar=('WIDTH', 'HEIGHT'),
+        help="initial window size in pixels, e.g., --window-size 1024 768")
```
heinezen, line 34:
```suggestion
        help="Initial window size in pixels")
```

see above

openage/main/main.py
```diff
@@ -28,6 +28,19 @@ def init_subparser(cli: ArgumentParser):
...
+
+    cli.add_argument(
+        "--vsync", action='store_true',
+        help="enable vertical synchronization")
```
heinezen, line 38:
```suggestion
        help="Enable vertical synchronization")
```

see above

openage/main/main.py
```diff
@@ -28,6 +28,19 @@ def init_subparser(cli: ArgumentParser):
...
+
+    cli.add_argument(
+        "--window-mode", choices=["fullscreen", "borderless", "windowed"], default="windowed",
+        help="set the window mode: fullscreen, borderless, or windowed (default)")
```
heinezen, line 42:
```suggestion
        help="Set the window mode")
```

see above

heinezen (changes requested):
Overall very good, thank you!

I just have a few requests.

ZzzhHe:
> Overall very good, thank you!
> 
> I just have a few requests.

Thanks!! done.

libopenage/renderer/opengl/window.cpp
```diff
@@ -57,6 +57,23 @@ GlWindow::GlWindow(const std::string &title,
 	this->window->setFormat(format);
 	this->window->create();
 
+	// set display mode
+	// Reset to a known state
+	this->window->setWindowState(Qt::WindowNoState);
+	switch (settings.mode) {
+	case window_mode::WINDOWED:
+		this->window->setFlags(this->window->flags() & ~Qt::FramelessWindowHint);
```
heinezen, line 65: I think you don't even need this line? Not having the `Qt::FramelessWindowHint` is the default I think.

heinezen (changes requested):
I found two small issues but then we are good to go I believe.

ZzzhHe:
> I found two small issues but then we are good to go I believe.

Thanks for pointing that out! I get it now—I'm still getting familiar with multi-threaded code.

libopenage/presenter/presenter.h
```diff
@@ -87,9 +89,9 @@ class Presenter {
 	/**
 	 * Start the presenter and initialize subsystems.
 	 *
-	 * @param debug_graphics If true, enable OpenGL debug logging.
+	 * @param window_settings The settings to customize the display window (e.g. size, display mode, vsync).
 	 */
-	void run(bool debug_graphics = false);
+	void run(renderer::window_settings window_settings = {});
```
heinezen, line 94:
```suggestion
	void run(const renderer::window_settings window_settings = {});
```

The passed argument should still be const as it is not supposed to be modified by the presenter.

heinezen (changes requested):

heinezen (approved):

libopenage/presenter/presenter.h
heinezen: fixed it myself, so we can merge :)

heinezen:
Look nice with `./run main --window-mode borderless --vsync --window-size 1600 900`

grafik

libopenage/presenter/presenter.h
ZzzhHe: Thanks!  I got the hint "Clang-Tidy: Parameter 'window_settings' is const-qualified in the function declaration; const-qualification of parameters only has an effect in function definitions" here, so then I removed this `const`.

PR #1734, opened 2024-12-27
Bind empty VAO to bufferless quad rendering

### LINKED ISSUE #1659: "`GL_INVALID_OPERATION`: Array object is not active" when rendering bufferless quad
### Description

OpenGL generates an error when a bufferless quad is drawn but no VAO is bound. This can happen in the renderer here (line 57):

https://github.com/SFTtech/openage/blob/29df8ac2a9ca164d52f27bbc05f7867ec0ba6560/libopenage/renderer/opengl/geometry.cpp#L56-L58

Even though `glDrawArrays` doesn't read from a buffer, it requires a VAO as stated in the  [OpenGL Wiki](https://www.khronos.org/opengl/wiki/Vertex_Rendering):

> A non-zero [Vertex Array Object](https://www.khronos.org/opengl/wiki/Vertex_Array_Object) must be bound (though no arrays have to be enabled, so it can be a freshly-created vertex array object).

### How To Reproduce

1. Create a `GlGeometry` object with a bufferless quad, e.g. with `renderer::opengl::GlRenderer::add_bufferless_quad()`
2. Create a renderable with the geometry
3. Add the renderable to a render pass
4. Render the render pass and observe the error

### Expected Behavior

An empty VAO should be created and bound when trying to render a bufferless quad.

ZzzhHe:
Hey! 👋  resolve #1659

I've made some initial fixes for the bufferless quad VAO issue, but wanted to get some advice before finalizing:

1. I tested my current code and noticed the VAO error was still there but occurred during shader validation in [`GlShaderProgram::use()`](https://github.com/SFTtech/openage/blob/beb894520e552cb6224cdd2f64697b2cf6e4ad8a/libopenage/renderer/opengl/shader_program.cpp#L295C1-L302C3). 

    This happens during `program->update_uniforms(in);` before `geom->draw()`; in `GlRenderer::render()`.

3. My current fix is binding the empty VAO in `geometry.cpp`, but I'm wondering:
  - Should we bind VAO before shader program validation?
  - Or should we handle/modify the validation error check in `GlShaderProgram::use()`?

appreciate your thoughts on the best approach!

### DIFF
--- a/libopenage/renderer/opengl/geometry.cpp
+++ b/libopenage/renderer/opengl/geometry.cpp
@@ -1,4 +1,4 @@
-// Copyright 2015-2024 the openage authors. See copying.md for legal info.
+// Copyright 2015-2025 the openage authors. See copying.md for legal info.
 
 #include "geometry.h"
 
@@ -54,6 +54,7 @@ void GlGeometry::update_verts_offset(std::vector<uint8_t> const &verts, size_t o
 void GlGeometry::draw() const {
 	switch (this->get_type()) {
 	case geometry_t::bufferless_quad:
+		// any VAO must be bound before this call
 		glDrawArrays(GL_TRIANGLE_STRIP, 0, 4);
 		break;
 
--- a/libopenage/renderer/opengl/renderer.cpp
+++ b/libopenage/renderer/opengl/renderer.cpp
@@ -1,4 +1,4 @@
-// Copyright 2017-2024 the openage authors. See copying.md for legal info.
+// Copyright 2017-2025 the openage authors. See copying.md for legal info.
 
 #include "renderer.h"
 
@@ -15,6 +15,7 @@
 #include "renderer/opengl/texture.h"
 #include "renderer/opengl/uniform_buffer.h"
 #include "renderer/opengl/uniform_input.h"
+#include "renderer/opengl/vertex_array.h"
 #include "renderer/opengl/window.h"
 #include "renderer/resources/buffer_info.h"
 
@@ -26,7 +27,8 @@ GlRenderer::GlRenderer(const std::shared_ptr<GlContext> &ctx,
 	gl_context{ctx},
 	display{std::make_shared<GlRenderTarget>(ctx,
                                              viewport_size[0],
-                                             viewport_size[1])} {
+                                             viewport_size[1])},
+	shared_quad_vao{std::make_shared<GlVertexArray>(ctx)} {
 	// color used to clear the color buffers
 	glClearColor(0.0f, 0.0f, 0.0f, 0.0f);
 
@@ -100,7 +102,7 @@ std::shared_ptr<UniformBuffer> GlRenderer::add_uniform_buffer(resources::Uniform
 		                     resources::UniformBufferInfo::get_size(input, info.get_layout()),
 		                     resources::UniformBufferInfo::get_stride_size(input.type, info.get_layout()),
 		                     input.count,
-							 input.name});
+		                     input.name});
 
 		offset += size;
 	}
@@ -168,6 +170,11 @@ void GlRenderer::check_error() {
 void GlRenderer::render(const std::shared_ptr<RenderPass> &pass) {
 	auto gl_target = std::dynamic_pointer_cast<GlRenderTarget>(pass->get_target());
 	gl_target->bind_write();
+
+	// ensure that an (empty) VAO is bound before rendering geometry
+	// a bound VAO is required to render bufferless quad geometries by OpenGL
+	// see https://www.khronos.org/opengl/wiki/Vertex_Rendering#Causes_of_rendering_failure
+	shared_quad_vao->bind();
 
 	glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
 
--- a/libopenage/renderer/opengl/renderer.h
+++ b/libopenage/renderer/opengl/renderer.h
@@ -1,4 +1,4 @@
-// Copyright 2017-2024 the openage authors. See copying.md for legal info.
+// Copyright 2017-2025 the openage authors. See copying.md for legal info.
 
 #pragma once
 
@@ -17,6 +17,7 @@ namespace opengl {
 class GlContext;
 class GlRenderPass;
 class GlRenderTarget;
+class GlVertexArray;
 class GlWindow;
 
 /// The OpenGL specialization of the rendering interface.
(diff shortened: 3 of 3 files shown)


heinezen:
Hey thanks a lot!

I read through this while on the train today, so I have not tried the code out yet. But here are my thoughts on the approach:

Defining `quad_vao` for every `GlGeometry` class is likely too much as the member is useless for anything that is not a bufferless quad. Usually having `std::optional` variables is not a problem, but the game will instance a lot of `GlGeometry` objects over its runtime. Also, OpenGL only allows a certain number of vertex arrays to be defined I think. So it's worth to think about a better solution here.

Making the VAO `static` would be nice, but since we need the OpenGL context I doubt it's possible. I think the least dirty solution would be:

1. Create a single VAO somewhere else (probably in the `GlRenderer` initialization)
2. Bind it at the start of every `render(..)` call
3. Keep the code in `GlGeometry` as-is, except for making a comment that a VAO needs to be bound in the switch statement branch for bufferless quad.

This ensures that a VAO is always bound during rendering and should also be much more storage/runtime efficient.

heinezen:
> I tested my current code and noticed the VAO error was still there but occurred during shader validation in [GlShaderProgram::use()](https://github.com/SFTtech/openage/blob/beb894520e552cb6224cdd2f64697b2cf6e4ad8a/libopenage/renderer/opengl/shader_program.cpp#L295C1-L302C3).
>
> This happens during program->update_uniforms(in); before geom->draw(); in GlRenderer::render().

This is normal btw ;) OpenGL does not raise errors because they happen on the GPU. Hence, we only learn about them during validity checks when we explicitly request the error state.

ZzzhHe:
Got it, thanks! Working on the updates now.

libopenage/renderer/opengl/geometry.h
heinezen, line 1: Since there are no changes here anymore, the date change is also unnecessary ;)

libopenage/renderer/opengl/geometry.cpp
```diff
@@ -54,6 +54,7 @@ void GlGeometry::update_verts_offset(std::vector<uint8_t> const &verts, size_t o
 void GlGeometry::draw() const {
 	switch (this->get_type()) {
 	case geometry_t::bufferless_quad:
+		// a VAO must be bound before this call. This is handled by GlRenderer.
```
heinezen, line 57: references to other classes outdated quickly when the code is refactored. Better just state what needs to happen.

```suggestion
		// any VAO must be bound before this call
```

libopenage/renderer/opengl/renderer.h
```diff
@@ -67,6 +68,8 @@ class GlRenderer final : public Renderer {
 
 	/// The main screen surface as a render target.
 	std::shared_ptr<GlRenderTarget> display;
+
+	std::shared_ptr<GlVertexArray> shared_quad_vao;
```
heinezen, line 72: You can make this member a `unique_ptr` I think, since it's only used by the GlRenderer class.

libopenage/renderer/opengl/renderer.cpp
```diff
@@ -26,7 +27,8 @@ GlRenderer::GlRenderer(const std::shared_ptr<GlContext> &ctx,
 	gl_context{ctx},
 	display{std::make_shared<GlRenderTarget>(ctx,
                                              viewport_size[0],
-                                             viewport_size[1])} {
+                                             viewport_size[1])},
+	shared_quad_vao(std::make_shared<GlVertexArray>(ctx)) {
```
heinezen, line 31: You should use a curly braced-initializer here and wherever possible. This prevents unintended type conversion.

libopenage/renderer/opengl/renderer.cpp
```diff
@@ -169,6 +171,9 @@ void GlRenderer::render(const std::shared_ptr<RenderPass> &pass) {
 	auto gl_target = std::dynamic_pointer_cast<GlRenderTarget>(pass->get_target());
 	gl_target->bind_write();
 
+	// bind the shared quad vao for all bufferless quad geometries
```
heinezen, line 174: I slightly reworded the comment so that it's more clear what's happening.

```suggestion
	// ensure that an (empty) VAO is bound before rendering geometry
	// a bound VAO is required to render bufferless quad geometries by OpenGL
	// see https://www.khronos.org/opengl/wiki/Vertex_Rendering#Causes_of_rendering_failure
```

libopenage/renderer/opengl/renderer.h
heinezen, line 72:
```suggestion
    /// An empty vertex array object (VAO).
    ///
    /// This VAO has to be bound at the start of a render pass to ensure
    /// that bufferless quad geometry can be drawn without errors. Drawing a
    /// bufferless quad requires any VAO to be bound
    /// see https://www.khronos.org/opengl/wiki/Vertex_Rendering#Causes_of_rendering_failure
	std::shared_ptr<GlVertexArray> shared_quad_vao;
```

heinezen (changes requested):

libopenage/renderer/opengl/renderer.h
ZzzhHe: I attempted to use `unique_ptr`, but encountered an issue since it requires the complete type definition. When I moved the include to the header file to replace forward declaration, it led to circular include dependencies. Did I do something wrong, or should I dive into the dependency cycle to fix it?

heinezen: No, you are right. I think you can change it back to `shared_ptr`.

I think the problem here is that `unique_ptr` only allows incomplete types under specific circumstances. I also didn't know the specifics for that :'D

> std::unique_ptr may be constructed for an [incomplete type](https://en.cppreference.com/w/cpp/language/incomplete_type) T, such as to facilitate the use as a handle in the [pImpl idiom](https://en.cppreference.com/w/cpp/language/pimpl). If the default deleter is used, T must be complete at the point in code where the deleter is invoked, which happens in the destructor, move assignment operator, and reset member function of std::unique_ptr. (In contrast, [std::shared_ptr](https://en.cppreference.com/w/cpp/memory/shared_ptr) can't be constructed from a raw pointer to incomplete type, but can be destroyed where T is incomplete). Note that if T is a class template specialization, use of unique_ptr as an operand, e.g. !p requires T's parameters to be complete due to [ADL](https://en.cppreference.com/w/cpp/language/adl).

https://en.cppreference.com/w/cpp/memory/unique_ptr

ZzzhHe: Thanks, I will change it back

heinezen (approved):

heinezen:
Very good!

PR #1735, opened 2024-12-27
Cython smp

jere8184:


jere8184:
should I add @9a24f0 to .mailmap or copying.md?

openage/convert/value_object/read/media/smx.pyx
```diff
@@ -59,6 +57,415 @@ cdef public dict LAYER_TYPES = {
...
+            self.chunk_pos = chunk_pos
+
+            self.pcolor.push_back(row_data)
+
+
+ctypedef fused SMXLayerVariant:
+    SMXMainLayer8to5
+    SMXMainLayer4plus1
+    SMXOutlineLayer
+    SMXShadowLayer
```
heinezen, line 117: The typedef should be moved to the top of the file next to the other declarations of structs/enums.

openage/convert/value_object/read/media/smx.pyx
```diff
@@ -1,7 +1,10 @@
-# Copyright 2019-2023 the openage authors. See copying.md for legal info.
+# Copyright 2019-2024 the openage authors. See copying.md for legal info.
 #
 # cython: infer_types=True
 
+from libcpp.vector cimport vector
+from libcpp cimport bool
+from libc.stdint cimport uint8_t, uint16_t
 from enum import Enum
```
heinezen, line 8: You should separate Python imports and cython imports with at least a space.

```suggestion
from libc.stdint cimport uint8_t, uint16_t

from enum import Enum
```

openage/convert/value_object/read/media/smx.pyx
```diff
@@ -49,8 +47,8 @@ class SMXLayerType(Enum):
     """
     SMX layer types.
     """
-    MAIN    = "main"
-    SHADOW  = "shadow"
+    MAIN = "main"
+    SHADOW = "shadow"
     OUTLINE = "outline"
```
heinezen, line 52: These spaces should not be formatted away. If autopep8 did that, you should configure its settings to exclude these changes. There's instructions in the docs on how to configure autopep8 in VSCode for our code style: https://github.com/SFTtech/openage/blob/master/doc/ide/vscode.md#opening-and-configuring-the-openage-workspace

openage/convert/value_object/read/media/smx.pyx
```diff
@@ -59,6 +57,415 @@ cdef public dict LAYER_TYPES = {
     1: SMXLayerType.SHADOW,
     2: SMXLayerType.OUTLINE,
 }
+cdef class SMXMainLayer8to5(SMXLayer):
```
heinezen, line 60: Declarations should be separated by at two spaces.

```suggestion
}


cdef class SMXMainLayer8to5(SMXLayer):
```

heinezen, line 60: You should declare all classes that inherit from `SMXLayer` **after** `SMXLayer` has been declared as you would do normally in Python/C++ code. Otherwise, this may create compilation problems in the future (even if it doesn't do now).

openage/convert/value_object/read/media/smx.pyx
```diff
@@ -59,6 +57,415 @@ cdef public dict LAYER_TYPES = {
...
+        super().__init__(layer_header, data)
+        # process cmd table
+        for i in range(self.row_count):
+            cmd_offset, color_offset, chunk_pos, row_data = create_color_row(
+                self, i)
+            self.cmd_offset = cmd_offset
+            self.color_offset = color_offset
+            self.chunk_pos = chunk_pos
+
+            self.pcolor.push_back(row_data)
```
heinezen, line 71: Shouldn't this code also be part of the parent class, since it is the same for all subclasses?

openage/convert/value_object/read/media/smx.pyx
heinezen: Also, you should probably not format Cython code with autopep8 as it doesn't understand a lot of Cython syntax.

openage/convert/value_object/read/media/smx.pyx
```diff
@@ -59,6 +57,415 @@ cdef public dict LAYER_TYPES = {
...
+                                                               const uint8_t[::1] & data_raw,
+                                                               vector[pixel] & row_data,
+                                                               Py_ssize_t rowid,
+                                                               Py_ssize_t first_cmd_offset,
+                                                               Py_ssize_t first_color_offset,
+                                                               int chunk_pos,
+                                                               size_t expected_size):
+    """
+    TODO: docstring
+    """
```
heinezen, line 196: Please address this TODO or at least add the original descriptions back.

openage/convert/value_object/read/media/smx.pyx
```diff
@@ -59,6 +57,415 @@ cdef public dict LAYER_TYPES = {
...
+            # shadows sometimes need an extra pixel at
+            # the end
+            if SMXLayerVariant is SMXShadowLayer and row_data.size() < expected_size:
+                # copy the last drawn pixel
+                # (still stored in nextbyte)
+                #
+                # TODO: confirm that this is the
+                #       right way to do it
+                row_data.push_back(pixel(color_shadow,
+                                         nextbyte, 0, 0, 0))
```
heinezen, line 256: You should always separe the fused checks from the rest of the code flow to make everything easier to read.

```suggestion
            if SMXLayerVariant is SMXShadowLayer:
                if row_data.size() < expected_size:
                    # copy the last drawn pixel
                    # (still stored in nextbyte)
                    #
                    # TODO: confirm that this is the
                    #       right way to do it
                    row_data.push_back(pixel(color_shadow,
                                             nextbyte, 0, 0, 0))
```

openage/convert/value_object/read/media/smx.pyx
```diff
@@ -103,14 +510,16 @@ class SMX:
...
         dbg("SMX")
         dbg(" version:                %s",   version)
         dbg(" frame count:            %s",   frame_count)
-        dbg(" file size compressed:   %s B", file_size_comp + 0x20)   # 0x20 = SMX header size
-        dbg(" file size uncompressed: %s B", file_size_uncomp + 0x40) # 0x80 = SMP header size
+        # 0x20 = SMX header size
+        dbg(" file size compressed:   %s B", file_size_comp + 0x20)
+        # 0x80 = SMP header size
+        dbg(" file size uncompressed: %s B", file_size_uncomp + 0x40)
         dbg(" comment:                %s",   comment.decode('ascii'))
```
heinezen, line 523: This change makes the code less readable, so please revert it.

openage/convert/value_object/read/media/smx.pyx
```diff
@@ -182,16 +593,20 @@ class SMX:
...
 
                 elif layer_type is SMXLayerType.SHADOW:
-                    self.shadow_frames.append(SMXShadowLayer(layer_header, data))
+                    self.shadow_frames.append(
+                        SMXShadowLayer(layer_header, data))
 
                 elif layer_type is SMXLayerType.OUTLINE:
-                    self.outline_frames.append(SMXOutlineLayer(layer_header, data))
+                    self.outline_frames.append(
+                        SMXOutlineLayer(layer_header, data))
```
heinezen, line 609: These changes are not necessary. The line length was previously wel below 100 lines. Make sure yoe set the max line length in your autopep8 config to 100 too.

openage/convert/value_object/read/media/smx.pyx
```diff
@@ -319,6 +734,21 @@ cdef class SMXLayer:
...
+    cdef int cmd_offset
+
+    # current color
+    cdef int color_offset
+
+    # current chunk position
+    cdef int chunk_pos
+
+    # rows
+    cdef size_t row_count
```
heinezen, line 750: Is there a reason why we would want to promote these variables to members instead of keeping them local? All of these members are never accessed again after the layer has been read. Keeping `data_raw` around is especially resource-intensive because we would store the whole original graphics file in memory well after it was used.

openage/convert/value_object/read/media/smx.pyx
```diff
@@ -492,557 +826,10 @@ cdef class SMXLayer:
...
-        return dpos, dpos, chunk_pos, row_data
-
-
 @cython.boundscheck(False)
 @cython.wraparound(False)
-cdef numpy.ndarray determine_rgba_matrix(vector[vector[pixel]] &image_matrix,
-                                         numpy.ndarray[numpy.uint8_t, ndim=2, mode="c"] palette):
+cdef numpy.ndarray determine_rgba_matrix(vector[vector[pixel]] & image_matrix,
+                                         numpy.ndarray[numpy.uint8_t, ndim= 2, mode = "c"] palette):
     """
```
heinezen, line 833: These changes are unecessary and might als break Cython syntax.

openage/convert/value_object/read/media/smx.pyx
```diff
@@ -1053,7 +840,7 @@ cdef numpy.ndarray determine_rgba_matrix(vector[vector[pixel]] &image_matrix,
     cdef size_t height = image_matrix.size()
     cdef size_t width = image_matrix[0].size()
 
-    cdef numpy.ndarray[numpy.uint8_t, ndim=3, mode="c"] array_data = \
+    cdef numpy.ndarray[numpy.uint8_t, ndim= 3, mode = "c"] array_data = \
         numpy.zeros((height, width, 4), dtype=numpy.uint8)
```
heinezen, line 844: same here

openage/convert/value_object/read/media/smx.pyx
```diff
@@ -1142,7 +929,7 @@ cdef numpy.ndarray determine_rgba_matrix(vector[vector[pixel]] &image_matrix,
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
-cdef numpy.ndarray determine_damage_matrix(vector[vector[pixel]] &image_matrix):
+cdef numpy.ndarray determine_damage_matrix(vector[vector[pixel]] & image_matrix):
```
heinezen, line 932: and here

openage/convert/value_object/read/media/smx.pyx
```diff
@@ -1152,7 +939,7 @@ cdef numpy.ndarray determine_damage_matrix(vector[vector[pixel]] &image_matrix):
     cdef size_t height = image_matrix.size()
     cdef size_t width = image_matrix[0].size()
 
-    cdef numpy.ndarray[numpy.uint8_t, ndim=3, mode="c"] array_data = \
+    cdef numpy.ndarray[numpy.uint8_t, ndim= 3, mode = "c"] array_data = \
```
heinezen, line 942: and here

heinezen (changes requested):
I've made comments to the SMX file only, but they apply to the SMP file in the same way. The graphics seem to convert fine, so there are only structural changes left to do.

heinezen:
> should I add @9a24f0 to .mailmap or copying.md?

If you based your changes on their code, then yes, you have to add them to copying.md.

openage/convert/value_object/read/media/smx.pyx
jere8184: honsestly I dont belive unessicary nesting makes this code more readable, espically when nesting effects control.

openage/convert/value_object/read/media/smx.pyx
jere8184: `create_color_row()` needs to be called from the derived classes in order for `SMXLayerVariant` to have the right value ASFIK.

jere8184: This also why promoted some of the local variables to member variables, as I dont think we can call create_color_row() from `SMXLayer` anymore, hence the data that it sets locally in its consturctor must be avilable to its dervied classes which themselves call `create_color_row()`.

If i'm correct a possible solution could be to just clear these variables once we no-longer need them.

openage/convert/value_object/read/media/smx.pyx
jere8184: I will revert the formatting changes :)

openage/convert/value_object/read/media/smx.pyx
heinezen: it does for me because I can easier see wha each layer does since it's in its own block.

heinezen: I'm also not sure if Cython can properly convert the if into an `if constexpr` if we mix compile-time and runtime checks.

heinezen:
@jere8184 I've managed to remove the inheritance from the classes used by the fused `SMXLayerVariant` which makes the whole thing a bit less complicated. Since the change was hard to explain, I just did and pushed it myself. While testing, I also found a small bug that got fixed too.

heinezen:
I'll do a few performance test before I merge. Hopefully we didn't make the code slower accidentally :D

jere8184:
> I'll do a few performance test before I merge. Hopefully we didn't make the code slower accidentally :D

Nice

heinezen:
DE2 media export time on this MR:

```
INFO [py] -- Exporting graphics files...
INFO [py] Dumping metadata files...
INFO [py] Finished export of modpack 'de2_base' v0.6.0 (75.19 seconds)
INFO [py] Finished export (75.19 seconds)
```

DE2 media export time on `master`:

```
INFO [py] -- Exporting graphics files...
INFO [py] Dumping metadata files...
INFO [py] Finished export of modpack 'de2_base' v0.6.0 (74.66 seconds)
INFO [py] Finished export (74.66 seconds)
```

It's probably safe to say that the difference is negligeable :P

heinezen (approved):
I added documentation to the classes, so it should be much easier to work on the cython files from now on :)

PR #1737, opened 2024-12-29
Suppress "--add_dll_search_path" in sub-parser

jere8184:


TheJJ:
why is this change needed? the parameter is added to the toplevel cli on windows only already?

jere8184:
> why is this change needed? the parameter is added to the toplevel cli on windows only already?

if you fail to provide a subcommand this parameter will be passed to the 'main' sub-parser causing it to fail with an error.
run locally  (on windows)
'python -m openage --add-dll-search-path <valid_path>' 
and see for yourself.  

you should get the following error
'openage main: error: unrecognized arguments: --add-dll-search-path <valid_path>' 
which is pretty confusing.

This change stops an error from occurring and invokes 'openage main' as if no arguments have been passed.

TheJJ:
ah, then the proper solution is to add this as a common parent to the subparsers that make use of the option. duplication is not the proper way, unfortunately.

jere8184:
@TheJJ i've went a slightly different, route see changes.

heinezen (approved):
Looks good to me

PR #1738, opened 2024-12-30
Configurable World Shader Commands System

### LINKED ISSUE #1541: Configurable world fragment shader
The *world shader* is the shader program used for the world render stage, which is responsible for drawing units, buildings, ambience, and everything else that is displayed by a sprite. Making it configurable gives us some leeway in how these objects are rendered.

Our planned approach to configuration is to allow encoding of "commands" in the alpha channels of pixels in a texture. In short, every even alpha value can represent a "command", which means the pixel should be processed using custom shader code. Odd alpha values are drawn without any processing. This gives us 128 possible commands to use in the shader.

The shader code that runs when such a command is encountered should be configurable by modders, e.g. to draw special pixels like player color or outlines. The final world shader would be assembled from the code snippets for each command individual command.  These snippets would probably be embedded in a simple switch statement in the GLSL code like this:

```glsl
# get command
int alpha = int(round(tex_val.a * 255));
switch (alpha) {
    case 254:
        # command 254
        col = vec4(1.0f, 0.0f, 0.0f, 1.0f);
        break;
    case 252:
        # command 252
        col = vec4(0.0f, 1.0f, 0.0f, 1.0f);
        break;
    default:
        # no command
        col = tex_val;
        break;
}
```

Tasks:
- [ ] Add method to define custom shader code for commands
- [ ] Dynamically assemble world shader from template + command code
- [ ] (later) Allow passing data to shader in uniforms for use in custom shader code

ZzzhHe:
Resolve #1541 

- Add WorldShaderCommands class to manage shader commands
- Support adding/removing commands with custom GLSL code
- Integrate command system into WorldRenderStage


heinezen:
Hey thanks a lot and cool that it already works :)

I think we need to design this a bit more to make it more in line in the way that it is supposed to be used, i.e. by putting these snippets as files somewhere and merging the shader code better.

I'll think about it and will come back to you with more ideas.

heinezen:
Okay, I finally have time to respond here :D As said before, I like the direction where this is going, but I have a few requests for how commands should be integrated.

1. If shader source code is supposed to be templated, I don't think it should be loaded as `ShaderSource` at the start as the code may be incomplete. Instead, there should probably be a `ShaderTemplate` class that the template is loaded into which can be fed with `ShaderSnippet`s to insert into placeholder in the template. Once `ShaderTemplate` is configured with snippets, it should be able to create the final source code which can then be passed to the `ShaderSource` constructor. This additional layer would probably prevent a lot of problems for templates that are not properly initialized. 
2. I think we can already make the feature more versatile and extendable by not having a generic `INSERT_POINT` in the shader source code and instead assigning placeholder IDs. Multiple shader snippets could then be inserted into different position in the shader template based on their ID.
3. I don't think the logic for creating shader code should be hardcoded, e.g. in `WorldShaderCommands`, because it defeats the point of making the commands configurable. The command code should be loadable from a file that the user can create. If we have to recompile the engine every time you want to change the commands, we might ass well just use the current solution.
4. It's probably better to create a [renderer demo](https://github.com/SFTtech/openage/blob/master/doc/code/renderer/demos.md) for this feature rather than using it in the actual renderer. This way, we can test the design and change it without having to worry about it affecting the main rendering loop just yet.

ZzzhHe:
Here are some updates:
1. Template system implemented via ShaderCommandTemplate class
2. Custom shader code for commands supported through .config files
3. Dynamic shader assembly with placeholder support (//@COMMAND_SWITCH@)
4. Add demo 7 to test the shader command template class.

But the current state is a very basic version, here are some current limitations
1. Config path is currently in shaders directory. Maybe I should make this more flexible or move to a new folder in future versions.
2. Missing direct way to set alpha values on entities (would be useful for testing). I'm considering consider adding set_alpha().
3. Only basic validation right now, I could add more robust error checking later.

ZzzhHe:
Because right now I didn't implement alpha setting function, so I hardcoded different alpha code into fragment shader to test.

demo_7_world.frag.glsl
```
void main() {
	vec4 tex_val = texture(tex, uv);
	int alpha = int(round(tex_val.a * 255));
	alpha = 252; // hardcoded for testing right now
	switch (alpha) {
	case 0:
```
world_commands.config
```
[COMMAND]
placeholder=COMMAND_SWITCH
alpha=252
description=Green tint
code={
col = vec4(0.0, 1.0, 0.0, 1.0) * tex_val;
}
```
image

assets/test/shaders/world_commands.config
heinezen, line 1: I think we could still make this much simpler :D

I'm not sure if we even need a custom config format that needs parsing. If we do placeholder replacement, we can skip most of the parsing logic while supporting even more use cases.

### Example

Template: `shader.template`

```glsl
#version 330

uniform float template_local;

// PLACEHOLDER: unifs_code.snippet

void main() {
    // PLACEHOLDER: x_code.snippet
    // PLACEHOLDER: y_code.snippet
    gl_Position = move * vec4(x, y, 0.0, 1.0);
}
```

Snippet: `unifs_code.snippet`

```
uniform float value;
```

Snippet: `x_code.snippet`

```
float x = value + 1.0 
```

Snippet: `y_code.snippet`

```
float y = value - 1.0 
```

End result:


```glsl
#version 330

uniform float template_local;

uniform float value;

void main() {
    float x = value + 1.0 
    float y = value - 1.0
    gl_Position = move * vec4(x, y, 0.0, 1.0);
}
```

libopenage/renderer/demo/demo_7.cpp
```diff
@@ -0,0 +1,107 @@
...
+	auto clock = std::make_shared<time::Clock>();
+	auto asset_manager = std::make_shared<renderer::resources::AssetManager>(
+		renderer,
+		path["assets"]["test"]);
+	auto cam_manager = std::make_shared<renderer::camera::CameraManager>(camera);
+
+	auto shaderdir = path / "assets" / "test" / "shaders";
+
+	std::vector<std::shared_ptr<RenderPass>>
+		render_passes{};
```
heinezen, line 51: This statement doesn't need to be multiline.

libopenage/renderer/demo/demo_7.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+	// Initialize world renderer with shader commands
+	auto world_renderer = std::make_shared<renderer::world::WorldRenderStage>(
+		window,
+		renderer,
+		camera,
+		shaderdir,
+		shaderdir, // Temporarily, Shader commands config has the same path with shaders for this demo
+		asset_manager,
+		clock);
```
heinezen, line 61: Demos should be minimal, so that we can show of specific features of the renderer without having to replicate the whole rendering pipeline. Ideally, you shouldn't initialize a whole world renderer and should just show the absolute basics, i.e.:

- Loading a shader template from file
- Adding placeholder snippets
- Getting the "complete" shader code (i.e. template with inserted placeholders)
- Rendering an object with the shader to the screen (this can be a bufferless quad with no texture)

libopenage/renderer/stages/world/render_stage.cpp
```diff
@@ -156,4 +181,37 @@ void WorldRenderStage::init_uniform_ids() {
...
+		resources::shader_stage_t::vertex,
+		vert_shader_file.read());
+	vert_shader_file.close();
+
+	auto frag_shader_file = (shaderdir / "demo_7_world.frag.glsl").open();
+	log::log(INFO << "Loading shader commands config from: " << (shaderdir / "demo_7_display.frag.glsl"));
+	this->shader_template = std::make_shared<ShaderCommandTemplate>(frag_shader_file.read());
+	if (not this->shader_template->load_commands(config_path / "world_commands.config")) {
+		log::log(ERR << "Failed to load shader commands configuration for world stage");
+		return;
```
heinezen, line 197: You should raise an Exception in this case because the stage cannot be initialized properly. It's in an unrecoverable state if the shader cannot be loaded.

libopenage/renderer/stages/world/world_shader_commands.cpp
```diff
@@ -0,0 +1,166 @@
...
+						reading_code = true;
+						code_block.clear();
+					}
+				}
+			}
+		}
+
+		return true;
+	}
+	catch (const std::exception &e) {
```
heinezen, line 97: You should almost never catch all C++ exceptions, only specific ones like `openage::Error` (and even that is already a stretch). If you let the program continue from an error state, it will definitly segfault at another place where error tracking will be much harder to do. If your code errors, you should fix the causes or raise errors by yourself (e.g. an exception if the file cannot be found).

assets/test/shaders/world_commands.config
ZzzhHe: Do you think using a JSON config file makes sense? The structure is clean and we only need to load the JSON file instead of scanning for snippets:
``` json
{
  "snippets": {
    "lighting": "snippets/lighting.glsl",
    "color_processing": "snippets/color.glsl",
    "effects": "snippets/effects.glsl"
  }
}
```

File structure would be:
``` bash
shaders/
  ├── world2d.frag.template    # Main shader template
  ├── world2d.vert.glsl       # Vertex shader
  ├── shader_config.json      # Snippet configuration
  └── snippets/               # User-modifiable snippets
      ├── lighting.glsl
      ├── color.glsl
      └── effects.glsl
```

libopenage/renderer/demo/demo_7.cpp
ZzzhHe: Oh, I get it. I was thinking about test it with the `WorldRenderStage`. Since the shader template seems to be standalone and doesn't rely on `WorldRenderStage` at all, would it make more sense to move it to the `util` namespace? So we can use it not only for world shader later.

ZzzhHe: Oh, I get it. I was thinking about testing it with the `WorldRenderStage`.

assets/test/shaders/world_commands.config
ZzzhHe: Alternatively, we could simply read all the snippets directly from the `snippets/` folder without setting a config file.

ZzzhHe:
I've simplified the shader template system to focus on its core functionality.

+ Shader templates now load snippets from a snippet folder that contains all snippet files.
+ Refactored demo7 to be a minimal demonstration of the template system only.

libopenage/renderer/stages/world/shader_template.cpp
```diff
@@ -0,0 +1,65 @@
...
+	snippets[name] = content;
+}
+
+std::string ShaderTemplate::generate_source() const {
+	std::string result = template_code;
+
+	// Process each placeholder
+	for (const auto &[name, snippet_code] : snippets) {
+		std::string placeholder = "// PLACEHOLDER: " + name;
+		size_t pos = result.find(placeholder);
```
heinezen, line 44: Idea for a faster approach: Instead of searching the whole source string for each snippet, we could search for the placeholder positions during initialization of the class and save the positions. Then when we want to insert the snippets later, we just have to look up the positions again.

Probably not that big of a deal right now, but it might be more performant on large template files.

libopenage/renderer/demo/demo_7.cpp
```diff
@@ -0,0 +1,85 @@
...
+	settings.width = 800;
+	settings.height = 600;
+	settings.debug = true;
+
+	opengl::GlWindow window("Shader Commands Demo", settings);
+	auto renderer = window.make_renderer();
+
+	auto shaderdir = path / "assets" / "test" / "shaders";
+
+	// Initialize shader templlalte
```
heinezen, line 32:
```suggestion
	// Initialize shader template
```

heinezen (dismissed):
Sorry for the long wait on the review :crying_cat_face: I was unfortunately very busy at work which made the open source reviews a bit difficult. But now that I have time I can, I must say I like your solution very much! I have a few small suggestions, but that is only really minor stuff. I would be happy to merge this soon :)

ZzzhHe:
> Sorry for the long wait on the review 😿 I was unfortunately very busy at work which made the open source reviews a bit difficult. But now that I have time I can, I must say I like your solution very much! I have a few small suggestions, but that is only really minor stuff. I would be happy to merge this soon :)

Thanks! I'll update the code later this week.

ZzzhHe:
Hi @heinezen , sorry for the long delay — I've been really busy with ongoing work these days.

I’ve now updated the code based on your suggestions, and it’s ready for another look when you have time 🙇‍♂️

heinezen (dismissed):
Hey, great job! I already know how to utilize this one in the future :D

I added documentation in the renderer docs to your PR, so that it's understandable what the feature is used for. Didn't want to bother you with that. Otherwise no complaints :)

heinezen (approved):

heinezen (approved):

ZzzhHe:
Thanks so much! 😊 Really appreciate you adding the docs. Glad to hear everything else looks good!

PR #1740, opened 2025-01-07
Gui Stencil Mask Optimization

### LINKED ISSUE #1538: Stencil tests for GUI elements
This could be a rather small optimization for the renderer. The idea is to save rendering time by not rendering pixels in the game world that would be hidden by GUI elements anyway. To do this, we would create a [stencil mask](https://learnopengl.com/Advanced-OpenGL/Stencil-testing) from opaque elements in the GUI and apply this mask in all other render passes. Doing this allows the GPU to discard fragments masked by the stencil mask early and save some time during rendering as a result.

The time save depends on a few conditions. For example, the stencil test gets more effective if the GUI covers large parts of the screen and the elements don't change positions. However, this should be at least the case for the old AoE1/AoE2/SWGB GUIs.

Tasks:
- [ ] Enable stencil tests in renderer
- [ ] Create stencil mask from GUI texture
- [ ] Upload stencil mask to renderer in other render stages (skybox, terrain, world, ....)

ZzzhHe:
Resolve #1538 

Changes:
- Modified `Presenter::render()` to use three-pass rendering:
  1. First pass: Create stencil mask from GUI
  2. Second pass: Render world with stencil test (skip GUI-covered pixels)
  3. Third pass: Render GUI normally

### DIFF
--- a/libopenage/renderer/demo/demo_8.cpp
+++ b/libopenage/renderer/demo/demo_8.cpp
@@ -98,14 +98,14 @@ void renderer_demo_8(const util::Path &path) {
 		bg_uniforms,
 		quad,
 		false,
-		true, // Enable depth test
+		false, // Disable depth test to make stencil testing more visible
 	};
 
 	Renderable gui_obj{
 		gui_uniforms,
 		quad,
 		false,
-		false, // No depth test needed
+		false,
 	};
 
 	Renderable display_obj{


heinezen:
Hey thanks for contributing again :) Unfortunately, I think the solution you made clashes a bit with our design of the renderer.

In general, the renderer code is separated into two levels (see our [renderer docs](https://github.com/SFTtech/openage/tree/master/doc/code/renderer) for details). The (low-level) level 1 interface handles the bare-metal stuff, e.g. interfacing with OpenGL to talk to the GPU, and should generally be inaccessible to other parts of the engine. Instead, the rest of the engine mainly interacts with the level 2 interface of the renderer which abstracts the low-level stuff away and implements systems that can be used for gameplay, e.g. playing sprite animations. The reason we have this level structure is mostly thread-safety, but also easier maintainance, since we can change or replace the level 1 implementaton without changing the entire engine. The latter is important for offering multiple renderer backends simultaneously to the user, e.g. OpenGL **and** Vulkan.

In your solution, there would be "bare-metal" OpenGL calls in the presenter, even though the presenter should only interact with the level 2 interface by default. If we do it like this, then the presenter does not work with a Vulkan renderer backend anymore, for example. Ideally, the presenter does not have to care what backend we initialize the renderer with, whether it's OpenGL or Vulkan.

heinezen:
For solving this task properly, we should try to add stencil tests as a more general feature to the level 1 renderer first. If we want to activate it for the GUI, the stencil test write code should be moved into the GUI subsystem. To use the stencil tests in the other render passes, we could add settings to the `RenderPass` implementations; maybe stencil test reading could be a flag... When the render pass is rendered, we could switch on/off stencil testing based on that flag.

So in summary my thoughts would be:

1. Move the stencil test writing into the `libopenage/renderer/gui` subsystem of the level 1 renderer
2. Add settings to a `RenderPass` that allows us to configure activating/deactivating stencil tests (and maybe other settings in the future)
3. Based on the `RenderPass` settings, activate stencil testing when the render pass becomes active in https://github.com/SFTtech/openage/blob/d6a4dd3b8c238ba642b4e987ea6677aae0827724/libopenage/renderer/opengl/renderer.cpp#L170

heinezen:
Apologies btw for askng for so many redesigns :smile:  For issues that are not labelled "good first issue", the issue description can sometimes be a bit vague because we are ourselves not sure of what we want as a solution yet. You can always ask me for where to look in the docs before you do another issue like this one, maybe that helps :)

ZzzhHe:
Hi, Thanks a lot for the detailed explanation! It makes sense now, I should do the separation between level 1 and level 2. I'll adjust my solution to follow this structure and work on it accordingly. Appreciate your guidance!

ZzzhHe:
Here are updates: 
1. Added three stencil states in render pass to control rendering behavior:`WRITE_STENCIL_MASK`,  `USE_STENCIL_TEST`,  `DISABLE_STENCIL`.
2. Implemented stencil state handling in `GlRenderer`.
3. Added `stencil_render_pass` functionality in the GUI class.
4. Updated the `Presenter` class to handle Level 2 rendering abstractly, without OpenGL details.

heinezen:
Finally had time to look at this :smile: I think we are getting there, but after looking at your code, I'm pretty sure that your stencil tests currently won't work. It's hard to test since we currently have deactivated the GUI, but here are my reasons:

1. Stencil testing is always done _per framebuffer_, so writing to a specific stencil buffer does not affect stencil testing in other framebuffers. In openage, the world render stage uses a different framebuffer, so you have to actively transfer the stencil buffer state from the GUI framebuffer before you can use stencil testing (i.e. read from the stencil values written by the GUI pass).
2. You have to actively attach a stencil buffer to the framebuffer in OpenGL which is currently not done in your code. The same applies to depth textures btw, they don't get assigned by default. You can see how we assign a depth buffer for the world render stage framebuffer here:

https://github.com/SFTtech/openage/blob/6d9d8d75a20ac42ae2e352f2d50888df7b0be512/libopenage/renderer/stages/world/render_stage.cpp#L138

3. You have introduced a lot of redundancy (e.g. the GUI is rendered twice now) which would counteract all our potential time saves from stencil testing :D

I would recommend that you don't implement this functionality in the presenter yet. A [renderer demo](https://github.com/SFTtech/openage/blob/master/doc/code/renderer/demos.md) is much better suited and allows you to test the level 1 implementation without having to worry about level 2 abstractions logic.

libopenage/presenter/presenter.cpp
```diff
@@ -313,11 +325,18 @@ void Presenter::render() {
...
-	for (auto &pass : this->render_passes) {
-		this->renderer->render(pass);
+	for (size_t i = 0; i < this->index_gui_stencil_pass; ++i) {
+		this->renderer->render(this->render_passes[i]);
 	}
+
+	this->gui->render();
+	this->renderer->render(this->render_passes[this->index_gui_render_pass]);
+
+	this->renderer->render(this->render_passes[this->index_final_render_pass]);
```
heinezen, line 339: The GUI is now rendered twice, which is not what we want if the goal is performance ;)

Also, code like this needs comment because otherwise it becomes very confusing to understand why the order of render passes is structured this way.

libopenage/renderer/gui/gui.cpp
```diff
@@ -98,6 +102,10 @@ void GUI::initialize_render_pass(size_t width,
 	// TODO: Rendering into the FBO is a bit redundant right now because we
 	// just copy the GUI texture into the output texture
 	this->render_pass = renderer->add_render_pass({display_obj}, fbo);
+
+	auto stencil_pass = renderer->add_render_pass({display_obj}, fbo);
+	stencil_pass->set_stencil_state(StencilState::WRITE_STENCIL_MASK);
```
heinezen, line 107: I think if you just set the stencil state for `this->render_pass`, you would effectively have the same result without the need for an additional render pass.

libopenage/renderer/opengl/renderer.cpp
```diff
@@ -162,6 +162,30 @@ void GlRenderer::optimize(const std::shared_ptr<GlRenderPass> &pass) {
 	}
 }
 
+void GlRenderer::setupStencilWrite() {
+	glEnable(GL_STENCIL_TEST);
+	glStencilFunc(GL_ALWAYS, 1, 0xFF);
+	glStencilOp(GL_KEEP, GL_KEEP, GL_REPLACE);
+	glStencilMask(0xFF);
+	glColorMask(GL_FALSE, GL_FALSE, GL_FALSE, GL_FALSE);
+	glDepthMask(GL_FALSE);
```
heinezen, line 171: Is this necessary? Running these two statement would mean that the only buffer you write to is the stencil buffer. However, I don't see a reason why we shouldn't be able to draw into the color and depth buffer simultaneously.

libopenage/renderer/opengl/renderer.cpp
```diff
@@ -162,6 +162,30 @@ void GlRenderer::optimize(const std::shared_ptr<GlRenderPass> &pass) {
...
+}
+
+void GlRenderer::setupStencilTest() {
+	glEnable(GL_STENCIL_TEST);
+	glStencilFunc(GL_NOTEQUAL, 1, 0xFF);
+	glStencilMask(0x00);
+	glColorMask(GL_TRUE, GL_TRUE, GL_TRUE, GL_TRUE);
+	glDepthMask(GL_TRUE);
```
heinezen, line 179: Same here, why is this necessary?

libopenage/renderer/opengl/renderer.cpp
```diff
@@ -162,6 +162,30 @@ void GlRenderer::optimize(const std::shared_ptr<GlRenderPass> &pass) {
...
+}
+
+void GlRenderer::disableStencilTest() {
+	glDisable(GL_STENCIL_TEST);
+	glStencilMask(0xFF);
+	glColorMask(GL_TRUE, GL_TRUE, GL_TRUE, GL_TRUE);
+	glDepthMask(GL_TRUE);
```
heinezen, line 186: same here (but `glStencilMask` would also have no effect)

libopenage/renderer/opengl/renderer.cpp
```diff
@@ -176,7 +200,12 @@ void GlRenderer::render(const std::shared_ptr<RenderPass> &pass) {
 	// see https://www.khronos.org/opengl/wiki/Vertex_Rendering#Causes_of_rendering_failure
 	shared_quad_vao->bind();
 
-	glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
+	// ensure that an (empty) VAO is bound before rendering geometry
+	// a bound VAO is required to render bufferless quad geometries by OpenGL
+	// see https://www.khronos.org/opengl/wiki/Vertex_Rendering#Causes_of_rendering_failure
+	shared_quad_vao->bind();
```
heinezen, line 206: This duplicates a comment from further above. Please remove it.

libopenage/renderer/opengl/renderer.cpp
```diff
@@ -198,6 +227,20 @@ void GlRenderer::render(const std::shared_ptr<RenderPass> &pass) {
...
+			break;
+
+		case StencilState::USE_STENCIL_TEST:
+			setupStencilTest();
+			break;
+
+		case StencilState::DISABLE_STENCIL:
+			disableStencilTest();
+			break;
+		}
```
heinezen, line 242: Doing stencil testing per layer is a bit overkill for now IMO, but it's a good idea :)

We have to make sure that the draw order is correct and should make performance tests for this.

libopenage/renderer/opengl/renderer.h
```diff
@@ -63,6 +63,12 @@ class GlRenderer final : public Renderer {
 	/// Optimize the render pass by reordering stuff
 	static void optimize(const std::shared_ptr<GlRenderPass> &pass);
 
+	void setupStencilWrite();
+
+	void setupStencilTest();
+
+	void disableStencilTest();
```
heinezen, line 70: These functions should use snake case naming, e.g. `setup_stencil_write`, to adhere to our code style conventions.

libopenage/renderer/render_pass.h
```diff
@@ -14,6 +14,18 @@ namespace openage {
 namespace renderer {
 class RenderTarget;
 
+/**
+ * Stencil states for the render pass.
+ */
+enum class StencilState {
```
heinezen, line 20: same here, enums should use snake case naming

libopenage/renderer/render_pass.h
```diff
@@ -14,6 +14,18 @@ namespace openage {
...
+	/// State for writing GUI elements to stencil buffer.
+	WRITE_STENCIL_MASK,
+	/// State for using the mask when rendering scene.
+	USE_STENCIL_TEST,
+	/// State for normal rendering (GUI rendering).
+	DISABLE_STENCIL
```
heinezen, line 26: These docstrings refer to the GUI, but stencil testing should not just be used for this purpose. So it's better to make the docstrings more agnostic.

heinezen (changes requested):

ZzzhHe:
Oh..  I was too focused on implementing the stencil test functions and completely missed the fundamentals of how framebuffers work in OpenGL. I'll start fresh with a renderer demo this week to re-write and test the core functionality properly. Thanks!

ZzzhHe:
Here are some new updates:

1. Added depth_stencil texture (pixel_format::depth24_stencil8) to framebuffer
2. Created StencilConfig struct to encapsulate stencil test parameters
3. Implemented stencil testing in render pass and renderer
4. Added demo8 with a GUI rectangle masking an animated background
    + Configured GUI pass to write to stencil buffer (ref=255, GL_REPLACE)  
    + Configured background pass to skip rendering where stencil=255 (GL_NOTEQUAL)

Then I verified it using RenderDoc:

Screenshot 2025-02-05 100818
Screenshot shows stencil buffer during background pass - white areas (255) indicate GUI regions where background rendering is skipped due to stencil test failure.

Current limitation:

GUI isn't visible because each render pass clears the color buffer (glClear in renderer::render). 
I'm thinking should we add clear flags to control buffer clearing per pass, or combine GUI and background into single render pass with layers, or maybe other better ideas?

PR #1743, opened 2025-01-18
Add Nix build result folder to gitignore

mtlynch:
The nix build step creates an output folder called /result. I've added this to the .gitignore file so that git doesn't try to add generated files/binaries to the source tree.


heinezen (approved):

heinezen:
Thanks a lot

Issue #1676, opened 2024-08-25
Cost stamps (pathfinder)


heinezen:
**Required Skills:** C++

**Difficulty:** Medium

Our current model of the flow field pathfinder doesn't account for *temporary* changes of the cost grid, e.g. a building being placed on terrain. In this example, the building would change the movement cost of the grid (making affected tiles impassable) only as long as it exists. Once it is removed, the cells should return to their original cost value (e.g. that of the underlying terrain).

For flow fields there already exists a suggested solution for this in the form of *cost stamps* (check out the [related article](http://www.gameaipro.com/GameAIPro/GameAIPro_Chapter23_Crowd_Pathfinding_and_Steering_Using_Flow_Field_Tiles.pdf)). Essentially, cost stamps record the original value of the cells when they change its value. Removing a stamp restores the original value.

Tasks:

- [ ] Add cost stamps storage for cells to the [`CostField`](https://github.com/SFTtech/openage/blob/master/libopenage/pathfinding/cost_field.h) implementation. For performance reasons, the stamps should be stored separately from the current cost values (so **not** in the `cells` vector). Stamps should record
    - Time of "stamping"
    - Original cell value
- [ ] Add new methods for cost stamps to  [`CostField`](https://github.com/SFTtech/openage/blob/master/libopenage/pathfinding/cost_field.h). You should at least implement these functions:
     - [ ] `CostField::stamp(..)`: Place a stamp down for a specific cell
     - [ ] `CostField::unstamp(..)`: Restore the original value for a specific cell

**Further Reading**

- [Pathfinder documentation](https://github.com/SFTtech/openage/blob/master/doc/code/pathfinding/README.md)
- [`openage::path` pathfinding module](https://github.com/SFTtech/openage/tree/master/libopenage/pathfinding)

anupriyakkumari:
Hi, I'd like to take this up.

heinezen:
@anupriyakkumari Hey, no problem go ahead :)

PR #1745, opened 2025-01-22
Add cost stamps

dmwever:
fixes https://github.com/SFTtech/openage/issues/1676.

### DIFF
--- a/libopenage/pathfinding/cost_field.cpp
+++ b/libopenage/pathfinding/cost_field.cpp
@@ -56,6 +56,27 @@ void CostField::set_costs(std::vector<cost_t> &&cells, const time::time_t &valid
 	this->valid_until = valid_until;
 }
 
+bool CostField::stamp(size_t idx, cost_t cost, const time::time_t &stamped_at) {
+	if (this->cost_stamps.contains(idx)) return false;
+
+	cost_t original_cost = this->get_cost(idx);
+	this->cost_stamps[idx].original_cost = original_cost;
+	this->cost_stamps[idx].stamp_time = stamped_at;
+
+	this->set_cost(idx, cost, stamped_at);
+	return true;
+}
+
+bool CostField::unstamp(size_t idx, const time::time_t &unstamped_at) {
+	if (!this->cost_stamps.contains(idx)) return false;
+	if (unstamped_at < this->cost_stamps[idx].stamp_time) return false;
+
+	cost_t original_cost = cost_stamps[idx].original_cost;
+
+	this->set_cost(idx, original_cost, unstamped_at);
+	return this->cost_stamps.erase(idx) != 0;
+}
+
 bool CostField::is_dirty(const time::time_t &time) const {
 	return time >= this->valid_until;
 }
--- a/libopenage/pathfinding/cost_field.h
+++ b/libopenage/pathfinding/cost_field.h
@@ -107,6 +107,27 @@ public:
 	void set_costs(std::vector<cost_t> &&cells, const time::time_t &changed);
 
 	/**
+	 * Stamp a cost field cell at a given time.
+	 * 
+	 * @param idx Index of the cell.
+	 * @param cost Cost to set.
+	 * @param stamped_at Time at which the cost cell is to be stamped.
+	 * 
+	 * @return True if the cell was successfully stamped, false if the cell was already stamped.
+	 */
+	bool stamp(size_t idx, cost_t cost, const time::time_t &stamped_at);
+
+	/**
+	 * Unstamp a cost field cell at a given time.
+	 * 
+	 * @param idx Index of the cell.
+	 * @param unstamped_at Time at which the cost cell is to be unstamped.
+	 * 
+	 * @return True if the cell was successfully unstamped, false if the cell was already not stamped.
+	 */
+	bool unstamp(size_t idx, const time::time_t &unstamped_at);
+
+	/**
 	 * Check if the cost field is dirty at the specified time.
 	 *
 	 * @param time Time of access to the cost field.
@@ -135,6 +156,11 @@ private:
 	 * Cost field values.
 	 */
 	std::vector<cost_t> cells;
+
+	/**
+	 * Cost field values.
+	 */
+	std::unordered_map<size_t, cost_stamp_t> cost_stamps;
 };
 
 } // namespace path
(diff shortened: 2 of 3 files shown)


libopenage/pathfinding/cost_field.h
```diff
@@ -135,6 +156,11 @@ class CostField {
 	 * Cost field values.
 	 */
 	std::vector<cost_t> cells;
+
+	/**
+	 * Cost field values.
+	 */
+	std::unordered_map<size_t, cost_stamp_t> cost_stamps;
```
jere8184, line 163: Would `vector<optional<cost_stamp_t>>` cost_stamps work better?

jere8184 (changes requested):

libopenage/pathfinding/cost_field.h
dmwever: Didn't know about the existence of optionals. Basically, have a vector the same size as the cost field, with optionals consisting of cost_stamp_t?

libopenage/pathfinding/cost_field.cpp
```diff
@@ -57,24 +57,25 @@ void CostField::set_costs(std::vector<cost_t> &&cells, const time::time_t &valid
...
 
 	this->set_cost(idx, cost, stamped_at);
 	return true;
 }
 
 bool CostField::unstamp(size_t idx, const time::time_t &unstamped_at) {
-	if (!this->cost_stamps.contains(idx)) return false;
-	if (unstamped_at < this->cost_stamps[idx].stamp_time) return false;
+	if (!this->cost_stamps[idx].has_value()) return false;
+	if (unstamped_at < this->cost_stamps[idx]->stamp_time) return false;
```
jere8184, line 72:
```suggestion
	if (!this->cost_stamps[idx].has_value() || unstamped_at < this->cost_stamps[idx]->stamp_time) return false;
```

libopenage/pathfinding/cost_field.h
```diff
@@ -160,7 +161,7 @@ class CostField {
 	/**
 	 * Cost field values.
```
jere8184, line 162: update

jere8184: btw dont feel like you need to impliment any of my suggestions, as they are just suggestions.

libopenage/pathfinding/cost_field.h
```diff
@@ -135,6 +157,11 @@ class CostField {
...
 	 */
 	std::vector<cost_t> cells;
+
+	/**
+	 * Cost stamp vector.
+	 */
+	std::vector<std::optional<cost_stamp_t>> cost_stamps;
```
jere8184, line 164: you need to make sure `cost_stamps` is initialised in the constructor with the correct size

jere8184 (changes requested):

libopenage/pathfinding/cost_field.cpp
```diff
@@ -56,6 +56,27 @@ void CostField::set_costs(std::vector<cost_t> &&cells, const time::time_t &valid
 	this->valid_until = valid_until;
 }
 
+bool CostField::stamp(size_t idx, cost_t cost, const time::time_t &stamped_at) {
+	if (this->cost_stamps[idx].has_value()) return false;
+
+	cost_t original_cost = this->get_cost(idx);
+	this->cost_stamps[idx]->original_cost = original_cost;
+	this->cost_stamps[idx]->stamp_time = stamped_at;
```
heinezen, line 64: I don't think this code will work, i.e. it could segfault because the optional value is not initialized. You have to explicitely assign it a `cost_stamp_t` struct before you can access the members directly. So you have to write something like

```cpp
this->cost_stamps[idx] = cost_stamp_t{
    orginal_cost,
    stamped_at,
};
```

libopenage/pathfinding/cost_field.cpp
```diff
@@ -56,6 +56,27 @@ void CostField::set_costs(std::vector<cost_t> &&cells, const time::time_t &valid
...
+
+	this->set_cost(idx, cost, stamped_at);
+	return true;
+}
+
+bool CostField::unstamp(size_t idx, const time::time_t &unstamped_at) {
+	if (!this->cost_stamps[idx].has_value() || unstamped_at < this->cost_stamps[idx]->stamp_time) return false;
```
heinezen, line 71: if conditions should always be multiline and wrapped in curly brackets, according to our code style conventions. Also, you should use `or`, `not`, and `and` instead of `||`, `!`, and `&&` for readability in the condition.

```suggestion
	if (not this->cost_stamps[idx].has_value() or unstamped_at < this->cost_stamps[idx]->stamp_time) {
	    return false;
	}
```

libopenage/pathfinding/cost_field.h
```diff
@@ -106,6 +107,27 @@ class CostField {
...
 	void set_costs(std::vector<cost_t> &&cells, const time::time_t &changed);
 
+	/**
+	 * Stamp a cost field cell at a given time.
+	 *
+	 * @param idx Index of the cell.
+	 * @param cost Cost to set.
+	 * @param stamped_at Time at which the cost cell is to be stamped.
+	 *
+	 * @return True if the cell was successfully stamped, false if the cell was already stamped.
```
heinezen, line 117: What is the benefit over overwriting the stamp? I'm just curious.

heinezen (changes requested):
Looks pretty good so far.

In addition to the fixes to the comments I made, you should also add unit tests for stamping and unstamping the cost field in `tests.cpp`.

libopenage/pathfinding/cost_field.h
dmwever: Good question. I was considering what would happen if one stamp was placed, and then if another stamp was placed while the first was active. If the stamp is overwritten, how do we communicate with the object that placed the first stamp? If that first object unstamps, it would unstamp the second object's stamp value. If instead we tell the second object that there is already a stamp in place there, we avoid data miscommunication. The downside is that we can't double-stamp, but that would be the downside with stamp overwriting as well.

libopenage/pathfinding/cost_field.cpp
dmwever: Thanks. Learning the ropes of c++ here.

libopenage/pathfinding/cost_field.h
heinezen: The communication is a good reason, although I don't think it's something the pathfinding system should have to worry about for now. Pathfinding is separate from the game logic and is not really concerned with what terrain or game objects are. Notification between objects would happen closer to the game simulation side.

I think we can take your solution for now, but it might be more convenient to just overwrite stamps in the future if that's more reliable.

PR #1746, opened 2025-01-22
Fix windows debug builds

jere8184:


heinezen:
I'm not gonna lioe, this is pretty cool :D

I'm abit worried about long-term maintainability, so let me think about this for a bit and wrap my head around the stackoverflow question and solution.

jere8184:
> I'm not gonna lioe, this is pretty cool :D
> 
> I'm abit worried about long-term maintainability, so let me think about this for a bit and wrap my head around the stackoverflow question and solution.

Thanks, I've changed the link in the docstring to https://github.com/python/cpython/issues/127619 as I think it provides  more clarity

heinezen (approved):

PR #1747, opened 2025-01-24
fix paring heap UB

jere8184:


heinezen:
I'm not really in favor of this change as making the iteration over the heap nodes non-const can have severe side effects. e.g. by changing the node values that are used for comparison, which would result in corruption of the entire datastructure.

I thought about this for a while and I think we should use a dedicated erasure method for `clear` and not use the `iter_all` method. A simple breadth-first or depth-first search would be enough for this. Then we don't have to worry about invalid pointer accesses too.

jere8184:
> e as making the iteration over the heap nodes non-const can have severe side effects. e.g. by changing the node values that are used for

Do you think it might be worth implementing an iterator for the tree?

heinezen:
> Do you think it might be worth implementing an iterator for the tree?

That would be cool! Currently we can get a problem with the recursion depth if the tree is too large. An iterator should solve that.

jere8184:
> > Do you think it might be worth implementing an iterator for the tree?
> 
> That would be cool! Currently we can get a problem with the recursion depth if the tree is too large. An iterator should solve that.

iterator added

heinezen (approved):
Very nice solution!

I changed the test code slightly at the end by changing the C array to a C++ `std::array` which is a bit more "stable" for test code.

Issue #1677, opened 2024-08-25
Vector Curve container


heinezen:
We should add a `curve::Vector` type to our curve container classes to record changes to sequences over time.

The new vector type should implement these methods:

**Read**

| Method          | Description                                                 |
| --------------- | ----------------------------------------------------------- |
| `get(t, i)`        | Get value of index `i` at time `t`                        |
| `get(t)`        | Get vector at time `t`                        |
| `size(t)`        | Get size of the vector at time `t`                        |
| `frame(t, i)`      | Get the previous keyframe (time and value) of index `i` before or at `t` |
| `next_frame(t, i)` | Get the next keyframe (time and value) of index `i` after `t`            |

There should also be an iterator type to iterate over the vector at time `t`.

**Modify**

Modify operations insert values for a specific point in time.

| Method                  | Description                                                                       |
| ----------------------- | --------------------------------------------------------------------------------- |
| `insert(t, i, value)` | Insert a new element at index `i` at time `t` |
| `erase(t, i)` | Erase element at index `i` at time `t` |
| `set_insert(t, i, value)`  | Insert a new keyframe value at time `t` at index `i`                                           |
| `set_last(t, i, value)`    | Insert a new keyframe value at time `t`; delete all keyframes after time `t` at index `i`      |
| `set_replace(t, i, value)` | Insert a new keyframe value at time `t` at index `i`; remove all other keyframes with time `t` at index `i` |

jere8184:
i'll take this aswell

PRDeving:
taking this one 👍

jere8184:
taking this back, implimenting as a std::vector of KeyFrameContainers.

How about publicly inheriting from std::vector<KeyFrameContainer<T>> ?

PR #1748, opened 2025-01-26
Pathfinder history

jere8184:


libopenage/pathfinding/cost_field.h
```diff
@@ -135,6 +161,17 @@ class CostField {
...
+	/**
+	 * Cost stamp vector.
+	 */
+	std::vector<std::optional<cost_stamp_t>> cost_stamps;
+
+
+	/**
+	 * Array curve recording cell cost history,
+	 */
+	curve::Array<cost_t, CHUNK_SIZE> cell_cost_history;
```
jere8184, line 174: How should `cell_cost_history` be exposed? A simple getter that returns a constant ref to the curve?

libopenage/pathfinding/cost_field.cpp
```diff
@@ -54,6 +55,28 @@ void CostField::set_costs(std::vector<cost_t> &&cells, const time::time_t &valid
 
 	this->cells = std::move(cells);
 	this->valid_until = valid_until;
+	this->cell_cost_history.set_insert_range(valid_until, this->cells.begin(), this->cells.end());
+}
+
+bool CostField::stamp(size_t idx, cost_t cost, const time::time_t &stamped_at) {
+	if (this->cost_stamps[idx].has_value()) return false;
```
heinezen, line 62: Please always make if clauses multi-line for consistent code style.

```suggestion
	if (this->cost_stamps[idx].has_value()) {
	    return false;
	}
```

libopenage/pathfinding/cost_field.cpp
```diff
@@ -54,6 +55,28 @@ void CostField::set_costs(std::vector<cost_t> &&cells, const time::time_t &valid
...
+	cost_t original_cost = this->get_cost(idx);
+	this->cost_stamps[idx]->original_cost = original_cost;
+	this->cost_stamps[idx]->stamp_time = stamped_at;
+
+	this->set_cost(idx, cost, stamped_at);
+	return true;
+}
+
+bool CostField::unstamp(size_t idx, const time::time_t &unstamped_at) {
+	if (!this->cost_stamps[idx].has_value() || unstamped_at < this->cost_stamps[idx]->stamp_time) return false;
```
heinezen, line 73: Same here and `not`/`and`/`or` should be used as operators for better readability.

```suggestion
	if (not this->cost_stamps[idx].has_value() or unstamped_at < this->cost_stamps[idx]->stamp_time) {
	    return false;
	}
```

libopenage/pathfinding/cost_field.h
```diff
@@ -14,6 +16,8 @@ namespace coord {
 struct tile_delta;
 } // namespace coord
 
+const unsigned int CHUNK_SIZE = 100;
```
heinezen, line 19: The value should probably not be hardcoded because then the cost field size is not configurable anymore. I think we either need to make `CostField` a templated class or wait for the implementation of `curve::Vector` in https://github.com/SFTtech/openage/issues/1677 . Both approaches would have pros and cons.

libopenage/pathfinding/cost_field.h
heinezen: Yes that would be enough I think.

libopenage/pathfinding/types.h
```diff
@@ -124,4 +125,19 @@ using cache_key_t = std::pair<portal_id_t, sector_id_t>;
  */
 using field_cache_t = std::pair<std::shared_ptr<IntegrationField>, std::shared_ptr<FlowField>>;
 
+/**
+ * Cost stamp for a given cost field cell.
+ */
```
heinezen, line 130: A small explanation of stamping for people who don't know what that is would be nice here. I added a suggesion below.

```suggestion
/**
 * Cost stamp for a cell on a cost field.
 * 
 * Stamps are used when cell costs are temporarily overwritten, e.g. when placing a game object
 * as an obstacle over terrain. Once the obstacle is removed, the cell can be reset to its original
 * value recorded by the stamp.
 */
```

heinezen (changes requested):
Sorry for the long wait, I was very busy with work :sweat_smile: Now that I have some time, I added a few suggestions to the code.

libopenage/pathfinding/cost_field.h
jere8184: Which way do you think we should go?

jere8184: Alot of classes contain a cost_field, plus cost_field itself is constructed with a size_t, should we also templatise the encapsulating classes and get rid of the constructor parameter?

jere8184: okay, didnt mean to request another review btw

heinezen: sry I couldn't help but take a look :sweat_smile:

libopenage/pathfinding/cost_field.h
```diff
@@ -19,6 +28,8 @@ namespace path {
 /**
  * Cost field in the flow-field pathfinding algorithm.
  */
+
+template <size_t N>
```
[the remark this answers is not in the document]
WillemKauf: Wouldn't it make more sense to have a `curve::Vector` here than an `Array`? `size` is not a compile time constant in `Grid/Sector`.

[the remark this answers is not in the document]
heinezen: @WillemKauf That's what my comment in https://github.com/SFTtech/openage/pull/1748#discussion_r2106328434 is related to ^^

libopenage/pathfinding/cost_field.h
jere8184: So just to be clear we're essentially templating the whole of `pathfinding`, including the `Pathfinder` class itself right.

Any ideas on dealing with the linking of the templatised member functions, aside from just putting them in the header, or is that the way to go. i.e. we could include the .cpp file containing the functions in the header file. Not sure if this is a good idea.

heinezen: > So just to be clear we're essentially templating the whole of pathfinding, including the Pathfinder class itself right.

Yes, I think that's necessary since the size of the fields propagates all the way up to `Grid` and `Pathfinder` classes.

> Any ideas on dealing with the linking of the templatised member functions, aside from just putting them in the header, or is that the way to go. i.e. we could include the .cpp file containing the functions in the header file. Not sure if this is a good idea.

You should put them in the header, at least that's what we do for other templated classes, e.g. for the curves.

libopenage/pathfinding/types.h
jere8184: Hi @heinezen, sorry for the delay. Long time no see.

At what point should I stop templatising? Should I templatise Map::Map?

heinezen: No Map::Map should not be templated I think. The templating should probably be constrained to the pathfinder system. But let's see how realistic that is...

PR #1759, opened 2025-03-17
[WIP] Update missing CMAKE_CXX_COMPILER env in ubuntu dockerfile

jsutcodes:


jsutcodes:
seems to be more than just the missing env var, or it did not pick up after the load. Further investigation needed.

jsutcodes:
Ubuntu 24.04 LTS image installs with base of g++-13 and gcc-13.

@heinezen  @TheJJ 
Unsure if this is the correct fix.

 Could either move forward with standard 13 (as currently proposed)
  or
can change the Docker build process to manually install the standard 11 versions.

Please advise, merge or changes. Thanks!

TheJJ (commented):
looks alright! we just want the standard version shipped with the distro :)

TheJJ (approved):

PR #1761, opened 2025-03-22
Add formations formulas reverse engineering

teaalltr:
This adds my results in clean-room reverse engineering the mechanics of choosing the number of lines in formations and so on

### DIFF
--- a/doc/reverse_engineering/game_mechanics/formations-formulas.md
+++ b/doc/reverse_engineering/game_mechanics/formations-formulas.md
@@ -1,0 +1,248 @@
+# Units formations
+
+Military units can be arranged in various formations. Here's an overview on how this may be implemented
+
+### Line formation
+
+If you table the number of units in the first row, in the second row and so on accordingly to the number of units in the group, you'll find a pattern in the units number intervals for a given row count: namely, the length of the first interval `(0...6)` is `6`, the second `(7...15)` is `9 = 6+3 + 0`, the third `(16...28)` is `13 = (6+3) + 3+1`, the fourth (maybe) `(29...46)` is `18 = ((6+3) + 3+1 ) + 3+2`.
+
+The value of `46` is chosen by observing that the difference between the last value for a certain row count and the first value for the following row count is always `3`; in this case, with `46` units `4` rows are formed: the first 3 with 12 units and the last with 10. For the next value, `47`, we guess `5` rows with `10, 10, 10, 10, 7`.
+
+Of course, since there's a limit of 40 to the number of selectable units, there's unfortunately no way to check this guess.
+
+##### Formulas
+
+If we neglect the first interval `(0...6)` from above, we can define a recurrence equation between the others.
+It looks like this:
+
+``` python
+def f(n):
+    if n == 0:
+    	return 9
+    return f(n-1) + n + 3
+```
+
+In order to find the nth last value before the row count changes, we just need to sum f(i) from 0 up to n. Then, in order to get the row count for a specified numOfUnits, we can iteratively check which of the values is the upper bound of the interval where our numOfUnits lies:
+
+``` python
+def findRowCount(numOfUnits):
+    # the last numOfUnits before the row count changes
+    lastBeforeInc = 6  
+    # iteration counter
+    i = 0
+
+    # for the first row
+    if numOfUnits <= 6:
+        return 1
+
+    while numOfUnits > lastBeforeInc: # is numOfUnits greater than last bound?
+        lastBeforeInc += f(i) # no, increase the bound
+        i += 1
+
+    rowCount = i + 1
+    return (lastBeforeInc, rowCount)
+```
+
+Finally, we have to find the number of units for each row. The first row from the list we got from experimentation leads to the following formula:
+
+``` python
+rows[0] = math.ceil(numOfUnits / findRowCount(numOfUnits))
+```
+
+This seems to agree with experimental data and nicely extends to values over 40.
+Rows other from the last one have the same value of the first. The last row is simply:
+
+``` python
+rows[rowCount - 1] = numOfUnits - (rowCount-1)*rows[0]
+```
+
+or, if you care about performance:
+
+``` python
+rows[rowCount - 1] = numOfUnits % (rowCount-1)
+```
+
+**Considerations** : for performance sake (or just out of curiosity), you may want to find a better, non recursive function for `f`. It turns out that the recurrence equation can be solved to
+
+        f(n) = (18 + 7*n + n^2)/2
+
+using Mathematica's
+
+``` mathematica
+        RSolve[{f[n] == f[n-1] + n + 3, f[0] == 9}, f[n], n]
+``` 
+
+We can separate `6 + f(0) + f(1) + f(2) + ... + f(n)` (the formula for `lastBeforeInc`) and make it a function of the upper bound `n`:
+
+``` python
+def findNthLastBeforeInc(n): 
+    return (90 + 65*n + 12*n^2 + n^3)/6
+``` 
+
+The above is the result of Mathematica's:
+
+``` mathematica
+        6 + Sum[1/2 (18+7 i+i^2), {i, 0, n}]
+``` 
+
+The `while` loop in `findRowCount` simply becomes:
+
+``` python
+    while numOfUnits > lastBeforeInc: # numOfUnits greater than last bound?
+        lastBeforeInc = findNthLastBeforeInc(i) # no, get the next bound
+        i += 1
+``` 
+
+This allows us to get rid of the **O(n)** recursive function. However, it is a third-degree function.
+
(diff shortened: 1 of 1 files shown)


doc/reverse_engineering/game_mechanics/formations-formulas.md
```diff
@@ -0,0 +1,248 @@
+# Units formations
+
+Military units can be arranged in various formations. Here's an overview on how this may be implemented
+
+### Line formation
+
+If you table the number of units in the first row, in the second row and so on accordingly to the number of units in the group, you'll find a pattern in the units number intervals for a given row count: namely, the length of the first interval `(0...6)` is `6`, the second `(7...15)` is `9 = 6+3 + 0`, the third `(16...28)` is `13 = (6+3) + 3+1`, the fourth (maybe) `(29...46)` is `18 = ((6+3) + 3+1 ) + 3+2`.
```
heinezen, line 7: I think the readability of your whole text suffers from "too much math lingo" :D I can understand what you are saying but someone implementing this mechanic will probably not have a great time. Especially someone without a mathematical background. You can also probably fix a lot of this with formatting things differently, e.g. by making a bullet point llist for the intervals or, even better, a table.

Here's an example of what I would like to see:

```suggestion
Units in a line formation are grouped into several lines. The number of lines and units per line is determined by the total number of units in the group.

The pattern for the number of lines is as follows:

| No. of Lines | Unit Count  |
|--------------|-------------|
| 1            | `(0...6)`   |
| 2            | `(7...15)`  |
| 3            | `(16...28)` |
| 4            | `(29...45)` |
| 5            | `(46...60)` | 
```

heinezen, line 7: Where does the 13 = (6+3) + 3+1` come from?

heinezen, line 7: The forth interval is `(29...45)` which you can actually check in HD Edition and DE2 since those have a selection limit of 60.

doc/reverse_engineering/game_mechanics/formations-formulas.md
```diff
@@ -0,0 +1,248 @@
...
+
+If we neglect the first interval `(0...6)` from above, we can define a recurrence equation between the others.
+It looks like this:
+
+``` python
+def f(n):
+    if n == 0:
+    	return 9
+    return f(n-1) + n + 3
+```
```
heinezen, line 23: Please give your functions and variables appropriate names, I have no idea what `f` does. `n` is the unit count I presume?

doc/reverse_engineering/game_mechanics/formations-formulas.md
```diff
@@ -0,0 +1,248 @@
...
+
+In order to find the nth last value before the row count changes, we just need to sum f(i) from 0 up to n. Then, in order to get the row count for a specified numOfUnits, we can iteratively check which of the values is the upper bound of the interval where our numOfUnits lies:
+
+``` python
+def findRowCount(numOfUnits):
+    # the last numOfUnits before the row count changes
+    lastBeforeInc = 6  
+    # iteration counter
+    i = 0
```
heinezen, line 32: If you write Python code, it should adhere to Python style guides, i.e. snake case function/variable names.

doc/reverse_engineering/game_mechanics/formations-formulas.md
```diff
@@ -0,0 +1,248 @@
...
+    return (lastBeforeInc, rowCount)
+```
+
+Finally, we have to find the number of units for each row. The first row from the list we got from experimentation leads to the following formula:
+
+``` python
+rows[0] = math.ceil(numOfUnits / findRowCount(numOfUnits))
+```
+
+This seems to agree with experimental data and nicely extends to values over 40.
```
heinezen, line 52: Instead of writing about "experimental data", I would recommend you just add an example for the calculation. You can also visualize this better using the ASCII approach from [the other formations docs](https://github.com/SFTtech/openage/blob/master/doc/reverse_engineering/game_mechanics/formations.md#subformations).

doc/reverse_engineering/game_mechanics/formations-formulas.md
```diff
@@ -0,0 +1,248 @@
...
+
+``` python
+rows[rowCount - 1] = numOfUnits - (rowCount-1)*rows[0]
+```
+
+or, if you care about performance:
+
+``` python
+rows[rowCount - 1] = numOfUnits % (rowCount-1)
+```
```
heinezen, line 63: I'm not sure if the second option is actually faster. In any case, I think the document should not be about coding recommendations and more about explaining the mechanic... and multiple code snippets can make this harder to understand.

I recommend you only put in one of these two code snippets. Preferrably the most readable one which for me is the second one.

doc/reverse_engineering/game_mechanics/formations-formulas.md
```diff
@@ -0,0 +1,248 @@
...
+
+**Considerations** : for performance sake (or just out of curiosity), you may want to find a better, non recursive function for `f`. It turns out that the recurrence equation can be solved to
+
+        f(n) = (18 + 7*n + n^2)/2
+
+using Mathematica's
+
+``` mathematica
+        RSolve[{f[n] == f[n-1] + n + 3, f[0] == 9}, f[n], n]
+```
```
heinezen, line 73: Same problem as above, i.e. you're now recommending more code. This time also in a language that we don't even use in our codebase :D I appreciate what you did here, but it's not really helpful for someone who wants to learn what how to replicate the formation formulas from AoE2. Same goes for the rest of this section which gets too much into details about mathematical specifics that are hard to understand.

I think you should do 2 things to improve this section:

1. Rewrite it so that this section only contains the formulas and their explanation. We don't need much derivation.
2. Create only a small example script which contains the **most readable** mock implmentation for these formulas.

doc/reverse_engineering/game_mechanics/formations-formulas.md
```diff
@@ -0,0 +1,248 @@
...
+and the other rows are unchanged.
+
+The final Python code may be:
+
+``` python
+import math
+
+def findRowCount(numOfUnits):
+    return math.ceil((math.sqrt(1 + 8*numOfUnits) - 3) / 4)
+```
```
heinezen, line 211: This should be at the start of the formulas section. We don't need the unoptimized solutions.

doc/reverse_engineering/game_mechanics/formations-formulas.md
```diff
@@ -0,0 +1,248 @@
...
+    -Quotient[N, Quotient[3 - Sqrt[BitShiftLeft[N, 3]], 4]]
+``` 
+
+which not contains the addition and uses a left shift to do the multiplication, thus being faster.
+Or one may use the following
+
+``` mathematica
+    -Quotient[N, Quotient[2 - Floor[Sqrt[BitShiftLeft[N, 3]]], 4]]
+``` 
+where `Floor[Sqrt[n]]` is the [integer square root function](https://en.wikipedia.org/wiki/Integer_square_root) and there exist plenty of optimized algorithms to do it.
```
heinezen, line 235: This whole section is interesting, but I would say it's not relevant at all in practice. Usually, any compiler or interpreter will optimize this code much better. See also our documentation on [optimization](https://github.com/SFTtech/openage/blob/master/doc/code/optimization.md), especially item no. 2

> - **readability comes first**: openage depends on outsiders being able to understand the engine code. If the code is super fast but totally incomprehensible, then nobody is going to maintain it.
> - **avoid premature optimization**: optimization without reason wastes time, reduces readability, and can sometimes lead to even worse performance. It's often better to implement a workable solution first and then use profiling to identify bottlenecks. Modern compilers are often smart enough to find small improvements during compilation anyway.
> - **prefer architecture improvements over speeding up specific use cases**: openage is a game engine, so we don't exactly know what people will use it for. We want the general case to be fast, not the edge cases.

doc/reverse_engineering/game_mechanics/formations-formulas.md
```diff
@@ -0,0 +1,248 @@
...
+
+The general formula is `M*i - (i-u)` with some constraints on `M` and `u`.
+This gives us:
+
+``` mathematica
+        Ceiling[N / Ceiling[(1-M-2*u + Sqrt[(-1+M)*(-9+M+ 8*N) + 4*(-1+M) u+4 u^2]) / (2*(-1+M))]]
+``` 
+
+The constraints on N, M and u are `(N | M | u) ∈ ℤ && u ≥ 1 && M ≥ 2 && N ≥ 2`.
+This is useful if we're not satisfied with the default game behavior :smile:
```
heinezen, line 248: What would this be used for?

heinezen (changes requested):
I like the overall idea of this document, but you should try to make this read less like someone's math homework :sweat_smile:

heinezen:
You also need to fix the failing CI checks. Click on them to see more details.

PR #1765, opened 2025-04-02
[ADD] fixed point operations pow and log ALSO fixed point test cases

PRDeving:
[Issue](https://github.com/SFTtech/openage/issues/1543)

- Implemented fixed point log function
- Implemented fixed point pow function
- Implemented missing testcase for sin
- Implemented missing testcase for cos
- Implemented missing testcase for tan

### DIFF
--- a/libopenage/util/fixed_point.h
+++ b/libopenage/util/fixed_point.h
@@ -465,6 +465,14 @@ public:
 	constexpr double tan() {
 		return std::tan(this->to_double());
 	}
+
+	constexpr double pow(const FixedPoint &exponent) {
+		return std::pow(this->to_double(), exponent.to_double());
+	}
+
+	constexpr double log() {
+		return std::log(this->to_double());
+	}
 };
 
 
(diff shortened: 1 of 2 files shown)


copying.md
```diff
@@ -165,6 +165,7 @@ _the openage authors_ are:
 | Ana Trias-Labellarte        | anatriaslabella             | ana dawt triaslabella à ufl dawt edu              |
 | Eelco Empting               | Eeelco                      | me à eelco dawt de                                |
 | Jordan Sutton               | jsutCodes                   | jsutcodes à gmail dawt com                        |
+| Pablo Roizo                 | PRDeving                    | pablo dawt deving à gmail dawt com                        |
```
heinezen, line 168:
```suggestion
| Pablo Roizo                 | PRDeving                    | pablo dawt deving à gmail dawt com                |
```

heinezen (changes requested):
Hey, this is not really what we want with the issue because the point is to have fixed-point values as the results (i.e. not double, float, etc.). We want these operations to be purely fixed-point based, so the game simulation stays deterministic.

I think no one working on this issue has actually understood that yet, so I tried to clarify it in the issue description now :D The code is still useful, so I would merge it, but it still doesn't really solve the issue.

PRDeving:
> Hey, this is not really what we want with the issue because the point is to have fixed-point values as the results (i.e. not double, float, etc.). We want these operations to be purely fixed-point based, so the game simulation stays deterministic.
> 
> I think no one working on this issue has actually understood that yet, so I tried to clarify it in the issue description now :D The code is still useful, so I would merge it, but it still doesn't really solve the issue.

Makes sense, I just sticked to the previous iterations, I'll redo all operators to comply with it

Issue #1766, opened 2025-04-04
Missing symbol on assets conversion.


Daniel-Wieczorek:
Hi guys, I want to contribute but I encountered an error while running the openage project. The error occurs when attempting to convert assets after launching. Below is the terminal output:

```
daniel@DanielPC:~/workareas/openage$ make run
cd bin && ./run main
INFO [py] launching openage 
INFO [py] compiled by GNU 13.3.0
INFO [py] running in DEVMODE
INFO [py] openage nyan API modpack is up to date
  Do you want to convert assets? [Y/n]
> Y

openage requires a local game installation for conversion
but no local installation could be found automatically.
  Do you want to download the AoC trial version? [Y/n]
> Y
Downloading AoC trial version from https://archive.org/download/AgeOfEmpiresIiTheConquerorsDemo/Age2XTrial.exe
Extracting game files to [Directory(/tmp/tmp17ck0xzt)]:...
converting from '/tmp/tmp17ck0xzt'
INFO [py] Looking for compatible games to convert...
INFO [py] Compatible game edition detected:
INFO [py]  * Age of Empires 2: The Conqueror's Trial Version
INFO [py] converting metadata
INFO [py] using cached wrapper: /tmp/AOC_empires2_x1.dat.pickle
INFO [py] Finished metadata read (1.20 seconds)
INFO [py] Starting conversion...
INFO [py] Extracting Genie data...
INFO [py] Creating API-like objects...
INFO [py] Linking API-like objects...
INFO [py] Generating auxiliary objects...
INFO [py] Creating nyan objects...
INFO [py] Creating requests for media export...
INFO [py] Finished data conversion (1.46 seconds)
INFO [py] Starting export...
INFO [py] Dumping info file...
INFO [py] Dumping data files...
INFO [py] Exporting media files...
INFO [py] -- Exporting graphics files...
Exception in thread Thread-3 (_handle_results):
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/usr/lib/python3.12/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
                    ^^^^^^^^^^^^^^^^^^^
  File "/home/daniel/workareas/openage/.bin/g++-debug-Oauto-sanitize-none/openage/convert/processor/export/media_exporter.py", line 819, in _export_texture
    texture = Texture(image, palettes)
              ^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1767, opened 2025-04-06
Fix assets conversion bug. Add running with docker docs

Daniel-Wieczorek:
Fix for bug described in: https://github.com/SFTtech/openage/issues/1766 

- Using Cython < 3.0.10 leads to the runtime errors on assets conversion. Cython >= 3.0.10 shall be used.
- Update docker build file and add chapter in documentation about running image with Docker.

Evidence:
- [Docker run log](https://github.com/user-attachments/files/19621330/docker_run.txt)
**Running inside docker image:**

image


.mailmap
```diff
@@ -23,3 +23,4 @@ Derek Frogget <fro22003@byui.edu> <114030121+derekfrogget@users.noreply.github.c
 Nikhil Ghosh <nghosh606@gmail.com>
 David Wever <dmwever@crimson.ua.edu> <56411717+dmwever@users.noreply.github.com>
 Ngô Xuân Minh <xminh.ngo.00@gmail.com>
+Daniel Wieczorek <daniel1223417@gmail.com>
```
TheJJ, line 26: i don't think you need an entry, this is just for mapping commits with accidentally wrong mails to the correct. you can just edit your commit directly.

doc/building.md
```diff
@@ -162,6 +162,42 @@ The reference package is [created for Gentoo](https://github.com/SFTtech/gentoo-
   which will then be packed/installed by your package manager.
 
 
+### Running with Docker
```
TheJJ, line 165: can we do a separate file for this please?

doc/building.md
```diff
@@ -162,6 +162,42 @@ The reference package is [created for Gentoo](https://github.com/SFTtech/gentoo-
...
+   ```bash
+   sudo docker build -t openage -f packaging/docker/devenv/Dockerfile.ubuntu.2404 .
+   ```
+
+3. **Run the Docker Container**
+
+   Run docker container and mount necessary X11 socket and .Xauthority file to
+   enable GUI for openage project.
+   ```bash
+   docker run -it -e DISPLAY=$DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix -v $HOME/.Xauthority:/root/.Xauthority --network host openage
```
TheJJ, line 194: i (and likely many others) have wayland. mind adding the forwards for it, too?

packaging/docker/devenv/Dockerfile.ubuntu.2404
```diff
@@ -36,4 +36,7 @@ RUN apt-get update && DEBIAN_FRONTEND="noninteractive" apt-get install -y sudo \
     qt6-multimedia-dev \
     qml6-module-qtquick3d-spatialaudio \
     && sudo apt-get clean \
-    && truncate -s 0 ~/.bash_history
\ No newline at end of file
+    && truncate -s 0 ~/.bash_history
+
+# At least cython >= 3.0.10 is required to avoid runtime errors
+RUN pip install cython==3.0.10 --break-system-packages
```
TheJJ, line 42: can we have a >=3.0.10 <4.0.0 here for future compatibility please?

TheJJ (changes requested):

packaging/docker/devenv/Dockerfile.ubuntu.2404
heinezen, line 42: Can you add a TODO here to remove this line once cython is upgraded. Should probably happen in 24.04.3 (around August 2025).

doc/build_instructions/ubuntu.md
```diff
@@ -7,6 +7,10 @@ Run the following commands:
 
 You will also need [nyan](https://github.com/SFTtech/nyan/blob/master/doc/building.md) and its dependencies.
 
+Notes:
+- The project requires at least **Cython 3.0.10** for successful compilation.
+  Using an older version may result in runtime errors!
+
 # Additional steps for Ubuntu 22.04 LTS
 
 The available system version of Cython is too old in Ubuntu 22.04. You have to get the correct version
```
heinezen, line 16: Adding installation instructions would be nice, but I think we can just merge this paragraph with the section below.

```suggestion
# Additional steps for Ubuntu 22.04 LTS & 24.04 LTS

The available system version of Cython is too old in Ubuntu 22.04 & 22.04. You have to get the correct version
```

README.md
```diff
@@ -145,6 +145,9 @@ To turn them off, use `./bin/run --dont-segfault --no-errors --dont-eat-dog`.
 If this still does not help, try our [troubleshooting guide](/doc/troubleshooting.md), the [contact section](#contact)
 or the [bug tracker](https://github.com/SFTtech/openage/issues).
 
+* **Running with Docker**
+You can build and run the project using Docker. See [Running with docker](/doc/building.md#running-with-docker)
+
```
heinezen, line 150: Adding this text to the end of the quickstart section seems a bit out-of-place. Maybe you can put it under "How do I get this to run on my box?" as an alternative method?

heinezen (changes requested):
Very nice!

.mailmap
Daniel-Wieczorek: Fixed - removed

doc/building.md
Daniel-Wieczorek: fixed: doc/build_instructions/docker.md

packaging/docker/devenv/Dockerfile.ubuntu.2404
Daniel-Wieczorek: Fixed

doc/build_instructions/ubuntu.md
Daniel-Wieczorek: Fixed

README.md
Daniel-Wieczorek: Fixed

doc/building.md
Daniel-Wieczorek: Fixed:

Daniel-Wieczorek:
CI is failing, I see few more changes will be needed here. I will add it before end of this week

Daniel-Wieczorek:
@TheJJ & @heinezen feel free to recheck 😉

kevinfile
```diff
@@ -14,6 +14,7 @@ sanity_check:
 configure:
 	- env: mode=debug   compiler=gcc            (? if job == "debian" ?)
 	- env: mode=debug   compiler=clang          (? if job == "debian-clang" ?)
+	pip3 install --upgrade "cython>=3.0.10" --break-system-packages
```
heinezen, line 17: This line should be removed because package installation is handled manually on the build system.

doc/build_instructions/docker.md
```diff
@@ -0,0 +1,77 @@
...
+     docker run -it \
+       -e DISPLAY=$DISPLAY \
+       -v /tmp/.X11-unix:/tmp/.X11-unix \
+       -v $HOME/.Xauthority:/root/.Xauthority \
+       --network host openage
+     ```
+
+   - **For Wayland**:
+     ```bash
+    docker run -it \
```
heinezen, line 53: If you don't indent properly, the formatting looks all wrong in rendered markdown.

```suggestion
     docker run -it \
```

doc/build_instructions/docker.md
```diff
@@ -0,0 +1,77 @@
...
+
+Docker simplifies the setup process by providing a consistent development environment. It allows you to build and run the project without manually installing dependencies. Currently provided [Dockerfile.ubuntu.2404](../../packaging/docker/devenv/Dockerfile.ubuntu.2404) supports Ubuntu 24.04 as the base operating system.
+
+#### Prerequisites
+
+- **Docker**: Ensure Docker is installed on your machine. Follow the [official documentation](https://docs.docker.com/) for installation instructions.
+- **Display Server**: This guide supports both **X11** and **Wayland** display servers for GUI applications.
+- **Tested Configuration**: These instructions were tested on Ubuntu 24.04 running on WSL2 on Windows 11.
+
+---
```
heinezen, line 11: Why make a section divider when you are starting a new chapter right after this line?

doc/build_instructions/docker.md
```diff
@@ -0,0 +1,77 @@
...
+      -v $XDG_RUNTIME_DIR/$WAYLAND_DISPLAY:/tmp/$WAYLAND_DISPLAY \
+      --user=$(id -u):$(id -g) \
+      --network host openage
+     ```
+
+4. **Follow the Regular Setup**
+
+Once inside the container, follow the regular setup described in the [Development](../building.md#development) chapter. You can skip dependency installation since the Docker image already includes all required dependencies.
+
+---
```
heinezen, line 66: same here

doc/build_instructions/docker.md
```diff
@@ -0,0 +1,77 @@
...
+
+#### Notes
+
+- **X11 vs. Wayland**: Ensure you know which display server your system is using. Most modern Linux distributions default to Wayland, but X11 is still widely used.
+- **Permissions**: For Wayland, you may need to adjust permissions for the Wayland socket (`/run/user/$(id -u)/wayland-0`) to allow Docker access.
+- **GUI Applications**: These configurations enable GUI applications to run inside the Docker container.
+
+---
```
heinezen, line 74: and here

kevinfile
Daniel-Wieczorek: Without it all CI checks fail. Default Cython version in image used on CI seems to be 0.29.x so `configure` scripts fails. How should we update it then?

heinezen: I don't see CI checks failing without it though. See https://github.com/SFTtech/openage/pull/1688 which just got updated and where the debian builds pass just fine.

doc/build_instructions/docker.md
Daniel-Wieczorek: For visual separation of main sections. Personal taste. Removed.

doc/build_instructions/docker.md
Daniel-Wieczorek: removed

doc/build_instructions/docker.md
Daniel-Wieczorek: removed

doc/build_instructions/docker.md
Daniel-Wieczorek: Fixed, should be fine now

kevinfile
Daniel-Wieczorek: Removed: https://github.com/SFTtech/openage/compare/d3c6e974a8feaf2af5b490f1452553397d9d576f..ecafbd40ce8e8f665963a92d262a5425c9ed6c3b. 

See the [CI](https://ci.sft.mx/?wsurl=wss://ci.sft.mx/ws&staticurl=https://cidata.sft.mx/&project=openage&hash=ecafbd40ce8e8f665963a92d262a5425c9ed6c3b&job=debian) checks, failing again:
image

heinezen: Ah that's now because of the CMake check. @TheJJ can you upgrade cython on the CI machines?

TheJJ: i can't do that for a while, but maybe i find someone who can. or we don't enforce the higher version check on Linux, since it apparently works with lesser versions on Ubuntu?

heinezen: it only happens for Cython versions 3.0.8 to 3.0.10, so 0.29 is okay. However, we can't specifiy this in CMake language.

heinezen: I'd say we introduce @Daniel-Wieczorek changes here as a dirty fix and add a TODO to change it in the future. The fix is probably more important right now than waiting for a CI upgrade.

TheJJ: we could do two version range checks https://cmake.org/cmake/help/latest/command/find_package.html

heinezen: So something like this?

```cmake
find_package(Cython 3.0.10 QUIET)
if (CYTHON_FOUND)
    message("Using Cython >3.0.10")
else()
    find_package(Cython 0.29.32 REQUIRED)
```

Daniel-Wieczorek: Okay, let me try

buildsystem/HandlePythonOptions.cmake
```diff
@@ -4,7 +4,17 @@
 
 # the Python version number requirement is in modules/FindPython_test.cpp
 find_package(Python ${PYTHON_MIN_VERSION} REQUIRED)
-find_package(Cython ${CYTHON_MIN_VERSION} REQUIRED)
+find_package(Cython 3.0.10 QUIET)
+if (CYTHON_FOUND AND CYTHON_VERSION VERSION_GREATER_EQUAL 3.0.10)
+    message("Using Cython >= 3.0.10 (found version ${CYTHON_VERSION})")
+else()
+    find_package(Cython 0.29.32 REQUIRED)
```
heinezen, line 11:
```suggestion
    find_package(Cython ${CYTHON_MIN_VERSION_FALLBACK}...${CYTHON_MAX_VERSION_FALLBACK} REQUIRED)
```

(the check is equivalent to `find_package(Cython 0.29.32...3.0.7 REQUIRED)`)

If you don't specify the range, Cython 3.0.8 would also be accepted.

buildsystem/HandlePythonOptions.cmake
```diff
@@ -4,7 +4,17 @@
 
 # the Python version number requirement is in modules/FindPython_test.cpp
 find_package(Python ${PYTHON_MIN_VERSION} REQUIRED)
-find_package(Cython ${CYTHON_MIN_VERSION} REQUIRED)
+find_package(Cython 3.0.10 QUIET)
+if (CYTHON_FOUND AND CYTHON_VERSION VERSION_GREATER_EQUAL 3.0.10)
+    message("Using Cython >= 3.0.10 (found version ${CYTHON_VERSION})")
```
heinezen, line 9:
```suggestion
if (CYTHON_FOUND)
    message("Using Cython >= ${CYTHON_MIN_VERSION} (found version ${CYTHON_VERSION})")
```

You don't need the second condition in the if as that is already checked by line 7. We should also still try to use the `CYTHON_MIN_VERSION` parameter.

buildsystem/HandlePythonOptions.cmake
```diff
@@ -4,7 +4,17 @@
...
+else()
+    find_package(Cython 0.29.32 REQUIRED)
+    if (CYTHON_VERSION VERSION_EQUAL 0.29.32)
+        message("Using Cython 0.29.32 (fallback version)")
+    else()
+        message(FATAL_ERROR "Cython version must be >= 3.0.10 or exactly 0.29.32. Found: ${CYTHON_VERSION}")
+    endif()
+endif()
```
heinezen, line 17:
```suggestion
endif()
```

The if block is not necessary as
1. `find_package` already prints a message similar to line 13.
2. the else will never be triggered because `find_package` will error earlier since it specifies `REQUIRED`

buildsystem/HandlePythonOptions.cmake
```diff
@@ -4,7 +4,17 @@
 
 # the Python version number requirement is in modules/FindPython_test.cpp
 find_package(Python ${PYTHON_MIN_VERSION} REQUIRED)
-find_package(Cython ${CYTHON_MIN_VERSION} REQUIRED)
+find_package(Cython 3.0.10 QUIET)
+if (CYTHON_FOUND AND CYTHON_VERSION VERSION_GREATER_EQUAL 3.0.10)
+    message("Using Cython >= 3.0.10 (found version ${CYTHON_VERSION})")
+else()
```
heinezen, line 10:
```suggestion
else()
    message("Cython version ${CYTHON_MIN_VERSION} not available. Searching for ${CYTHON_MIN_VERSION_FALLBACK} instead")
```

You can add a failback message before the next `find_package` call here.

CMakeLists.txt
```diff
@@ -39,7 +39,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 # Python and Cython requirements
 set(PYTHON_MIN_VERSION 3.9)
-set(CYTHON_MIN_VERSION 0.29.31)
+set(CYTHON_MIN_VERSION 3.0.10)
```
heinezen, line 43:
```suggestion
set(CYTHON_MIN_VERSION 3.0.10)
set(CYTHON_MIN_VERSION_FALLBACK 0.29.32)
set(CYTHON_MAX_VERSION_FALLBACK 3.0.7)

```

I recommend you also add a `CYTHON_MIN_VERSION_FALLBACK` and `CYTHON_MAX_VERSION_FALLBACK` parameters.

heinezen (changes requested):

buildsystem/HandlePythonOptions.cmake
Daniel-Wieczorek: Can you please attach a log where the problem is because I tested it locally with 3.0.8 in venv at it worked as expected:
image

Daniel-Wieczorek: Also range seems to be not supported for Cython
image
I will check it this weekend.

Daniel-Wieczorek: Also range seems to be not supported for Cython
image
I will check it this weekend.

heinezen: Ah my bad, disregard my comment about the range here. When i wrote the comment, I thought that CMake would accept all newer versions after 0.29.32 with that call... Which actually makes no sense as it should only accept versions with minor prefix. So 3.0.8 is excluded from that.

No changes necessary here then.

CMakeLists.txt
heinezen: On second thought, the max fallback version is not necessary I think.

buildsystem/HandlePythonOptions.cmake
Daniel-Wieczorek: Ad.1 I wll move.
Ad.2 It will be triggered for versions lower than 3.0.10 and not 0.29.32:
image

buildsystem/HandlePythonOptions.cmake
Daniel-Wieczorek: Fixed

buildsystem/HandlePythonOptions.cmake
Daniel-Wieczorek: Added

CMakeLists.txt
Daniel-Wieczorek: Updated

buildsystem/HandlePythonOptions.cmake
```diff
@@ -4,7 +4,24 @@
...
+    message("Using Cython >= ${CYTHON_MIN_VERSION} (found version ${CYTHON_VERSION})")
+else()
+    message("Cython version ${CYTHON_MIN_VERSION} not available. Searching for fallback version ${CYTHON_MIN_VERSION_FALLBACK} ")
+    find_package(Cython QUIET)
+    if (CYTHON_FOUND AND CYTHON_VERSION VERSION_EQUAL ${CYTHON_MIN_VERSION_FALLBACK})
+        message("Using Cython ${CYTHON_VERSION} (fallback version)")
+    else()
+        message(FATAL_ERROR "Cython version must be >= ${CYTHON_MIN_VERSION} or exactly ${CYTHON_MIN_VERSION_FALLBACK}. Found: ${CYTHON_VERSION}")
+    endif()
+endif()
```
Daniel-Wieczorek, line 24: For reviewer:
Updated logic and messages. All conditions checked locally. From my understanding we need here version at least 3.0.10  or specific 0.29.32 (for CI).Messages are verbose now:

---

---

---

---

heinezen: Ah the 0.29 version does not have to be specific, i think that's a misunderstanding. It just has to be >= 0.29.31 because after this version, it's forward-compatible with Cython 3.0

heinezen: So no equality check necessary.

Daniel-Wieczorek: Hmmm... but we still need version >=3.0.10 because of the bug with asset covnversion in 3.0.8, right? Build with cython 3.0.8 shall fail on config. I don't want to add support for all versions of Cython >=0.29.31, because that won't fix the problem this bug was created for. I figured that compatibility with 0.29.31 is needed as a temporary fix for CI, but in general version 3.0.10 will be used by users. I am not sure if we are on the same page... 🤔

heinezen: I think it's just another misunderstanding :D

Cython 0.29.31 is still the minimum supported version. That shouldn't change, especially since not all distros have Cython 3 in their packages. Our issue that we are trying to solve is that compilation works with all versions since 0.29.31 **except** 3.0.8/3.0.9 (because that's were the bug occurs). So 0.29.31 is not a temporary fix, it's actually compatible and fine.

Ideally, the build system should allow

- All versions in 0.29 after 0.29.31
- All versions in 3.0 after 3.0.10

Daniel-Wieczorek: @heinezen [updated](https://github.com/SFTtech/openage/compare/f5b92e3b177d389b56291aa2ce507cc11e8d493b..4c86e1f2f96be1c897622da159abb289267af25a) to follow your expectations. Also I fixed indentation in file, it did not look correct. Feel free to recheck.

heinezen (approved):

TheJJ:
thanks for the patience and your contribution!

PR #1770, opened 2025-04-13
Add pure FixedPoint sqrt implementation and tests

bytegrrrl:
This implements a pure fixed point square root function per #1543.

Notes:
I've just replaced the existing function with a pure fixed point implementation (can't overload when only return type differs). It should be trivial to make it return an actual FixedPoint instead, but that would requiring updating prior test cases and I wasn't sure if that was okay to do.

There's a small loss in precision depending on the value of `fractional_bits` and the position of the most significant bit: if the integer portion is very large, we won't have as much (absolute) precision. Ideally you would want the `intermediate_type` to be twice the size of `raw_type` to avoid any losses.

### DIFF
--- a/libopenage/util/fixed_point.h
+++ b/libopenage/util/fixed_point.h
@@ -3,6 +3,7 @@
 #pragma once
 
 #include <algorithm>
+#include <bit>
 #include <climits>
 #include <cmath>
 #include <iomanip>
@@ -446,8 +447,37 @@ public:
 		return is;
 	}
 
+	/**
+	 * Pure FixedPoint sqrt implementation using Heron's Algorithm.
+	 *
+	 * Note that this function is undefined for negative values.
+	 */
 	constexpr double sqrt() {
-		return std::sqrt(this->to_double());
+		// Zero can cause issues later, so deal with now.
+		if (this->raw_value == 0) {
+			return 0.0;
+		}
+
+		// A greater shift = more precision, but can overflow the intermediate type if too large.
+		size_t max_shift = std::countl_zero((unsigned_intermediate_type)this->raw_value) - 1;
+		size_t shift = max_shift > fractional_bits ? fractional_bits : max_shift;
+		shift &= ~1;
+
+		// We can't use the safe shift since the shift value is unknown at compile time.
+		intermediate_type n = (intermediate_type)this->raw_value << shift;
+		intermediate_type guess = (intermediate_type)1 << fractional_bits;
+
+		for (size_t _i = 0; _i < fractional_bits; _i++) {
+			intermediate_type prev = guess;
+			guess = (guess + n / guess) / 2;
+			if (guess == prev)
+				break;
+		}
+
+		// The sqrt operation halves the number of bits, so we'll we'll have to calculate a shift back
+		size_t unshift = fractional_bits - (shift + fractional_bits) / 2;
+
+		return from_raw_value(guess << unshift).to_double();
 	}
 
 	constexpr double atan2(const FixedPoint &n) {
--- a/libopenage/util/fixed_point.h
+++ b/libopenage/util/fixed_point.h
@@ -451,11 +451,11 @@ public:
 	 * Pure FixedPoint sqrt implementation using Heron's Algorithm.
 	 *
 	 * Note that this function is undefined for negative values.
-         *
-         * There's a small loss in precision depending on the value of fractional_bits and the position of
-         * the most significant bit: if the integer portion is very large, we won't have as much (absolute)
-         * precision. Ideally you would want the intermediate_type to be twice the size of raw_type to avoid
-         * any losses.
+	 *
+	 * There's a small loss in precision depending on the value of fractional_bits and the position of
+	 * the most significant bit: if the integer portion is very large, we won't have as much (absolute)
+	 * precision. Ideally you would want the intermediate_type to be twice the size of raw_type to avoid
+	 * any losses.
 	 */
 	constexpr FixedPoint sqrt() {
 		// Zero can cause issues later, so deal with now.
@@ -464,16 +464,16 @@ public:
 		}
 
 		// Check for negative values
-		ENSURE(std::is_unsigned<raw_type>() or std::is_signed<raw_type>() and this->raw_value > 0, "FixedPoint::sqrt() is undefined for negative values.");
+		ENSURE(std::is_unsigned<raw_type>() or (std::is_signed<raw_type>() and this->raw_value > 0), "FixedPoint::sqrt() is undefined for negative values.");
 
 		// A greater shift = more precision, but can overflow the intermediate type if too large.
 		size_t max_shift = std::countl_zero(static_cast<unsigned_intermediate_type>(this->raw_value)) - 1;
 		size_t shift = max_shift > fractional_bits ? fractional_bits : max_shift;
 
-                // shift + fractional bits must be an even number
-                if ((shift + fractional_bits) % 2) {
-                    shift -= 1;
-                }
+		// shift + fractional bits must be an even number
+		if ((shift + fractional_bits) % 2) {
+			shift -= 1;
+		}
 
 		// We can't use the safe shift since the shift value is unknown at compile time.
 		intermediate_type n = static_cast<intermediate_type>(this->raw_value) << shift;
@@ -484,8 +484,7 @@ public:
 			guess = (guess + n / guess) / 2;
 			if (guess == prev) {
 				break;
-
-                        }
+			}
 		}
 
 		// The sqrt operation halves the number of bits, so we'll we'll have to calculate a shift back
--- a/libopenage/util/fixed_point.h
+++ b/libopenage/util/fixed_point.h
@@ -460,11 +460,11 @@ public:
 	constexpr FixedPoint sqrt() {
 		// Zero can cause issues later, so deal with now.
 		if (this->raw_value == 0) {
-			return 0.0;
+			return zero();
 		}
 
-                // Check for negative values
-                ENSURE(std::is_unsigned<raw_type>() or std::is_signed<raw_type>() and this->raw_value > 0, "FixedPoint::sqrt() is undefined for negative values.");
+		// Check for negative values
+		ENSURE(std::is_unsigned<raw_type>() or std::is_signed<raw_type>() and this->raw_value > 0, "FixedPoint::sqrt() is undefined for negative values.");
 
 		// A greater shift = more precision, but can overflow the intermediate type if too large.
 		size_t max_shift = std::countl_zero(static_cast<unsigned_intermediate_type>(this->raw_value)) - 1;
--- a/libopenage/util/fixed_point.h
+++ b/libopenage/util/fixed_point.h
@@ -464,7 +464,9 @@ public:
 		}
 
 		// Check for negative values
-		ENSURE(std::is_unsigned<raw_type>() or (std::is_signed<raw_type>() and this->raw_value > 0), "FixedPoint::sqrt() is undefined for negative values.");
+		if constexpr (std::is_signed<raw_type>()) {
+			ENSURE(this->raw_value > 0, "FixedPoint::sqrt() is undefined for negative values.");
+		}
 
 		// A greater shift = more precision, but can overflow the intermediate type if too large.
 		size_t max_shift = std::countl_zero(static_cast<unsigned_intermediate_type>(this->raw_value)) - 1;
--- a/libopenage/util/fixed_point.h
+++ b/libopenage/util/fixed_point.h
@@ -451,33 +451,47 @@ public:
 	 * Pure FixedPoint sqrt implementation using Heron's Algorithm.
 	 *
 	 * Note that this function is undefined for negative values.
+         *
+         * There's a small loss in precision depending on the value of fractional_bits and the position of
+         * the most significant bit: if the integer portion is very large, we won't have as much (absolute)
+         * precision. Ideally you would want the intermediate_type to be twice the size of raw_type to avoid
+         * any losses.
 	 */
-	constexpr double sqrt() {
+	constexpr FixedPoint sqrt() {
 		// Zero can cause issues later, so deal with now.
 		if (this->raw_value == 0) {
 			return 0.0;
 		}
 
+                // Check for negative values
+                ENSURE(std::is_unsigned<raw_type>() or std::is_signed<raw_type>() and this->raw_value > 0, "FixedPoint::sqrt() is undefined for negative values.");
+
 		// A greater shift = more precision, but can overflow the intermediate type if too large.
-		size_t max_shift = std::countl_zero((unsigned_intermediate_type)this->raw_value) - 1;
+		size_t max_shift = std::countl_zero(static_cast<unsigned_intermediate_type>(this->raw_value)) - 1;
 		size_t shift = max_shift > fractional_bits ? fractional_bits : max_shift;
-		shift &= ~1;
+
+                // shift + fractional bits must be an even number
+                if ((shift + fractional_bits) % 2) {
+                    shift -= 1;
+                }
 
 		// We can't use the safe shift since the shift value is unknown at compile time.
-		intermediate_type n = (intermediate_type)this->raw_value << shift;
-		intermediate_type guess = (intermediate_type)1 << fractional_bits;
+		intermediate_type n = static_cast<intermediate_type>(this->raw_value) << shift;
+		intermediate_type guess = static_cast<intermediate_type>(1) << fractional_bits;
 
-		for (size_t _i = 0; _i < fractional_bits; _i++) {
+		for (size_t i = 0; i < fractional_bits; i++) {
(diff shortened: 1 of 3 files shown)


libopenage/util/fixed_point.h
```diff
@@ -446,8 +447,37 @@ class FixedPoint {
 		return is;
 	}
 
+	/**
+	 * Pure FixedPoint sqrt implementation using Heron's Algorithm.
+	 *
+	 * Note that this function is undefined for negative values.
+	 */
 	constexpr double sqrt() {
```
heinezen, line 455: It's probably better that we make the return type `FixedPoint` by default. If we want a floating point sqrt, there's already the std::sqrt overload in line 606.

libopenage/util/fixed_point.h
```diff
@@ -446,8 +447,37 @@ class FixedPoint {
 		return is;
 	}
 
+	/**
+	 * Pure FixedPoint sqrt implementation using Heron's Algorithm.
+	 *
+	 * Note that this function is undefined for negative values.
```
heinezen, line 453: Can you add an assert for this case? There is an `ENSURE` macro defined by openage that you can use for asserts.

heinezen, line 453: Would also be nice if you add the other limitation from your PR description to the docstring:

> There's a small loss in precision depending on the value of fractional_bits and the position of the most significant bit: if the integer portion is very large, we won't have as much (absolute) precision. Ideally you would want the intermediate_type to be twice the size of raw_type to avoid any losses.

libopenage/util/fixed_point.h
```diff
@@ -446,8 +447,37 @@ class FixedPoint {
...
+	 */
 	constexpr double sqrt() {
-		return std::sqrt(this->to_double());
+		// Zero can cause issues later, so deal with now.
+		if (this->raw_value == 0) {
+			return 0.0;
+		}
+
+		// A greater shift = more precision, but can overflow the intermediate type if too large.
+		size_t max_shift = std::countl_zero((unsigned_intermediate_type)this->raw_value) - 1;
```
heinezen, line 462: use `static_cast` instead of C casts here to convert to a different type.

```suggestion
		size_t max_shift = std::countl_zero(static_cast<unsigned_intermediate_type>(this->raw_value)) - 1;
```

libopenage/util/fixed_point.h
```diff
@@ -446,8 +447,37 @@ class FixedPoint {
...
+		size_t shift = max_shift > fractional_bits ? fractional_bits : max_shift;
+		shift &= ~1;
```
heinezen, line 464: You should AND with either a hex or binary value for better readability (and also ensure that the types match).

libopenage/util/fixed_point.h
```diff
@@ -446,8 +447,37 @@ class FixedPoint {
...
+
+		// We can't use the safe shift since the shift value is unknown at compile time.
+		intermediate_type n = (intermediate_type)this->raw_value << shift;
+		intermediate_type guess = (intermediate_type)1 << fractional_bits;
```
heinezen, line 468: same here for casting: use `static_cast`

libopenage/util/fixed_point.h
```diff
@@ -446,8 +447,37 @@ class FixedPoint {
...
+
+		for (size_t _i = 0; _i < fractional_bits; _i++) {
+			intermediate_type prev = guess;
+			guess = (guess + n / guess) / 2;
+			if (guess == prev)
+				break;
+		}
```
heinezen, line 475: The code is fine, I just recommend adjusting to our code style. i.e. remove the underscores from variables and always use brackets for if-clauses.

```suggestion
		for (size_t i = 0; i < fractional_bits; i++) {
			intermediate_type prev = guess;
			guess = (guess + n / guess) / 2;
			if (guess == prev) {
				break;
			}
		}
```

libopenage/util/fixed_point_test.cpp
```diff
@@ -156,6 +156,44 @@ void fixed_point() {
...
+			TESTEQUALS_FLOAT(T(value * value).sqrt(), value, 1e-7);
+		}
+
+		for (size_t i = 0; i < 100; i++) {
+			double value = 14.25 * i;
+			TESTEQUALS_FLOAT(T(value * value).sqrt(), value, 1e-7);
+		}
+
+		// This one can go up to 2^63, but that would take years.
+		for (uint32_t i = 0; i < (1u << 16); i++) {
```
heinezen, line 180: I suggest writing the actual number here instead of the shift. Makes it a bit more readable.

```suggestion
		for (uint32_t i = 0; i < 65536; i++) {
```

libopenage/util/fixed_point_test.cpp
```diff
@@ -156,6 +156,44 @@ void fixed_point() {
...
+			T value = T::from_raw_value(i * i);
+			TESTEQUALS_FLOAT(value.sqrt(), std::sqrt(value.to_double()), 1e-7);
+		}
+
+		// We lose some precision when raw_type == intermediate_type
+		for (uint64_t i = 1; i < (1ul << 63); i = (i << 1) ^ i) {
```
heinezen, line 186: Same here, I think the `std::numeric_limits` definitions are more readable. Also, shifting by one is just multiplication by 2 and the compiler can probably optimize that out on its own :D

```suggestion
		for (uint64_t i = 1; i < std::numeric_limits<uint64_t>::max(); i = (i * 2) ^ i) {
```

libopenage/util/fixed_point_test.cpp
```diff
@@ -156,6 +156,44 @@ void fixed_point() {
...
+			T value = T::from_raw_value(i * i);
+			TESTEQUALS_FLOAT(value.sqrt(), std::sqrt(value.to_double()), 1e-4);
+		}
+
+		using FP16_16 = FixedPoint<uint32_t, 16, uint64_t>;
+		for (uint32_t i = 0; i < (1u << 16); i++) {
```
heinezen, line 192: Same here

```suggestion
		for (uint32_t i = 0; i < 65536; i++) {
```

heinezen (changes requested):
Very nice, thanks a lot :)

I have a few comments about the code, but nothing too serious.

libopenage/util/fixed_point.h
bytegrrrl: This was to ensure the shift was an even value, but it turns out shift + fractional_bits needs to be even. I've updated this to reflect that - hopefully it's more readable now

libopenage/util/fixed_point.h
bytegrrrl: Sorry - I've been working with rust recently, and I must have accidentally picked this up!

libopenage/util/fixed_point_test.cpp
bytegrrrl: Fixed!

bytegrrrl:
Thank you for all the feedback! I think I've incorporated everything. If this all looks good, I'd be happy to work on some of the other math functions too. (Hopefully I'll be able to break my C habits!)

---

I ran into an interesting bug / edge case with`TESTEQUALS_FLOAT` by the way:

An unsigned fixed point 0.0 and a double 0.0 fail the equality check.
So, part of the macro tests for inequality with `lhs < (rhs - epsilon)`.  The result of `rhs - epsilon` is `static_cast` back into its raw type for the actual comparison. Now, if `rhs - epsilon` is negative and we cast into an unsigned type, then this behavior is undefined. On x86, the value underflows so the inequality condition is met. On arm, the value is set to 0. (Which was why I didn't catch the failing test). 

Here's an example case that fails on x86:
```
TESTEQUALS_FLOAT((FixedPoint<uint32_t, 16>::zero()), 0.0, 1e-4);
```
I think it should be possible to resolve this by checking for `lhs + epsilon < rhs` instead. But floats are cursed and aren't associative, so this could cause unexpected behavior down the line. I've chosen to just omit the 0 case for now.

libopenage/util/fixed_point.h
```diff
@@ -451,33 +451,47 @@ class FixedPoint {
...
+         * any losses.
 	 */
-	constexpr double sqrt() {
+	constexpr FixedPoint sqrt() {
 		// Zero can cause issues later, so deal with now.
 		if (this->raw_value == 0) {
 			return 0.0;
 		}
 
+                // Check for negative values
```
heinezen: Your whole indentation is off now :D I recommend you run over the file with clang-format to fix it and activate the autoformatter for whatever IDE you are using.

libopenage/util/fixed_point.h
```diff
@@ -451,33 +451,47 @@ class FixedPoint {
...
+                ENSURE(std::is_unsigned<raw_type>() or std::is_signed<raw_type>() and this->raw_value > 0, "FixedPoint::sqrt() is undefined for negative values.");
```
heinezen: parenthesis around the second condition would be nice for readability.

libopenage/util/fixed_point_test.cpp
```diff
@@ -177,23 +177,31 @@ void fixed_point() {
...
 		// This one can go up to 2^63, but that would take years.
-		for (uint32_t i = 0; i < (1u << 16); i++) {
+		for (uint32_t i = 0; i < 65536; i++) {
 			T value = T::from_raw_value(i * i);
 			TESTEQUALS_FLOAT(value.sqrt(), std::sqrt(value.to_double()), 1e-7);
 		}
 
 		// We lose some precision when raw_type == intermediate_type
-		for (uint64_t i = 1; i < (1ul << 63); i = (i << 1) ^ i) {
+		for (uint64_t i = 1; i < std::numeric_limits<uint64_t>::max(); i = (i * 2) ^ i) {
```
heinezen, line 186: There's a risk that this loop goes on forever because it only stops if `i == std::numeric_limits<uint64_t>::max()`. Maybe we should try a lower limit.

heinezen: tests run through though ^^

heinezen (changes requested):
I found a few new issues

libopenage/util/fixed_point.h
bytegrrrl: Whoops, sorry! I had fixed the broken test case on a separate machine (one without clang-format set up properly)

libopenage/util/fixed_point.h
```diff
@@ -604,7 +618,7 @@ namespace std {
 
 template <typename I, unsigned F, typename Inter>
 constexpr double sqrt(openage::util::FixedPoint<I, F, Inter> n) {
-	return n.sqrt();
+	return static_cast<double>(n.sqrt());
```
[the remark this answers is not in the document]
bytegrrrl: The `constexpr explicit operator double()` on [line 343](https://github.com/SFTtech/openage/blob/1e6d3b92b699d9d0260cd733dd9a5ef46f902124/libopenage/util/fixed_point.h#L343) ought to handle this for us. I can change this to use `to_double()` directly if you'd prefer though.

libopenage/util/fixed_point_test.cpp
bytegrrrl: This generates the binary sequence `0b1`, `0b11`, `0b111`, `0b1111`, ...  so we should get to the max value after 63 iterations. I can add this as a comment, or I can generate the sequence in a different way if that would be better?

bytegrrrl: This generates a sequence that ends with the maximum value set.  (I had originally intended a slightly different sequence, but this one is neat). This is the sequence, (including the final value of `i`):
```
0000000000000000000000000000000000000000000000000000000000000001
0000000000000000000000000000000000000000000000000000000000000011
0000000000000000000000000000000000000000000000000000000000000101
0000000000000000000000000000000000000000000000000000000000001111
0000000000000000000000000000000000000000000000000000000000010001
0000000000000000000000000000000000000000000000000000000000110011
0000000000000000000000000000000000000000000000000000000001010101
0000000000000000000000000000000000000000000000000000000011111111
0000000000000000000000000000000000000000000000000000000100000001
0000000000000000000000000000000000000000000000000000001100000011
0000000000000000000000000000000000000000000000000000010100000101
0000000000000000000000000000000000000000000000000000111100001111
0000000000000000000000000000000000000000000000000001000100010001
0000000000000000000000000000000000000000000000000011001100110011
0000000000000000000000000000000000000000000000000101010101010101
0000000000000000000000000000000000000000000000001111111111111111
0000000000000000000000000000000000000000000000010000000000000001
0000000000000000000000000000000000000000000000110000000000000011
0000000000000000000000000000000000000000000001010000000000000101
0000000000000000000000000000000000000000000011110000000000001111
0000000000000000000000000000000000000000000100010000000000010001
0000000000000000000000000000000000000000001100110000000000110011
0000000000000000000000000000000000000000010101010000000001010101
0000000000000000000000000000000000000000111111110000000011111111
0000000000000000000000000000000000000001000000010000000100000001
0000000000000000000000000000000000000011000000110000001100000011
0000000000000000000000000000000000000101000001010000010100000101
0000000000000000000000000000000000001111000011110000111100001111
0000000000000000000000000000000000010001000100010001000100010001
0000000000000000000000000000000000110011001100110011001100110011
0000000000000000000000000000000001010101010101010101010101010101
0000000000000000000000000000000011111111111111111111111111111111
0000000000000000000000000000000100000000000000000000000000000001
0000000000000000000000000000001100000000000000000000000000000011
0000000000000000000000000000010100000000000000000000000000000101
0000000000000000000000000000111100000000000000000000000000001111
0000000000000000000000000001000100000000000000000000000000010001
0000000000000000000000000011001100000000000000000000000000110011
0000000000000000000000000101010100000000000000000000000001010101
0000000000000000000000001111111100000000000000000000000011111111
0000000000000000000000010000000100000000000000000000000100000001
0000000000000000000000110000001100000000000000000000001100000011
0000000000000000000001010000010100000000000000000000010100000101
0000000000000000000011110000111100000000000000000000111100001111
0000000000000000000100010001000100000000000000000001000100010001
0000000000000000001100110011001100000000000000000011001100110011
0000000000000000010101010101010100000000000000000101010101010101
0000000000000000111111111111111100000000000000001111111111111111
0000000000000001000000000000000100000000000000010000000000000001
0000000000000011000000000000001100000000000000110000000000000011
0000000000000101000000000000010100000000000001010000000000000101
0000000000001111000000000000111100000000000011110000000000001111
0000000000010001000000000001000100000000000100010000000000010001
0000000000110011000000000011001100000000001100110000000000110011
0000000001010101000000000101010100000000010101010000000001010101
0000000011111111000000001111111100000000111111110000000011111111
0000000100000001000000010000000100000001000000010000000100000001
0000001100000011000000110000001100000011000000110000001100000011
0000010100000101000001010000010100000101000001010000010100000101
0000111100001111000011110000111100001111000011110000111100001111
0001000100010001000100010001000100010001000100010001000100010001
0011001100110011001100110011001100110011001100110011001100110011
0101010101010101010101010101010101010101010101010101010101010101
1111111111111111111111111111111111111111111111111111111111111111
```

libopenage/util/fixed_point.h
bytegrrrl: Fixed!

libopenage/util/fixed_point.h
```diff
@@ -464,16 +464,16 @@ class FixedPoint {
 		}
 
 		// Check for negative values
-		ENSURE(std::is_unsigned<raw_type>() or std::is_signed<raw_type>() and this->raw_value > 0, "FixedPoint::sqrt() is undefined for negative values.");
+		ENSURE(std::is_unsigned<raw_type>() or (std::is_signed<raw_type>() and this->raw_value > 0), "FixedPoint::sqrt() is undefined for negative values.");
```
heinezen, line 467: I think we could also simplify this with a `constexpr` if-clause to check the type at compile time. Then the ENSURE macro is only checked if `raw_type` is signed. Saves us a few calls in debug mode too :)

```suggestion
		if constexpr (std::is_signed<raw_type>()) {
			ENSURE(this->raw_value > 0, "FixedPoint::sqrt() is undefined for negative values.");
		}
```

heinezen (commented):
I have a very small suggestion but everything else looks nice.

libopenage/util/fixed_point.h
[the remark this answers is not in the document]
heinezen: oh no I forgot about this :D It's probably time to remove the implicit conversion, so that we don't accidentally use floating points somewhere. That's something for a follow-up PR though.

libopenage/util/fixed_point_test.cpp
heinezen: Ah that makes sense. then let's leave it in :)

libopenage/util/fixed_point.h
bytegrrrl: Great idea, thanks!

heinezen (approved):

heinezen:
Thanks  again!

bytegrrrl:
Thank you for all the guidance! I might take a look at some of the others soon.

PR #1773, opened 2025-04-24
Add pure FixedPoint implementations and tests for sin(), cos(), and tan()

bytegrrrl:
Per #1543.

Notes:
This uses a power series approximation to calculate sine and cosine. Tangent is just calculated using `sin()` and `cos()`. The input value is first shifted into the interval `(-pi, pi)` for `sin()` and `cos()`and `(-pi/2, pi/2)` for `tan()`. At the asymptotes for tan, an exception will be thrown because the FixedPoint doesn't implement `inf` (yet). Near the asymptotes absolute precision is a bit poor.

It's possible to use another expansion for `tan()`, but it's not as nice as those for `sin()` and `cos()`. I'd be happy to implement that down the road if that would be preferred. (It will still have problems at the asymptotes though).

### DIFF
--- a/libopenage/util/fixed_point.h
+++ b/libopenage/util/fixed_point.h
@@ -176,6 +176,13 @@ public:
 	}
 
 	/**
+	 * FixedPoint value that is preinitialized to one.
+	 */
+	static constexpr FixedPoint one() {
+		return FixedPoint::from_int(1);
+	}
+
+	/**
 	 * Math constants represented in FixedPoint
 	 */
 	// naming, definition and value are kept compatible with `math_constants.h`
@@ -366,6 +373,26 @@ public:
 			static_cast<FixedPoint::unsigned_int_type>(this->raw_value) & std::integral_constant<int_type, FixedPoint::fractional_part_bitmask()>::value);
 	}
 
+	/**
+	 * Converter to retrieve the integral (pre-decimal) part of the number.
+	 */
+	constexpr FixedPoint get_integral_part() const {
+		// similar to the get_fractional_part() implementation, but the opposite bits.
+		return FixedPoint::from_raw_value(this->raw_value & ~std::integral_constant<int_type, FixedPoint::fractional_part_bitmask()>::value);
+	}
+
+	/**
+	 * Round to the nearest integer.
+	 */
+	constexpr FixedPoint round() const {
+		if (this->get_fractional_part() < same_type_but_unsigned(0.5)) {
+			return this->get_integral_part();
+		}
+		else {
+			return this->get_integral_part() + one();
+		}
+	}
+
 	// Comparison operators for comparison with other
 	constexpr auto operator<=>(const FixedPoint &o) const = default;
 
@@ -531,7 +558,7 @@ public:
 		// Ensure we're in the interval (-pi, pi)
 		// Since this is a series expansion approximation around 0, this interval is where we are most accurate
 		if (x > FixedPoint::pi()) {
-			size_t n = round((x / FixedPoint::tau()).to_double());
+			int_type n = (x / FixedPoint::tau()).round().to_int();
 			x -= FixedPoint::tau() * n;
 		}
 
@@ -568,7 +595,7 @@ public:
 		// Ensure we're in the interval (-pi, pi)
 		// Since this is a series expansion approximation around 0, this interval is where we are most accurate
 		if (x > FixedPoint::pi()) {
-			size_t n = round((x / FixedPoint::tau()).to_double());
+			int_type n = (x / FixedPoint::tau()).round().to_int();
 			x -= FixedPoint::tau() * n;
 		}
 
@@ -605,7 +632,7 @@ public:
 
 		// Ensure we are in the interval (-pi/2, pi/2) for maximum accuracy
 		if (x > FixedPoint::pi_2()) {
-			size_t n = round((x / FixedPoint::pi()).to_double());
+			int_type n = (x / FixedPoint::pi()).round().to_int();
 			x -= FixedPoint::pi() * n;
 		}
 
--- a/libopenage/util/fixed_point.h
+++ b/libopenage/util/fixed_point.h
@@ -414,6 +414,18 @@ public:
 	}
 
 	/**
+	 * FixedPoint *= FixedPoint
+	 *
+	 * This shares the same caveats as FixedPint * FixedPoint.
+	 *
+	 * Use a larger intermediate type to avoid overflow.
+	 */
+	constexpr FixedPoint operator*=(const FixedPoint &rhs) {
+		*this = *this *rhs;
+		return *this;
+	}
+
+	/**
 	 * FixedPoint /= N
 	 */
 	template <typename N>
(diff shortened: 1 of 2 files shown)


libopenage/util/fixed_point.h
```diff
@@ -499,16 +511,111 @@ class FixedPoint {
...
+		// Sine is an odd function, so we can pull out the sign.
+		bool negative = x < 0;
+		if (negative) {
+			x = -x;
+		}
+
+		// Ensure we're in the interval (-pi, pi)
+		// Since this is a series expansion approximation around 0, this interval is where we are most accurate
+		if (x > FixedPoint::pi()) {
+			size_t n = round((x / FixedPoint::tau()).to_double());
```
heinezen, line 534: This uses floating point math which we want to avoid. I think we need a native fixed-point rounding function.

libopenage/util/fixed_point.h
```diff
@@ -499,16 +511,111 @@ class FixedPoint {
...
-		return std::sin(this->to_double());
+	/**
+	 * Calculate pure FixedPoint sine using a power series approximation.
+	 *
+	 * This may lose absolute precision when the fractional part is small. Because trig functions are
+	 * cyclic, we don't lose precision for large integer values. For best results, you would want
+	 * intermediate_size to be twice the size of raw_type.
+	 */
+	constexpr FixedPoint sin() {
+		size_t order = 10;
```
heinezen, line 522: If this is a constant, it should probably be defined outside of the function.

libopenage/util/fixed_point_test.cpp
```diff
@@ -203,6 +203,73 @@ void fixed_point() {
 		TESTNOEXCEPT((FixedPoint<int64_t, 32>::from_float(3.25).sqrt()));
 		TESTNOEXCEPT((FixedPoint<uint64_t, 32>::from_float(-3.25).sqrt()));
 	}
+
+	// Pure FixedPoint trig tests
+	{
+		using TrigType = FixedPoint<int64_t, 32, __int128>;
```
heinezen, line 209: `__int128` might not be supported on all architectures or compilers. Maybe we should use a smaller type here.

heinezen (changes requested):
Very nice, I only have a few requests

libopenage/util/fixed_point.h
bytegrrrl: I actually wanted to ask about this. It might be helpful to have it as a tunable parameter since not all applications require that much accuracy. But I'm not sure how best to implement that. I had been considering as an optional function argument, but that doesn't really make sense semantically.

libopenage/util/fixed_point.h
bytegrrrl: Ah, good catch!

libopenage/util/fixed_point_test.cpp
bytegrrrl: We do lose some precision with a smaller type, but let me see if I can figure out a workaround.

bytegrrrl: So I've been able to write an optimized multiplication function that doesn't lose precision, but I've run into another issue: the shift in the division algorithm erases the integral part of any FixedPoint where the fractional part is half the number of bits.

Here: 
```
Inter ret = div((static_cast<Inter>(lhs.get_raw_value()) << F), static_cast<Inter>(rhs.get_raw_value()));
```

libopenage/util/fixed_point.h
bytegrrrl: Okay, so I've implemented a `round()` function for FixedPoint and have changed the implementation to use that. 

`round()` uses a helper function to get the integral portion of the number (essentially the dual of the `get_fractional_part()`).

libopenage/util/fixed_point.h
bytegrrrl: I think I've found a solution. I'm using the value of `this->approx_decimal_places`, which should translate nicely to different sizes of FixedPoints. Let me know what you think?

libopenage/util/fixed_point_test.cpp
bytegrrrl: So I think I may have gone a bit overboard, but I've changed the `__int128` back to an `int64_t`. To preserve the precision, I've implemented multiplication and division algorithms that don't overflow while calculating.

The multiplication multiplies the upper and lower bits independently then sums them together, which means the uppermost bits won't be lost. 

The division just implements bitwise long division to avoid the shift that lops off the integral bits of the number. There's almost certainly a more efficient solution here, but this works pretty well for the time being. 

It'll always be faster to use a larger datatype though. So maybe it's worth doing a check for the existence of __int128 during compilation?

bytegrrrl: I added a better division algorithm. (Basically another power series expansion). Should be somewhat faster now. It'll still be best to try to use an int128 if possible though.

libopenage/util/fixed_point.h
heinezen: If you want to make it configurable just for the function, you can define an additional template parameter. You can even default initialize it to some value. That would look something like this:

```cpp
template <size_t ORDER = 10>
constexpr FixedPoint sin() {
    ...
    for (size_t i = 0; i < ORDER; i++) {
        ...
    }
    ...
};

File: libopenage/util/fixed_point_test.cpp [hunk 4]
```diff
@@ -127,7 +127,8 @@ void fixed_point() {
 		using T = FixedPoint<uint16_t, 7U, uint64_t>;
 
 		auto a = S::from_int(16U);
-		TESTNOTEQUALS((a*a).to_int(), 256U);
+		// This test case is now equal with the improved multiplication algorithm
```
heinezen, line 130: You should probably remove the comment since no one will know what "improved multiplication algorithm" refers to in a few months :D

libopenage/util/fixed_point_test.cpp
```diff
@@ -139,7 +140,8 @@ void fixed_point() {
 		using S = FixedPoint<int32_t, 12U>;
 		auto a = S::from_int(256);
 		auto b = S::from_int(8);
-		TESTNOTEQUALS((a/b).to_int(), 32);
+		// This test case is now equal with the improved division algorithm
```
heinezen, line 143: same here

libopenage/util/fixed_point_test.cpp
```diff
@@ -203,6 +205,74 @@ void fixed_point() {
 		TESTNOEXCEPT((FixedPoint<int64_t, 32>::from_float(3.25).sqrt()));
 		TESTNOEXCEPT((FixedPoint<uint64_t, 32>::from_float(-3.25).sqrt()));
 	}
+
+	// Pure FixedPoint trig tests
+	{
+		using TrigType = FixedPoint<int64_t, 32, int64_t>;
+
+		// Testing sin() and cos()
+		for (int i = -100'000; i <= 100'000; i++) {
```
heinezen, line 214: the loop runs for a very long time. Can we go with lower precision for speeding these up a bit?

heinezen (changes requested):
Very nice changes with the multiplication/division

libopenage/util/fixed_point.h
```diff
@@ -366,6 +373,26 @@ class FixedPoint {
 			static_cast<FixedPoint::unsigned_int_type>(this->raw_value) & std::integral_constant<int_type, FixedPoint::fractional_part_bitmask()>::value);
 	}
 
+	/**
+	 * Converter to retrieve the integral (pre-decimal) part of the number.
+	 */
+	constexpr FixedPoint get_integral_part() const {
```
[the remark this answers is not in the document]
bytegrrrl: Yep! I'll try to get those added soon!

libopenage/util/fixed_point.h
```diff
@@ -366,6 +373,26 @@ class FixedPoint {
...
+		return FixedPoint::from_raw_value(this->raw_value & ~std::integral_constant<int_type, FixedPoint::fractional_part_bitmask()>::value);
+	}
+
+	/**
+	 * Round to the nearest integer.
+	 */
+	constexpr FixedPoint round() const {
+		if (this->get_fractional_part() < same_type_but_unsigned(0.5)) {
+			return this->get_integral_part();
+		}
```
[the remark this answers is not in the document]
bytegrrrl: Good call, I'll take a look soon!

libopenage/util/fixed_point.h
```diff
@@ -575,10 +709,56 @@ typename std::enable_if<std::is_arithmetic<N>::value, FixedPoint<I, F, Inter>>::
...
-	Inter ret = static_cast<Inter>(lhs.get_raw_value()) * static_cast<Inter>(rhs.get_raw_value());
-	ret >>= F;
+	using uInter = typename std::make_unsigned<Inter>::type;
+
+	// An optimization that can prevent overflows.
+	// This only helps overflows from the actual operations happening here, but not an ordinary overflow of int_type
+	// This is essentially just Karatsuba
+	if constexpr (sizeof(I) == sizeof(Inter)) {
+		constexpr int hwidth = sizeof(Inter) * 4;
+		uInter lower_mask = ~static_cast<uInter>(0) >> hwidth;
```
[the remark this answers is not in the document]
bytegrrrl: The `~` takes precedence over the `>>` I thought? But I do agree your first solution is cleaner - thanks! I'll add those tests soon!

libopenage/util/fixed_point.h
```diff
@@ -587,8 +767,53 @@ constexpr FixedPoint<I, F, Inter> operator*(const FixedPoint<I, F, Inter> lhs, c
...
+		Inter lhs_upper = lhs_lower >> F;
+		lhs_lower = lhs_lower & lower_mask;
+
+		bool r_pos = rhs > 0;
+		uInter rhs_lower = static_cast<uInter>(std::abs(rhs.get_raw_value()));
+		Inter rhs_upper = rhs_lower >> F;
+		rhs_lower = rhs_lower & lower_mask;
+
+		// special case when integeral part of rhs is 0
+		if (rhs_upper == 0) {
```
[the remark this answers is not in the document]
bytegrrrl: There is - just not explicitly. It took me ages to figure out why my prior tests weren't passing because of this 😅

I'll add an explicit set of tests that call this out though!

libopenage/util/fixed_point.h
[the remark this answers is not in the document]
heinezen: I tested it in the debugger and it did not take precendence (even though it should I agree). I think the cast introduced some weird behavior that the compiler couldn't resolve properly.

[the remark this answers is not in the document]
heinezen: e.g. for `uint16_t` the value of `lower_mask` was 65535 even though it should have been 255.

libopenage/util/fixed_point.h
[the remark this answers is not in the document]
heinezen: ah i see it now being used by `sin(..)` for example. An explicit test would be nice though :)

TheJJ:
I'll close this for now, when it gets activity again we can always reopen :)

PR #1777, opened 2025-05-18
Fix macos build and run issues

Skosulor:
The goal of this pull request is to fix build and run issues for macOs.

- Run presenter in the main thread (macos starts a hissy fit unless GUI related stuff is not run in the main thread)
- Remove guards stopping QT events to be processed on macos
- Fix shaders that for an unknown reason crashes when glLinkProgram is called. 

Note: Requires some additional tweeks

### DIFF
--- a/assets/shaders/identity.vert.glsl
+++ b/assets/shaders/identity.vert.glsl
@@ -1,16 +1,14 @@
-#version 120
+#version 330 core
 
-// no-transformation texture mapping vertex shader
+// Input vertex position
+layout(location = 0) in vec4 vertex_position;
 
-// the position of this vertex
-attribute vec4 vertex_position;
-
-// interpolated texture coordinates sent to fragment shader
-varying vec2 tex_position;
+// Output to fragment shader
+out vec2 tex_position;
 
 void main(void) {
-	gl_Position = vertex_position;
+    gl_Position = vertex_position;
 
-	// convert from screen coordinates (-1, 1) to texture coordinates (0, 1)
-	tex_position = (vertex_position.xy + vec2(1.f)) / 2.f;
+    // Convert from clip space (-1, 1) to texture space (0, 1)
+    tex_position = (vertex_position.xy + vec2(1.0)) * 0.5;
 }
--- a/assets/shaders/maptexture.frag.glsl
+++ b/assets/shaders/maptexture.frag.glsl
@@ -1,10 +1,10 @@
 #version 330 core
 
-uniform sampler2D tex_sampler;
+uniform sampler2D texture_;
 
 in vec2 tex_position;
 out vec4 frag_color;
 
 void main(void) {
-    frag_color = texture(tex_sampler, tex_position);
+    frag_color = texture(texture_, tex_position);
 }
--- a/assets/shaders/maptexture.frag.glsl
+++ b/assets/shaders/maptexture.frag.glsl
@@ -1,13 +1,10 @@
-#version 120
-// total basic standard texture drawing fragment shader
+#version 330 core
 
-// the texture data
-uniform sampler2D texture;
+uniform sampler2D tex_sampler;
 
-// interpolated texture coordinates received from vertex shader
-varying vec2 tex_position;
+in vec2 tex_position;
+out vec4 frag_color;
 
-void main (void) {
-	// this sets the fragment color to the corresponding texel.
-	gl_FragColor = texture2D(texture, tex_position);
+void main(void) {
+    frag_color = texture(tex_sampler, tex_position);
 }
--- a/assets/shaders/world2d.frag.glsl
+++ b/assets/shaders/world2d.frag.glsl
@@ -19,25 +19,18 @@ vec2 uv = vec2(
 void main() {
 	vec4 tex_val = texture(tex, uv);
 	int alpha = int(round(tex_val.a * 255));
-	switch (alpha) {
-		case 0:
-			col = tex_val;
-			discard;
-
-			// do not save the ID
-			return;
-		case 254:
-			col = vec4(1.0f, 0.0f, 0.0f, 1.0f);
-			break;
-		case 252:
-			col = vec4(0.0f, 1.0f, 0.0f, 1.0f);
-			break;
-		case 250:
-			col = vec4(0.0f, 0.0f, 1.0f, 1.0f);
-			break;
-		default:
-			col = tex_val;
-			break;
+	if (alpha == 0) {
+		col = tex_val;
+		discard;
+		return;
+	} else if (alpha == 254) {
+		col = vec4(1.0f, 0.0f, 0.0f, 1.0f);
+	} else if (alpha == 252) {
+		col = vec4(0.0f, 1.0f, 0.0f, 1.0f);
+	} else if (alpha == 250) {
+		col = vec4(0.0f, 0.0f, 1.0f, 1.0f);
+	} else {
+		col = tex_val;
 	}
 	id = u_id;
 }
--- a/assets/test/shaders/demo_2_obj.frag.glsl
+++ b/assets/test/shaders/demo_2_obj.frag.glsl
@@ -10,22 +10,18 @@ layout(location=1) out uint id;
 void main() {
 	vec4 tex_val = texture(tex, v_uv);
 	int alpha = int(round(tex_val.a * 255));
-	switch (alpha) {
-		case 0:
-		discard;
-		break;
-		case 254:
-		col = vec4(1.0f, 0.0f, 0.0f, 1.0f);
-		break;
-		case 252:
-		col = vec4(0.0f, 1.0f, 0.0f, 1.0f);
-		break;
-		case 250:
-		col = vec4(0.0f, 0.0f, 1.0f, 1.0f);
-		break;
-		default:
+
+	if (alpha == 0) {
+	    discard;
+	} else if (alpha == 254) {
+	    col = vec4(1.0, 0.0, 0.0, 1.0);
+	} else if (alpha == 252) {
+	    col = vec4(0.0, 1.0, 0.0, 1.0);
+	} else if (alpha == 250) {
+	    col = vec4(0.0, 0.0, 1.0, 1.0);
+	} else {
 	    col = tex_val;
-		break;
 	}
+
 	id = u_id;
 }
--- a/assets/test/shaders/demo_4_obj.frag.glsl
+++ b/assets/test/shaders/demo_4_obj.frag.glsl
@@ -10,22 +10,18 @@ layout(location=1) out uint id;
 void main() {
 	vec4 tex_val = texture(tex, v_uv);
 	int alpha = int(round(tex_val.a * 255));
-	switch (alpha) {
-		case 0:
-		discard;
-		break;
-		case 254:
-		col = vec4(1.0f, 0.0f, 0.0f, 1.0f);
-		break;
-		case 252:
-		col = vec4(0.0f, 1.0f, 0.0f, 1.0f);
-		break;
-		case 250:
-		col = vec4(0.0f, 0.0f, 1.0f, 1.0f);
-		break;
-		default:
+
+	if (alpha == 0) {
+	    discard;
+	} else if (alpha == 254) {
+	    col = vec4(1.0, 0.0, 0.0, 1.0);
+	} else if (alpha == 252) {
+	    col = vec4(0.0, 1.0, 0.0, 1.0);
+	} else if (alpha == 250) {
+	    col = vec4(0.0, 0.0, 1.0, 1.0);
+	} else {
 	    col = tex_val;
-		break;
 	}
+
 	id = u_id;
 }
--- a/libopenage/engine/engine.cpp
+++ b/libopenage/engine/engine.cpp
@@ -20,6 +20,7 @@ Engine::Engine(mode mode,
 	running{true},
 	run_mode{mode},
 	root_dir{root_dir},
+	window_settings{window_settings},
 	threads{} {
 	log::log(INFO
 	         << "launching engine with root directory"
@@ -53,30 +54,39 @@ Engine::Engine(mode mode,
 		this->time_loop.reset();
 	});
 
-	// if presenter is used, run it in a separate thread
-	if (this->run_mode == mode::FULL) {
-		this->threads.emplace_back([&]() {
-			this->presenter->run(window_settings);
-
-			// Make sure that the presenter gets destructed in the same thread
-			// otherwise OpenGL complains about missing contexts
-			this->presenter.reset();
-			this->running = false;
-		});
-	}
-
 	log::log(INFO << "Using " << this->threads.size() + 1 << " threads "
 	              << "(" << std::jthread::hardware_concurrency() << " available)");
 }
 
 void Engine::loop() {
-	// Run the main game simulation loop:
-	this->simulation->run();
 
-	// After stopping, clean up the simulation
-	this->simulation.reset();
-	if (this->run_mode != mode::FULL) {
+	// if presenter is used, run the simulation in a separate thread
+	if (this->run_mode == mode::FULL) {
+
+		this->threads.emplace_back([&]() {
+			// Run the main game simulation loop:
+			this->simulation->run();
+			// After stopping, clean up the simulation
+			this->simulation.reset();
+			if (this->run_mode != mode::FULL) {
+				this->running = false;
+			}
+		});
+
+		this->presenter->run(this->window_settings);
+		// Make sure that the presenter gets destructed in the same thread
+		// otherwise OpenGL complains about missing contexts
+		this->presenter.reset();
 		this->running = false;
+	}
+	else {
+		// Run the main game simulation loop:
+		this->simulation->run();
+		// After stopping, clean up the simulation
+		this->simulation.reset();
+		if (this->run_mode != mode::FULL) {
+			this->running = false;
+		}
 	}
 }
 
--- a/libopenage/engine/engine.cpp
+++ b/libopenage/engine/engine.cpp
@@ -1,4 +1,4 @@
-// Copyright 2023-2024 the openage authors. See copying.md for legal info.
+// Copyright 2023-2025 the openage authors. See copying.md for legal info.
 
 #include "engine.h"
 
--- a/libopenage/engine/engine.cpp
+++ b/libopenage/engine/engine.cpp
@@ -64,30 +64,30 @@ void Engine::loop() {
 	if (this->run_mode == mode::FULL) {
 
 		this->threads.emplace_back([&]() {
-			// Run the main game simulation loop:
-			this->simulation->run();
-			// After stopping, clean up the simulation
-			this->simulation.reset();
-			if (this->run_mode != mode::FULL) {
-				this->running = false;
-			}
+			this->loop_simulation();
 		});
 
 		this->presenter->run(this->window_settings);
+		log::log(MSG(info) << "presenter exited");
 		// Make sure that the presenter gets destructed in the same thread
 		// otherwise OpenGL complains about missing contexts
 		this->presenter.reset();
 		this->running = false;
 	}
 	else {
-		// Run the main game simulation loop:
-		this->simulation->run();
-		// After stopping, clean up the simulation
-		this->simulation.reset();
-		if (this->run_mode != mode::FULL) {
-			this->running = false;
-		}
+		this->loop_simulation();
 	}
+}
+
+void Engine::loop_simulation() {
+	// Run the main game simulation loop:
+	this->simulation->run();
+	// After stopping, clean up the simulation
+	this->simulation.reset();
+	if (this->run_mode != mode::FULL) {
+		this->running = false;
+	}
+
 }
 
 } // namespace openage::engine
--- a/libopenage/engine/engine.h
+++ b/libopenage/engine/engine.h
@@ -100,6 +100,12 @@ public:
 	bool running;
 
 private:
+
+	/**
+	 * Run the simulation loop
+	 */
+	void loop_simulation();
+
 	/**
 	 * The run mode to use.
 	 */
--- a/libopenage/engine/engine.h
+++ b/libopenage/engine/engine.h
@@ -112,7 +112,7 @@ private:
 
 	/**
 	 * Window settings
-	 */ 
+	 */
 	const renderer::window_settings window_settings;
 
 	/**
--- a/libopenage/engine/engine.h
+++ b/libopenage/engine/engine.h
@@ -1,4 +1,4 @@
-// Copyright 2023-2024 the openage authors. See copying.md for legal info.
+// Copyright 2023-2025 the openage authors. See copying.md for legal info.
 
 #pragma once
 
--- a/libopenage/engine/engine.h
+++ b/libopenage/engine/engine.h
@@ -111,6 +111,11 @@ private:
 	util::Path root_dir;
 
 	/**
+	 * Window settings
+	 */ 
+	const renderer::window_settings window_settings;
+
+	/**
 	 * The threads used by the engine.
 	 */
 	std::vector<std::jthread> threads;
--- a/libopenage/renderer/gui/gui.cpp
+++ b/libopenage/renderer/gui/gui.cpp
@@ -119,7 +119,7 @@ void GUI::resize(size_t width, size_t height) {
 		auto fbo = renderer->create_texture_target({output_texture});
 
 		// pass the new texture to shader uniform
-		this->texture_unif->update("texture", this->texture);
+		this->texture_unif->update("texture_", this->texture);
 
 		this->render_pass->set_target(fbo);
 	}
--- a/libopenage/renderer/gui/gui.cpp
+++ b/libopenage/renderer/gui/gui.cpp
@@ -1,4 +1,4 @@
-// Copyright 2015-2024 the openage authors. See copying.md for legal info.
+// Copyright 2015-2025 the openage authors. See copying.md for legal info.
 
 #include "gui.h"
 
--- a/libopenage/renderer/gui/gui.cpp
+++ b/libopenage/renderer/gui/gui.cpp
@@ -87,7 +87,7 @@ void GUI::initialize_render_pass(size_t width,
 		resources::Texture2dInfo(width, height, resources::pixel_format::rgba8));
 	auto fbo = this->renderer->create_texture_target({output_texture});
 
-	this->texture_unif = maptex_shader->new_uniform_input("texture", this->texture);
+	this->texture_unif = maptex_shader->new_uniform_input("texture_", this->texture);
 	Renderable display_obj{
 		this->texture_unif,
 		quad,
--- a/libopenage/renderer/gui/gui.cpp
+++ b/libopenage/renderer/gui/gui.cpp
@@ -16,6 +16,7 @@
 #include "renderer/window.h"
 #include "util/path.h"
 
+static constexpr const char* texture_ = "texture_";
 
 namespace openage::renderer::gui {
 
@@ -87,7 +88,7 @@ void GUI::initialize_render_pass(size_t width,
 		resources::Texture2dInfo(width, height, resources::pixel_format::rgba8));
 	auto fbo = this->renderer->create_texture_target({output_texture});
 
-	this->texture_unif = maptex_shader->new_uniform_input("texture_", this->texture);
+	this->texture_unif = maptex_shader->new_uniform_input(texture_, this->texture);
 	Renderable display_obj{
 		this->texture_unif,
 		quad,
@@ -119,7 +120,7 @@ void GUI::resize(size_t width, size_t height) {
 		auto fbo = renderer->create_texture_target({output_texture});
 
 		// pass the new texture to shader uniform
-		this->texture_unif->update("texture_", this->texture);
+		this->texture_unif->update(texture_, this->texture);
 
 		this->render_pass->set_target(fbo);
 	}
--- a/assets/shaders/maptexture.vert.glsl
+++ b/assets/shaders/maptexture.vert.glsl
@@ -1,21 +1,16 @@
-//total basic standard texture mapping vertex shader
+#version 330 core
 
-//modelview*projection matrix
+// Input attributes (set via glVertexAttribPointer)
+layout(location = 0) in vec4 vertex_position;
+layout(location = 1) in vec2 tex_coordinates;
+
+// Output to fragment shader
+out vec2 tex_position;
+
+// Uniform matrix
 uniform mat4 mvp_matrix;
 
-//the position of this vertex
-attribute vec4 vertex_position;
-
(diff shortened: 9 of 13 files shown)


Skosulor:
Changing the structure of engine.loop have resulted in some new issues that need to be fixed, error output:

```
INFO [T2] Loading .nyan file: swgb_base/data/game_entity/generic/frigate/frigate.nyan
INFO [T2] Loading .nyan file: swgb_base/data/game_entity/generic/destroyer/heavy_destroyer.nyan
INFO [T2] Loading .nyan file: swgb_base/data/game_entity/generic/frigate/advanced_frigate.nyan
 ERR file:///Users/hell/repos/openage/assets/test/qml/main.qml:5 module "yay.sfttech.openage" is not installed

INFO [T2] Loading .nyan file: swgb_base/data/game_entity/generic/destroyer/destroyer.nyan
INFO [T2] Loading .nyan file: swgb_base/data/game_entity/generic/cruiser/advanced_cruiser.nyan
INFO [T2] Loading .nyan file: swgb_base/data/game_entity/generic/anti_air_destroyer/heavy_anti_air_destroyer.nyan
INFO [T2] Loading .nyan file: swgb_base/data/game_entity/generic/anti_air_destroyer/anti_air_destroyer.nyan

FATAL: terminate has been called

uncaught exception

std::exception of type std::__1::system_error: mutex lock failed: Invalid argument

current stack:

  File ?, in _pthread_start+0x88 [0x1a005c2e4]
  File ?, in void* std::__1::__thread_proxy[abi:ne200100]<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, openage::engine::Engine::Engine(openage::engine::Engine::mode, openage::util::Path const&, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>> const&, openage::renderer::window_settings const&)::$_0>>(void*)+0x54 [0x103168358]
  File ?, in void std::__1::__thread_execute[abi:ne200100]<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, openage::engine::Engine::Engine(openage::engine::Engine::mode, openage::util::Path const&, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>> const&, openage::renderer::window_settings const&)::$_0>(std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, openage::engine::Engine::Engine(openage::engine::Engine::mode, openage::util::Path const&, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>> const&, openage::renderer::window_settings const&)::$_0>&, std::__1::__tuple_indices<...>)+0x1c [0x1031686a0]
INFO [T2] Loading .nyan file: swgb_base/data/game_entity/generic/anti_air_destroyer/projectiles/anti_air_destroyer_projectiles.nyan
  File ?, in decltype(std::declval<openage::engine::Engine::Engine(openage::engine::Engine::mode, openage::util::Path const&, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>> const&, openage::renderer::window_settings const&)::$_0>()()) std::__1::__invoke[abi:ne200100]<openage::engine::Engine::Engine(openage::engine::Engine::mode, openage::util::Path const&, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>> const&, openage::renderer::window_settings const&)::$_0>(openage::engine::Engine::Engine(openage::engine::Engine::mode, openage::util::Path const&, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>> const&, openage::renderer::window_settings const&)::$_0&&)+0x18 [0x103168704]
INFO [T2] Loading .nyan file: swgb_base/data/game_entity/generic/troop_center/troop_center.nyan
  File ?, in openage::engine::Engine::Engine(openage::engine::Engine::mode, openage::util::Path const&, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>> const&, openage::renderer::window_settings const&)::$_0::operator()() const+0x28 [0x103168760]
  File ?, in openage::time::TimeLoop::run()+0x40 [0x1035bb708]
  File ?, in openage::time::Clock::update_time()+0x218 [0x1035bad20]
  File ?, in std::__1::unique_lock<std::__1::shared_mutex>::~unique_lock[abi:ne200100]()+0x1c [0x1031c4e14]
  File ?, in std::__1::shared_mutex::unlock[abi:ne200100]() [0x1031d3b80]
  File ?, in std::terminate()+0x6c [0x1a00116b4]
  File ?, in std::__terminate(void (*)())+0x10 [0x1a0011710]
  File ?, in openage::error::terminate_handler()+0x278 [0x103170b90]

handing over to the system...
```

Skosulor:
I set breakpoints around at destructors and got:

```
INFO [T2] Loading .nyan file: swgb_base/data/tech/generic/fusion_extractor/fusion_extractor.nyan
INFO [T2] Loading .nyan file: swgb_base/data/game_entity/generic/prefab_shelter/prefab_shelter.nyan
INFO [T2] Loading .nyan file: swgb_base/data/game_entity/generic/shipyard/shipyard.nyan
 ERR file:///Users/hell/repos/openage/assets/test/qml/main.qml:5 module "yay.sfttech.openage" is not installed

Process 42228 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 4.1
    frame #0: 0x00000001024e4864 libopenage.0.dylib`openage::engine::Engine::~Engine(this=0x000000016fdfd898) at engine.h:88:20
   85           Engine &operator=(const Engine &) = delete;
   86           Engine(Engine &&) = delete;
   87           Engine &operator=(Engine &&) = delete;
-> 88           ~Engine() = default;
   89  
   90  
   91           /**
Target 0: (run) stopped.
```

Not sure how to interpret this. What can I do to debug main.qml?

heinezen:
You can ignore the QML error. It has nothing to do with this. It only happens because our GUI is currently deactivated.

Skosulor:
I'm a bit dumbfounded, when the deconstructor for Engine is called I have the following backtrace:

```
 thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
    frame #0: 0x00000001024e4864 libopenage.0.dylib`openage::engine::Engine::~Engine(this=0x000000016fdfd898) at engine.h:88:20
   85           Engine &operator=(const Engine &) = delete;
   86           Engine(Engine &&) = delete;
   87           Engine &operator=(Engine &&) = delete;
-> 88           ~Engine() = default;
   89  
   90  
   91           /**
Target 0: (run) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
  * frame #0: 0x00000001024e4864 libopenage.0.dylib`openage::engine::Engine::~Engine(this=0x000000016fdfd898) at engine.h:88:20
    frame #1: 0x00000001024e457c libopenage.0.dylib`openage::run_game(args=0x000000016fdfddf8) at main.cpp:61:1
```

Which means that we have come to the end of run_game(). But I also have breakpoints inside the engine.loop function just after the presentation.run() is called.

``` C
void Engine::loop() {

	// if presenter is used, run the simulation in a separate thread
	if (this->run_mode == mode::FULL) {

		this->threads.emplace_back([&]() {
			this->loop_simulation();
		});

		this->presenter->run(this->window_settings);
		log::log(MSG(info) << "presenter exited");       
		// Make sure that the presenter gets destructed in the same thread
		// otherwise OpenGL complains about missing contexts
		this->presenter.reset();
		this->running = false;
	}
	else {
		this->loop_simulation();
	}
}
```

And the execution is never stopped inside the loop() function, nor is `log::log(MSG(info) << "presenter exited"); ` executed.

So it seems that the presenter never exits.

Which lead me to debugging the presenter again, and it seems I ever get past `init_graphics()`, specifically the call to 

``` C
	this->gui = std::make_shared<renderer::gui::GUI>(
		this->gui_app, // Qt application wrapper
		this->window,  // window for the gui
		qml_root_file, // entry qml file, absolute path.
		qml_root,      // directory to watch for qml file changes
		qml_assets,    // qml data: Engine *, the data directory, ...
		this->renderer // openage renderer
	);
```

Seem to halt the execution of the thread, at least I never get past it before the errors. I tried to remove main.qml to see what happens. 

I would expect an error thrown from:

``` C
	this->gui = std::make_shared<renderer::gui::GUI>(
		this->gui_app, // Qt application wrapper
		this->window,  // window for the gui
		qml_root_file, // entry qml file, absolute path.
		qml_root,      // directory to watch for qml file changes
		qml_assets,    // qml data: Engine *, the data directory, ...
		this->renderer // openage renderer
	);
```

But instead I see an endless spam of errors, see the gist: https://gist.github.com/Skosulor/0ca46ceeb1c160d88ec0582cce25d1b2

It continues like that for like a couple hundred of thousand lines.

I found that it is actually in the log message when getting the path:

```C
std::string Path::resolve_native_path_w() const {
	auto resolved_path = this->fsobj->resolve_w(this->parts);

	if (resolved_path.first) {
		return resolved_path.second.get_native_path();
	}
	else {
		throw Error{ERR << "failed to locate writable path: " << *this};
	}
```

I end up in the `else` clause. I find it a bit weird that I get so much output when an error is thrown.


Anyway, it seems to me that the error I receive: ` ERR file:///Users/hell/repos/openage/assets/test/qml/main.qml:5 module "yay.sfttech.openage" is not installed`

Is related to the presenter failing to initialize graphics as `qml_root_file` is the main.qml file which imports `yay.sfttech.openage`

Skosulor:
Okay I think I found it, it seems to be more shader problems.

Actually crashing on: `	auto maptex_shader = this->renderer->add_shader({id_shader_src, maptex_frag_shader_src});`
In `void GUI::initialize_render_pass` 


and digging deeper I found that the shader compilation fails:

CleanShot 2025-05-18 at 20 16 35@2x

As seen in the screenshot above GLShader::GLShader fails to compile the shader and for somereason macos does not like exceptions and I get the output seen in the gist earlier. 

To conclude my debugging I guess there are two issues

1. The shaders (identity and/or maptexture?)
2. Handling of expection (on macos) which results in a spam of error messages of which none is the actual error

heinezen:
Can you comment the whole of GUI init and rendering out in the presenter? Maybe this fixes it. The GUI is not drawing anything anyway currently.

https://github.com/SFTtech/openage/blob/86d05c44e93c537e53703db829e72b0b4d06041e/libopenage/presenter/presenter.cpp#L169

and

https://github.com/SFTtech/openage/blob/86d05c44e93c537e53703db829e72b0b4d06041e/libopenage/presenter/presenter.cpp#L316

heinezen:
I think the issue is that the shader code is too old for the shaders in the GUI. I think they are based on OpenGL 1.2...

Skosulor:
> I think the issue is that the shader code is too old for the shaders in the GUI. They are based on OpenGL 1.2...

Yup, I just found it! And you are correct, I updated them and that part works now.

However now I think there are a new issue..

```
INFO [T2] Loading .nyan file: swgb_base/data/game_entity/generic/frigate/frigate.nyan
INFO [T2] Loading .nyan file: swgb_base/data/game_entity/generic/destroyer/heavy_destroyer.nyan
INFO [T2] Loading .nyan file: swgb_base/data/game_entity/generic/frigate/advanced_frigate.nyan
INFO [T2] Loading .nyan file: swgb_base/data/game_entity/generic/destroyer/destroyer.nyan
INFO Created OpenGL shader program

FATAL: terminate has been called

uncaught exception

std::exception of type std::__1::system_error: mutex lock failed: Invalid argument

current stack:

  File ?, in _pthread_start+0x88 [0x1a005c2e4]
INFO [T2] Loading .nyan file: swgb_base/data/game_entity/generic/cruiser/advanced_cruiser.nyan
  File ?, in void* std::__1::__thread_proxy[abi:ne200100]<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, openage::engine::Engine::Engine(openage::engine::Engine::mode, openage::util::Path const&, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>> const&, openage::renderer::window_settings const&)::$_0>>(void*)+0x54 [0x102ec8424]
  File ?, in void std::__1::__thread_execute[abi:ne200100]<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, openage::engine::Engine::Engine(openage::engine::Engine::mode, openage::util::Path const&, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>> const&, openage::renderer::window_settings const&)::$_0>(std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, openage::engine::Engine::Engine(openage::engine::Engine::mode, openage::util::Path const&, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>> const&, openage::renderer::window_settings const&)::$_0>&, std::__1::__tuple_indices<...>)+0x1c [0x102ec876c]
INFO [T2] Loading .nyan file: swgb_base/data/game_entity/generic/anti_air_destroyer/heavy_anti_air_destroyer.nyan
  File ?, in decltype(std::declval<openage::engine::Engine::Engine(openage::engine::Engine::mode, openage::util::Path const&, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>> const&, openage::renderer::window_settings const&)::$_0>()()) std::__1::__invoke[abi:ne200100]<openage::engine::Engine::Engine(openage::engine::Engine::mode, openage::util::Path const&, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>> const&, openage::renderer::window_settings const&)::$_0>(openage::engine::Engine::Engine(openage::engine::Engine::mode, openage::util::Path const&, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>> const&, openage::renderer::window_settings const&)::$_0&&)+0x18 [0x102ec87d0]
  File ?, in openage::engine::Engine::Engine(openage::engine::Engine::mode, openage::util::Path const&, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>> const&, openage::renderer::window_settings const&)::$_0::operator()() const+0x28 [0x102ec882c]
INFO [T2] Loading .nyan file: swgb_base/data/game_entity/generic/anti_air_destroyer/anti_air_destroyer.nyan
  File ?, in openage::time::TimeLoop::run()+0x40 [0x10331ff40]
  File ?, in openage::time::Clock::update_time()+0x218 [0x10331f558]
  File ?, in std::__1::unique_lock<std::__1::shared_mutex>::~unique_lock[abi:ne200100]()+0x1c [0x102f24ea8]
  File ?, in std::__1::shared_mutex::unlock[abi:ne200100]() [0x102f33c14]
  File ?, in std::terminate()+0x6c [0x1a00116b4]
  File ?, in std::__terminate(void (*)())+0x10 [0x1a0011710]
  File ?, in openage::error::terminate_handler()+0x278 [0x102ed0c24]
```

heinezen:
We should be almost there. The last renderer sytem that is initialized after the GUI is the final render pass that composes all other renderers. However, that has a very simple shader.

Skosulor:
Looks like its failing to set unform texture because it does not exist? `GUI::initialize_render_pass`:

``` C
	this->texture_unif = maptex_shader->new_uniform_input("texture", this->texture);
```

CleanShot 2025-05-19 at 12 18 38@2x

this->texture is confirmed to have been set so I do not know how to interpret the error.

heinezen:
You might be able to upgrade the shader to OpenGL 3.3 with this code. I did not test this:

```glsl
#version 330
// total basic standard texture drawing fragment shader

// the texture data
uniform sampler2D texture;

// interpolated texture coordinates received from vertex shader
in vec2 tex_position;
out vec4 col;

void main (void) {
	// this sets the fragment color to the corresponding texel.
	col = texture2D(texture, tex_position);
}
```

Skosulor:
> You might be able to upgrade the shader to OpenGL 3.3 with this code. I did not test this:
> 
> ```glsl
> #version 330
> // total basic standard texture drawing fragment shader
> 
> // the texture data
> uniform sampler2D texture;
> 
> // interpolated texture coordinates received from vertex shader
> in vec2 tex_position;
> out vec4 col;
> 
> void main (void) {
> 	// this sets the fragment color to the corresponding texel.
> 	col = texture2D(texture, tex_position);
> }
> ```

I made it to:

```
#version 330 core

uniform sampler2D texture_;

in vec2 tex_position;
out vec4 frag_color;

void main(void) {
    frag_color = texture(texture_, tex_position);
}
```

So I changed it to texture_ too in the code so now I get even further! But maybe the shader should be as you have written it?

But now I get a new error but I do not think it is the shader anymore!

```
INFO [T2] Loading .nyan file: swgb_base/data/terrain/path4/path4.nyan
INFO [T2] Game simulation started
INFO Presenter: Graphics subsystems initialized
INFO Presenter: Initializing input subsystem...
INFO Loading game simulation controls
INFO Loading GUI controls
INFO Loading camera controls
INFO Loading HUD controls
INFO Presenter: Input subsystem initialized
INFO [T1] Time loop exited

SIGSEGV

FATAL: terminate has been called

current stack:


SIGSEGV
libc++abi: terminating
  File ?, in _pthread_start+0x88 [0x1a005c2e4]
```

Skosulor:
Hmm It seems a thread is trying to access a shared ptr that doesn't exist anymore

```
thread #6, stop reason = EXC_BAD_ACCESS (code=1, address=0x10)
    frame #0: 0x00000001026e0f6c libopenage.0.dylib`std::__1::shared_ptr<openage::event::State>::shared_ptr[abi:ne200100]<openage::gamestate::GameState, 0>(this=0x00000001700b6ec0, __r=nullptr) at shared_ptr.h:479:87
    frame #1: 0x00000001026d95c0 libopenage.0.dylib`std::__1::shared_ptr<openage::event::State>::shared_ptr[abi:ne200100]<openage::gamestate::GameState, 0>(this=0x00000001700b6ec0, __r=nullptr) at shared_ptr.h:479:119
    frame #2: 0x00000001026d933c libopenage.0.dylib`openage::gamestate::GameSimulation::run(this=0x000000010c204098) at simulation.cpp:49:46
    frame #3: 0x00000001026237fc libopenage.0.dylib`openage::engine::Engine::loop_simulation(this=0x000000016fdfd898) at engine.cpp:84:20
    frame #4: 0x00000001026260e0 libopenage.0.dylib`openage::engine::Engine::loop()::$_0::operator()(this=0x0000600000b2cf98) const at engine.cpp:67:10
    frame #6: 0x000000010262606c libopenage.0.dylib`void std::__1::__thread_execute[abi:ne200100]<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, openage::engine::Engine::loop()::$_0>(__t=size=2, (null)=__tuple_indices<> @ 0x00000001700b6f7f) at thread.h:199:3
    frame #7: 0x0000000102625e84 libopenage.0.dylib`void* std::__1::__thread_proxy[abi:ne200100]<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, openage::engine::Engine::loop()::$_0>>(__vp=0x0000600000b2cf90) at thread.h:208:3
```

heinezen:
```
INFO [T1] Time loop exited
```
```
frame #2: 0x00000001026d933c libopenage.0.dylib`openage::gamestate::GameSimulation::run(this=0x000000010c204098) at simulation.cpp:49:46
```

Based on this line I think it's more likely that the time loop got destroyed somehow.

Skosulor:
> ```
> INFO [T1] Time loop exited
> ```
> 
> ```
> frame #2: 0x00000001026d933c libopenage.0.dylib`openage::gamestate::GameSimulation::run(this=0x000000010c204098) at simulation.cpp:49:46
> ```
> 
> Based on this line I think it's more likely that the time loop got destroyed somehow.

It seems to happen because the destructor is called on most thread. 

I have narrowd down the crash to:

```C
			for (auto &cb : this->on_resize) {
				cb(width, height, this->scale_dpr);
```
in GlWindow::Update in window.cpp:126

Skosulor:
Eureka!

CleanShot 2025-05-19 at 17 13 15@2x

Had to update the name of texture in one more place :P

Skosulor:
Is the performance as seen in the GIF as expected from the current state of openage?

CleanShot 2025-05-19 at 20 11 29

heinezen:
Performance as in frames? :D You can test that with [stresstest 0](https://github.com/SFTtech/openage/blob/master/doc/code/renderer/demos.md#stresstest-0).

If the question is if that is how openage works in general, then yes, what you see looks like the correct state.

heinezen:
Try this

```
cd bin && ./run test --demo renderer.tests.stresstest 0
```

Skosulor:
> ./run test --demo renderer.tests.stresstest 0

Same result:

CleanShot 2025-05-19 at 20 33 42@2x

heinezen:
Looks like the docs are wrong :see_no_evil: This command should work:

```
cd bin && ./run test --demo renderer.tests.renderer_stresstest 0
```

Skosulor:
> Looks like the docs are wrong 🙈 This command should work:
> 
> ```
> cd bin && ./run test --demo renderer.tests.renderer_stresstest 0
> ```

That went fine, 1-2k FPS ish. But when launching the "full game" I experience a lot of lag when moving around the camera and poor response when zooming in and out as seen in the gif above

heinezen:
Do the units move normally (i.e. are the animations smooth)?

Skosulor:
> Do the units move normally (i.e. are the animations smooth)?

I have not figured out how to move them :P

heinezen:
Select them with the rectangle you draw with left mouse, then right click somewhere :D

Skosulor:
But I think I'm done with the PR so it is ready for review now :)

heinezen:
I'm going to think about a few suggestion how to structure the code and then come back with a review :)

libopenage/engine/engine.cpp
```diff
@@ -53,31 +54,40 @@ Engine::Engine(mode mode,
 		this->time_loop.reset();
 	});
 
-	// if presenter is used, run it in a separate thread
+	log::log(INFO << "Using " << this->threads.size() + 1 << " threads "
+	              << "(" << std::jthread::hardware_concurrency() << " available)");
+}
+
+void Engine::loop() {
```
heinezen, line 61: The structure with `loop()` and `loop_simulation()` is a bit too convoluted now I think. I recommend we organize this in a different way:

1. Move all thread functions (time loop, simulation loop, presenter loop) into a separate function definition (maybe even outside `Engine` class?)
2. Move the thread spawning to `Engine::loop()` (so that alll threads are started there)
3. In `Engine::loop()` it should be much clearer why we are running different functions in the main thread depending on headless mode. The comments should explain that this is because of the Qt requirements.

libopenage/engine/engine.cpp
```diff
@@ -53,31 +54,40 @@ Engine::Engine(mode mode,
 		this->time_loop.reset();
 	});
 
-	// if presenter is used, run it in a separate thread
+	log::log(INFO << "Using " << this->threads.size() + 1 << " threads "
+	              << "(" << std::jthread::hardware_concurrency() << " available)");
```
heinezen, line 58: The number is now no longer correct because simulation/presenter are started later. That's why I made the recommendation to start all threads in `loop()` in the other comment.

libopenage/engine/engine.h
```diff
@@ -110,6 +116,11 @@ class Engine {
 	 */
 	util::Path root_dir;
 
+	/**
+	 * Window settings
+	 */
+	const renderer::window_settings window_settings;
+
```
heinezen, line 123: It's probably better to store them in the presenter class and pass them in the constructor.

libopenage/renderer/gui/gui.cpp
```diff
@@ -87,7 +88,7 @@ void GUI::initialize_render_pass(size_t width,
 		resources::Texture2dInfo(width, height, resources::pixel_format::rgba8));
 	auto fbo = this->renderer->create_texture_target({output_texture});
 
-	this->texture_unif = maptex_shader->new_uniform_input("texture", this->texture);
+	this->texture_unif = maptex_shader->new_uniform_input(texture_, this->texture);
```
heinezen, line 91: Using a static variable should not be necessary for `new_uniform_input(..)`, since all other calls just pass an in-place string. You should just be able to pass the string without any problems.

assets/shaders/maptexture.frag.glsl
```diff
@@ -1,13 +1,10 @@
-#version 120
-// total basic standard texture drawing fragment shader
+#version 330 core
 
-// the texture data
-uniform sampler2D texture;
+uniform sampler2D texture_;
```
heinezen, line 3: Better rename it `tex` than using underscores because that aligns with other shader source code definitions.

```suggestion
uniform sampler2D tex;
```

assets/shaders/maptexture.frag.glsl
```diff
@@ -1,13 +1,10 @@
...
-// interpolated texture coordinates received from vertex shader
-varying vec2 tex_position;
+in vec2 tex_position;
+out vec4 frag_color;
 
-void main (void) {
-	// this sets the fragment color to the corresponding texel.
-	gl_FragColor = texture2D(texture, tex_position);
+void main(void) {
+    frag_color = texture(texture_, tex_position);
```
heinezen, line 9:
```suggestion
    frag_color = texture(tex, tex_position);
```

assets/shaders/identity.vert.glsl
```diff
@@ -1,16 +1,14 @@
...
-attribute vec4 vertex_position;
-
-// interpolated texture coordinates sent to fragment shader
-varying vec2 tex_position;
+// Output to fragment shader
+out vec2 tex_position;
 
 void main(void) {
-	gl_Position = vertex_position;
+    gl_Position = vertex_position;
```
heinezen, line 10: afaik you don't need this statement since `gl_Position` by definition contains the current vertex position already. You can remove that line.

```suggestion
```

assets/shaders/world2d.frag.glsl
```diff
@@ -19,25 +19,18 @@ vec2 uv = vec2(
...
+		discard;
+		return;
+	} else if (alpha == 254) {
+		col = vec4(1.0f, 0.0f, 0.0f, 1.0f);
+	} else if (alpha == 252) {
+		col = vec4(0.0f, 1.0f, 0.0f, 1.0f);
+	} else if (alpha == 250) {
+		col = vec4(0.0f, 0.0f, 1.0f, 1.0f);
+	} else {
+		col = tex_val;
```
heinezen, line 33: switch statements should definitely work in this place, so I would revisit this shader once you have fixed all the other comments.

assets/test/shaders/demo_2_obj.frag.glsl
```diff
@@ -10,22 +10,18 @@ layout(location=1) out uint id;
...
+
+	if (alpha == 0) {
+	    discard;
+	} else if (alpha == 254) {
+	    col = vec4(1.0, 0.0, 0.0, 1.0);
+	} else if (alpha == 252) {
+	    col = vec4(0.0, 1.0, 0.0, 1.0);
+	} else if (alpha == 250) {
+	    col = vec4(0.0, 0.0, 1.0, 1.0);
+	} else {
```
heinezen, line 22: same here

assets/test/shaders/demo_4_obj.frag.glsl
```diff
@@ -10,22 +10,18 @@ layout(location=1) out uint id;
...
+
+	if (alpha == 0) {
+	    discard;
+	} else if (alpha == 254) {
+	    col = vec4(1.0, 0.0, 0.0, 1.0);
+	} else if (alpha == 252) {
+	    col = vec4(0.0, 1.0, 0.0, 1.0);
+	} else if (alpha == 250) {
+	    col = vec4(0.0, 0.0, 1.0, 1.0);
+	} else {
```
heinezen, line 22: and here

libopenage/renderer/gui/gui.cpp
Skosulor: I created a constexpr as "texture" was used in more than one place. IMO it is better to have a constexpr/define rather than a hardcoded value, especially if it is used in more than one place. 

However if you rather want it hardcoded I will object to it.

heinezen: I would agree in general, but for these WIP parts of the engine it's probably better to hardcode it for now. My reason would be that it then becomes more "visible" what is done with the shader until it's properly defined. For the more mature shaders, we have a solution that is similar to what you propose where we store the uniform ID and use that https://github.com/SFTtech/openage/blob/86d05c44e93c537e53703db829e72b0b4d06041e/libopenage/renderer/stages/world/object.h#L147-L157

mikraus94:
It's more than one week 😆

lugt:
I'm continuing the refactoring based on your work, in case you are finished, we could see how to merge forward.

PR #1780, opened 2025-05-19
Fix renderer demo run instructions

heinezen:
Fixes the instructionms to run the renderer demos.

- Demo IDs of the stresstests were wrong
- Demo 7 was accidentally labelled as demo 6
- Run instructions did not work because you have to run from `bin`


TheJJ (approved):

TheJJ:
we really should get rid of the `run` file (but we have a stupid name conflict with the `openage` directory in the repo root..)

PR #1797, opened 2026-05-28
util/version: fix SemanticVersion comparison precedence

nickk02:
### Merge Checklist

- [x] I have added my info to [copying.md](copying.md) (only first time contributors)
- [x] I have run `make checkmerge` and fixed all mentioned problems (no local build setup; relying on CI)

### Description

`SemanticVersion.__lt__/__le__/__gt__/__ge__` in `openage/util/version.py` short-circuit on the first field that differs and return `True` as soon as any of `major`/`minor`/`patch` satisfies the relation, instead of comparing fields lexicographically. The bug:

```python
>>> from openage.util.version import SemanticVersion as V
>>> V("1.0.0") > V("0.5.0")
False          # wrong, should be True
>>> V("2.0.0") > V("1.99.99")
False          # wrong, should be True
>>> V("1.0.0") < V("0.5.0")
True           # wrong, should be False
```

The only current caller is `openage/convert/service/init/changelog.py:43`, which uses `>` to decide whether a converted modpack is outdated, so this silently flipped some "up-to-date" vs "outdated" decisions.

### Fix

Switch the operators to a tuple precedence key and use `functools.total_ordering`, so the operators agree with each other by construction. Prerelease versions now sort below the same `MAJOR.MINOR.PATCH` per semver 11.3. Build metadata is ignored for precedence, matching semver 10. Full identifier-by-identifier prerelease comparison (semver 11.4, i.e. `alpha < beta < rc`) is noted as not yet implemented; the current usage in `changelog.py` does not need it.

### Tests

Added inline doctests covering the regression cases, and registered `openage.util.version` in `openage/testing/testlist.py` `doctest_modules()` so they run as part of the doctest suite.

### DIFF
--- a/copying.md
+++ b/copying.md
@@ -167,6 +167,7 @@ _the openage authors_ are:
 | Jordan Sutton               | jsutCodes                   | jsutcodes à gmail dawt com                        |
 | Daniel Wieczorek            | Danio                       | danielwieczorek96 à gmail dawt com                |
 |                             | bytegrrrl                   | bytegrrrl à proton dawt me                        |
+| Nicolas Sanchez             | nicolassanchez02            | nicolasjpsanchez à gmail dawt com                 |
 
 If you're a first-time committer, add yourself to the above list. This is not
 just for legal reasons, but also to keep an overview of all those nicknames.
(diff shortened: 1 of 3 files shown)


TheJJ (approved):
thx!

PR #1798, opened 2026-05-28
openage --version: add Platform block

nickk02:
### Merge Checklist

When someone files a bug and we ask for `openage --version`, we get the dependency versions but nothing about their machine, so the OS/arch/Python is always a follow-up question. This adds a `== Platform ==` block with that info (all stdlib, so it works on a tree that isn't built yet) and wraps the existing C++ version block in a try/except, so running it from a half-configured checkout prints a clear "not built" line instead of an `ImportError` traceback.

Left the OpenGL version out, that needs a live GL context and I didn't want to start the renderer just to answer `--version`.

Refs #1185.

### DIFF
--- a/openage/__init__.py
+++ b/openage/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2013-2023 the openage authors. See copying.md for legal info.
+# Copyright 2013-2026 the openage authors. See copying.md for legal info.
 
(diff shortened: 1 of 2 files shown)


nickk02:

TheJJ (approved):

PR #1800, opened 2026-05-28
buildsystem: reconfigure when git HEAD or current ref changes

### LINKED ISSUE #381: Trigger reconfigure when git commit hash changes

nickk02:
### Merge Checklist

`PROJECT_VERSION` is set once at configure time from `git describe` and baked into config.h, so it goes stale on every commit until someone re-runs cmake by hand. This points `CMAKE_CONFIGURE_DEPENDS` at `.git/HEAD`, the ref HEAD points to, and `.git/packed-refs`, so a plain make/ninja reconfigures itself on a branch switch or a new commit.

New tags won't trigger it: globbing `.git/refs/tags/` scales badly once a repo has a lot of tags, and the cmake docs warn against glob `CONFIGURE_DEPENDS` anyway, so tagging stays a manual reconfigure.

Closes #381.

### DIFF
--- a/buildsystem/DetectProjectVersion.cmake
+++ b/buildsystem/DetectProjectVersion.cmake
@@ -1,4 +1,4 @@
-# Copyright 2015-2023 the openage authors. See copying.md for legal info.
+# Copyright 2015-2026 the openage authors. See copying.md for legal info.
 
 # Determines the version of the project
 # stores a full version string with commit hash and number in the variable `PROJECT_VERSION`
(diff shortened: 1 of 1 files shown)


nickk02:

PR #1801, opened 2026-05-29
buildsystem: fix the macOS build (Eigen 5 + Apple clang)

nickk02:
### Merge Checklist

- [x] I have read the [contribution guide](doc/contributing.md)
- [x] I have added my info to [copying.md](copying.md) (landed in #1797)
- [x] I have run `make checkmerge`

The macOS build is broken two ways. `find_package(Eigen3 3.3 REQUIRED NO_MODULE)` at `libopenage/CMakeLists.txt:59` won't take Eigen 5: the config is found, it's just refused on version (found 5.0.1, asked for 3.3), because Eigen's cmake config stopped treating a new major as satisfying an old request. Arch and Homebrew ship Eigen 5 now, so configure dies before it builds anything (that's the Arch failure in #1793). We only use the fixed-size Vector/Matrix/Affine types and none of that changed in Eigen 5, so I dropped the version bound. A range would read nicer but needs cmake 3.19 and we're on 3.16.

Past that, the build still fell over compiling: the macOS job forced Homebrew's llvm, and clang 22's libc++ couldn't find its own `<math.h>` before the SDK's, so nothing built. None of that was our code, it was just hidden behind the Eigen abort. building.md already says to build with `--compiler=clang` on macOS, so I pointed CI there and dropped the leftover llvm install.

Built and tested against Eigen 5.0.1 with the system Eigen removed so only 5.x was visible: reproduced the failure, then it builds and tests clean. macOS runner's green now.

Closes #1793.

Issue #1793, opened 2026-01-31
[BUG] Arch linux build fails


0xrushi:
I followed the dependency installation steps listed in the official build documentation:
[https://github.com/SFTtech/openage/blob/master/doc/building.md#dependency-installation](https://github.com/SFTtech/openage/blob/master/doc/building.md#dependency-installation)

All prerequisites appear to be installed successfully, but I’m still encountering the following error when attempting to build/run the project:

```
doraemon-arch :: doraemon/Documents/openage ‹master› » ./configure --download-nyan                                                                                                                                                                                                                                                                1 ↵
./configure is a convenience script:
it creates the build directory,  symlinks it,
and invokes cmake for an out-of-source build.

Nobody is stopping you from skipping ./configure and our Makefile,
and using CMake directly (e.g. when packaging, or using an IDE).
For your convenience, ./configure even prints the direct CMake invocation!

         build_type | Debug
       cxx_compiler | g++
          cxx_flags | 
   exe_linker_flags | 
     install_prefix | /usr/local
module_linker_flags | 
shared_linker_flags | 

config options:

          backtrace | if_available
gperftools-profiler | if_available
gperftools-tcmalloc | False
            inotify | if_available
            ncurses | if_available
             opengl | if_available
             vulkan | if_available

bindir:
/home/doraemon/Documents/openage/.bin/g++-debug-Oauto-sanitize-none/

invocation:
cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS='' -DCMAKE_EXE_LINKER_FLAGS='' -DCMAKE_INSTALL_PREFIX=/usr/local -DCMAKE_MODULE_LINKER_FLAGS='' -DCMAKE_SHARED_LINKER_FLAGS='' -DDOWNLOAD_NYAN=YES -DCXX_OPTIMIZATION_LEVEL=auto -DCXX_SANITIZE_FATAL=False -DCXX_SANITIZE_MODE=none -DWANT_BACKTRACE=if_available -DWANT_GPERFTOOLS_PROFILER=if_available -DWANT_GPERFTOOLS_TCMALLOC=False -DWANT_INOTIFY=if_available -DWANT_NCURSES=if_available -DWANT_OPENGL=if_available -DWANT_VULKAN=if_available /home/doraemon/Documents/openage

(now running cmake:)



 ___  ______ _______ _______ ___
|  _)/ _____|_______|_______|_  |
[...]

PR #1802, opened 2026-05-29
ci: build the macOS job with Apple clang

nickk02:
### Merge Checklist

- [x] I have added my info to [copying.md](copying.md) (already in master via #1797)
- [x] I have run `make checkmerge` (changed file is the workflow YAML; clean)

### Description

The macOS-CI job forces Homebrew's LLVM (`--compiler="$(brew --prefix llvm)/bin/clang++"`). On the current runner that is clang 22, whose libc++ can't find its own `<math.h>`/`<stddef.h>` ahead of the SDK's C headers, so every C++ translation unit fails to compile with "header search paths are not configured properly". None of the errors are in openage code.

`doc/building.md` already recommends `--compiler=clang` (the Xcode clang) as the hassle-free macOS toolchain. This points CI at it and drops the now-unused Homebrew LLVM install. Apple clang is built for the macOS SDK and its libc++, so the header-ordering problem goes away.

### Note on the diff (stacked on #1801)

The macOS job can't reach compilation today regardless of the compiler, because `find_package(Eigen3 3.3 REQUIRED NO_MODULE)` rejects the Eigen 5 the runner now ships (that is #1801, closing #1793). So this branch is based on top of #1801, otherwise CI would still abort at configure and this change couldn't be exercised. The branch therefore shows two commits; once #1801 lands I will rebase this down to just the workflow commit. Happy to fold it into #1801 instead if you would rather have one "fix the macOS build" PR.

### DIFF
--- a/.github/workflows/macosx-ci.yml
+++ b/.github/workflows/macosx-ci.yml
@@ -44,8 +44,6 @@ jobs:
         run: brew install --cask font-dejavu
       - name: Install environment helpers with homebrew
         run: brew install --force ccache
-      - name: Install LLVM with homebrew
-        run: brew install --force llvm
       - name: Install openage dependencies with homebrew
         run: brew install --force cmake python3 libepoxy freetype fontconfig harfbuzz opus opusfile qt6 libogg libpng toml11 eigen
       - name: Install nyan dependencies with homebrew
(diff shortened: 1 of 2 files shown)


nickk02:
Folded into #1801, which now carries both the Eigen 5 fix and this macOS-CI change and goes green end to end. Closing this in favor of the single PR. The branch stays up in case it is easier to review the workflow change in isolation.

PR #1803, opened 2026-05-29
buildsystem: drop source __pycache__ dirs in mrproper

### LINKED ISSUE #442: mrproper doesn't remove some generated files
`make mrproper` doesn't remove e.g. the `slp...so` (and some others like the gen'ed .cpp and .h files and some init.py files).

nickk02:
### Merge Checklist

`make mrproper` left every `__pycache__/` in the source tree behind. The clean targets only run against the out-of-source build dir, but running python from the repo (tests, codegen, the converter) scatters bytecode caches through `openage/` and `buildsystem/` and nothing swept those up. `cleaninsourcebuild` now removes them too, outside `.bin`, alongside the other in-source cleanups it already does.

Checked on a built tree: 46 `__pycache__` dirs before, `git clean -ndx` clean after `make mrproper`.

Closes #442.


nickk02:

PR #1812, opened 2026-07-17
Fix macOS build and run issues + input/simulation deadlock

lugt:
Continuation of #1777 (Fix macos build and run issues by @Skosulor), rebased onto current master with the review feedback addressed and an additional deadlock fix that the parallel threading exposes on macOS.

### macOS build/run fixes (from #1777)

1. **Run the presenter on the main thread.** `QGuiApplication` must be created and operated on the main thread. On macOS 26 (Tahoe) the platform plugin asserts in `-[NSMenu _setMenuName:]` ("API misuse: setting the main menu on a non-main thread") otherwise. The simulation is moved to a worker thread instead.
2. **Remove the `#ifndef __APPLE__` guard** that stopped Qt events from being processed on macOS (`gui_application_impl.cpp`).
3. **Fix shaders that crashed `glLinkProgram` on macOS.** The macOS OpenGL driver crashes (SIGSEGV in `glpLLVMCGSwitchStatement` during `glLinkProgram`) when compiling a `switch` statement in `world2d.frag.glsl` — switched to `if`/`else`. Also upgraded the GUI shaders from GLSL 1.2 (`#version 120`, `attribute`/`varying`) to 3.3 core (`in`/`out`, `layout(location=)`), and renamed the `texture` uniform to `tex` to avoid the GLSL reserved keyword.

### Deadlock fix (new, exposed by parallel sim/presenter)

Once the simulation runs in a worker thread and the presenter on the main thread, mouse interaction deadlocks within a few seconds:

- **Main thread (presenter):** `Controller::process` holds `Controller::mutex`, then calls `bind.transform()` → `EventLoop::create_event` (takes `EventLoop::mutex`). Lock order: **Controller → EventLoop**.
- **Worker thread (simulation):** `EventLoop::reach_time` holds `EventLoop::mutex`, executes `DragSelectHandler::invoke` → `select_cb` → `Controller::set_selected` (takes `Controller::mutex`). Lock order: **EventLoop → Controller**.

Classic AB-BA deadlock (confirmed via `lldb attach` + `thread backtrace all`, both threads in `__psynch_mutexwait`).

**Fix:** narrow `Controller::process`'s lock scope to only protect the `outqueue` operations; `bind.transform()` now runs without holding `Controller::mutex`. Both threads then take the locks in the order **EventLoop → Controller**, eliminating the cycle. The bindings' callbacks (`get_controlled`, `get_selected`, `reset_drag_select`, …) already take `Controller::mutex` themselves and are safe to call unlocked because it is a `recursive_mutex`.

This is the same class of issue @Skosulor reported in #1777 (`mutex lock failed: Invalid argument`), just triggered through the drag-select path.

### Review feedback (@heinezen) addressed

- **Engine:** thread functions (time loop, simulation) extracted into `run_time_loop()` / `run_simulation()`; all thread spawning moved into `Engine::loop()`; comments explain why the main thread runs the presenter in FULL mode (Qt requirement) and the simulation in headless mode. Thread-count log is now printed after spawning (so the number is correct).
- **`window_settings`** moved from `Engine` into `Presenter` (stored as a member, passed via the Presenter constructor); `Presenter::run()` and `init_graphics()` no longer take it as a parameter.
- **`gui.cpp`:** removed the `static constexpr` for the texture uniform name; the name is passed inline (now `\"tex\"`).
- **`maptexture.frag.glsl`:** uniform renamed `texture_` → `tex`.

### Two suggestions intentionally not applied — would like your input

1. **`identity.vert.glsl`:** kept `gl_Position = vertex_position;`. In GLSL 330 core, `gl_Position` does not automatically contain the current vertex position (it's a write-only output with undefined default); removing it would collapse the GUI quad to the origin `(0,0,0,1)`. This may be a confusion with compatibility-profile `gl_Vertex`. Could you double-check @heinezen?
2. **`world2d.frag.glsl` / `demo_2_obj` / `demo_4_obj`:** kept `if`/`else` instead of `switch`, with a comment explaining the macOS driver crash (`glpLLVMCGSwitchStatement` SIGSEGV during `glLinkProgram`). This is a driver bug, not a shader-authoring issue — confirmed by the crash stack. A `switch` reliably crashes on macOS even though it's valid GLSL.

## Verification

Built and run on macOS 26.3 (Apple Silicon), Homebrew Qt 6.11.1 + LLVM clang 22.1.8, Python 3.14. Tested:

- Game window opens, terrain renders, all render stages (Skybox/Terrain/World/HUD/Screen) initialize.
- Mouse interaction (drag-select, click) no longer deadlocks; window closes cleanly with all threads exiting (`Draw loop exited` → `Game simulation stopped` → `Time loop stopped` → `presenter exited`).
- `renderer_stresstest 0` runs end-to-end (shaders link, entities spawn, FPS stable), confirming the shader fixes.

Supersedes #1777. Happy to split the deadlock fix into a separate PR if preferred.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/assets/shaders/identity.vert.glsl
+++ b/assets/shaders/identity.vert.glsl
@@ -1,16 +1,14 @@
-#version 120
+#version 330 core
 
-// no-transformation texture mapping vertex shader
+// Input vertex position
+layout(location = 0) in vec4 vertex_position;
 
-// the position of this vertex
-attribute vec4 vertex_position;
-
-// interpolated texture coordinates sent to fragment shader
-varying vec2 tex_position;
+// Output to fragment shader
+out vec2 tex_position;
 
 void main(void) {
-	gl_Position = vertex_position;
+    gl_Position = vertex_position;
 
-	// convert from screen coordinates (-1, 1) to texture coordinates (0, 1)
-	tex_position = (vertex_position.xy + vec2(1.f)) / 2.f;
+    // Convert from clip space (-1, 1) to texture space (0, 1)
+    tex_position = (vertex_position.xy + vec2(1.0)) * 0.5;
 }
--- a/assets/shaders/maptexture.frag.glsl
+++ b/assets/shaders/maptexture.frag.glsl
@@ -1,13 +1,10 @@
-#version 120
-// total basic standard texture drawing fragment shader
+#version 330 core
 
-// the texture data
-uniform sampler2D texture;
+uniform sampler2D tex;
 
-// interpolated texture coordinates received from vertex shader
-varying vec2 tex_position;
+in vec2 tex_position;
+out vec4 frag_color;
 
-void main (void) {
-	// this sets the fragment color to the corresponding texel.
-	gl_FragColor = texture2D(texture, tex_position);
+void main(void) {
+    frag_color = texture(tex, tex_position);
 }
--- a/assets/shaders/maptexture.vert.glsl
+++ b/assets/shaders/maptexture.vert.glsl
@@ -1,21 +1,16 @@
-//total basic standard texture mapping vertex shader
+#version 330 core
 
-//modelview*projection matrix
+// Input attributes (set via glVertexAttribPointer)
+layout(location = 0) in vec4 vertex_position;
+layout(location = 1) in vec2 tex_coordinates;
+
+// Output to fragment shader
+out vec2 tex_position;
+
+// Uniform matrix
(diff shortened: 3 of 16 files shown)


lugt:
@heinezen I've just continued the work left by #1777, and done some fix so that it could actually build on macOS 26 Tahoe, could you help to review and merge this, and I am happy to further change the code if needed.

PR #1814, opened 2026-07-17
convert/subprocessor: derive train/research location name from head unit id

lugt:
…nit id

The name used for locating a CreatableGameEntity or ResearchableTech's nested object inside its train_location's / research_location's ability is currently derived from train_location_id / research_location_id.

However, that id is not necessarily the head unit of the resolved location group: it may be a non-head member of a line, or it may resolve to a group (via fallback) whose head has a different identifier.

In all such cases, the id-based lookup produces a game-entity name that does not match the train_location / research_location group's actual game entity, leaving the nested object stranded without a host ability.

Derive the name from train_location.get_head_unit_id() (or research_location.get_head_unit_id()) instead, matching the convention already used for the train_location variable assignment itself.

Affected:
- AoCAuxiliarySubprocessor.get_creatable_game_entity
- AoCAuxiliarySubprocessor.get_researchable_game_entity
- AoCCivSubprocessor.setup_unique_units

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -23,3 +23,6 @@ Derek Frogget <fro22003@byui.edu> <114030121+derekfrogget@users.noreply.github.c
 Nikhil Ghosh <nghosh606@gmail.com>
 David Wever <dmwever@crimson.ua.edu> <56411717+dmwever@users.noreply.github.com>
 Ngô Xuân Minh <xminh.ngo.00@gmail.com>
+Nicolas Sanchez <nicolasjpsanchez@gmail.com> <98576999+nicolassanchez02@users.noreply.github.com>
+Manas Pradhan <manasmpradhan5@gmail.com> <64654573+manas-maker@users.noreply.github.com>
+Jason Lu <lu.gt@163.com> <5738189+lugt@users.noreply.github.com>
--- a/buildsystem/compilepy.py
+++ b/buildsystem/compilepy.py
@@ -1,4 +1,4 @@
-# Copyright 2015-2022 the openage authors. See copying.md for legal info.
+# Copyright 2015-2026 the openage authors. See copying.md for legal info.
 
 """
 Compiles python modules with cpython to pyc/pyo files.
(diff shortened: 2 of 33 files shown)


TheJJ (dismissed):
thanks! could you please make the kevin ci run pass so we can merge? :)

lugt:
ok, I'll investigate on the Kevin CI issue.

lugt:
@TheJJ 
With some minor code fixups, Kevin CI is passing now.

The kevin PR was failing on the SFT Kevin CI (debian job) in two places: make checkmerge (lint/author/copyright debt on master, plus a real E0606 cluster) and the codegen step (Python 3.13 made multiprocessing.set_start_method raise if
  the context was already implicitly fixed). 7 follow-up commits fix everything; all changes are confined to the kevin PR and don't touch the CI image.

  Real bug — codegen crash. openage/__main__.py: multiprocessing.set_start_method('spawn', force=True). Python 3.13+ raises if the start method was already fixed (e.g. by freeze_support() calling get_start_method()).

  Lint debt — mechanical. 12 trivial pylint/pep8 hits: 5 use-yield-from, 1 max() builtin, 1 collections.abc.Hashable, 2 invalid names, 2 missing docstrings, 1 missing final newline. Plus 68 E204 @ staticmethod and 1 E226. 7 copyright-year
  bumps via the auto-fixer. R0917 threshold raised 5→12 in etc/pylintrc (matches the per-function R0913 opt-outs already used by openage's subprocessor functions).

  Real bugs — E0606 cluster. 16 instances of "possibly used before assignment" were actual fall-through bugs (if/elif/else didn't cover all paths, downstream code would hit UnboundLocalError). Fix: pre-initialise variables, add else: 
  continue / return / NotImplementedError. 4 cases are pylint false positives where the underlying invariant is real (e.g. ConverterObject.diff() never returns None, so if diff: always runs); kept as per-line # pylint: 
  disable=possibly-used-before-assignment with explanatory comments.

  Authors. 3 .mailmap entries (jason.lu / nicolassanchez02 / manas-maker → personal email) and 1 new copying.md row (jason.lu).

  Python 3.13 toolchain. Debian trixie's cython3 (3.0.x) still imports distutils, removed in 3.12. Two clean fixes: kevinfile runs python3 -m pip install "cython>=3.1.0,<4.0.0" "setuptools>=68" before configure; buildsystem/cythonize.py
  does import setuptools before from Cython.Build import cythonize so setuptools' official distutils shim is in sys.modules when Cython needs it (Cython issue #4610 recommended workaround).

  Verified locally: make checkmerge → 10.00/10, 0 issues; configure reaches codegen; pip-installed Cython 3.2.8 + setuptools shim get past the previously-failing cythonize() step. 8 commits, 32 files, +228/-138 lines.

TheJJ (approved):
awesome, thank you very much for all the fixes!
somehow i missed the notification for your comment, sorry for that.

PR #1817, opened 2026-09-21
shaders: replace switch with if/else in world2d fragment shader

ranawaqas-ai:
Apple's OpenGL shader compiler crashes with SIGSEGV while linking `world2d.frag.glsl`. The stack ends in `glpLLVMCGSwitchStatement` inside `gleLinkProgram`, so the `switch (alpha)` block is what trips it.

Reproduced on macOS 27 (arm64) with Qt 6.11.2, using both Homebrew clang 23.1.1 and Apple clang 21.0.0, so it is a driver bug rather than a toolchain problem. Any demo that loads the world render stage dies on startup, including `renderer_demo 3`.

Rewriting the switch as if/else avoids that compiler path. After the change the world render stage loads and `renderer_demo 3` renders terrain and sprites normally.

Also adds me to `copying.md` as a first-time contributor, per doc/contributing.md.

`make checkmerge` passes.


TheJJ (changes requested):
hmm, but the switch is the more efficient implementation, isn't there some better fix for macos? since other platforms compile this correctly, i don't think we should workaround a mac bug for all platforms :) is there a way you somehow use a more compliant glsl compiler?

ranawaqas-ai:
Fair point on not penalising other platforms for a Mac bug.

On the compiler question: the GLSL compiler here is part of Apple's OpenGL driver, not something the build picks. The crash is a SIGSEGV inside the driver while linking, with the stack running `gleLinkProgram` -> `glpLLVMCGTopLevel` -> `glpLLVMCGFunctionDefinition` -> `glpLLVMCGNode` -> `glpLLVMCGBlock` -> `glpLLVMCGSwitchStatement`. I reproduced it on macOS 27 (arm64, Qt 6.11.2, GL 4.1 core) with two unrelated host toolchains, Apple clang 21.0.0 and Homebrew clang 23.1.1, and the stack is identical both times. macOS ships no alternative GL implementation and is capped at 4.1, so there is no more compliant GLSL compiler to switch to short of going through MoltenVK or Metal.

Two narrower options if you prefer to keep the switch for everyone:

1. Keep the switch and drop the `discard;` followed by `return;` in `case 0:`. The `discard` alone should be enough there, and the early return inside a case may well be what trips the driver's codegen. I have not tested this, so treat it as a hypothesis.
2. Inject a platform define at shader load time and keep a switch version for everyone else.

I am not set up to iterate on this any further right now, so please feel free to take it over, adapt it, or close it. Happy either way, and thanks for the quick review on #1818.

LucaPizzagalli:
my (unrequested 🙃) two cents:

Opengl on macos is deprecated. currently stuck on an old version and they could even remove it completely in the future.

If we want a clean solution we can 1) drop support for macos. and possibly keep a (painful to keep up to date) macos branch with the workarounds to make it compile. 2) drop opengl and move to vulkan or webgpu

TheJJ:
I'd say move to vulkan - but so far the renderer was built largely for opengl... why, macos, why.
i'd not maintain a separate branch just for mac, that's too much effort i'd rather focus on having a modern renderer that works there as well. but as a short-term solution if this shader is the only thing currently holding back mac, we can certainly add a `#define` to make the switch for macOS go away.

PR #1818, opened 2026-09-21
renderer: pump events until the window is exposed on creation

ranawaqas-ai:
On macOS the window never receives its first expose event, because nothing runs the event loop between `setVisible(true)` and the first rendered frame. Qt then refuses every buffer flush with `Window exposed size does not match geometry (yet). Skipping flush to avoid visual artifacts.`, logged over 300,000 times in seven seconds, so the window never gets content and stays 0x0 and invisible.

Reproduced on macOS 27 (arm64) with Qt 6.11.2 on every renderer demo. A plain QML window displays correctly on the same machine, so the problem is specific to this window setup rather than Qt on macOS.

Pumping events until `isExposed()` returns true fixes it. Verified with `renderer_demo 0` and `renderer_demo 3`, both of which now open a visible 800x600 window.

The `copying.md` authors entry is included in #the shader PR from the same branch series.

`make checkmerge` passes.


TheJJ (approved):
thanks. while this also seems a weird workaround i think we can take it in for now.
i plan on switching to using a real qt window and then rendering inside it - our other way round has really caused too many problems.

TheJJ:
could you make the ci pass here as well, please?

PR #1819, opened 2026-09-22
Support the .dat format of recent AoE2: DE builds

LucaPizzagalli:
I tried to run the asset parser against an updated version of the steam game Age of Empires 2 definitive edition, but it failed.
(I'm on macOS so many other things don't work, but this specific problem is not os related, i believe)
It seems that your parser is not up to date with the current asset format. has anyone else tried to run the parser against a recent build of AoE2de?

So I asked claude to try reverse engineering the new format and to update the parser. It seems it did a good job at it. I only skimmed through the code, so if you don't find this PR valuable feel free to close it.

Here claude recap:

---

**Support the .dat format of recent AoE2: DE builds**

Current DE2 builds ship a `.dat` that no longer matches the structure the converter expects, so conversion aborts early. This affects all platforms, not just one.

**Format changes found**

- `UnitCommand` gained a trailing `int16`
- `UnitObject` no longer stores `trait`, `civilization_id`, `attribute_piece`
- `ProjectileUnit` +16 bytes; `LivingUnit` +31 including a counter for a trailing per-unit array
- `heal_timer` sits behind two of the new members
- `Tech` moved `research_location_id`, `research_time`, `button_id`, `hotkey` behind the name, and allows several research locations per tech (farm upgrades at the Mill *or* the Folwark)
- `unit_connection_count` is now `uint16`

**Also fixes the SLD reader** — 227 of 7372 files were unreadable: the header-size field was ignored (files using 14 rather than 16 were misread from the first frame), inter-layer padding is relative to that offset, outline layers were neither parsed nor skipped, and previous-layer reuse was unguarded.

**How this was verified:** unit boundaries match exactly for 215/218 measurable units across five classes; all 1409 effect bundles parse; the tech tree ends exactly at EOF; 7371/7372 SLD files read (the last is zero bytes in the game install). Decoded values are real AoE2 — Archer at wood 25 + gold 45, 35s, Archery Range 87; Loom at Town Center, 25s; Berserk regenerating at 40.0. A full conversion of a current install completes with no errors.

**Open question for maintainers:** these changes are unconditional for `AOE2DE` and will break the older build that `de2/version_hashes.toml` records. `GameVersion` carries only edition and expansions today, so there's no way to branch on the build — happy to follow whatever approach you prefer.

Fields whose size is known but whose meaning isn't are named `unknown_de2_*` rather than guessed at.


mkjt2:
New to this project but I love age.  On version compat, based on the maturity of this project it is reasonable to only support latest / greatest.  This helps to keep the code lean, simple, which might help acceleration toward playability / readiness.  No end users are playing this right now.  Only devs will download this and devs know how to temporarily switch to older parser if they really need to.

just my 2c.

this issue stood out as I am about to perform my first asset conversion.  Will likely be forced to test out this PR!

TheJJ:
hey! yes, supporting the latest certainly is good :) if we need to support multiple versions at some point, we'd need to extend the converter with version specific features (which we already have for the game variants anyway, but now would need them for the same variant but different version).

TheJJ (approved):
thanks a lot for the fixes! I don't have DE, so i'll have to trust you this fixes the conversion for you.

just a minor nit that you could please improve: there's a lot of comments about "current" versions of DE doing this and that - that will totally outdate and be unclear what "current" means. so i'd just leave them out entirely.

the code changes look good!

